PR #10, opened 2022-08-24
feat: allow tuples as return values in templates

mrnaveira:
Description
---
* Boilerplate code for supporting tuple types. Followed the existing `Type::Other` in ABI definition.
* Replacement, in tuples, of `Self`-type fields for the corresponding `component_id`.

Motivation and Context
---
Templates should allow Rust tuples as return types. Specially in constructor functions, as it will be needed to allow the implementation of the permission system using _badges_ (we have [some examples](https://github.com/tari-project/wasm-examples/blob/master/emoji_id/src/lib.rs#L29) already).

The most simple example of this scenario can be seen below:
```
#[template]
mod tuple_template {
    pub struct Tuple {
        pub value: u32,
    }

    impl Tuple {
        pub fn new() -> (Self, String) {
            (Self { value: 0 }, "Hello World!".to_string())
        }
    }
}
```

How Has This Been Tested?
---
New test template `tuples` with scenarios for regular tuples as well as tuples as part of a constructor.


sdbondi (approved):
LGTM

sdbondi (approved):

stringhandler (approved):

PR #15, opened 2022-09-05
feat: transaction manifest

mrnaveira:
Description
---
* New `transaction_manifest` crate that parses a subset of Rust code into a list of engine instructions.
* The instruction definition is currently a custom one, not integrated with the actual wasm engine. This is to be able to rapidly iterate and identify the set of instruction types needed. Afterwards they can be integrated into the engine and gRPC interfaces

Motivation and Context
---
The client side of templates should have a human readable way of building transactions. That transaction would then be sent to the wasm engine to execute atomically.

How Has This Been Tested?
---
Unit test with an example transaction of a NFT sell.


dan_layer/transaction_manifest/src/instruction.rs
```diff
@@ -0,0 +1,104 @@
...
+//  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
+//  INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+//  DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+//  SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+//  SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
+//  WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
+//  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Instruction {
```
jorgeantonio21, line 24: Could we add documentation sketching how one uses `Instructions` in a simple example ?

dan_layer/transaction_manifest/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+
+    // TODO: component address, etc
+
+    let args = build_call_args(expr.args);
+
+    Instruction::CallMethod {
+        package_address: String::new(),
+        component_address: String::new(),
+        method,
+        proofs: vec![],
```
jorgeantonio21, line 154: We don't require any proofs for `CallMethod` ?

dan_layer/transaction_manifest/src/instruction.rs
mrnaveira: We have [a repo with some template examples](https://github.com/tari-project/wasm-examples). In each template we wrote unit tests, and each statement in a test would be roughly translated into an instruction.

For example, in a NFT template test we have this sentence:
```let (picture, _) = picture_seller.buy(payment).unwrap();```
This would translate into a `Instruction::CallMethod`. The whole test would be a single transaction with as many instructions as needed to translate the operations in the test.

Take in account that those examples are just purely for showcase and establish a direction, for the most part they do not compile/execute.

dan_layer/transaction_manifest/src/lib.rs
mrnaveira: The idea of this PR was to create a first version of the transaction manifest. In this case, `proofs` are just a placeholder but the idea was to include them whenever a badge was required by a method. If no badge is required, then we do not need to include a proof in the instruction call.

mrnaveira:
Leaving this PR as draft for now, as we still need to decide the direction we are going with the client-side manifest API.

sdbondi (approved):
I like it

PR #38, opened 2022-09-27
feat: allow submitting transactions from jsonrpc

stringhandler:
This breaks the "PR does one thing rule", so sorry ahead of time. 

* hook up submit transaction JSON-Rpc to mempool handle
* convert new_tx receiver in mempool to broadcast channel. This was needed to allow cloning of the MempoolHandle. Perhaps there's a better way to do this since it is only Hotstuff that cares about new transactions for now
* created a basic HTTP UI (hello there, world) server for future use

I also fixed a bug where the JSON RPC  (and I presume the comms services) would not start because the base layer scanner was awaited instead of spawning


applications/tari_validator_node/src/p2p/services/mempool/initializer.rs
```diff
@@ -20,22 +20,25 @@
...
 };
 
 pub fn spawn(
-    new_transactions: mpsc::Receiver<(CommsPublicKey, Transaction)>,
+    new_transactions: mpsc::Receiver<Transaction>,
+    new_transactions_sender: mpsc::Sender<Transaction>,
     outbound: OutboundMessaging,
 ) -> MempoolHandle {
-    let (tx_valid_transactions, rx_valid_transactions) = mpsc::unbounded_channel();
+    let (tx_valid_transactions, rx_valid_transactions) = broadcast::channel(100);
```
sdbondi, line 39: Only concern here is that we will drop transactions if we cant process them fast enough and we should account for a massive amount of transactions on the 2nd layer. We should probably separate the inbound tx channel from the handle because (1) there is only going to be one receiver of transactions and (2) so that the mempool handle can be copied.

sdbondi (approved):
LGTM

applications/tari_validator_node/src/p2p/services/mempool/initializer.rs
stringhandler: Yeah was also thinking this

PR #39, opened 2022-09-27
feat: new CLI command to register templates in the base layer

mrnaveira:
Description
---
* The validator node CLI now has two main commands:
    * `vn` with the subcommand: `register`
    * `template` with the subcommand: `register`
* New `template register` command (allows both command line arguments as well as interactive input) for template registration
* Added JSON-RPC method (`register_template`) for template registration in the validator node. The existing validator node registration was renamed to `register_validator_node`

Motivation and Context
---
Up until now, to register templates in the base layer the only option was to craft a gRPC request to the wallet. This PR aims to provide an easy to use CLI command to avoid manually creating gRPC requests.

How Has This Been Tested?
---
Manually tested:
1. Start the `tari_base_node` and `tari_console_wallet` from the feature-dan branch.
2. Start the validator node
3. Invoke the validator node CLI to register the `hello-world` template
4. Mine some blocks for the template registration transaction to be included and confirmed
5. Use the gRPC method in the base node to retrieve the published template


sdbondi:
I think we should have the commands in the form `resource` `subcommand` e.g
`vn_cli templates publish ...`
`vn_cli vn register`
`vn_cli mempool show`

mrnaveira:
> If you agree, I think we should have the commands in the form `resource` `subcommand` e.g `vn_cli templates publish ...` `vn_cli vn register` `vn_cli mempool show`

Sounds good to me, I will adapt this PR to this

stringhandler (approved):

PR #45, opened 2022-10-04
feat: epoch manager stores/fetches vns from database

mrnaveira:
Description
---
* Created SQL migrations and `diesel` boilerplate for storing validator nodes in sqlite.
* Added utility functions to store and fetch validator nodes, associated with a specific epoch.
* Refactored the `EpochManager` to use the persisted vns instead of in-memory storage as until now.

Motivation and Context
---
The epoch manager should persist in database the validator nodes retrieved from the base layer, so that after restart the information is not lost.

**Important:** I made the assumption that the `shard_key` of a validator node may be different in different epochs. If that's not the case then there may be some database storage optimisations to be done.

How Has This Been Tested?
---
Testing still in progress (draft PR)


stringhandler (commented):
Looks correct so far

stringhandler (approved):

PR #49, opened 2022-10-05
fix: negative begin index in committee

Cifko:
Description
---
Rust modulo on negative numbers is negative, that leads to wrong begin index.
Epoch valid had a similar issue with negative number.

How Has This Been Tested?
---
Manually, by running 17 VNs and checking their committees.


stringhandler (approved):

stringhandler:
Interesting, I didn't know that

PR #68, opened 2022-10-05
feat: move tooling into main lib for reuse outside of package

stringhandler:
Move tooling so that the Test classes can be used in repos outside of this code base.

The tooling is hidden behind a feature `tooling` so that it isn't always included


sdbondi (approved):
LGTM, Maybe test-tooling

sdbondi (approved):

PR #69, opened 2022-10-06
refactor: add templates to global db, storage crate

sdbondi:
Description
---
- tari_dan_storage crate for abstract interface to storage
- remove chaindb
- move templates to global db

Motivation and Context
---
chaindb is no longer used. Templates should live in global db 

How Has This Been Tested?
---
Manually


stringhandler (approved):
Nice

PR #70, opened 2022-10-06
feat: enhance vn web-gui

Cifko:
Description
---
Add committees to the VN web-gui.
- you see an interval representation to see if the shard space covered by the committee wraps around
- there is a show/hide committee button


sdbondi (approved):
LGTM - think we should move the html/css stuff out of src into something like `tari_validator_node/web_ui`

PR #81, opened 2022-10-12
feat: make jrpc rust client

sdbondi:
Description
---
JRPC rust client
Remove old and unused validator node grpc code
Use default jrpc port for cli

Motivation and Context
---

How Has This Been Tested?
---


stringhandler (approved):
agree

PR #84, opened 2022-10-12
feat: web gui react+typescript

Cifko:
Description
---
Rewrite the web gui into react+typescript.
Features:
- register button, but it does get only partial feedback from the vn. 
- auto refresh (2 minutes)
- react web is included in the binary
- jrpc port is provided by the VN

Motivation and Context
---
It's easier to maintain react code than vanilla js.


applications/tari_validator_node/src/http_ui/server.rs
```diff
@@ -22,15 +22,24 @@
...
+use tower_http::services::ServeDir;
 
 const LOG_TARGET: &str = "tari_validator_node::http_ui::server";
 
 pub async fn run_http_ui_server(address: SocketAddr, json_rpc_address: Option<String>) -> Result<(), anyhow::Error> {
-    let router = Router::new().route("/", get(index)).layer(Extension(Arc::new(
-        json_rpc_address.unwrap_or_else(|| "127.0.0.1:18145".to_string()),
-    )));
+    let serve_dir_service = get_service(ServeDir::new(
+        "d:/code/work/tari-dan/applications/tari_validator_node_web_ui/build",
```
sdbondi: Can we use a relative path here?
We could also include the built assets directly in the binary as per
https://github.com/tari-project/tari-dan/issues/86

Cifko: Oh, I forgot about that. I will look into the `include_dir`. I also need to pass somehow the json_rpc port from the VN to the react app.

sdbondi: Yeah that's tricky, maybe create a config struct, serve that struct as json at `/config.json` and have the js load that. Or you could string replace e.g. `// $CONFIG$` with `const CONFIG = {....}` in the compiled js

Cifko: I added a special route `/json_rpc_address` to the http server that provides the port back to the react ui. This could be avoided if we decide to merge jrpc and http to the same server, I don't see a reason why not. The jrpc can be served from a special route e.g. all request to /jrpc could be served by that handlers. But I'm not sure if we have a reason to run two servers.

stringhandler (approved):
utACK

Issue #66, opened 2022-10-05
[state-sync] Add state sync p2p RPC method


stringhandler:
Acceptance criteria:

1. Add a state sync RPC method that can stream back the current state required to bootstrap a VN to the shard committee
2. Add a basic unit test for this method.

stringhandler:
- [ ] Add RPC to retrieve data from sync peer
- [ ] Add client code to retrieve data on epoch change and start up

PR #91, opened 2022-10-13
feat: add p2p state sync for validator nodes (see issue #66)

jorgeantonio21:
Description
---
Add p2p state sync for validator nodes

Motivation and Context
---

How Has This Been Tested?
---


applications/tari_validator_node/proto/dan/network.proto
```diff
@@ -46,3 +46,24 @@ message GetPeersResponse {
   repeated bytes addresses = 2;
   IdentitySignature identity_signature = 3;
 }
+
+message VNStateSyncRequest {
+  // Range
+  tari.dan.common.ShardId start_shard_id = 1;
```
stringhandler: These are probably not needed

stringhandler (commented):
Looks great so far. Please check the conflicts, specifically in the database. 

I don't think the start and end shards are needed for the sync request, since the list of shards should be sufficient. 

I'll want to review again after the conflicts are resolved, but looking on the right track so far

dan_layer/storage_sqlite/src/sqlite_shard_store_factory.rs
```diff
@@ -712,6 +715,79 @@ impl ShardStoreTransaction<PublicKey, TariDanPayload> for SqliteShardStoreTransa
...
+            })
+            .unwrap();
+
+        if let Some(substate_states) = substate_states {
+            substate_states
+                .iter()
+                .map(|ss| deserialize::<ShardId>(ss.shard_id.as_slice()))
+                .collect::<Result<Vec<_>, _>>()
+        } else {
+            return Err(Self::Error::NotFound {
```
SWvheerden: I would argue that we should rather return `Vec::new()` and not an `Error::NotFound`
Because we are looking at a slice of the shard space, it is possible that it could be empty. 
If where we looking a specific shard, then yes `notFound` is a good error to return, but we are just searching a slice of shard space.

dan_layer/storage_sqlite/src/sqlite_shard_store_factory.rs
```diff
@@ -712,6 +715,79 @@ impl ShardStoreTransaction<PublicKey, TariDanPayload> for SqliteShardStoreTransa
         Ok(())
     }
 
+    fn get_state_inventory(&self, start_shard: ShardId, end_shard: ShardId) -> Result<Vec<ShardId>, Self::Error> {
```
SWvheerden, line 718: Should we not try and combine these two functions?

The actual sqlite query is the same, we just wrap the data differently? 
Either we need to wrap it differently upstream or keep them the same and just change how we use the data?

dan_layer/storage_sqlite/src/sqlite_shard_store_factory.rs
```diff
@@ -712,6 +715,79 @@ impl ShardStoreTransaction<PublicKey, TariDanPayload> for SqliteShardStoreTransa
...
+                    "DOWN" => Ok(SubstateState::Down {
+                        deleted_by: PayloadId::try_from(ss.deleted_by_payload_id.clone().unwrap_or_default())?,
+                    }),
+                    _ => Err(StorageError::InvalidSubStateType {
+                        substate_type: ss.substate_type.clone(),
+                    }),
+                })
+                .collect::<Result<_, _>>()
+        } else {
+            return Err(Self::Error::NotFound {
```
SWvheerden: Same argument as above

SWvheerden (commented):
Looks good. 
I just have a philosophical change on what to do with no shards. But that's a discussion point.

applications/tari_validator_node/proto/dan/network.proto
jorgeantonio21: Hey @stringhandler, I removed these from use. Let me know if you think it is fine !

dan_layer/storage_sqlite/src/sqlite_shard_store_factory.rs
jorgeantonio21: Refactored this piece of code.

dan_layer/storage_sqlite/src/sqlite_shard_store_factory.rs
jorgeantonio21: Refactored this piece of code.

dan_layer/storage_sqlite/src/sqlite_shard_store_factory.rs
jorgeantonio21: I agree, let me change it

applications/tari_validator_node/src/p2p/rpc/service_impl.rs
```diff
@@ -119,4 +143,90 @@ where TPeerProvider: PeerProvider + Clone + Send + Sync + 'static
...
+        let missing_shard_ids = msg
+            .missing_shard_state_ids
+            .iter()
+            .map(|s| {
+                ShardId::try_from(s.bytes.as_slice())
+                    .expect("Invalid gRPC request: failed to parse shard id's request data")
+            })
+            .collect::<Vec<ShardId>>();
+
+        if missing_shard_ids.is_empty() {
```
SWvheerden, line 190: Edge case here, and if the network is running very likely that won't ever be empty after a small period. 
But when the network just started, and the first nodes boot up, they will fail to sync as there will be no shards available.

applications/tari_validator_node/proto/dan/common.proto
```diff
@@ -115,6 +112,24 @@ message Arg {
...
+message SubstateState {
+  enum SubstateStateType {
+    DOES_NOT_EXIST = 0;
+    UP = 1;
+    DOWN = 2;
+  }
+  SubstateStateType substate_state_type = 1;
+  PayloadId created_by = 2;
+  bytes data = 3;
+  PayloadId deleted_by = 4;
```
stringhandler, line 132:
```suggestion
  PayloadId deleted_by = 4;
  }
```

stringhandler (approved):

PR #95, opened 2022-10-14
feat: vn re-registers each epoch

brianp:
Description
---
- Removes auto registration on startup
- Adds auto registration when config option `auto_register` is set to true and the epoch changes

Caveats:
- If auto register is set it'll register on the next epoch change regardless of if it's ever been registered before. VN doesn't currently track its own registered state.
- ~~The current scanning for epoch change is done on a 2-minute interval which is likely a bad interval type to use. We should probably use the block height, which is being scanned more frequently anyway.~~ The current scanning happens at the same interval as the epoch manager scans the base layer. Might be good to set up a channel to pub/sub epoch changes.

Motivation and Context
---
To keep the VN registered as the epochs change.

How Has This Been Tested?
---
Manually


brianp:
Oddities I didn't fix in this PR:

There's a delay between registering (confirmed tx) and when the VN web interface shows the vn as registered.

~~There's something re-adding the VN as a registered VN each epoch locally. As in the VN looks like it has re-registered each epoch via the web interface and DB even if it has not sent a new registration. Potentially scanning is being done from the wrong height.~~
This turns out to be a feature. The VN timeout is 100 blocks. After a registration occurs we don't need to re-register again for longer than I would have initially expected.

sdbondi (approved):
utAck

CjS77:
ACK

PR #96, opened 2022-10-17
feat: implement waiting for transaction result in jrpc

sdbondi:
Description
---

- adds `wait_for_result` flag to submit transaction request
- adds HotstuffEvent broadcast channel which notifies caller OnCommit or OnError
- adds optional `--wait-for-result` flag submit-call-function cli command

Motivation and Context
---
Allow created components to be displayed once transaction executes.

Currently, it more often than not will fail.

How Has This Been Tested?
---
Manually


stringhandler (approved):
utACK

PR #100, opened 2022-10-18
feat: use default port

Cifko:
Description
---
Use default per network ports for grpc for base node and wallet.


stringhandler (approved):

stringhandler:
utACK

PR #104, opened 2022-10-18
feat: better sqlite object pledging

stringhandler:
Description
---
* remove ObjectId. Only use shards, and one object per shard
* remove Objects table 
* rename substate_changes table to substates
* add columns to substates for tracking pledges. Of note is the column `is_draft` which signifies the pledging of objects

Motivation and Context
---
Objects must exist in order to be pledged, and the current state should be read from previous data.

How Has This Been Tested?
---
manually


dan_layer/storage_sqlite/migrations/2022-10-03-178235_create_shard_store/up.sql
```diff
@@ -90,21 +90,19 @@ create table leader_proposals (
...
-create table substate_changes (
+create table substates (
     id integer not null primary key AUTOINCREMENT,
+    substate_type text not null,
     shard_id blob not NULL,
-    substate_change text not null,
-    tree_node_hash blob not NULL
+    node_height bigint not NULL,
+    data blob null,
+    created_by_payload_id blob not null,
```
jorgeantonio21, line 100: As a curiosity: can't we store a serialized byte buffer from `SubstateState`, and recover the data from `create_by_payload_id` and `deleted_by_payload_id`, etc by deserializing it ? why adding these as fields of the table of their own ?

dan_layer/core/src/storage/error.rs
```diff
@@ -66,6 +66,8 @@ pub enum StorageError {
     LockError,
     #[error("Error converting to or from json: {0}")]
     SerdeJson(#[from] serde_json::Error),
+    #[error("Error converting substate type:{substate_type}")]
```
jorgeantonio21, line 69: Add a space after `:`

jorgeantonio21:
LGTM ! :)

dan_layer/core/src/models/payload.rs
```diff
@@ -63,11 +63,11 @@ impl Payload for (String, Vec<ShardId>) {
         self.1.clone()
     }
 
-    fn objects_for_shard(&self, shard: ShardId) -> Vec<(ObjectId, SubstateChange, ObjectClaim)> {
+    fn objects_for_shard(&self, shard: ShardId) -> Option<(SubstateChange, ObjectClaim)> {
```
sdbondi, line 66: `object_for_shard`?

dan_layer/storage_sqlite/src/sqlite_shard_store_factory.rs
```diff
@@ -610,28 +651,87 @@ impl ShardStoreTransaction<PublicKey, TariDanPayload> for SqliteShardStoreTransa
...
-            };
-
-            diesel::insert_into(substate_changes)
-                .values(&new_row)
-                .execute(&self.connection)
-                .map_err(|e| Self::Error::QueryError {
-                    reason: format!("Save substate change: {}", e),
-                })?;
+
+            let rows_affected = diesel::update(
```
sdbondi, line 661: Looks like upserts are a Diesel 2.0 feature. A raw query is probably easiet `REPLACE OR INSERT INTO` 
and would greatly shorten this code. Though happy to leave as is

sdbondi (approved):
utAck

PR #122, opened 2022-10-20
feat: add sqlite message logging

stringhandler:
Description
---
Adds a Sqlite log of all incoming and outgoing messages, stored in `data/peer_db/message_log.sqlite`.

Also saves the payload before sending out a NewView message. This is not 100% ideal, since nodes may not receive the payload via mempool sync.

Motivation and Context
---
The main aim here is to help debugging. This code should eventually be removed via config, since it is not implemented efficiently and will hinder performance. Until then, it helps with two things: Was something sent. Was it received. And do they match on both sides.

How Has This Been Tested?
---
Manually

### DIFF
--- a/.license.ignore
+++ b/.license.ignore
@@ -1,5 +1,6 @@
 ./CODEOWNERS
 ./LICENSE
+./comms/tari_comms_logging/src/schema.rs
 ./dan_layer/storage_sqlite/src/schema.rs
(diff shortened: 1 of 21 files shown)


sdbondi (approved):
Nice

sdbondi:
utACK

SWvheerden (commented):
utACK

Issue #50, opened 2022-10-05
Bug: VN should use it's own config file, not the base layer default


stringhandler:
The config file generated when starting the validator node has all the settings from the base layer. These are unnecessary and misleading

stringhandler:
Some work on this has been done in #97, but creating a new directory and then running `cargo run --bin tari_validator_node -- -b .` results in:

```
ExitError { exit_code: ConfigError, details: Some("Invalid value for `validator_node`: unknown field `constitution_management_polling_interval`, expected one of `override_from`, `shard_key_file`, `identity_file`, `tor_i
dentity_file`, `public_address`, `base_node_grpc_address`, `wallet_grpc_address`, `scan_base_layer`, `base_layer_scanning_interval`, `data_dir`, `p2p`, `grpc_address`, `json_rpc_address`, `http_ui_address`, `auto_regist
er`") }
```

Also, the `config.toml` file created in the current directory still has `base_node` sections in it (e.g. `[dibbler.base_node]`)

PR #126, opened 2022-10-21
fix: vn should use its own config file, not base layer config's (see #50)

jorgeantonio21:
Description
---
In #50, it was described that VN config files contain the whole set of base layer configurations. This is not intended, and ideally, VNs will have their own config files.

Motivation and Context
---
We remove this dependency on the base layer configurations, while we keep only the `e_validator_node.toml` configuration set. 

How Has This Been Tested?
---
Manually

### DIFF
--- a/applications/tari_validator_node/config/a_validator_node.toml
+++ b/applications/tari_validator_node/config/a_validator_node.toml
@@ -1,0 +1,42 @@
+
+########################################################################################################################
+#                                                                                                                      #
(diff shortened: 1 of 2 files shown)


applications/tari_validator_node/config/a_validator_node.toml
```diff
@@ -0,0 +1,42 @@
+
+########################################################################################################################
```
stringhandler, line 2: this file can be deleted

PR #128, opened 2022-10-21
fix: fix saving votes to db

stringhandler:
Description
---
The commit statement in sql was layered in an If statement, so was basically never triggered. I've moved it out of the if so `received_votes` should now be populated in sqlite.

I also fixed the incorrect conversion from VoteMessage to its protobuf implementation.


sdbondi (approved):

sdbondi:
ACK

Issue #107, opened 2022-10-19
Base layer scanning should scan `num_confirmations` blocks behind


stringhandler:

jorgeantonio21:
This issue was addressed in #107, let's close it.

PR #129, opened 2022-10-21
fix: base layer scanner should scan fixed number blocks behind (see #107)

jorgeantonio21:
Description
---
In order to mitigate reorgs, we should scan the base layer blockchain not up to the longest chain tip, but instead up to it minus a configurable consensus constant.

Motivation and Context
---
Address issue #107.

How Has This Been Tested?
---
Manually


applications/tari_validator_node/src/consensus_constants.rs
```diff
@@ -0,0 +1,33 @@
...
+//  USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+pub struct ConsensusConstants {
+    pub base_layer_confirmations: u64,
+}
+
+impl ConsensusConstants {
+    pub const fn devnet() -> Self {
+        Self {
+            base_layer_confirmations: 100, // 3 days
```
sdbondi: Think for now, we should scan something like 3 blocks behind to make it easier to get started on a new chain. The first epoch won't come into existence until after this.
```suggestion
            base_layer_confirmations: 3, 
```

jorgeantonio21: It makes sense. I presume that in the future we will want to increase this number, for devnet as well as mainnet.

stringhandler (approved):

sdbondi:
ACK

PR #132, opened 2022-10-21
fix: use detach HSflag for tor

sdbondi:
Description
---
Adds the DETACH hidden service flag for the tor transport on DAN layer

Motivation and Context
---
Currently we're starting and stopping the VNs a lot, each time the hidden service is recreated because it is associated with the control port session. DETACH allows the onion address to remain registered when the vn is offline. This causes a failed assertion when the VN attempts to reregister the onion, however the onion still works and the assertion is ignored.

How Has This Been Tested?
---

### DIFF
--- a/applications/tari_validator_node/src/comms/initializer.rs
+++ b/applications/tari_validator_node/src/comms/initializer.rs
@@ -32,16 +32,29 @@ use tari_comms::{
     pipeline,
     pipeline::SinkService,
     protocol::{messaging::MessagingProtocolExtension, NodeNetworkInfo},
+    tor,
+    transports::{predicate::FalsePredicate, MemoryTransport, SocksConfig, SocksTransport, TcpWithTorTransport},
     types::CommsPublicKey,
     utils::cidr::parse_cidrs,
     CommsBuilder,
+    CommsNode,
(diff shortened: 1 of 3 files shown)


applications/tari_validator_node/src/comms/initializer.rs
```diff
@@ -194,3 +207,89 @@ async fn add_seed_peers(
...
+                .spawn_with_transport(transport)
+                .await?
+        },
+        TransportType::Tor => {
+            let tor_config = transport_config.tor;
+            debug!(target: LOG_TARGET, "Building TOR comms stack ({:?})", tor_config);
+            let mut hidden_service_ctl = initialize_hidden_service(tor_config).await?;
+            // Set the listener address to be the address (usually local) to which tor will forward all traffic
+            let transport = hidden_service_ctl.initialize_transport().await?;
+            debug!(target: LOG_TARGET, "Comms and DHT configured");
```
SWvheerden, line 253: Nit:
Should this not move to after the match to catch all types?

SWvheerden (approved):
utACK

stringhandler (approved):

PR #133, opened 2022-10-21
fix: allow compile without npm

### LINKED ISSUE #121: [web-ui] build failures without npm
Currently we build the web-ui in the build.rs, and this requires npm to be setup in the environment.
This means that a node environment is required to be setup to build the vn which can be annoying when running the VN from a different shell (like intelij run) leading to confusing build failures.

I suggest rather that we have to manually run npm to build the web-ui to be included in the binary.
There are a few options from here:
1. Let that be that, if you dont build the webui, it will display a blank page/help page.
2. add an optional cli flag `--webui-root ./whatever` that overrides the content served on the web-ui port.
3. stringhandler: Only run npm if a certain env is set / nodejs is setup
4. Cifko: ignore the error in the build.rs and emit a [build warning](https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargo-warning)

no 1. is simpler, ~~we'll have to allow CORS in development because you may want to dev on the web-ui using `npm serve`~~
no 2. allows the updated content to be served directly from the VN, making dev easier. But you can serve from a node server anyway so YAGNI

Both are fine for me, but 2 is a feature that needs to be added that may/may not be generally useful.

For production builds, we add the npm build to the process (build script or whatever).

Cifko:
Description
---
Allow compilation without the npm installed. The compiler returns warning on any error returned by the `npm ci` and `npm run build`. Also prints a warning that the ui will not be present. The default web ui returns page with explanation that the node.js should be installed.

How Has This Been Tested?
---
Manually.

Fixes #121

### DIFF
--- a/applications/tari_validator_node/build.rs
+++ b/applications/tari_validator_node/build.rs
@@ -33,15 +33,20 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("cargo:rerun-if-changed=../tari_validator_node_web_ui/public");
     let npm = if cfg!(windows) { "npm.cmd" } else { "npm" };
 
-    Command::new(npm)
+    if let Err(error) = Command::new(npm)
         .arg("ci")
         .current_dir("../tari_validator_node_web_ui")
         .status()
-        .unwrap();
-    Command::new(npm)
+    {
+        println!("cargo:warning='npm ci' error : {:?}", error);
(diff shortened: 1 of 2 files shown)


sdbondi (approved):
Awesome thanks!

stringhandler (approved):

PR #138, opened 2022-10-21
fix: fix only propose once each node

stringhandler:
Fixes a bug where a leader would add a new node and new proposal for every NewView message


sdbondi (approved):

sdbondi:
ACK

PR #140, opened 2022-10-24
fix(base-layer-scanner): takes confirmations into account when detecting chain progress

sdbondi:
Description
---
takes confirmations into account when detecting chain progress

Motivation and Context
---
Base layer scanner would try sync when there was no chain progress due to confirmations offset.

How Has This Been Tested?
---
Manually


SWvheerden (commented):
utACK

stringhandler (approved):

PR #145, opened 2022-10-25
feat: base layer scanner should read consensus constants from base layer (see #115)

### LINKED ISSUE #115: Base layer scanner should read current ConsensusConstants from the base node, according to the version in the blockheader
Specificallly, we need:

- VnConfirmationTime (To be applied to base layer scanning)
- `validator_node_timeout` (To be used to know when to reregister)

jorgeantonio21:
Description
---
Base layer scanner should read consensus constants, via gRPC client.

Motivation and Context
---
Fixes #115.

How Has This Been Tested?
---
Cargo check


applications/tari_validator_node/src/base_layer_scanner.rs
```diff
@@ -284,7 +284,23 @@ impl BaseLayerScanner {
                 })?;
                 match sidechain_feature {
                     SideChainFeature::ValidatorNodeRegistration(reg) => {
-                        self.register_validator_node_registration(current_height, reg).await?;
+                        let validator_node_timeout = self
+                            .base_node_client
+                            .get_consensus_constants(tip.height_of_longest_chain)
```
jorgeantonio21: Is this preferrable to use `end_height`, instead ?

applications/tari_validator_node/src/consensus_constants.rs
```diff
@@ -22,12 +22,14 @@
 
 pub struct ConsensusConstants {
     pub base_layer_confirmations: u64,
+    pub validator_node_timeout: u64,
 }
 
 impl ConsensusConstants {
-    pub const fn devnet() -> Self {
+    pub const fn devnet(validator_node_timeout: u64) -> Self {
```
sdbondi: I wouldnt mix the DAN constants and the base layer constants, as they are independent of each other. This value may change as the base layer progresses anyhow.
I think get the base layer constants as needed in the vn registration code for now

jorgeantonio21: Refactored this

mrnaveira:
LGTM, the only thing I want to comment is that the variable name `validator_node_timeout` can be a bit confusing, as it's not about time but about block confirmations. So something like `validator_node_timeout_blocks` may be better

sdbondi:
Yeah I've never liked that name either. A timeout is not correct, it's more of an expiry - `validator_node_registration_expiry` or `validator_node_registration_expiry_blocks`. Time is typically measured in blocks so happy if that is implicit.

applications/tari_validator_node/src/base_layer_scanner.rs
```diff
@@ -284,7 +285,23 @@ impl BaseLayerScanner {
                 })?;
                 match sidechain_feature {
                     SideChainFeature::ValidatorNodeRegistration(reg) => {
-                        self.register_validator_node_registration(current_height, reg).await?;
+                        let validator_node_timeout = self
```
sdbondi: Here we are acknowledging that we received a validator node registration on chain so can be called unconditionally once the UTXO is received. 

The `validator_node_registration_expiry` is related to 
- when we should reregister (I'd say the logic for that belongs applications/tari_validator_node/src/registration.rs)
- whether other nodes are still registered. I don't think this is implemented so we can leave that out for now in this PR because we need a good strategy to manage other peers. This logic does not belong in the base layer scanner.

applications/tari_validator_node/src/base_layer_scanner.rs
sdbondi: Currently the base layer uses the tip so let's use the tip for now

applications/tari_validator_node/src/base_layer_scanner.rs
jorgeantonio21: Thanks for the explanation. From the comment, I understood that `validator_node_registration_expiry` alone is not enough to execute a re-registration of the current VN. What should be the re-registration criteria for the current VN on the registration status of other VN's ?

applications/tari_validator_node/src/base_layer_scanner.rs
jorgeantonio21: Perfect, thanks for the hint !

applications/tari_validator_node/src/registration.rs
```diff
@@ -140,8 +148,17 @@ async fn start(
             Ok(event) = rx.recv() => {
                 match event {
                     EpochManagerEvent::EpochChanged(current_epoch) => {
-                        if let Ok(Some(next_registration_epoch)) = epoch_manager.next_registration_epoch().await {
-                            if current_epoch == next_registration_epoch {
+                        if let Ok(Some(last_registration_epoch)) = epoch_manager.last_registration_epoch().await {
+                            let mut base_node_client = GrpcBaseNodeClient::new(config.validator_node.base_node_grpc_address.unwrap_or_else(|| {
```
stringhandler, line 152: Made an issue for this to be cleaned up in future: https://github.com/tari-project/tari-dan/issues/168

stringhandler (approved):
Happy. Clean up can happen later

applications/tari_validator_node/src/registration.rs
jorgeantonio21: Great ! Will probably tackle that one, as well :)

PR #146, opened 2022-10-25
feat: send the whole payload with first message

Cifko:
Description
---
Send whole payload with first message, not just the id.

### DIFF
--- a/applications/tari_validator_node/proto/dan/consensus.proto
+++ b/applications/tari_validator_node/proto/dan/consensus.proto
@@ -52,7 +52,7 @@ message HotStuffTreeNode {
   uint64 epoch = 7;
   bytes proposed_by = 8;
   QuorumCertificate justify = 9;
-  optional TariDanPayload payload = 10;
+  TariDanPayload payload = 10;
 }
 
 message ValidatorSignature {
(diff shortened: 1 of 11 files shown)


stringhandler (commented):
Looks like the right track. It seems to include the payload in every node, not just the first one though

Cifko:
> Looks like the right track. It seems to include the payload in every node, not just the first one though

It should be fixed now.

sdbondi (commented):
utACK

applications/tari_validator_node/proto/dan/consensus.proto
```diff
@@ -44,14 +44,15 @@ message QuorumCertificate {
...
-  bytes payload = 2;
+  bytes payload_id = 2;
   uint64 height = 3;
   bytes shard = 4;
   uint64 payload_height = 5;
   repeated ObjectPledge local_pledges = 6;
   uint64 epoch = 7;
   bytes proposed_by = 8;
   QuorumCertificate justify = 9;
+  optional TariDanPayload payload = 10;
```
sdbondi: this is proto2 syntax, in proto3 all fields are optional

sdbondi (approved):

sdbondi:
utACK

stringhandler (approved):
utACK

PR #150, opened 2022-10-25
feat(template_manager): validate binary hash for template downloads

### LINKED ISSUE #106: Validate binary hash for template downloads
Acceptance criteria:
1. After validator node downloads the wasm binary, it should check the hash that was specified in the base layer
2. If the hash is not valid, then it should be marked "invalid" in the database
3. When a node calls that template, it should fail and not execute any code in the binary

### LINKED ISSUE #131: Bug: Template is still tries to run even if it is not downloaded
Acceptance criteria:
1. Template fails to download
2. Submit transaction with the failed template hash
3. The validator node does not attempt to run the template and returns an error saying that the template has DownloadFailed status

mrnaveira:
Description
---
* DRYed up the template binary hash calculation function.
* The template manager now checks that the downloaded hash corresponds the one specified in the registration UTXO, and sets the corresponding status in the template database row.
* When trying to fetch a template from db, the template manager returns an error if the template was marked as "invalid" previously, so instruction invocations will not be possible

Motivation and Context
---
After a validator node downloads the template WASM binary, it should check the hash matches the one specified in the base layer template registration UTXO.

Also, the template should be marked as "invalid" in the database and forbid transactions for that template.

Fixes #106 
Fixes #131 

How Has This Been Tested?
---
Manually registered templates for both scenarios: valid and invalid binary hashes


applications/tari_validator_node/src/p2p/services/template_manager/manager.rs
```diff
@@ -108,6 +108,10 @@ impl TemplateManager {
             .get_template(address)?
             .ok_or(TemplateManagerError::TemplateNotFound { address: *address })?;
 
+        if matches!(template.status, TemplateStatus::Invalid) {
+            return Err(TemplateManagerError::TemplateCodeHashMismatch);
+        }
```
sdbondi: We should only return ok if the template is available for use. Let's create a new error variant for this case.

```suggestion
        if !matches!(template.status, TemplateStatus::Active | TemplateStatus::Deprecated) {
            return Err(TemplateManagerError::TemplateUnavailable);
        }
```

applications/tari_validator_node/src/p2p/services/template_manager/service.rs
```diff
@@ -168,14 +181,22 @@ impl TemplateManagerService {
     async fn handle_add_template(&mut self, template: TemplateRegistration) -> Result<(), TemplateManagerError> {
         let address = template.template_address;
         let url = template.registration.binary_url.to_string();
+        let expected_binary_sha = template.registration.binary_sha.clone();
```
sdbondi: I think we should convert this to a hash, the binary_sha is wrong (my bad :smile: ) should be binary_hash and be a FixedHash type. But for now, we can use it as a hash in the DAN layer:
```suggestion
        let expected_binary_hash = FixedHash::try_from(template.registration.binary_sha.into_vec())?;
```

applications/tari_validator_node/src/p2p/services/template_manager/service.rs
```diff
@@ -142,15 +143,27 @@ impl TemplateManagerService {
...
             Ok(bytes) => {
-                // TODO: Validate binary against hash
                 info!(
                     target: LOG_TARGET,
                     "✅ Template {} downloaded successfully", download.template_address
                 );
+
+                // validation of the downloaded template binary hash
+                let actual_binary_sha = calculate_template_binary_hash(&bytes);
+                let template_status = if actual_binary_sha.to_vec() == download.expected_binary_sha.to_vec() {
```
sdbondi: to_vec allocates on the heap which is unnecessary for the purposes of comparison
```suggestion
                let template_status = if actual_binary_hash.as_slice() == download.expected_binary_hash.as_slice() {
```

sdbondi (commented):
LGTM

Some minor comments

stringhandler (approved):

PR #153, opened 2022-10-26
feat: check base node grpc connection on startup

Cifko:
Description
---
Check that the grpc connection to the base node is alive at the startup.

How Has This Been Tested?
---
Manually with the both status of BN.


stringhandler (approved):

sdbondi:
ACK

PR #160, opened 2022-10-28
feat: execute payload to determine which vote to cast

sdbondi:
Description
---

Motivation and Context
---

How Has This Been Tested?
---

### DIFF
--- a/dan_layer/core/src/workers/hotstuff_waiter.rs
+++ b/dan_layer/core/src/workers/hotstuff_waiter.rs
@@ -23,9 +23,9 @@
 use std::collections::HashMap;
 
 use futures::future::join_all;
-use log::{debug, error, info};
+use log::{debug, error, info, warn};
 use tari_dan_common_types::{Epoch, PayloadId, ShardId, SubstateState};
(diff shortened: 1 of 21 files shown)


dan_layer/core/src/workers/hotstuff_waiter.rs
```diff
@@ -388,63 +407,90 @@ where
...
+                        target: LOG_TARGET,
+                        "payload changeset: {}",
+                        serde_json::to_string(&changes).unwrap()
+                    );
+                    self.on_commit(node, &changes, &mut tx)?;
+                },
+                None => {
+                    warn!(
+                        target: LOG_TARGET,
+                        "No finalize result for payload {}. Not enough votes?",
```
stringhandler: This is actually an error. It should never get to this point without a finalize result, so there is another cause. (I'm not sure what it is)
```suggestion
                        "No finalize result for payload {}. ",
```

sdbondi: Thanks, wasn't sure

stringhandler (approved):

PR #161, opened 2022-10-28
feat: save payload on first message

Cifko:
Description
---
Save payload on first received message.


dan_layer/core/src/workers/hotstuff_waiter.rs
```diff
@@ -480,6 +480,8 @@ where
 
         let shard = node.shard();
         let payload = if let Some(node_payload) = node.payload() {
+            let mut tx = self.shard_store.create_tx()?;
+            tx.set_payload(node_payload.clone()).map_err(|e| e.into())?;
```
sdbondi, line 484: `tx.commit()?`

stringhandler (approved):

Issue #154, opened 2022-10-26
On Epoch change, I must sync state from the other nodes


stringhandler:

sdbondi:
Closed in #166

PR #166, opened 2022-10-31
feat: state sync on epoch change (see #154)

jorgeantonio21:
Description
---
At each epoch change, current validator node must synchronize state from VNs in the same committee, with missing state.

Motivation and Context
---
Address issue #154.

How Has This Been Tested?
---


applications/tari_validator_node/src/p2p/services/epoch_manager/base_layer_epoch_manager.rs
```diff
@@ -94,9 +106,42 @@ impl BaseLayerEpochManager {
 
         // If the committee size is bigger than vns.len() then this function is broken.
         let mut base_node_client = self.base_node_client.clone();
-        let mut vns = base_node_client.get_validator_nodes(epoch.0 * 10).await?;
+        let mut vns = base_node_client.get_validator_nodes(epoch.to_height()).await?;
         vns.sort_by(|a, b| a.shard_key.partial_cmp(&b.shard_key).unwrap());
 
+        let current_shard_key = vns
```
stringhandler: Logic looks like it's on the right track, but design-wise, I would prefer a separate service, `SyncManagerService`, that listens for a `broadcast` event on a `channel` and then runs this logic. Finally, the epoch manager can publish that event here

jorgeantonio21: Perfect, I really like it!

applications/tari_validator_node/src/p2p/rpc/service_impl.rs
```diff
@@ -144,68 +142,61 @@ where TPeerProvider: PeerProvider + Clone + Send + Sync + 'static
         Ok(Streaming::new(rx))
     }
 
-    async fn get_vn_state_inventory(
+    async fn vn_state_sync(
```
jorgeantonio21, line 145: See #180.

jorgeantonio21:
It also addresses issue #180.

applications/tari_validator_node/src/lib.rs
```diff
@@ -129,6 +130,11 @@ pub async fn run_validator_node_with_cli(config: &ApplicationConfig, cli: &Cli)
     );
 
     let (base_node_client, wallet_client) = create_base_layer_clients(config).await?;
+    let connectivity = initialize(node_identity.clone(), config, shutdown.to_signal())
```
sdbondi: You're initializing comms twice. Rather create the `TariCommsValidatorNodeClientFactory` inside spawn_services. You can use `comms.connectivity()` to get the handle there.

jorgeantonio21: Nice ! Definitely, refactored the code accordingly :)

applications/tari_validator_node/src/p2p/rpc/service_impl.rs
```diff
@@ -144,68 +142,61 @@ where TPeerProvider: PeerProvider + Clone + Send + Sync + 'static
...
+        if request_inventory.is_empty() {
             return Err(RpcStatus::bad_request(
                 "Invalid gRPC request: request should contain at least one shard id available",
             ));
         }
 
-        let shard_store = self.shard_state_store.clone();
+        let missing_shard_ids = current_inventory
+            .into_iter()
+            .filter(|sid| !request_inventory.contains(sid))
```
mrnaveira: LGTM. The only thing is that here calculating the `missing_shard_ids` has quadratic time complexity (`contains` on a `Vec` does a linear search). If the number of shards to be requested is big (which I think it may be) it would be more optimal for `request_inventory` to be a `HashSet` instead. That or even better implementing getting the `missing_shard_ids`  on the database layer in a new method, so it leverages db indexes

sdbondi (approved):

sdbondi:
utAck

Not sure if it works but we can fix up any issues as we find them

mrnaveira (approved):

applications/tari_validator_node/src/p2p/rpc/service_impl.rs
jorgeantonio21: I think the second approach is the more sensible one, at this point.

jorgeantonio21: Refactored it

sdbondi (approved):

sdbondi:
On second thoughts, this needs to happen once scanning is complete - currently, it will try sync for each epoch while the VN is processing the blockchain. 

I suggest we change the epoch manager to only call update_epoch on the final block height that the scanner scans

applications/tari_validator_node/src/p2p/services/epoch_manager/base_layer_epoch_manager.rs
```diff
@@ -93,10 +108,31 @@ impl BaseLayerEpochManager {
...
+        let peer_sync_service_manager = PeerSyncManagerService::new(
+            self.validator_node_config.clone(),
+            self.validator_node_client_factory.clone(),
+        );
+
+        // synchronize state with committee validator nodes
+        peer_sync_service_manager
+            .sync_peers_state(committee_vns, start_shard_id, end_shard_id, vn_shard_key)
+            .await?;
+
```
sdbondi, line 154: Move this code to after the epoch is processed (line 151)

```rust
// TODO: sync should be triggered by EpochChanged event
```

Add a scope (`{ ... }`) from line 138 to line 146 so that the transaction and db are cleaned up

sdbondi (changes requested):
Two changes are needed so that the VN does not break

1. base layer sanner: call update_epoch after all blocks have synced ` self.epoch_manager.update_epoch(tip.height_of_longest_chain).await?;` 
2. Move the sync code to after the current epoch has been committed (see comment)

applications/tari_validator_node/src/p2p/services/epoch_manager/base_layer_epoch_manager.rs
jorgeantonio21: Makes sense, thanks for the remarks and explanations !

sdbondi (approved):

sdbondi:
ACK

Tested and the node starts and syncs the base layer - thanks :rocket:

PR #167, opened 2022-10-31
feat(tests): create a cucumber test to call Counter::new() function on a VNC

### LINKED ISSUE #147: Create a cucumber test to call Counter::new() function on a VNC
There should be at least 2 VNs in the committee

mrnaveira:
Description
---
* Allow for multiple base nodes, wallets, miners and validator nodes
* New test that tries to register two VNs, registers the "counter" template and calls the "new" function on it. 

Motivation and Context
---
We want a full test where multiple validator nodes are registered, a template is registered and a call to the template is performed.

Fixes #147 

How Has This Been Tested?
---
**Currently the integration test is not fully functional, as the base node does not return the list of registered VNs**, so the template call fails with an epoch error. But the template is registered and scanned by the VNs.


sdbondi:
ACK

sdbondi (approved):

PR #171, opened 2022-11-01
fix: vote reject when wasm panics

stringhandler:
Description
---
When the WASM panics or execution fails, vote as reject instead of exiting

Motivation and Context
---
Some old WASMs no longer work after recent commits. Invoking them causes the hotstuff waiter to exit instead of voting as a reject

How Has This Been Tested?
---
Manually


applications/tari_validator_node_web_ui/public/index.html
```diff
@@ -2,7 +2,8 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
+    <link rel="icon" href="data:;base64,=">
```
stringhandler, line 5: Removes the log in the console

applications/tari_validator_node/src/grpc/services/base_node_client.rs
```diff
@@ -117,7 +116,7 @@ impl BaseNodeClient for GrpcBaseNodeClient {
                     });
                 },
                 Ok(None) => {
-                    info!(target: LOG_TARGET, "No new validator nodes for this epoch");
+                    trace!(target: LOG_TARGET, "No new validator nodes for this epoch");
```
stringhandler, line 119: removes an extra log in the console

sdbondi (approved):

sdbondi:
utACK

PR #172, opened 2022-11-02
feat: return QC to client for submit-transaction calls

sdbondi:
Description
---
Return QC back to client when submitting the transaction

Motivation and Context
---
Allows client to validate finalization

How Has This Been Tested?
---
Manually (cli)


stringhandler (approved):
Nice addition

PR #177, opened 2022-11-02
fix(integration_test): cucumber-rs improvements

mrnaveira:
Description
---
* Validator nodes are now able to register (depends on [tari#4879](https://github.com/tari-project/tari/pull/4879))
* Adding new cucumber steps to facilitate manual testing: printing node information and pausing execution

Motivation and Context
---
Previously, the cucumber integration test didn't allow VN registration, due to an invalid consensus constant in the `localnet` network.

Also, manual inspection (gRPC, json-rpc, database files, etc.) should be easier to do, so new cucumber steps are added to enable it.

Depends on [tari#4879](https://github.com/tari-project/tari/pull/4879)

How Has This Been Tested?
---
The integration tests now allows registration of VNs and asserts that they are indeed registered. **The template function call does still fail, so it's commented out.**


sdbondi:
tari#4879 merged

sdbondi (approved):
LGTM

PR #179, opened 2022-11-03
feat: sign/check votemessage

Cifko:
Description
---
Sign the votemessage, and check the votemessage for valid signature.

How Has This Been Tested?
---
cargo test --package tari_dan_tests --lib --all-features -- --nocapture


Cifko:
Couple questions:
- I'm not quite sure why the `ValidatorSignature` was created this way. So I didn't want to change it for now.
- What should be the behavior in case the VoteMessage is not valid?

PR #184, opened 2022-11-04
fix: remove finalized transactions from mempool (see issue #111)

### LINKED ISSUE #111: When submitting a transaction, it is not removed from the mempool when processed
To reproduce:
1. Register a template on the base layer.

// use tari_template_macros::template;
//
// #[template]
// mod counter {
//     pub struct Counter {
//         value: u32,
//     }
//
//     impl Counter {
//         pub fn new() -> Self {
//             Self { value: 0 }
//         }
//
//         pub fn value(&self) -> u32 {
//             self.value
//         }
//
//         pub fn increase(&mut self) {
//             self.value += 1;
//         }
//     }
// }

2. Submit an instruction CallFunction <template_address>, `new`

Expected: transaction is processed
Expected: transaction is removed from mempool
Actual: transaction is still in mempool.

jorgeantonio21:
Description
---
Removed Finalized transactions  from the mempool.

Motivation and Context
---
Fixes #111.

How Has This Been Tested?
---


applications/tari_validator_node/src/json_rpc/handlers.rs
```diff
@@ -170,9 +170,11 @@ impl JsonRpcHandlers {
         let hash = *mempool_tx.hash();
 
         let subscription = self.hotstuff_events.subscribe();
+        // create a mempool request out of mempool_tx
+        let mempool_request = MempoolRequest::SubmitTransaction(Box::new(mempool_tx));
```
jorgeantonio21: In this case, we want to submit a transaction, I think it is safe to assume we have to store it on the mempool, and not have to consider the case in which the transaction is possibly finalized.

sdbondi (approved):
LGTM

I pushed some changes that revert the proto changes, as the mempool request is an internal service request and not to be exposed as a p2p message.

sdbondi:
ACK

PR #185, opened 2022-11-06
fix: remove generic from bucket

stringhandler:
WIP - Attempt to make a generic Account.


dan_layer/template_abi/src/lib.rs
```diff
@@ -27,7 +27,9 @@
 //! This library provides types and encoding that allow low-level communication between the Tari WASM runtime and the
 //! WASM modules.
 
+#[cfg(target_arch = "wasm32")]
 mod abi;
+#[cfg(target_arch = "wasm32")]
```
sdbondi: I think we just need to stub out the call_engine function for non-wasm targets and then we won't need to have the cfg(wasms) on the rest of the code. But can clean this up fairly quickly once this PR is merged

sdbondi (commented):
Awesome, will merge once tests and clippes are fixed

sdbondi (approved):

sdbondi:
ACK

sdbondi:
ACK

?

PR #187, opened 2022-11-07
fix: add committee size constant, to remove duplicate code (see #174)

jorgeantonio21:
Description
---
Address #174. Even though we don't add this constant to `ConsensusConstants`, as it is not used as inner type of `EpochManager`, we remove duplicate instances of the value `committee_size`.

Motivation and Context
---

How Has This Been Tested?
---

### DIFF
--- a/applications/tari_validator_node/src/p2p/services/epoch_manager/base_layer_epoch_manager.rs
+++ b/applications/tari_validator_node/src/p2p/services/epoch_manager/base_layer_epoch_manager.rs
@@ -26,6 +26,7 @@ use tari_comms::{types::CommsPublicKey, NodeIdentity};
 use tari_crypto::tari_utilities::ByteArray;
 use tari_dan_common_types::{Epoch, ShardId};
 use tari_dan_core::{
+    consensus_constants::ConsensusConstants,
     models::{Committee, ValidatorNode},
     services::{
         epoch_manager::{EpochManagerError, ShardCommitteeAllocation},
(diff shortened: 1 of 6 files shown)


applications/tari_validator_node/src/p2p/services/epoch_manager/base_layer_epoch_manager.rs
```diff
@@ -247,7 +247,7 @@ impl BaseLayerEpochManager {
         // retrieve the validator nodes for this epoch from database
         let vns = self.get_validator_nodes_per_epoch(epoch)?;
 
-        let half_committee_size = 4; // total committee = 7
+        let half_committee_size = COMMITTEE_SIZE.div(2) + 1; // middle point of COMMITTEE_SIZE = 7
```
sdbondi: If COMMITTEE_SIZE is even, it will be half + 1 .

```rust

        let half_committee_size = {
             // Ceiling div
             let v = COMMITTEE_SIZE / 2;
            let rem = COMMITTEE_SIZE % 2;
            if rem > 0 {
                v + 1
            } else {
                v
            } 
};
```

sdbondi (commented):
I would say it should go into ConsensusConstants

applications/tari_validator_node/src/p2p/services/epoch_manager/base_layer_epoch_manager.rs
jorgeantonio21: You are right @sdbondi. For some reason, I assumed we wanted to work always with odd numbers, as the range `[0, num]` always have a midpoint. My bad, and thanks for the remark !

sdbondi (approved):
Nice

sdbondi:
utACK

PR #192, opened 2022-11-08
feat: add sorting to recent TX in webgui

Cifko:
Description
---
Add sorting to recent transaction in web gui. Added indicator (I hope this indicator works in all browser normally).


sdbondi (approved):
LGTM - we can add a local timestamp at some point

sdbondi:
utACK

PR #199, opened 2022-11-14
fix: reduces number of executes per transaction

sdbondi:
Description
---
- executes payload at payload_height 0
- extracts changes from QC for parent nodes
- adds more code comments and reorders functions in step order
- get_leaf_node returns `LeafNode` type

Motivation and Context
---
Nodes should execute at the first round of consensus (node 0) not every round.

We still execute when deciding on the vote and again before committing. I think we need to persist the pending changeset (committed to in the QC) and retrieve it later when we commit (Pending -> Commit). 

How Has This Been Tested?
---
Manually


stringhandler (approved):
utACK

Looks like the failing test is related

sdbondi:
Thanks - tests updated

PR #206, opened 2022-11-16
fix: vns should not store transactions for shards outside committee range (see issue #193)

### LINKED ISSUE #193: VNs should not store transaction if not for this shard

jorgeantonio21:
Description
---
Validator nodes should only store transactions that contain substates to be processed within its shard range. Moreover, the validator node should only broadcast the transaction for other validator nodes in the same committee. We refactor the code to address these two issues.

Motivation and Context
---
Tackle issue #193 and broadcast transactions only for VNs in the same committee.

Fixes #193 

How Has This Been Tested?
---

### DIFF
--- a/applications/tari_validator_node/src/p2p/services/mempool/service.rs
+++ b/applications/tari_validator_node/src/p2p/services/mempool/service.rs
@@ -129,18 +129,17 @@ impl MempoolService {
                     "🎱 Transaction {} already in mempool",
                     transaction.hash()
                 );
-                // TODO: return an error
-                return Err(MempoolError::TransactionNotProcessedByCurrentVN);
+                return Err(MempoolError::TransactionAlreadyExists);
             }
-            // TODO: add here
+
             let current_node_pubkey = self.node_identity.public_key().clone();
             let mut should_process_txn = false;
+
             for sid in &shards {
                 if self
                     .epoch_manager
                     .is_validator_in_committee_for_current_epoch(*sid, current_node_pubkey.clone())
-                    .await
-                    .unwrap()
+                    .await?
                 {
                     should_process_txn = true;
                     break;
@@ -169,16 +168,23 @@ impl MempoolService {
         Ok(())
     }
 
-    pub async fn propagate_transaction(&mut self, transaction: &Transaction, shards: &[ShardId]) {
+    pub async fn propagate_transaction(
+        &mut self,
+        transaction: &Transaction,
+        shards: &[ShardId],
+    ) -> Result<(), MempoolError> {
         // TODO: unwrap !
-        let epoch = self.epoch_manager.current_epoch().await.unwrap();
-        let committees = self.epoch_manager.get_committees(epoch, shards).await.unwrap();
+        let epoch = self.epoch_manager.current_epoch().await?;
+        let committees = self.epoch_manager.get_committees(epoch, shards).await?;
 
         let msg = DanMessage::NewTransaction(transaction.clone());
 
         // propagate over the involved shard ids
-        let committees_set =
-            HashSet::<RistrettoPublicKey>::from_iter(committees.into_iter().flat_map(|x| x.committee.unwrap().members));
+        let committees_set = HashSet::<RistrettoPublicKey>::from_iter(committees.into_iter().flat_map(|x| {
+            x.committee
+                .expect("mempool_service::propagate_transaction::shard committee should be available")
+                .members
+        }));
         let committees = committees_set.into_iter().collect::<Vec<_>>();
 
         if let Err(err) = self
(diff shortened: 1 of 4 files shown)


applications/tari_validator_node/src/p2p/services/mempool/service.rs
```diff
@@ -53,43 +60,51 @@ impl MempoolService {
...
             mempool_requests,
             outbound,
             tx_valid_transactions,
+            epoch_manager,
+            node_identity,
         }
     }
 
-    pub async fn run(mut self) {
+    pub async fn run(mut self) -> Result<(), MempoolError> {
```
sdbondi: Any error will "crash" the mempool service. This run loop should remain active until the new_transactions  and the mempool_requests channels are closed

applications/tari_validator_node/src/p2p/services/mempool/service.rs
```diff
@@ -53,43 +60,51 @@ impl MempoolService {
...
         }
+
+        Ok(())
     }
 
-    async fn handle_request(&mut self, request: MempoolRequest) {
+    async fn handle_request(&mut self, request: MempoolRequest) -> Result<(), MempoolError> {
         match request {
-            MempoolRequest::SubmitTransaction(transaction) => self.handle_new_transaction(*transaction).await,
+            MempoolRequest::SubmitTransaction(transaction) => self.handle_new_transaction(*transaction).await?,
```
sdbondi: All errors from these functions should be logged and not propagated to the service run loop

see `hotstuff_service.rs`

applications/tari_validator_node/src/p2p/services/mempool/service.rs
```diff
@@ -101,35 +116,99 @@ impl MempoolService {
...
+                    should_process_txn = true;
+                    break;
+                }
+            }
+
+            let mut access = self.transactions.lock().unwrap();
+
+            if should_process_txn {
+                access.push((transaction.clone(), None));
+            } else {
```
sdbondi, line 155: We still want to broadcast the transaction even if it's not for this node

sdbondi (commented):
LGTM - just need fixes for:
- mempool service "crash on error", and
- propagate the transaction even if it is not processed by this node

applications/tari_validator_node/src/p2p/services/mempool/service.rs
jorgeantonio21: I see, will refactor shortly !

applications/tari_validator_node/src/p2p/services/mempool/service.rs
jorgeantonio21: Cool !

applications/tari_validator_node/src/p2p/services/mempool/service.rs
jorgeantonio21: @sdbondi, only for the VNs committee, is that right ?

jorgeantonio21: Currently, in L164

```rs
 match self.propagate_transaction(&transaction, &shards).await {
            Ok(()) => (),
            Err(e) => error!(
                target: LOG_TARGET,
                "Unable to propagate transaction among peers: {}",
                e.to_string()
            ),
        }
```

we broadcast the transaction to peer VNs in the same committee as the current VN.

sdbondi: Yup, that's right. 

We have three steps
1. validate and store the transaction if we're involved
2. propagate/broadcast the valid transaction to the shards involved (whether we are involved or not)
3. Send the transactions that need to be processed to the hotstuff worker (line 173)

For number 3, I think filter out to only the shards this node is involved with so that it only sends those to the worker (there could be 0 shards involved  which is still correct)

applications/tari_validator_node/src/p2p/services/mempool/service.rs
```diff
@@ -132,6 +180,49 @@ impl MempoolService {
...
+                .committee
+                .expect("mempool_service::propagate_transaction::shard committee should be available")
+                .members;
+            for x in xs {
+                if !v.contains(&x) {
+                    v.push(x);
+                }
+            }
+            v
+        });
```
sdbondi, line 213: nit(perf): 
```suggestion
  let unique_members = committees
            .into_iter()
            .filter_map(|s| s.committee)
            .map(|s| s.members)
            .flatten()
            .collect::<HashSet<_>>();
        let committees = unique_members.into_iter().collect::<Vec<_>>();
```

sdbondi (approved):

PR #208, opened 2022-11-17
feat: add ability to debug local WASM templates

stringhandler:
* add config for adding a debug replacement WASM for a template to allow hotswapping for development
* add a provisional nonfungible builder
* add ChainTime stub for contracts to use to measure time passed
* add more detail to template macros that help show why it could not generate
* add WorkspaceManager with list_buckets action

Motivation and Context
---
Sorry for adding in a few features in one. The main feature here is to allow debugging of wasms without having to mine it into the base layer.

How Has This Been Tested?
---
local debugging


sdbondi (approved):
Awesome LGTM

PR #209, opened 2022-11-17
fix: minimal changes to get vn reg working with rfc-0313

sdbondi:
Description
---
- `get_consensus_constants` grpc updates
- update vn reg signature
- update cargo.lock
- move OnEpochChange handler code into function
- fix bug where if the validator node is not registered it would return an error 
- fix epoch not found bug
- fix  genesis newview QC is always rejected 

Motivation and Context
---
[RFC-0313](https://rfc.tari.com/RFC-0313_VNRegistration.html)

Ref https://github.com/tari-project/tari/pull/4928

How Has This Been Tested?
---
Manually - registering 3 validators from clean igor chain


brianp (approved):
utAck

PR #215, opened 2022-11-22
feat: get transaction result via payload id

mrnaveira:
Description
---
* Added a new field in the Sqlite table `payloads` to store the result of the transaction
* After the payload is processed, the result of the transaction is stored in the database
* New JSON-RPC method to get a transaction result via the `playload_id`
* New CLI command to get a transaction result

Motivation and Context
---
If calling a transaction via the CLI times out, the user should be able to get the result an execution by calling `get_transaction_result(<payload_id>)`

There should be a new command in the validator node CLI: `transactions get <payload_id>`

Solves https://github.com/tari-project/tari-dan/issues/212

How Has This Been Tested?
---
Manual testing


dan_layer/storage_sqlite/src/sqlite_shard_store_factory.rs
```diff
@@ -1226,6 +1241,27 @@ impl ShardStoreTransaction<PublicKey, TariDanPayload> for SqliteShardStoreTransa
...
     }
 
+    fn update_payload_result(
+        &self,
+        requested_payload_id: &PayloadId,
+        result: FinalizeResult,
+    ) -> Result<(), StorageError> {
+        use crate::schema::payloads;
+
+        let result_bytes = bincode::serialize(&result).map_err(|_| StorageError::EncodingError)?;
```
sdbondi: Could we use JSON here to make debugging easier?

sdbondi (approved):
Nice - going to fix the conflicts and merge

PR #228, opened 2022-11-28
refactor: move QC etc types into common crate

sdbondi:
Description
---
Moves a number of types into tari_dan_common_types

Motivation and Context
---
The validator node client lib requires QC and therefore everything in core

How Has This Been Tested?
---
Code compiles and no lints


stringhandler (approved):
Nice

PR #242, opened 2022-12-06
feat: new dry run transaction processor

mrnaveira:
Description
---
**This PR is a work in progress, please do not merge yet**
* Created a new `DryRunTransactionProcessor` service to isolate dry run execution
    * Reads the shard store database for local pledge retrieval
    * Queries other VNS to retrieve pledges that are not local
    * Processes the transaction and returns the result as if it was processed for real
* Modified the shard database service:
    * The old method `get_substate_states` is renamed to `get_substate_states_by_range`
    * Made the shard range in the SQL query totally inclusive, to be able to retrieve individual shard substates 
    * Added a new method to retrieve all substates from a list of specific shard ids, to avoid querying one by one or specify a big range

Motivation and Context
---
We need a _dry run_ functionality added to the `sumbit_transaction` JSON-RPC method and corresponding CLI. The goal is to be able to simulate and get the results of executing a transaction (using the latest substate for each involved shard) without side effects (i.e. database writes) or running consensus.

Example of a CLI call (it works for both `call-function` and `call-method`):
```
$ tari_validator_node_cli transactions submit --num-outputs 1 -w --is-dry-run call-function c3a...dbf new
```

Solves https://github.com/tari-project/tari-dan/issues/219

How Has This Been Tested?
---
Manually. Transactions that involve remote shards are not yet fully tested


stringhandler:
looking good!

sdbondi:
Nice! 

We may run into issues with the pledges not being synced currently (due to a panic TODO: include pledges in state sync) but @stringhandler is currently working on improvements to pledges so we can fix that after.

sdbondi (approved):
Nice

Fixed up merge conflicts and fixed the node trying to fetch states from itself (Peer not found error)

Tested and works! Think this can go in

stringhandler (approved):

PR #246, opened 2022-12-07
feat: better substate db structure

stringhandler:
Separate substates and pledges into 2 tables .

WIP: Initial create of contract works, but currently unable to DOWN and UP a substate

Also added timestamps to every SQL table. This will help with identifying the time taken at each step


applications/tari_validator_node/proto/dan/rpc.proto
```diff
@@ -92,10 +92,15 @@ message VNStateSyncRequest {
...
-  tari.dan.consensus.SubstateState substate_state = 2;
-  uint64 node_height = 3;
-  bytes tree_node_hash = 4;
-  bytes payload_id = 5;
-  tari.dan.consensus.QuorumCertificate certificate = 6;
+  bytes shard_id = 1;
+  uint32 version = 2;
+  bytes substate =3;
+  uint64 created_height = 4;
+  optional uint64 destroyed_height = 5;
```
sdbondi:
```suggestion
  uint64 destroyed_height = 5;
```

stringhandler: yeah saw this as well, but 0 is technically a valid height. I suppose it would never be destroyed at genesis

sdbondi: sorry should have been more specific, optional isn't a keyword in proto3

applications/tari_validator_node/src/p2p/services/epoch_manager/state_sync.rs
```diff
@@ -149,7 +109,7 @@ impl<TShardStore: ShardStore> PeerSyncManagerService<TShardStore> {
                 }
 
                 // increase node inventory
-                inventory.push(sync_vn_shard.into());
+                // inventory.push(sync_vn_shard.into());
```
sdbondi, line 112: Should this be commented out?

sdbondi (approved):
Nice LGTM - utACK

sdbondi (approved):

PR #247, opened 2022-12-07
feat(templates): improve template panic messages, return err in exec result and engine calls

sdbondi:
Description
---
- improves template errors (template_lib and template_macros)
- improves argument errors, detailing the offending function and arg 
- invoke result returns error responses to the caller (incomplete)
- WASM panic messages are propagated through to the FinalizeResult
- decode_exact checks that the encoding as completely used to mitigate invalid data being misinterpreted and decoding not failing

Motivation and Context
---
If the transaction contains invalid argument data or the template panics, it can be extremely difficult to figure out what has gone wrong. This PR aims to substantially improve that. 

How Has This Been Tested?
---
Manaully, adding panics to templates, new engine integration test


dan_layer/engine/tests/templates/errors/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[workspace]
+[package]
+name = "account"
```
stringhandler, line 3: not a trainsmash

stringhandler (approved):

PR #253, opened 2022-12-12
fix: shard ids are generated by the object address and version and change on each transaction

sdbondi:
Description
---

- derive shard address from `H(H(tx_hash||n) || version)`
- remove input_refs from the transaction builder
- Newtypes for component, resource and vault addresses

Motivation and Context
---
Changes the mapping from shard id to component, resource, vault addresses so that they are no longer always the same. 
Inputs are always DOWNed and outputs are UPd. 

How Has This Been Tested?
---
Manually


stringhandler (commented):
Looks good so far. Interested to see the ComponentHeader vs body.

Also must remember to put the validation of pledges back

sdbondi:
Changes here included in #267

PR #254, opened 2022-12-12
feat: create account template builtin

mrnaveira:
Description
---
* Created a new `template_builtin` crate:
    * On compile time detects changes to the internal `templates` directory and compiles the wasm if necessary
    * The wasm is copied on build time to the root of the corresponding template, so it's added to version control. The wasm will be refreshed automatically on compile time if the template source code is modified.
    * Includes the wasm in the binary (via `include_bytes!` macro)
    * Exports a function to get the account wasm.
    * New builtin templates can be easily added just by adding them to the `templates` folder and exporting them on `lib.rs`. An even more dynamic approach could be implemented in the future, where Rust code is dynamically generated for each template found.
* Updated the `TemplateManager`:
    * On startup it loads the account builtin template code. The builtin mechanism is flexible so more builtin templates can be easily added in the future.
    * Updated `fetch_template` and `fetch_template_metadata` to include the builtin templates.

Motivation and Context
---
We want a way to add the Account template builtin in the project:
* Template address should be hardcoded
* Template code should be kept in separate crate and the `.wasm` file committed to the repo
* VN should include the wasm bytes in the binary

Solves https://github.com/tari-project/tari-dan/issues/244

How Has This Been Tested?
---
Manually ran a validator node. The account template is preloaded and we can call the `new` function on it via CLI.


sdbondi (approved):

sdbondi:
utACK

sdbondi (approved):

PR #261, opened 2022-12-14
fix: cucumber errors for templates and transactions

mrnaveira:
Description
---
* Added a time sleep before checking if a template is registered, to allow VNs to catch up
* Added the `num_outputs` param when calling a template function

Motivation and Context
---
Some steps in cucumber scenarios were commented out because they failed:
* VN2 did not receive the template
* The `submit_transaction` request failed

How Has This Been Tested?
---
The cucumber scenario steps now work

### DIFF
--- a/applications/tari_validator_node/tests/cucumber.rs
+++ b/applications/tari_validator_node/tests/cucumber.rs
@@ -39,7 +39,7 @@ use tari_validator_node::GrpcBaseNodeClient;
 use tari_validator_node_client::types::{GetIdentityResponse, GetTemplateRequest, TemplateRegistrationResponse};
 use utils::{
     miner::{mine_blocks, register_miner_process},
-    template::send_template_transaction,
+    template::send_call_function_transaction,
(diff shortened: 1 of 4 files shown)


applications/tari_validator_node/tests/cucumber.rs
```diff
@@ -155,6 +155,9 @@ async fn assert_vn_is_registered(world: &mut TariWorld, vn_name: String) {
 
 #[then(expr = "the template \"{word}\" is listed as registered by the validator node {word}")]
 async fn assert_template_is_registered(world: &mut TariWorld, template_name: String, vn_name: String) {
+    // give it some time for the template tx to be picked up by the VNs
+    tokio::time::sleep(Duration::from_secs(4)).await;
```
sdbondi, line 159: Should probably implement a way to poll JRPC (or any future) until some condition == true or we timeout

sdbondi (approved):

sdbondi:
utACK

PR #263, opened 2022-12-14
feat: cucumber tests construct components

mrnaveira:
Description
---
* Added the proper `public_address` config parameter to the VN, so now multiple VNs are able to sync. **The test is still flaky when using multiple VNs**, so I commented out the second VN from the cucumber scenario.
* Refactored the call function cucumber step as a component initialization, and the component id is stored in the test state for future use (i.e. method calling)

How Has This Been Tested?
---
The existing test scenarios pass


sdbondi (approved):
Good progress LGTM

PR #267, opened 2022-12-15
fix: fixes for consensus,mempool and more

sdbondi:
Description
---

Motivation and Context
---

How Has This Been Tested?
---


dan_layer/core/src/workers/hotstuff_waiter.rs
```diff
@@ -205,21 +217,21 @@ where
 
             let high_qc = tx.get_high_qc_for(shard).optional()?.unwrap_or_else(|| {
                 // TODO: sign genesis
```
stringhandler: Just a note, I don't think the genesis needs to be signed

dan_layer/core/src/workers/hotstuff_waiter.rs
```diff
@@ -445,97 +473,115 @@ where
...
+                    },
+                    None => {
+                        info!(
+                            target: LOG_TARGET,
+                            "🔥 Not enough proposals to vote on the message. num involved_shards: {}",
+                            involved_shards.len()
+                        );
+
+                        tx.commit()?;
+                        self.update_nodes(node)?;
```
stringhandler: More a note for myself: This might be wrong (although it's probably not different from the previous implementation),  but I need to confirm whether this could cause a node to be executed before receiving all the votes for it

dan_layer/core/src/workers/hotstuff_waiter.rs
```diff
@@ -643,25 +689,26 @@ where
...
-            let num_new_views = self.count_new_views_for(shard)?;
-            if num_new_views >= min_required_new_views {
-                self.on_propose(shard, payload_id).await?;
-            } else {
-                info!(
-                    target: LOG_TARGET,
-                    "🔥 Waiting for more NEWVIEW messages ({}/{})", num_new_views, min_required_new_views
-                );
-            }
+            // TODO: Fix count_new_views_for - I think it needs to count new views for the shard and payload id
```
stringhandler: correct, current code would not work with multiple payloads for the same shard

dan_layer/core/src/workers/hotstuff_waiter.rs
```diff
@@ -852,27 +929,63 @@ where
...
-                tx.save_substate_changes(changes, node)?;
-                tx.complete_pledges(node.shard(), node.payload_id(), node.hash())?;
-            }
-            tx.set_last_executed_height(shard, node.height())?;
+    fn on_reject(
+        &self,
+        node: &HotStuffTreeNode<TAddr, TPayload>,
+        tx: &mut TShardStore::Transaction<'_>,
+    ) -> Result<(), HotStuffError> {
+        if *node.parent() != TreeNodeHash::zero() {
```
stringhandler: You actually need to check the results of each QC here and commit or reject depending on the QC. This step of recursively applying the result to the parent is for nodes that have missed consensus rounds and are catching up. It might never really happen in Cerberus, but it's part of the original Hotstuff implementation

dan_layer/core/src/workers/hotstuff_waiter.rs
```diff
@@ -1115,29 +1222,24 @@ where
...
-                    .or_default()
-                    .push(SubstateState::Up {
-                        created_by: payload_id,
-                        data: substate.clone(),
-                    });
-            }
-        },
-        TransactionResult::Reject(ref reason) => return Err(HotStuffError::TransactionRejected(reason.clone())),
+    // down first, then up
+    for (address, version) in diff.down_iter() {
```
stringhandler, line 1345: I think it's better (in a future PR) to do these in a linear order

stringhandler (commented):
Only reviewed the hotstuff section. Seems good so far

dan_layer/core/src/workers/hotstuff_waiter.rs
sdbondi: Yeah we might have to do that, for multiple mutations of any object there engine gives you one down and one up so I suspect downing and upping is ok

applications/tari_validator_node/proto/dan/consensus.proto
```diff
@@ -71,10 +71,11 @@ message TariDanPayload {
 
 message VoteMessage {
   bytes local_node_hash = 1;
-  bytes shard_id = 2;
-  QuorumDecision decision = 3;
-  repeated ShardVote all_shard_nodes = 4;
-  ValidatorMetadata validator_metadata = 5;
+  bytes payload_id = 2;
```
stringhandler, line 74: The payload id should not technically be needed here because it is implied by the local node hash. The shard is also not necessary. It might have crept in for convenience, but it might mean that these fields need to be validated for mismatch on the node

applications/tari_validator_node/src/dry_run_transaction_processor.rs
```diff
@@ -192,18 +193,22 @@ impl DryRunTransactionProcessor {
...
                         continue;
                     },
                 };
             }
         }
 
-        // The shard was not found in any VN
-        Ok(None)
+        // The shard does not exist on any VN, so we pledge it to be created in this payload
+        Ok(ObjectPledge {
```
stringhandler, line 207: This could be wrong or potentially lead to errors in future. I suppose it's not technically incorrect because the shard doesn't exist.

I am not 100% familiar with the dry-run logic though so I might be speaking out of place. I am not completely against this logic, so maybe just an extra comment

dan_layer/core/src/models/hot_stuff_tree_node.rs
```diff
@@ -84,41 +83,47 @@ impl<TAddr: NodeAddressable, TPayload: Payload> HotStuffTreeNode<TAddr, TPayload
...
+        local_pledge: Option<ObjectPledge>,
+    ) -> Self {
+        Self {
+            // A genesis node is special, and therefore has a zero hash to indicate this
+            hash: TreeNodeHash::zero(),
             parent: TreeNodeHash::zero(),
-            payload_id: PayloadId::zero(),
+            shard: shard_id,
+            height: NodeHeight(0),
+            payload_id,
```
stringhandler, line 99: This should be 000000

dan_layer/core/src/models/hot_stuff_tree_node.rs
```diff
@@ -84,41 +83,47 @@ impl<TAddr: NodeAddressable, TPayload: Payload> HotStuffTreeNode<TAddr, TPayload
...
+        shard_id: ShardId,
+        proposed_by: TAddr,
+        local_pledge: Option<ObjectPledge>,
+    ) -> Self {
+        Self {
+            // A genesis node is special, and therefore has a zero hash to indicate this
+            hash: TreeNodeHash::zero(),
             parent: TreeNodeHash::zero(),
-            payload_id: PayloadId::zero(),
+            shard: shard_id,
```
stringhandler, line 97: This should be 00000

stringhandler (approved):
There's a lot in this PR, but I've spent time offline with @sdbondi going through it and agree with the changes. Happy to merge and apply further changes in subsequent prs

applications/tari_validator_node/src/dry_run_transaction_processor.rs
sdbondi: I also wasnt sure, but when we pledge we do this, use the current payload id and pledge DoesNotExist. I guess the question is: what does a pledge of None mean vs DoesNotExist?

sdbondi: If there is a difference, I'll explain that in a comment and revert the commit (this change was done in one commit incase it was wrong)

applications/tari_validator_node/proto/dan/consensus.proto
sdbondi: Yup absolutely correct, looked in the code and it's just used debug data in the db which we could get by looking up the node. Will remove in subsequent PR

PR #268, opened 2022-12-16
feat: integrating cucumber tests with validator_node_cli

mrnaveira:
Description
---
* Refactored the `validator_node_cli` to export utilities through a library (new `lib.rs` file)
* The cucumber tests now have steps to:
    * Create a file-based account (needed for creating subsequent transactions) in a temp directory
    * Initialize components, storing the component id associated with a name
    * Call component methods, and (optionally) assert the returned value
* The `Counter` template integration test now does method invocations

Motivation and Context
---
We want to integrate cucumber steps with the `validator_node_cli` to reuse transaction submission logic as well as covering the CLI itself with the tests.

Solves https://github.com/tari-project/tari-dan/issues/265

How Has This Been Tested?
---
The cucumber scenario works

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6349,6 +6349,7 @@ dependencies = [
  "tari_template_builtin",
  "tari_template_lib",
  "tari_test_utils",
+ "tari_validator_node_cli",
  "tari_validator_node_client",
  "tari_wallet",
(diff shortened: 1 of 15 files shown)


sdbondi (approved):
utACK

PR #269, opened 2023-01-10
fix: re-add pledge validation

sdbondi:
Description
---
- adds validate_pledges back 
- bugfix: commits all decided nodes

Motivation and Context
---
Pledges received from the leader are validated against the local execution change set.

In multi-shard executions, only one of the state changes for a shard is persisted. This change collects the decided nodes and persists all of them. 

How Has This Been Tested?
---
Manually, creating account and faucet components and depositing faucet coins in the account


stringhandler (approved):

sdbondi:
Closing in favor of #270 which includes these changes

PR #270, opened 2023-01-10
fix: commit substates in one db transaction after all shards have decided

sdbondi:
Description
---
- waits for all shards to decide before committing the state changes in single db transaction

Motivation and Context
---
In a transaction where some shards ACCEPT and some REJECT, the ACCEPT shards would be persisted even though the transaction as a whole is rejected. This PR ensures that either all or none of the changes local to the VN are persisted 

How Has This Been Tested?
---
Manually, token transfer manifest

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,8 +35,7 @@ report
 *.sqlite3
 keys.json
 node_modules
-/integration_tests/temp
-/integration_tests/cucumber_output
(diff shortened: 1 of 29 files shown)


stringhandler (approved):
tested locally and seems fine. could be some other bugs, but we'll find them in time

PR #272, opened 2023-01-11
docs: add openrpc spec for jsonrpc interface

stringhandler:
Description
---
Adds the beginnings of a JSONRPC spec

Motivation and Context
---
This allows invocation of methods using an interface like https://playground.open-rpc.org/ and allows mocking of the JSONRPC (for example to dev the frontend) using https://github.com/open-rpc/mock-server

How Has This Been Tested?
---
using the above tools


sdbondi (approved):
Nice

PR #274, opened 2023-01-11
fix(consensus): reuse previous execution result for a payload

sdbondi:
Description
---
Reuses a previous payload execution result for all shards

Motivation and Context
---
This greatly improves performance for multi-shard payloads as the payload is only executed once

How Has This Been Tested?
---
Manually, multi-shard transaction takes only slightly more time than a single shard transaction


dan_layer/core/src/workers/hotstuff_waiter.rs
```diff
@@ -1064,6 +1064,14 @@ where
     }
 
     fn execute(&self, payload: TPayload, object_pledges: &[ShardPledge]) -> Result<FinalizeResult, HotStuffError> {
+        let maybe_finalize_result = self
```
stringhandler: maybe we need to have a hash of the pledges used, to check that it's still the same

sdbondi: The payload id should have the pledges in it (we'll need to validate the transaction hash as part of mempool validation) - maybe that's all the security needed?

stringhandler: The actual data in the pledges might not be correct. This could be a way for a malicious node to invoke a different result earlier

sdbondi: Good point, will try work out the best way to do that check in this PR

stringhandler (approved):

PR #275, opened 2023-01-11
refactor(shard_store): split read and write operations in shard store

sdbondi:
Description
---
- Splits up read and write transactions.
- Fixes bug where write transactions using with_write_tx are committed even if the closure fails
- Adds WARN log message for uncommitted write transactions

Motivation and Context
---
API changes that allow implementations to take advantage of read/write semantics of the implemented database.

It is fairly easy to forget to commit the transaction, this PR will emit a warning log in this case. 

This also helps make it clearer when state is written.

How Has This Been Tested?
---
Manually, consensus still works


stringhandler (approved):
Nice. Can't believe how quickly you wrote this

PR #277, opened 2023-01-11
fix: exit validator node process on panic

sdbondi:
Description
---
- Custom panic hook that exits when any panic is encountered
- fixes panic in HTTP ui if the socket is in use by using an OS-assigned port

Motivation and Context
---
A panic may fairly silently occur in a tokio task. Panics should end the process so that the problem is immediately noticed, in a production environment the process may be restarted.

How Has This Been Tested?
---
By running the below, the process exits after 1s
```rust
    tokio::task::spawn(async move {
        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
        panic!("Testing panic hook");
    });
```


CjS77:
Please try keep your PRs focused on one thing

CjS77:
Is this really what we want? 
You can make tasks not fail silently by inspecting the `join*` result and handling from there

stringhandler (approved):
Nice

PR #280, opened 2023-01-11
fix(consensus): validate pledges are correct payload

stringhandler:
Currently replicas vote Accept even if a shard is pledged to a different payload:

image

In this case, the first and third shards are pledged to 4de... 
image

With this PR, the nodes reject the payload for the right reason, but they still seem to go all the way to Decide. (WIP: Not sure why yet)

image


sdbondi:
image

Nice, I think we should check the existing pledge from the `pledge_object` function matches the current payload and vote reject earlier if it doesn't.

sdbondi (approved):

PR #285, opened 2023-01-12
fix(consensus): any execution error will result in REJECT decision

sdbondi:
Description
---
Changes behaviour to reject decision for any payload execution failures

Motivation and Context
---
The payload processor returns a REJECT for the payload, however REJECT results were not persisted causing consensus to fail in an unexpected way and not complete, which led to active pledges being re-used in subsequent payloads (fixed in #280 ).

This also votes to reject for any execution failures, currently this is only possible if the template does not exist which is an invariant at this point (template addresses are checked before the transaction gets to consensus)

How Has This Been Tested?
---
Manually, create a counter component, call an invalid method on the counter component, then check that no pledges are dangling and REJECT consensus is completed


stringhandler (approved):
Tested locally and working

PR #287, opened 2023-01-12
fix(consensus): check payload matches on proposed and local pledges, else REJECT

sdbondi:
Description
---
checks payload matches on proposed and local pledges, else REJECT

Motivation and Context
---
If a leader sends a proposal pledged to another payload, fail proposal validation
If a node has locally pledged to another payload, REJECT

How Has This Been Tested?
---
Not explicitly tested (TODO), Previous manual tests still work


stringhandler (approved):
Yeah this is better

PR #292, opened 2023-01-13
feat(consensus): abandon early if shard is rejected

sdbondi:
Description
---
Abandons payload early if any shard has a REJECT QC

Motivation and Context
---
Since we know that at least one foreign shard has QC REJECT, the entire payload will never be able to be committed so no further votes are required, pledges can be released and the entire payload is REJECTED. This decreases TTR (time to result) and excess network traffic from erroneous payloads.

How Has This Been Tested?
---
A multishard payload with ~~one missing pledge (single shard REJECT) and separately a payload with~~ an incorrect method node (all shards REJECT), followed by an ACCEPT with the correct payload

### DIFF
--- a/dan_layer/core/src/workers/hotstuff_waiter.rs
+++ b/dan_layer/core/src/workers/hotstuff_waiter.rs
@@ -729,7 +729,7 @@ where
 
                 // If an active pledge already exists for another payload, we REJECT this payload.
                 if pledge.pledged_to_payload != node.payload_id() {
-                    let finalize_result = FinalizeResult::errored(
+                    let finalize_result = FinalizeResult::reject(
                         node.payload_id().into_array().into(),
                         RejectReason::ShardPledgedToAnotherPayload(format!(
                             "Shard {} is pledged to another payload {}",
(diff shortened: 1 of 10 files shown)


dan_layer/core/src/workers/hotstuff_waiter.rs
```diff
@@ -425,6 +425,13 @@ where
                 return Ok(());
             }
 
+            // All proposals are for different shards and have the same height, so if one of them is REJECT then there
+            // is no point in continuing to vote.
+            if let Some(reject_node) = proposed_nodes.iter().find(|n| n.justify().decision().is_reject()) {
+                self.abandon_payload(reject_node, &proposed_nodes)?;
```
stringhandler: Seems correct. My only concern is that maybe we should vote to our local leader and wait for the local  QC reject before abandoning

sdbondi: Thanks I like that better, and then we should always expect the local pledge to exist when abandoning.

stringhandler (approved):

stringhandler:
Seems correct. utACK

PR #294, opened 2023-01-13
fix(template)!: remove extra unused version field

sdbondi:
Description
---
removes extra unused version field from component header

Motivation and Context
---
Version is contained within the substate value, not on the template-level objects.

How Has This Been Tested?
---
Code compiles


stringhandler (commented):
nice

stringhandler (approved):

PR #295, opened 2023-01-17
fix(consensus): fixes panic when generating a merkle proof

sdbondi:
Description
---
Removed panics when generating the VN Merkle Proof. 

Motivation and Context
---
The panic may occur due to incorrect API usage. For debugging purposes, I previously added a call `verify` after generating the proof, but I should have used `verify_leaf`. Removed the debugging code. 

To investigate: A panic inside the MerkleProof::check_roots was encountered.

How Has This Been Tested?
---
N/A temporary debug code removed

### DIFF
--- a/dan_layer/core/src/models/vote_message.rs
+++ b/dan_layer/core/src/models/vote_message.rs
@@ -92,28 +92,11 @@ impl VoteMessage {
         let challenge = self.construct_challenge();
         let signature = signing_service.sign(&*challenge).ok_or(HotStuffError::FailedToSignQc)?;
         // construct the merkle proof for the inclusion of the VN's public key in the epoch
-        let leaf_index = vn_mmr
-            .find_leaf_index(&*vn_mmr_node_hash(signing_service.public_key(), &shard_id))
-            .expect("Unexpected Merkle Mountain Range error")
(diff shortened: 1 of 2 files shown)


stringhandler:
I'd like to keep this one open for a bit so that I can replicate the cause and see where to fix the MMR crate

sdbondi:
Makes sense, perhaps we should dump the parameters and merkle proof state somewhere (logs) so that we can replicate it in a test

Cifko:
The panic inside the check_root was caused by mistaking the leaf index for a node index. We send non-peak node to the check_root which makes it fail. This was caused by generating path for one index and then sending the path for another node to the verify. E.g. one of the cases I debugged. If you have MMR of size 23, and you get leaf 10 (node index 18), you generate path to the root of length 2 (root of 18 is 21). But then you send it for index 10 (this is the mistake node vs leaf index). The path to the root for node 10 is length 3. So it eventually end up on node 13 thinking this is the peak and it send it to the `check_root` function. I really suggest we should wrap one of the indices e.g. `struct NodeIndex(usize)` this way we will not mix them again.

sdbondi:
Let's close this one

PR #296, opened 2023-01-17
fix: minor len fix

stringhandler:
Fix to use the vn length instead of the Hashmap length when setting the capacity


sdbondi:
Nice, Minor: the capacity still isn't correct because it isn't the size of the MMR nodes, would be nice if the tari_mmr lib could calculate that based on the number of leaf nodes we're inserting

sdbondi (approved):

PR #299, opened 2023-01-18
fix(integration_tests): base node, wallet and template compilation is broken

mrnaveira:
Description
---
* Fixed broken configuration on base node and wallet test process spawning
* Updated old `cargo.lock `file in the `counter` template inside the integration test
* Updated assert on the `counter` test after increasing it: it now works as expected

Motivation and Context
---
The integration test was broken:
* The wallet and base node did not spawn correctly (couldn't communicate between them)
* The counter template was not able to be compiled due to a cargo error

How Has This Been Tested?
---
The integration test now pass

### DIFF
--- a/applications/tari_validator_node/tests/features/basic.feature
+++ b/applications/tari_validator_node/tests/features/basic.feature
@@ -38,8 +38,9 @@ Feature: Basic scenarios
 
     # Increase the counter
     When I invoke on VAL_1 on component COUNTER_1 the method call "increase" with 1 outputs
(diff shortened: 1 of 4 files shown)


sdbondi (approved):
I think what might have happened to me is that the DHT memory DB didnt work which crashed everything, and maybe the ports were conflicting with local nodes. 

As you alluded to in the FIXME, the assigned port shouldn't matter as long as it's open. 

Will check if things are running locally for me after merging but thanks for fixing this.

sdbondi:
utACK

PR #300, opened 2023-01-18
chore(integration_tests): use proper port initialization in base node and remove old CI tasks

mrnaveira:
Description
---
* Used the `get_os_assigned_ports` function in the base node spawning
* Removed old CI tasks for JS cucumber tests

Motivation and Context
---
As part of a previous fix, we didn't use the `get_os_assigned_ports` function in the base node spawning

Also, the old JavaScript cucumber CI task is shown Github Actions but it does not make sense anymore

How Has This Been Tested?
---
The test pass


sdbondi (approved):

sdbondi:
utACK

Thanks!

PR #301, opened 2023-01-19
feat(templates)!: add support for minting of NFTs

sdbondi:
Description
---
- adds a basic example NFT template and test
- implements NFT support in template engine
- adds NFT token substate
- updates cli to displayNFT token substate changes

BREAKING CHANGES: all previous WASMS are not compatible with changes to the template_lib

Motivation and Context
---
Adds support for non-fungible resources in templates

```rust

    pub struct SparkleNft {
        address: ResourceAddress,
    }

    impl SparkleNft {
        pub fn new() -> Self {
            // Create the non-fungible resource 
            let address = ResourceBuilder::non_fungible()
                .with_token_symbol("SPKL")
                .build();
            Self { address }
        }

        pub fn mint(&mut self) -> Bucket {
            // Mint a new token with a random ID
            let id = NftTokenId::random();
      
            // These are characteristic of the NFT and are immutable
            let mut immutable_data = Metadata::new();
            immutable_data
                .insert("name", format!("Sparkle{}", id))
                .insert("image_url", format!("https://nft.storage/sparkle{}.png", id));
            // TODO: Custom data that is mutable by the token owner (probably a serialized struct)
            // Mint the NFT, this will fail if the token ID already exists
            ResourceManager::get(self.address).mint_non_fungible(id, NftToken::new(immutable_data, Vec::new()))
        }
    }
}
```

How Has This Been Tested?
---
New NFT template functional test


dan_layer/engine_types/src/substate.rs
```diff
@@ -158,6 +205,7 @@ pub enum SubstateValue {
     Component(ComponentHeader),
     Resource(Resource),
     Vault(Vault),
+    NonFungible(NftToken),
```
mrnaveira, line 208: Same here: `Resource` -> `Fungible`

mrnaveira (approved):
Looks good to me

dan_layer/engine_types/src/substate.rs
```diff
@@ -75,6 +81,7 @@ pub enum SubstateAddress {
     Component(ComponentAddress),
     Resource(ResourceAddress),
     Vault(VaultId),
+    NonFungible(ResourceAddress, NftTokenId),
```
mrnaveira, line 84: IMO the `Resource` variant should be named `Fungible` now to avoid confusion

dan_layer/engine_types/src/substate.rs
mrnaveira, line 208: Same here: `Resource` -> `Fungible`

dan_layer/engine_types/src/substate.rs
sdbondi: A resource is still either fungible or non fungible. It is the parent substate of a token, for nfts you need a separate substate to represent its data (vault contains IDs referencing them)  but fungible can be represented straight in the vault as a value.

PR #302, opened 2023-01-19
fix(consensus): abandon pledges if not voting on rejected tx

sdbondi:
Description
---
Abandons pledges if all nodes have voted REJECT and we will never reach DECIDE phase because voting has stopped.

Motivation and Context
---
Ref #292 

When all shards are rejected, pledges remain active. This PR fixes this bug.

How Has This Been Tested?
---
Manually, transaction that should be rejected, no pledges are active


stringhandler (approved):

sdbondi:
ACK

PR #304, opened 2023-01-20
fix(cli): create key-pair automatically if one doesnt exist

sdbondi:
Description
---
- renames accounts command to keys
- create a new key automatically if one doesnt exist

Motivation and Context
---
accounts command was misnamed as it creates a keypair. Avoid having to explicitly create a keypair.

How Has This Been Tested?
---
Manually: Running the cli without previously creating a key


mrnaveira (approved):
Nice

PR #306, opened 2023-01-20
fix(consensus): improve handling of all shards rejected

sdbondi:
Description
---
- correctly abandons pledges in conflicting transaction case, where some pledges may be pledged to another payload
- emits HotstuffEvent::Failed for payload in the above case

Motivation and Context
---
When testing conflicting transactions, the pledges for one or both transactions are correctly abandoned however the abandon call may fail as the pledge may correctly have not been made in the first place. The correct fail event will be emitted which is used currently to provide feedback to the JSONRPC user.

How Has This Been Tested?
---
Manually, running concurrent conflicting multishard transactions where pledges for different shards same payload are made for one or either payload - either one transaction passes or both fail (race condition)


stringhandler (approved):
utACK - happy to merge once conflicts are resolved

PR #307, opened 2023-01-20
feat: implementation of a leader failure pacemaker

jorgeantonio21:
Description
---
Implements a simple pacemaker to deal with waiting for leader failure on sending hotstuff messages

Motivation and Context
---

How Has This Been Tested?
---


dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -0,0 +1,225 @@
...
+
+#[derive(Debug)]
+pub struct LeaderFailurePacemaker {
+    rx_waiter_signal: Receiver<PacemakerSignal>,
+    tx_waiter_status: Sender<(WaitOver, PacemakerWaitStatus)>,
+    max_timeout: u64,
+    wait_over_set: HashSet<WaitOver>,
+}
+
+impl LeaderFailurePacemaker {
```
Cifko: Lets call it just Pacemaker. Btw you never actually know that any leader failed.

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -0,0 +1,225 @@
...
+        rx_waiter_signal: Receiver<PacemakerSignal>,
+        tx_waiter_status: Sender<(WaitOver, PacemakerWaitStatus)>,
+        max_timeout: u64,
+        shutdown: ShutdownSignal,
+    ) -> JoinHandle<Result<(), HotStuffError>> {
+        let pacemaker = Self::new(rx_waiter_signal, tx_waiter_status, max_timeout);
+        tokio::spawn(pacemaker.run(shutdown))
+    }
+
+    pub fn new(
```
Cifko: Should this be `pub`?

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -0,0 +1,225 @@
...
+                                if self.wait_over_set.contains(&wait_over) {
+                                    // we already start a waiting process for this
+                                    // payload and shard id
+                                    continue;
+                                }
+                                info!(target: LOG_TARGET,
+                                    "Waiting possible leader failure for payload_id = {}, shard_id = {}, hotstuff_phase = {:?}",
+                                    wait_over.0, wait_over.1, wait_over.2
+                                );
+                                tokio::time::sleep(Duration::from_secs(self.max_timeout)).await;
```
Cifko: This will lock this `loop` and you will never receive the `StopWait` because you will always send the timeout.
What you should do is have vector of futures, where each future is a tokio::select (or futures::select) over a block {timeout, send message} and other one is checking kill signal. Add all of the into `FuturesUnordered` and than add it here.

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -0,0 +1,225 @@
...
+            .unwrap();
+
+        // tokio::time::sleep(Duration::from_secs(2)).await;
+
+        let msg = tester.rx_waiter_status.recv().await.unwrap();
+        assert_eq!(msg.0 .0, payload);
+        assert_eq!(msg.0 .1, shard_id);
+        assert_eq!(msg.0 .2, phase);
+        assert_eq!(msg.1, PacemakerWaitStatus::WaitTimeOut);
+    }
```
Cifko, line 298: Add simple test for timeouts. 
Also add a one more complex e.g. a test where you add 3 waiter signals. You stop the middle one, and the other two trigger timeouts in order.

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -0,0 +1,225 @@
...
+
+use super::hotstuff_error::HotStuffError;
+use crate::models::pacemaker::{PacemakerSignal, PacemakerWaitStatus, WaitOver};
+
+const LOG_TARGET: &str = "tari::dan_layer::pacemaker_worker";
+
+#[derive(Debug)]
+pub struct LeaderFailurePacemaker {
+    rx_waiter_signal: Receiver<PacemakerSignal>,
+    tx_waiter_status: Sender<(WaitOver, PacemakerWaitStatus)>,
```
Cifko: Maybe it's not necessary but can we do a generic waitover? And in the future if we need to waitover something else it could come handy.

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    async fn send_timeout_msg(
+        &mut self,
+        wait_over: WaitOver,
+        status: PacemakerWaitStatus,
+    ) -> Result<(), HotStuffError> {
+        info!(
+            target: LOG_TARGET,
+            "Leader is not responsive, failed to communicate with replicas,  for payload_id = {}, shard_id = {}, \
```
Cifko: You can delete this log message. This will be handled in the receive of the message. Also if we change it to generic you will not have the info.

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -0,0 +1,225 @@
...
+            self.tx_waiter_status
+                .send((wait_over, status))
+                .await
+                .map_err(|_| HotStuffError::SendError)?;
+        }
+        Ok(())
+    }
+
+    pub async fn run(mut self, mut shutdown: ShutdownSignal) -> Result<(), HotStuffError> {
+        let (tx_status, mut rx_status) = tokio::sync::mpsc::channel::<(WaitOver, PacemakerWaitStatus)>(10);
```
Cifko: This is not necessary, you are using it just to send message to yourself.

dan_layer/core/src/workers/pacemaker_worker.rs
jorgeantonio21: Cool, let's rename it. Yes, we just timeout on the receiving of `StartWait` messages.

dan_layer/core/src/workers/pacemaker_worker.rs
jorgeantonio21: Let's change to private

dan_layer/core/src/workers/pacemaker_worker.rs
jorgeantonio21: I did refactored it accordingly

Cifko: I unresolved couple comments above, and now I realized that you forgot to push the changes.

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -0,0 +1,270 @@
...
+        }
+    }
+
+    pub async fn run(mut self, mut shutdown: ShutdownSignal) -> Result<(), HotStuffError> {
+        let max_timeout = self.max_timeout;
+        loop {
+            tokio::select! {
+                msg = self.rx_start_signal.recv() => {
+                    if let Some(wait_over) = msg {
+                        let (tx, mut rx) = channel::<bool>(1);
```
Cifko: use `oneshot::channel::<()>()`
also give it a proper naming (e.g. `rx_stop_timeout` or whatever, I'm not good with naming, but it should have a name)

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -0,0 +1,270 @@
...
+                        self.tx_inner_map.insert(wait_over, tx);
+                        info!(
+                            target: LOG_TARGET,
+                            "Received start wait signal for value: {:?}", wait_over
+                        );
+                        let tx_waiter_status = self.tx_waiter_status.clone();
+                        let _join = tokio::spawn(async move {
```
Cifko, line 127: I see that you switch from the `FuturesUnordered` approched to this. Was it not feasible to do it without the `tokio::spawns`? I'm wondering what's the limit of the `tokio::spawn`. I remember when there are too much thread you will not be notified at all, it will just not run. But maybe that's the case for `FuturesUnordered` as well. Don't know... So you will never receive a trigger when it happens. That's okay as long as you don't need the timeout. But maybe the number is high. I'm just thinking out loud. You don't have to change this if the timings and limitations are enough for us.

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -0,0 +1,270 @@
...
+                                _ = rx.recv() => {
+                                    info!("The wait signal for wait_over = {:?} has been shutted down", wait_over);
+                                }
+                            }
+                        });
+                    }
+                },
+                msg = self.rx_shutdown_signal.recv() => {
+                    if let Some(wait_over) = msg {
+                        if let Some(tx) = self.tx_inner_map.get(&wait_over) {
```
Cifko: you can call remove here and it will provide the value

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -0,0 +1,270 @@
...
+///   2. Receiving [`stop_wait`] messages for an already existing [`start_wait`].
+///   3. If it receives a stop message then it should stop the waiting and do nothing else;
+///   4. Otherwise, it should notify the sender of [`start_wait`] message that the max_timeout duration has passed.
+#[derive(Debug)]
+pub struct Pacemaker<T> {
+    rx_start_signal: Receiver<T>,
+    rx_shutdown_signal: Receiver<T>,
+    tx_inner_map: HashMap<T, Sender<bool>>,
+    tx_waiter_status: Sender<(T, PacemakerWaitStatus)>,
+    max_timeout: u64,
```
Cifko: The `max_` in the name suggests that the timeout can be different, but in reality it's always this value.
Can we change this directly to `Duration`? If not, than provide better description (e.g. `timeout_millis`).

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -0,0 +1,270 @@
...
+use super::hotstuff_error::HotStuffError;
+
+const LOG_TARGET: &str = "tari::dan_layer::pacemaker_worker";
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum PacemakerWaitStatus {
+    WaitTimeOut,
+}
+
+async fn send_timeout_message<T: Debug + Send>(
```
Cifko, line 50: Does this have to be in a function?

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -0,0 +1,270 @@
...
+                        let (tx, mut rx) = channel::<bool>(1);
+                        self.tx_inner_map.insert(wait_over, tx);
+                        info!(
+                            target: LOG_TARGET,
+                            "Received start wait signal for value: {:?}", wait_over
+                        );
+                        let tx_waiter_status = self.tx_waiter_status.clone();
+                        let _join = tokio::spawn(async move {
+                            tokio::select! {
+                                _ = tokio::time::sleep(Duration::from_millis(max_timeout)) => {
```
Cifko: When the timeout happens, you have to remove the `wait_over` from the `tx_inner_map`.

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -0,0 +1,270 @@
...
+}
+
+/// A pacemaker service that is responsible for:
+///   1. Receiving [`start_wait`] messages parametrized over a type `T` and start wait up to a max_timeout value.
+///   2. Receiving [`stop_wait`] messages for an already existing [`start_wait`].
+///   3. If it receives a stop message then it should stop the waiting and do nothing else;
+///   4. Otherwise, it should notify the sender of [`start_wait`] message that the max_timeout duration has passed.
+#[derive(Debug)]
+pub struct Pacemaker<T> {
+    rx_start_signal: Receiver<T>,
```
Cifko, line 71: Sorry for nitpicking, but can you add a comment to each field what's the purpose. To me and you it's obvious, but for future generations :-)

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -0,0 +1,270 @@
...
+                    if let Some(wait_over) = msg {
+                        if let Some(tx) = self.tx_inner_map.get(&wait_over) {
+                            tx.send(true).await.unwrap();
+                            // remove the entry from the tx_inner_map
+                            self.tx_inner_map.remove(&wait_over);
+                        }
+                    }
+                },
+                _ = shutdown.wait() => {
+                    info!("Shutting down pacemaker service..");
```
Cifko, line 153: I'm not sure if it's necessary, but maybe you should trigger all shutdowns. So nothing is left behind running.

dan_layer/core/src/workers/pacemaker_worker.rs
Cifko: So I've tested this, and tokio doesn't spawn real threads (my test spawned 10k threads) and there were only 20 system threads. I don't know how tokio does it, but it's good.

jorgeantonio21:
Will close this PR, in favor of #321.

PR #308, opened 2023-01-20
feat(integration_test): allow to submit manifests

mrnaveira:
Description
---
* Added new steps to submit embedded manifests
* Added a new scenario to test fungible token mint and deposit using the new manifest step

Motivation and Context
---
In cucumber integration tests, we want a way to easily test transaction manifests. Something of the sort of:
```
When I submit a transaction manifest on VN with 4 outputs
        ```
        use template_faucet as TestFaucet;
        use template_0000000000000000000000000000000000000000000000000000000000000000 as Account;

        fn main() {
            let faucet = global!["FAUCET"];
            let mut acc1 = global!["ACC_1"];

            // get tokens from the faucet
            let faucet_bucket = faucet.take_free_coins();
            acc1.deposit(faucet_bucket);
        }
        ```
```
Where template addresses (e.g. `use template_faucet as`) and component ids (e.g. `global!["FAUCET"]`) are automatically resolved from the test context variables

Solves [#258](https://github.com/tari-project/tari-dan/issues/258)

How Has This Been Tested?
---
New cucumber scenario that mints fungible tokens and transfers them to an account


sdbondi (approved):
Awesome!

PR #313, opened 2023-01-25
feat(test): add more NFT template examples

mrnaveira:
Description
---
Created new templates around NFTs:
* **emoji_id**: a classic, first come first served, constant price NFT drop
* **tickets**: example of a ticket sale for event attendances. The same logic could be used for supply chain tracking of products
* **marketplace**: users can sell and buy NFTs directly as well as through english-like auctions.

**TBD: a gaming example**

The templates do not compile yet as they present new features.

Motivation and Context
---
We have some basic functionality implemented for NFTs, but we need some more advanced, real life scenarios to visualise which features we still lack.

So this PR is a proposal of implementation of multiple templates that may offer real usage of NFTs. These templates showcase features that are still not implemented.

How Has This Been Tested?
---
There new templates showcase features that are still not implemented, so testing does not apply yet.


sdbondi:
These are great! Looks like we're not too far from making these work :rocket:

sdbondi (approved):
:rocket:  - we can work on getting these working in subsequent PRs

PR #319, opened 2023-01-27
feat: add hardcoded network latency constant to hotstuff waiter

jorgeantonio21:
Description
---
Adds network latency to `HotstuffWaiter` as an hardcoded constant. Eventually, this value should be able to be (pre)computed on every hotstuff round, by averaging/max, depending on nodes responsiveness.

Motivation and Context
---

How Has This Been Tested?
---


jorgeantonio21:
Closed in favor of the most extensive PR #321

PR #320, opened 2023-01-27
feat(test): implement the emoji_id NFT template example

mrnaveira:
Description
---
* Adapted the `nft/emoji_id` template example for the engine tests to be able to be compiled and executed
* Added a new unit test to mint 
* `NonFungibleId:: try_from_string` now does **not** check char range, only length, to be able to support emoji UTF-8 characters

Motivation and Context
---
On https://github.com/tari-project/tari-dan/pull/313 we added some examples of advanced NFT scenarios. We want those scenarios to compile and execute. This PR achieves that for one of them (`emoji_id`)

How Has This Been Tested?
---
New unit test for `emoji_id` minting


dan_layer/engine/tests/templates/nft/emoji_id/src/lib.rs
```diff
@@ -65,35 +85,37 @@ mod emoji_id {
             }
         }
 
-        // TODO: how do we ensure that the payment is indeed in Thaums?
-        pub fn mint(&mut self, emojis: Vec<Emoji>, payment: Bucket) -> (Bucket, Bucket<Thaum>) {
+        // TODO: return change
```
sdbondi, line 88: Maybe instead of returning change we just check the value `bucket.value() == required_amount`

sdbondi (approved):
Nice LGTM

PR #321, opened 2023-01-27
feat: integrate pacemaker service across codebase

jorgeantonio21:
Description
---
Implements pacemaker functionality, whose purpose is to be used on Hotstuff leader response failure. 

Motivation and Context
---
In OCerberus, any VN will request messages from both its local leader as well as foreign leaders. Leader responses are timed, using a pacemaker. If after a well defined timeout duration, our VN hasn't received a new (local or foreign) leader message, it will proceed to get the current consensus state via other VN's (if applicable), or simply request a new leader election. See #311 for further details.

How Has This Been Tested?
---
Added suitable unit tests.


dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -0,0 +1,299 @@
...
+    /// Duration of timeout period
+    max_timeout: Duration,
+}
+
+impl<T: Clone + Debug + PartialEq + Eq + Hash + Send + Sync + 'static> Pacemaker<T> {
+    pub fn spawn(
+        rx_start_signal: Receiver<T>,
+        rx_shutdown_signal: Receiver<T>,
+        tx_waiter_status: Sender<(T, PacemakerWaitStatus)>,
+        max_timeout: Duration,
```
Cifko: I think we have to remove this timeout, instead add it to the channel. Because it will be different for some type of messages.

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -0,0 +1,299 @@
...
+            tx_waiter_status,
+            max_timeout,
+        }
+    }
+
+    pub async fn run(mut self, mut shutdown: ShutdownSignal) -> Result<(), HotStuffError> {
+        let max_timeout = self.max_timeout;
+        loop {
+            tokio::select! {
+                msg = self.rx_start_signal.recv() => {
```
Cifko: we have to add a check if the pacemaker is already running for `msg`, and if it is, ignore the message.

dan_layer/core/src/workers/pacemaker_worker.rs
jorgeantonio21: So it is part of the overall message?  In that case, we need a check to see if the message contains an inner `Duration` type, so that we can use it to timeout

Cifko: What do you mean by inner `Duration`? The channel will be over `(T, Duration)`. The duration is not in the `T`

jorgeantonio21: Add this refactor

dan_layer/integration_tests/src/harness.rs
```diff
@@ -152,11 +153,13 @@ impl HsTestHarness {
             tx_broadcast,
             tx_vote_message,
             tx_events,
+            pacemaker,
             payload_processor,
             shard_store.clone(),
-            shutdown.to_signal(),
+            shutdown.to_signal().clone(),
```
sdbondi: unnecessary clone
```suggestion
            shutdown.to_signal()
```

sdbondi (approved):
LGTM - utACK

PR #322, opened 2023-01-27
feat: add example of feature gating

brianp:
Description
---
This utilizes the introduced network type feature gate to remove a controversial piece of code from mainnet. 

Motivation and Context
---
Utilize the feature gate on a practical example

How Has This Been Tested?
---
Local compilation


applications/tari_validator_node/src/main.rs
```diff
@@ -39,11 +46,14 @@ async fn main() {
...
-        process::exit(1);
-    }));
+    #[cfg(tari_feature_crash_on_thread_panic)]
+    {
+        let default_hook = panic::take_hook();
+        panic::set_hook(Box::new(move |info| {
+            default_hook(info);
+            process::exit(1);
+        }));
+    }
```
brianp, line 56: The use here of `#[cfg(tari_feature_crash_on_thread_panic)]` instead of `#[tari_feature(crash_on_thread_panic)]` is because I could not successfully re-write the macro to manage syn::Item's as well as expressions. It may yet be possible but I thought it more important to push the PR than to spend more time debugging that detail.

PR #332, opened 2023-01-31
feat: add LFU template cache

sdbondi:
Description
---
Adds template cache using [mini-moka](https://crates.io/crates/mini-moka)

Motivation and Context
---
Shortcuts the WASM module and ABI loading using a little over 200MiB of system memory maximum (configurable with `validator_node.templates.max_cache_size_bytes`)

How Has This Been Tested?
---
Manually


stringhandler (commented):
Like the idea. Will need to review properly to make sure there's no chance of cache problems

stringhandler (approved):

PR #335, opened 2023-02-01
feat(engine): get the current epoch inside template code

mrnaveira:
Description
---
* Created a new `Consensus` type accessible inside templates, with a `current_epoch` method. The name _Consensus_ was chosen because it will likely serve in the future as an interface into other committee consensus protocols (timestamps, random generator, etc.). 
* The calls to `Consensus` map into a new engine operation, that relies into a new `ConsensusProvider` trait with all the needed operations, for now only the current epoch number. The trait is needed to avoid dependencies from the `engine` to outside (right now, the `EpochManager` in the validator node).
* The `ConsensusProvider` trait is implemented in the validator node application to access the `EpochManager` for retrieving the current epoch.

Motivation and Context
---
The NFT marketplace example showcases the need to allow time-sensitive features. This PR aims to provide a first version of that.

How Has This Been Tested?
---
New engine unit test with a new template (`consensus`) that calls the new engine operation to get the current epoch


applications/tari_validator_node/src/consensus_provider.rs
```diff
@@ -0,0 +1,45 @@
...
+
+impl TariDanConsensusProvider {
+    pub fn new(epoch_manager: EpochManagerHandle) -> Self {
+        Self { epoch_manager }
+    }
+}
+
+impl ConsensusProvider for TariDanConsensusProvider {
+    fn current_epoch(&self) -> u64 {
+        let res = block_on(self.epoch_manager.current_epoch());
```
sdbondi: Nice and neat. I have a concern: what happens if the epoch changes during execution? I think in principle the epoch should be provided by consensus as a (possibly pledged) substate as all nodes have to agree on this. The engine can read that substate from the state store. I suspect this will also simplify the code a bit and decouple the runtime from an additional dependency.

We can consider passing in the epoch here (something like `ConsensusContext`) or adding it to pledges, which in turn make that substate available to the state store.
https://github.com/tari-project/tari-dan/blob/b9b42a608cf0a19e61a96bfe08912b7c8302defa/dan_layer/core/src/workers/hotstuff_waiter.rs#L1313

@stringhandler any thoughts?

mrnaveira: Sure, for now we can pass the `ConsensusContext` to the `proces_payload` method as you suggest. If later we think it's better to include in in pledges, the change should be simple enough.

sdbondi (approved):
Nice, thanks!

sdbondi:
Tests ran locally - let's merge!

PR #338, opened 2023-02-02
fix: removes pacemaker pending wait overs that already have timed out

jorgeantonio21:
Description
---
The PR #321 introduces a pacemaker service. As opposed to a previous implementation (see PR #307), we were not removing pending wait overs, after they timed out, which increased linearly over time the size of the collection storing these. This was not intended, and this PR fixes this problem.

Motivation and Context
---

How Has This Been Tested?
---
Manually, by printing pending_timeouts in current units and check that those that have timed out are removed.


Cifko (approved):
LGTM

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -104,11 +110,15 @@ where T: Clone + Debug + PartialEq + Eq + Hash + Send + Sync + 'static
...
-                    if self.pending_timeouts.insert(wait_over.clone(), tx).is_none() {
+                    let pending_timeouts = self.pending_timeouts.clone();
+                    if pending_timeouts.lock().unwrap().insert(wait_over.clone(), tx).is_none() {
                         self.waiting_futures.push(Box::pin(async move {
                             tokio::select! {
                                 _ = tokio::time::sleep(duration_timeout) => {
                                     info!("The wait signal for value = {:?} has timeout", wait_over);
+                                    // remove the wait_over from the pending_timeouts
+                                    // so that this data structure doesn't grow every time
+                                    pending_timeouts.lock().unwrap().remove(&wait_over);
```
sdbondi: Sync locks in async code are not good.

You shouldnt need a lock, the waiting_future will resolve one way or another so you should be able to  clean up pending timeouts then

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -104,11 +110,15 @@ where T: Clone + Debug + PartialEq + Eq + Hash + Send + Sync + 'static
...
                                     Some(wait_over)
```
sdbondi:
```suggestion
                                    (wait_over, true)

```
and replace None with `(wait_over, false)`

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -123,9 +133,9 @@ where T: Clone + Debug + PartialEq + Eq + Hash + Send + Sync + 'static
                 },
                 Some(msg) = self.waiting_futures.next() => {
                     if let Some(t) = msg {
```
sdbondi:
```suggestion
                    let (wait_over, is_resolved) = msg;
                    self.pending_timeouts.remove(wait_over);
                    if is_resolved {
                     //....
```

sdbondi (commented):
Let's avoid locks if we can. Async locks (e.g. `tokio::sync::Mutex`) should be used in async code but we should be able to avoid them altogether.

dan_layer/core/src/workers/pacemaker_worker.rs
jorgeantonio21: Oh I see, that is actually a clever approach. Let me refactor it, to put these considerations in place.

dan_layer/core/src/workers/pacemaker_worker.rs
jorgeantonio21: Addressed this

dan_layer/core/src/workers/pacemaker_worker.rs
jorgeantonio21: Addressed this

Cifko (commented):
Do you mind adding test for this?

sdbondi (approved):
Failing tests were fixed in #339 - so happy with this change, can add tests in another PR

PR #340, opened 2023-02-03
feat!: initial component access rules and ownership

sdbondi:
Description
---
- feat(engine): adds component access rules
- feat(template): adds support for access rules
- feat(template-macro): adds support for specifying access rules
- feat(account-builtin): adds access rules to account component
- feat(test-tooling): update and add tests for account ownership
- feat(cli): add support for ownership
- feat(cli): add `account create` command which creates an account for the owner token based on the cli signing key
- feat(validator-node): integrates payload processor changes

Motivation and Context
---

Enforces basic ownership rules to component methods. The transaction signer is granted a virtual token that represents their permission to access component methods that permit that token. 

The only difference that should be expected from previous executions is if you attempt to withdraw from an account that was created from a different key than the tx signer.

An Account component now takes in an NF token. This is typically set to the virtual token generated from the owner's public key, but could be any NFT token. At the moment, setting it to any token would mean that no one would be able to withdraw from the Account because this implementation only provides the transaction signer public key ownership token to the auth system without including any other owned tokens that may be pledged to the transaction.

How Has This Been Tested?
---
New account template test, Manually by creating an account with key1, minting an NFT and depositing it to the account, creating another account with key2 and attempting to submit a transaction to transfer the NFT to the key2 account (which fails with Access denied error)

BREAKING CHANGES:
- Account::new renamed to Account::create (both to satisfy clippy and to differentiate now that a parameter is required)
- Template ABI changes mean some operations are not compatible with previous templates (templates need to be rebuilt)
- ComponentHeader substate includes access_rules so will not decode from previous component state


applications/tari_validator_node/src/payload_processor.rs
```diff
@@ -102,6 +101,28 @@ where TTemplateProvider: TemplateProvider<Template = LoadedTemplate>
...
+            .get_template_module(&addr)
+            .map_err(|err| PayloadProcessorError::FailedToLoadTemplate(err.into()))?;
+        builder.add_template(addr, template);
+    }
+
+    Ok(builder.build())
+}
+
+fn get_auth_token(transaction: &Transaction) -> NonFungibleAddress {
+    let public_key = Ed25519PublicKey::from_bytes(transaction.sender_public_key().as_bytes())
```
stringhandler: why not ristretto?

stringhandler (commented):
Happy in general, just wondering about the Ed25519 instead of Ristretto

applications/tari_validator_node/src/payload_processor.rs
sdbondi: Ristretto is the group so thought it better to specify the curve since ristretto is more of a detail. It refers to the Resource more than the actual contents. We could, for example, use a hash of the key and/or part of the entropy of the key for the NFT ID just as long as it's derived from an ED25519 curve. Though happy to use Ristretto as the name, just a decision I made :)

stringhandler: The problem with reading an Ed25519 from bytes is that there could be 4 private keys because of the co-factor. I think let's stick to Ristretto for everything for now and we can expand as the need arises

stringhandler (approved):

stringhandler:
Tested locally. I think the structure might evolve depending on security, but this is usable. Nice one!

PR #341, opened 2023-02-03
test: add foreignly only failure test

Cifko:
Description
---
fixes two pacemakers bugs
- when the stop/start are called one after another 
- when you add existing timer it will work properly

How Has This Been Tested?
---
The cerberus tests are no longer failing.

### DIFF
--- a/dan_layer/core/src/workers/pacemaker_worker.rs
+++ b/dan_layer/core/src/workers/pacemaker_worker.rs
@@ -104,24 +104,26 @@ where T: Clone + Debug + PartialEq + Eq + Hash + Send + Sync + 'static
                             self.handle_stop_signal(wait_over);
                         }
                         PacemakerTimer::Start(wait_over, duration_timeout) => {
-                    info!(
-                        target: LOG_TARGET,
-                        "Received start wait signal for value: {:?}", wait_over
-                    );
-                    let (tx, rx_stop_signal) = oneshot::channel();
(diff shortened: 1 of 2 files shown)


dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -104,34 +104,38 @@ where T: Clone + Debug + PartialEq + Eq + Hash + Send + Sync + 'static
                             self.handle_stop_signal(wait_over);
                         }
                         PacemakerTimer::Start(wait_over, duration_timeout) => {
-                    info!(
-                        target: LOG_TARGET,
-                        "Received start wait signal for value: {:?}", wait_over
-                    );
-                    let (tx, rx_stop_signal) = oneshot::channel();
-                    if self.pending_timeouts.insert(wait_over.clone(), tx).is_none() {
```
jorgeantonio21, line 112: Was this the source of the bug ? That you always inserted a new pending signal, independently if it was already there ?

jorgeantonio21 (approved):

sdbondi (approved):
Makes sense. Thanks

dan_layer/core/src/workers/pacemaker_worker.rs
Cifko: One of them. What this was doing is best described by the documentation `If the map did have this key present, the value is updated, and the old value is returned` from that you can imagine the implications.

PR #344, opened 2023-02-03
feat: import burned utxos

stringhandler:
Requires https://github.com/tari-project/tari/pull/5169

Description
---
Adds a step to the base layer scanner that will import burnt utxos. All utxos are stored, regardless of the shard space the VN is responsible for. The utxos are automatically created as substates with a unique address so they cannot be spent twice. In the current code, only the engine and consensus are responsible for creating substates, so this is slightly different. If anyone has a better suggestion I'm all ears. The reason I created it as a substate is so that it can be pledged immediately and the consensus layer code already handles that management of it

Future TODO: Add test to prove you can't claim it twice
Future TODO: Prove another account cannot claim it


sdbondi (commented):
Looks great

applications/tari_validator_node/src/base_layer_scanner.rs
```diff
@@ -267,8 +282,8 @@ impl BaseLayerScanner {
             for output in utxos.outputs {
                 let output_hash = output.hash();
                 if output.is_burned() {
-                    warn!(target: LOG_TARGET, "Burned output encountered. Not yet implemented");
-                    // self.register_burnt_utxo(output.commitment);
+                    info!(target: LOG_TARGET, "Found burned output: {}", output_hash);
+                    self.register_burnt_utxo(output.commitment)?;
```
sdbondi: Maybe only the shards @ H(commitment) should save this substate, so that it can be DOWNed on claim

sdbondi (commented):
To summarize the flow: 
1. Burnt on L1, Vns pick this up
2. All VNs persist a substate for the burnt commitment 
3. Claim: User provides burnt L1 substate as input to transaction
4. Claim: Bucket is created for L2 TARI resource
5. Claim: Deposited into L2 TARI Vault 

L1 substate remains after claim, this is because not all VNs are privy to the claim transaction so it currently isn't possible to do otherwise. Perhaps only shards @ `H(commitment)` persist this substate. Otherwise it can be claimed twice since the destination is a vault which has an independent address. 

Alternatively, a claim could 'mint' the commitment with a shard address: `H(H(L2 Tari resource, commitment), version)` similar to non-fungibles, so that if claimed twice, the destination address would be UP or previously DOWN and the transaction will fail.

sdbondi:
So perhaps I misunderstood, the double claim is prevented by the shards that own the burnt L1 commitment downing it. Which means the next time it's claimed, it will be down in those shards. Think that works, though still think shards at the commitment address `H(commit_xxx || version(0))` should only store it in the substate store as an optimisation (doesnt have to be in this PR). 

I think some code is missing from the `finalize` method that adds the down to the finalize result, and therefore commits the down.

stringhandler:
Yeah good catch, I meant to add it but forgot to. Let me add some cucumber steps as well to verify that it is downed

sdbondi (approved):

PR #347, opened 2023-02-04
feat(integration_test): new cucumber NFT scenario

mrnaveira:
Description
---
* Created a new cucumber scenario that mints, mutates and burns NFTs.
* Needed to add an ad-hoc way to manifests to easily declare the new NFT substates of a transaction, so I specify them through special comments (`$mint` and `$mint_specific`). In the future, when the substate calculation is delegated to the VNs, this workaround will be not necessary.

Motivation and Context
---
Integration tests must cover the new NFT functionality

How Has This Been Tested?
---
The new cucumber scenario passes


applications/tari_validator_node/tests/cucumber.rs
```diff
@@ -260,6 +260,30 @@ async fn call_template_constructor(
...
+        template_name,
+        vn_name,
+        function_call,
+        vec![],
+        num_outputs,
+    )
+    .await;
+
+    // give it some time between transactions
+    tokio::time::sleep(Duration::from_secs(4)).await;
```
sdbondi, line 284: Is there a reason we wait here?

sdbondi (approved):
Nice, like the way you allowed NFT outputs to be defined in comments.

Failing test is unrelated and has been fixed in #346

PR #348, opened 2023-02-06
fix: multiple pledges for the same committee

Cifko:
Description
---
When leader fails after sending a proposal. The new leader will sends proposal for the same height. And then the committee votes for both of them, which will result in incorrect pledges hash.

How Has This Been Tested?
---
I'm working on a proper test for the local leader failure. Which will trigger local leader failure on every step. And this fixed a part of the problem that is currently there.


sdbondi:
If there are duplicate proposals for a shard ID, surely the proposal sent by the previous leader could be returned?

Perhaps we need to further qualify the query with the leader public key.

Cifko:
> If there are duplicate proposals for a shard ID, surely the proposal sent by the previous leader could be returned?
> 
> Perhaps we need to further qualify the query with the leader public key.

There are two possible ways we can do this. Ignore the old one, or discard the new one.
The new leader have to send something to the committee once he is elected. So the committee know they can vote again. Or when we do re-election we can send the votes as well. But that feels wrong.
Btw the multiple proposal for the same shard and height can also happen for foreign committee proposal.

Cifko:
> > If there are duplicate proposals for a shard ID, surely the proposal sent by the previous leader could be returned?
> > Perhaps we need to further qualify the query with the leader public key.
> 
> There are two possible ways we can do this. Ignore the old one, or discard the new one. The new leader have to send something to the committee once he is elected. So the committee know they can vote again. Or when we do re-election we can send the votes as well. But that feels wrong. Btw the multiple proposal for the same shard and height can also happen for foreign committee proposal.

This will actually not work. Because we have a check that the current leader validates only votes for his own proposal.
https://github.com/tari-project/tari-dan/blob/development/dan_layer/core/src/workers/hotstuff_waiter.rs#L1265

And I can't just send local leader to the query. I will change this PR to query all proposal and choose the correct ones.

sdbondi (approved):
Nice - utACK

PR #350, opened 2023-02-06
feat: add a remove pending timeout unit test to pacemaker

jorgeantonio21:
Description
---
We test that pending timeouts are being removed correctly, within the internal state of a `Pacemaker` struct. Ideally, this test would make use of raw/smart pointers while `Pacemaker` run on the background (within a spawned thread) as a service. But this turns out to be complex, as we don't know which new memory address is the data pointing to. For this reason, we encapsulated most of the logic of the `run` function into proper segments, so that we can test these in a unit test environment.

Motivation and Context
---
Related to a previous bug, that we had before.

How Has This Been Tested?
---
Add a specific unit test.

### DIFF
--- a/dan_layer/core/src/workers/pacemaker_worker.rs
+++ b/dan_layer/core/src/workers/pacemaker_worker.rs
@@ -301,15 +301,13 @@ mod tests {
         let (_, rx_signal) = channel::<PacemakerTimer<u8>>(10);
         let (tx_waiter_status, _) = channel::<u8>(10);
 
-        let mut pacemaker = Box::new(Pacemaker::new(rx_signal, tx_waiter_status));
+        let mut pacemaker = Pacemaker::new(rx_signal, tx_waiter_status);
 
         // loop over start wait messages
         for i in 0..100 {
             pacemaker.handle_signal(PacemakerTimer::Start(i, Duration::from_millis(2000)));
             assert_eq!(pacemaker.pending_timeouts.len(), i as usize + 1);
         }
-
-        tokio::time::sleep(Duration::from_millis(1)).await;
 
         // stop waiting messages that are indexed by even numbers, and check that the corresponding
         // pending timeouts are removed from the pending_timeouts hash map
@@ -320,7 +318,6 @@ mod tests {
 
         // assert that timeouts occur and they are removed from pending_timeouts hash map,
         // for each odd number in 0..100
-        tokio::time::sleep(Duration::from_millis(100)).await;
         for i in (0..100).filter(|i| i % 2 == 1) {
             pacemaker.handle_timeout(i, true).await;
             assert_eq!(pacemaker.pending_timeouts.len(), 50 - (i / 2) as usize - 1);
--- a/dan_layer/core/src/workers/pacemaker_worker.rs
+++ b/dan_layer/core/src/workers/pacemaker_worker.rs
@@ -86,6 +86,34 @@ where T: Clone + Debug + PartialEq + Eq + Hash + Send + Sync + 'static
         }
     }
 
+    fn handle_start_signal(&mut self, wait_over: T, duration: Duration) {
+        info!(
+            target: LOG_TARGET,
+            "Received start wait signal for value: {:?}", wait_over
+        );
+        if self.pending_timeouts.contains_key(&wait_over) {
+            info!(
+                target: LOG_TARGET,
+                "Already received an existing wait timer for value = {:?}", wait_over
+            );
+        } else {
+            let (tx, rx_stop_signal) = oneshot::channel();
+            self.pending_timeouts.insert(wait_over.clone(), tx);
+            self.waiting_futures.push(Box::pin(async move {
+                tokio::select! {
+                    _ = tokio::time::sleep(duration) => {
+                        info!("The wait signal for value = {:?} has timeout", wait_over);
+                        (wait_over, true)
+                    },
+                    _ = rx_stop_signal => {
+                        info!("The wait signal for wait_over = {:?} has been shut down", wait_over);
+                        (wait_over, false)
+                    }
+                }
+            }));
+        }
+    }
+
     fn handle_stop_signal(&mut self, wait_over: T) {
         if let Some(signal) = self.pending_timeouts.remove(&wait_over) {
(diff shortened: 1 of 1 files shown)


dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -95,53 +96,61 @@ where T: Clone + Debug + PartialEq + Eq + Hash + Send + Sync + 'static
         }
     }
 
+    fn handle_signal(&mut self, signal: PacemakerTimer<T>) {
+        match signal {
+            PacemakerTimer::Stop(wait_over) => {
+                self.handle_stop_signal(wait_over);
+            },
+            PacemakerTimer::Start(wait_over, duration_timeout) => {
+                info!(
```
Cifko: If you want to go for this refactoring, then factor this ti `handle_start_signal` as well

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -95,53 +96,61 @@ where T: Clone + Debug + PartialEq + Eq + Hash + Send + Sync + 'static
...
+                                (wait_over, false)
+                            }
+                        }
+                    }));
+                }
+            },
+        }
+    }
+
+    async fn handle_timeout(&mut self, wait_over: T, has_timed_out: bool) {
```
Cifko: Change the name, it's not always a timeout as the name suggests. Something like handle_pending_timeout_ended, I'm open to any name that indicates that the timer ended either by signal or timing out.

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -287,6 +296,37 @@ mod tests {
             .is_err());
     }
 
+    #[tokio::test]
+    async fn test_remove_pending_timeouts() {
+        let (_, rx_signal) = channel::<PacemakerTimer<u8>>(10);
+        let (tx_waiter_status, _) = channel::<u8>(10);
+
+        let mut pacemaker = Box::new(Pacemaker::new(rx_signal, tx_waiter_status));
```
sdbondi:
```suggestion
        let mut pacemaker = Pacemaker::new(rx_signal, tx_waiter_status);
```

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -287,6 +296,37 @@ mod tests {
...
+
+        // loop over start wait messages
+        for i in 0..100 {
+            pacemaker.handle_signal(PacemakerTimer::Start(i, Duration::from_millis(2000)));
+            assert_eq!(pacemaker.pending_timeouts.len(), i as usize + 1);
+        }
+
+        tokio::time::sleep(Duration::from_millis(1)).await;
```
sdbondi: Dont need to sleep here

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -287,6 +296,37 @@ mod tests {
...
+        // stop waiting messages that are indexed by even numbers, and check that the corresponding
+        // pending timeouts are removed from the pending_timeouts hash map
+        for i in (0..100).filter(|i| i % 2 == 0) {
+            pacemaker.handle_signal(PacemakerTimer::Stop(i));
+            assert_eq!(pacemaker.pending_timeouts.len(), 100 - (i / 2) as usize - 1);
+        }
+
+        // assert that timeouts occur and they are removed from pending_timeouts hash map,
+        // for each odd number in 0..100
+        tokio::time::sleep(Duration::from_millis(100)).await;
```
sdbondi: Sleep not needed

sdbondi (commented):
LGTM - should be able to remove the sleeps in the tests now

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -287,6 +296,37 @@ mod tests {
...
+        for i in 0..100 {
+            pacemaker.handle_signal(PacemakerTimer::Start(i, Duration::from_millis(2000)));
+            assert_eq!(pacemaker.pending_timeouts.len(), i as usize + 1);
+        }
+
+        tokio::time::sleep(Duration::from_millis(1)).await;
+
+        // stop waiting messages that are indexed by even numbers, and check that the corresponding
+        // pending timeouts are removed from the pending_timeouts hash map
+        for i in (0..100).filter(|i| i % 2 == 0) {
```
Cifko: Maybe use (0..100).step_by(2)

dan_layer/core/src/workers/pacemaker_worker.rs
```diff
@@ -287,6 +296,37 @@ mod tests {
...
+            pacemaker.handle_signal(PacemakerTimer::Stop(i));
+            assert_eq!(pacemaker.pending_timeouts.len(), 100 - (i / 2) as usize - 1);
+        }
+
+        // assert that timeouts occur and they are removed from pending_timeouts hash map,
+        // for each odd number in 0..100
+        tokio::time::sleep(Duration::from_millis(100)).await;
+        for i in (0..100).filter(|i| i % 2 == 1) {
```
Cifko: As above (1..100).step_by(2)

dan_layer/core/src/workers/pacemaker_worker.rs
jorgeantonio21: Ups, this was left from our previous conversation. Addressed 👍

sdbondi (approved):

dan_layer/core/src/workers/pacemaker_worker.rs
jorgeantonio21: Good suggestion !

dan_layer/core/src/workers/pacemaker_worker.rs
jorgeantonio21: We can leave it as `handle_pending_timeout_ended`.

dan_layer/core/src/workers/pacemaker_worker.rs
jorgeantonio21: Cool !

PR #353, opened 2023-02-06
refactor!: remove address from the substate data

sdbondi:
Description
---
Removes redundant address data from substates and removes component address from component header

BREAKING CHANGES:
- Substates schema changed (address removed from data, up variant includes address)
- Sqlite db added field without a new migration

Motivation and Context
---
Address information should not be included in the substate data. The only place this is used was in the template macro, but this usage was removed in #340 

image

How Has This Been Tested?
---
Existing tests and manually


stringhandler (approved):
Happy, but this is a breaking change, correct?

sdbondi:
Yip it is updated the title/description

PR #354, opened 2023-02-06
fix: get consensus context epoch from tree node

sdbondi:
Description
---
Gets the epoch used in payload execution from the node that all validators have agreed on instead of from the epoch manager.

Motivation and Context
---
The epoch may change from what is agreed on in consensus.

How Has This Been Tested?
---
Existing tests


stringhandler (approved):
Yip, otherwise it might change halfway

PR #357, opened 2023-02-07
fix(consensus): avoid task starvation when when executing payload

sdbondi:
Description
---
avoid task starvation when when executing payload

Motivation and Context
---
Execution can take anywhere from 200ms to a few seconds while blocking the thread and any tasks queued for execution on that thread. This PR signals to tokio that this will happen so that tokio requeues other pending tasks on another thread.

This is still far from ideal but _should be_ better until we decide on the concurrency model for consensus.	

How Has This Been Tested?
---
Manually

### DIFF
--- a/dan_layer/core/src/workers/hotstuff_waiter.rs
+++ b/dan_layer/core/src/workers/hotstuff_waiter.rs
@@ -56,6 +56,7 @@ use tokio::{
         broadcast,
         mpsc::{Receiver, Sender},
     },
+    task,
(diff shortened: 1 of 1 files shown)


stringhandler (approved):
Neat

stringhandler:
Good start, as you say. I think it should be spawned, but let's not optimize just yet

sdbondi:
Yeah definitely,  spawn_blocking would require a largish refactor so left it for now

PR #358, opened 2023-02-07
fix: address hotstuff waiter todos pertaining leader failure

jorgeantonio21:
Description
---
We address some of the TODOs left in the `hostuff_waiter.rs` file, regarding leader failure.

Motivation and Context
---

How Has This Been Tested?
---


dan_layer/core/src/workers/hotstuff_waiter.rs
```diff
@@ -149,9 +149,14 @@ pub struct HotStuffWaiter<
...
     current_leader_round: HashMap<(ShardId, PayloadId), u32>,
-    // We have to store if we are in the middle of an election, so that when someone sends a recovery message, we can
-    // responde that we are voting and he should wait.
+    // We have to store if we are in the middle of an election, so that when we receive a recovery message
+    // from another VN, we can answer that we are electing and it should wait.
     election_in_progress: HashSet<(ShardId, PayloadId)>,
+    // We keep track of the foreign leader committee messages, so that we abandon payloads in which
+    // we trigger foreign leader failure, multiple times
+    foreign_leader_failures: HashSet<PacemakerEvents>,
+    // Stores the number of election rounds for a given pacemaker event
```
sdbondi: cleanup: Make these into doc comments `///` and remove alllow(dead_code) attrs where applicable

dan_layer/core/src/workers/hotstuff_waiter.rs
```diff
@@ -607,14 +614,28 @@ where
...
-        // be thrown away.
+        // Check if this is the second time we are triggering a foreign leader request
+        if self
+            .foreign_leader_failures
+            .contains(&PacemakerEvents::ForeignCommittee(epoch, shard_id, payload_id))
+        {
+            // This is the second time we are triggering a foreign leader request, for the same triplet
+            // (epoch, shard_id, payload_id). This means that they probably can't reach consensus, so we throw
+            // away this transaction
+            self.shard_store.with_write_tx(|tx| {
```
sdbondi, line 646: Doing this on the second foreign leader failure seems arbitrary. In other rejections, we end up voting REJECT once the proposals come through. In this case, we assume (because likely at this stage we've voted already) that other local replicas have come to the same conclusion, which may not be the case because e.g our network is bad - so we must still allow for enough other replicas to vote ACCEPT because they received the proposals (I think anyway). 

This is a tricky problem that probably is addressed in the HotStuff/Cerberus papers. Perhaps let's leave this one out for now.

dan_layer/core/src/workers/hotstuff_waiter.rs
```diff
@@ -698,14 +752,30 @@ where
...
         shard_id: ShardId,
         payload_id: PayloadId,
     ) -> Result<(), HotStuffError> {
         // TODO: if there is a byzantine replica, he can always send election in progress. We should count the election
         // rounds. Be aware that the election rounds are based on delta time, and this trigger is 2*delta.
+        let election_rounds = self.elections_rounds.entry((shard_id, payload_id)).or_default();
+        *election_rounds += 1;
+
+        // For now, we hardcode the max num of election rounds as 4, before we cancel the transaction
+        if *election_rounds >= 4 {
```
sdbondi, line 829: A byzantine node may send 4 messages to get the payload rejected. This can currently also happen without shenanigans if `f + 1 >= 4` sent in response to `f + 1`  `MissingProposal`s. We need to check that we sent the MissingProposal in the first place and this is a response to it which is a bit more complex (Req/Resp semantics, keeping sent request ids and checking that the response matches).  

To fail the transaction, we should set the payload result to rejected and notify using `self.publish_event(HotstuffEvent::Failed(...)` as well as abandon the pledges. 

```rust
 let finalize_result = FinalizeResult::reject(
                    payload_id.into_array().into(),
                    RejectReason::ForeignLeaderFailure(...),
                );

                tx.update_payload_result(payload_id, PayloadResult {
                    finalize_result,
                    pledge_hash,
                })?;
```

dan_layer/core/src/workers/hotstuff_waiter.rs
```diff
@@ -661,35 +696,68 @@ where
...
+        let payload = match self.shard_store.create_read_tx()?.get_payload(&payload_id) {
+            Ok(payload) => payload,
+            Err(e) => {
+                info!(
+                    target: LOG_TARGET,
+                    "Payload = {} is missing from node shard store {}", payload_id, e
+                );
+                return Ok(());
+            },
+        };
```
sdbondi: Use the `Optional` trait to change a not found error to an Option, this means that other errors are still handled correctly.

```suggestion
        // If the current replica has not stored the payload, we simply don't do anything else and return
        let payload = match self.shard_store.create_read_tx()?.get_payload(&payload_id).optional()? {
            Some(payload) => payload,
            None => {
                info!(
                    target: LOG_TARGET,
                    "Payload = {} is missing from node shard store", payload_id
                );
                return Ok(());
            },
        };
```

sdbondi (commented):
Some tricky problems, in these cases, IMO it's better not to handle them until we understand them fully esp because we know there are no byzantine nodes so we can focus on getting leader failure to work in the legit case.

dan_layer/core/src/workers/hotstuff_waiter.rs
jorgeantonio21: Definitely, thank you for catching up this !

dan_layer/core/src/workers/hotstuff_waiter.rs
jorgeantonio21: If I understand it correctly, the logic is that if I am not receiving a foreign leader message the pacemaker will trigger after delta time. In that case, I request `f + 1` nodes from that committee the latest QC they have stored, while I start a pacemaker timer for this. If it times out, it means that neither the leader, neither the requested `f + 1` nodes have answered, in which case, they would not be able to progress local consensus (transaction malformed, etc). In this case, we should remove the local substates we have stored, because we are basically rejecting the transaction. @Cifko

Cifko: The substate is not written at this state, is it? In hotstuff, the substate (command) is executed once we are at decide phase. So there is no rollbacking DB. I don't know how the REJECT voting works, but I assume similar to ACCEPT. So there should never be a case when the pacemaker trigger twice, right? Because one of the nodes should send the REJECT. You should double check the code if the reject is send as a proposal to all committees. If so we can be sure that one of the good replicas will send the REJECT. And it that case you can leave it as is (starting the pacemaker again). But make sure the reject is send back as a response.

dan_layer/core/src/workers/hotstuff_waiter.rs
jorgeantonio21: The logic to make this work is more involved than I anticipated. The tricky part is to get the `pledge_hash` to the `PayloadResult`. A possible way is to check local storage contents for each `involved_shard` in the `payload`, as well as the latest `payload_height`. This should be reviewed one more time, I reckon.

dan_layer/core/src/workers/hotstuff_waiter.rs
jorgeantonio21: `Because one of the nodes should send the REJECT. `

Here you mean, a local node for the foreign committee ? So, if the consensus cannot be reached at that foreign committee (malformed transaction, etc), at least one good replica will send REJECT to all other replicas (for each involved shard), including the local node we started with. Is this correct ?

Cifko: I think this is already implemented, because the rejection is in the QC, so when you ask for missing proposal. You will get it from at least one honest replica. And the replica will send you the QC with the rejection on the tx_leader/rx_leader channel. So it should be handled in the `on_receive_proposal` function.

PR #360, opened 2023-02-08
fix: add mmr check for votes

Cifko:
Description
---
Uncomment the mmr check. Fix the index. 

Motivation and Context
---
It's very easy to confuse the leaf index with the node index. Maybe we can wrap one of them in struct to differentiate between them.

How Has This Been Tested?
---
I just run one of the test that were failing because of this check.

### DIFF
--- a/dan_layer/core/src/models/vote_message.rs
+++ b/dan_layer/core/src/models/vote_message.rs
@@ -33,7 +33,7 @@ use tari_dan_common_types::{
     TreeNodeHash,
     ValidatorMetadata,
 };
-use tari_mmr::{common::node_index, MerkleProof};
+use tari_mmr::MerkleProof;
 
 use crate::{services::SigningService, workers::hotstuff_error::HotStuffError, TariDanCoreHashDomain};
 
--- a/dan_layer/core/src/models/vote_message.rs
+++ b/dan_layer/core/src/models/vote_message.rs
@@ -33,7 +33,7 @@ use tari_dan_common_types::{
     TreeNodeHash,
     ValidatorMetadata,
 };
-use tari_mmr::MerkleProof;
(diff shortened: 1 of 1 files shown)


dan_layer/core/src/models/vote_message.rs
```diff
@@ -99,22 +99,22 @@ impl VoteMessage {
...
-        //     log::warn!(
-        //         target: "tari::dan_layer::votemessage",
-        //         "Merkle proof verification failed for validator node {:?} at index {:?} with error: {}",
-        //         hash,
-        //         idx,
-        //         err
-        //     );
-        // }
+        if let Err(err) =
+            merkle_proof.verify::<tari_core::ValidatorNodeMmrHasherBlake256>(&root, &*hash, leaf_node_index as usize)
```
sdbondi: You could use verify_leaf (see description in #295)

jorgeantonio21 (dismissed):
utAck ! I agree that it is a good idea to have proper structures to differentiate between `leaf_node_index` and `leaf_index`

dan_layer/core/src/models/vote_message.rs
Cifko: I see. Done

dan_layer/core/src/models/vote_message.rs
```diff
@@ -33,7 +33,7 @@ use tari_dan_common_types::{
     TreeNodeHash,
     ValidatorMetadata,
 };
-use tari_mmr::MerkleProof;
+use tari_mmr::{common::node_index, MerkleProof};
```
sdbondi: Just the one clippy

```suggestion
use tari_mmr::MerkleProof;
```

sdbondi (approved):

PR #361, opened 2023-02-09
feat: addresses part of left todo's in `hotstuff_waiter.rs`

jorgeantonio21:
Description
---
This is a follow-up to PR #358. We prefer to leave it unchanged as it contains discussions that might be relevant in the future, regarding some possible attacks to the network. Instead, we address less critical left todo's here.

Motivation and Context
---

How Has This Been Tested?
---

### DIFF
--- a/dan_layer/core/src/workers/hotstuff_waiter.rs
+++ b/dan_layer/core/src/workers/hotstuff_waiter.rs
@@ -148,10 +148,10 @@ pub struct HotStuffWaiter<
     /// Network latency
     #[allow(dead_code)]
     network_latency: Duration,
-    // We store what round is for next leader selection, default is 0.
+    /// We store what round is for next leader selection, default is 0.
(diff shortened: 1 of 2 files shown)


dan_layer/core/src/workers/hotstuff_waiter.rs
```diff
@@ -662,18 +666,53 @@ where
...
+                "Missing proposal sent by node = {} which is not processing current payload = {}", from, payload_id
+            );
+            return Ok(());
+        }
+
         let leader_proposals =
             self.shard_store
                 .create_write_tx()?
                 .get_leader_proposals(payload_id, last_height + NodeHeight(1), &[shard_id]);
-        // We know the last payload, so we check if we have higher height
```
sdbondi, line 670: Just noticed that in the above code a read-only transaction could be used.

sdbondi (approved):
Looks good

PR #365, opened 2023-02-13
feat: dial local shard peers on startup

sdbondi:
Description
---
Eagerly dials local shard peers on startup.

Motivation and Context
---
Validator nodes dial seed peers on startup, but they will most commonly interact with their local shard peers.
A node will lazily dial peers as needed as it sends messages. This PR adds a request to dial the local shard peers on startup
to reduce the time it takes to first interact with local shard peers.

How Has This Been Tested?
---
Manually


stringhandler (approved):
Nice. I think the todo can even be removed, as it will sort itself out

PR #366, opened 2023-02-13
fix: crash node if any service exits with an error

sdbondi:
Description
---
Checks all JoinHandles from services for error responses and crashes the node if this is the case.

Motivation and Context
---
If a service exits due to an error, this would previously be logged but the validator node would remain started.
This PR collects the join handles and polls them in the main event loop. If any service exits with an error, the error is propagated and shuts down the node. Services may exit cleanly `Ok(())` early, in most cases this will be because of the shutdown signal triggering.

How Has This Been Tested?
---
Manually. Adding a handle that errors after a few seconds.


stringhandler (approved):
Nice. Would have done it the same way

PR #368, opened 2023-02-14
feat: new indexer application to retrieve substate data

mrnaveira:
Description
---
**Additions**: 
* New application (`applications/tari_indexer`):
    * Exposes a JSON-RPC API for
        * Querying substates based on the substate address and (optionally) a version.
        * Managing the watchlist of addresses to be polled from the dan layer network.
    * Polls the dan network every `dan_layer_scanning_internal` seconds for newer versions of stored substates.
* New client (`applications/tari_indexer_client`) for the JSON-RPC API of the indexer
* New cucumber scenario (and related utils) to test indexers

**Changes**:
* Fixed an inconsistency of the `Display` trait for `NonFungibleAddress`, it now displays the same way it expects when parsing a full substate address.
* Moved the `proto` files and conversions from the `tari_validator_node` app into a new `tari_dan_app_grpc` crate, in similar fashion to the base layer. Both the `tari_validator_node` and the new `tari_indexer` now reuse the same `proto` files.
* Moved common code (`base_layer_scanner`, `epoch_manager_handle`, etc) between validator node and indexer into a new `tari_dan_app_utilities` crate. Both the `tari_validator_node` and the new `tari_indexer` now reuse this new crate.

**Current limitations** (to be addressed in future PRs):
* It does not have a web UI, it only exposes a JSON-RPC interface.
* It does not perform QC validation.
* Currently there is no direct way of linking related substates together, you can only query individual substate addresses. For example you cannot list all the resources on an account, or all the individual NFTs of a non fungible resource. This seems more like a limitation of the network itself. A workaround for accounts could be to decode the hash map of the vaults to get a list of the vaults and query them from the indexer (we could provide a separate JSON-RPC method for accounts that do this ).

**Ideas for features**:
* Retrieve a list of ordered transactions for a substate address
* If we incorporate a dry run processor like in the validator nodes, the indexers could be used for invoking template read only methods with no fee cost.

Motivation and Context
---
Up until now, the only way of access to the second layer substate data was being a registered VN into the specific committee for a shard. We want an application that allow us easy access to the latest version (or a specific version) of a substate, without being a registered VN and regardless of the shard.

This PR creates a new application (`applications/tari_indexer`) that exposes a JSON-RPC method for querying substates based on the substate address and (optionally) a version. If a `null` value is provided for the version, then the indexer will try to find the latest version.

An example of query:
```
curl --location --request POST 'localhost:JSON_RPC_PORT' \
--header 'Content-Type: application/json' \
--data-raw '{
    "jsonrpc": "2.0",
    "method": "get_substate",
    "params": { "address": "component_11a59a06b382b6fb85e00bcf4b22717e56b3df9a2f3ca6e82b8cdc70f2ecced6", "version": null },
    "id": 1
}'
```
It supports all types of substate addresses (components, resources, vaults and NFTs). The returning value is the serialisation of the rust `Substate` struct. If the address (or version) does not exist in the network, the result is an error.

It will try first to get the substate from the database, and only query the network if the substate (or version) is not present. The list of stored substates can be managed in multiple ways:
* Config file's `address_watchlist` property.
* CLI `-a` param
* Dynamically on runtime through the JSON-RPC API (methods `add_address`, `delete_address`, `clear_addresses`, `get_addresses`)

The stored substates will be periodically updated by polling the network every `dan_layer_scanning_internal` seconds for newer versions.

How Has This Been Tested?
---
New cucumber scenario for querying different substate types to an indexer


applications/tari_indexer/proto/dan/common.proto
```diff
@@ -0,0 +1,15 @@
+// Copyright 2022 The Tari Project
+// SPDX-License-Identifier: BSD-3-Clause
+
+syntax = "proto3";
```
stringhandler: Let's rather move the proto files to a single `tari_dan_app_grpc` crate like we do in the base layer

applications/tari_indexer/src/base_layer_scanner.rs
```diff
@@ -0,0 +1,353 @@
...
+    last_scanned_tip: Option<FixedHash>,
+    last_scanned_hash: Option<FixedHash>,
+    next_block_hash: Option<FixedHash>,
+    base_node_client: GrpcBaseNodeClient,
+    epoch_manager: EpochManagerHandle,
+    shutdown: ShutdownSignal,
+    consensus_constants: ConsensusConstants,
+}
+
+impl BaseLayerScanner {
```
stringhandler: This can also move to common crate

stringhandler (dismissed):

sdbondi (dismissed):

stringhandler (approved):

PR #370, opened 2023-02-14
fix: dont dial local shards if epoch manager has not synced

sdbondi:
Description
---
no auto-register when wallet has insufficient funds
Use correct consensus constants when checking for reregistration

Motivation and Context
---
Validator node will error when no funds are available. This PR checks the wallet balance before attempting to submit a transaction.

Depends on: https://github.com/tari-project/tari/pull/5181 and https://github.com/tari-project/tari/pull/5183

How Has This Been Tested?
---
Manually on fresh VN, cucumber tests also test this


stringhandler (dismissed):

sdbondi:
Will keep this as draft until  tari-project/tari#5183 is merged

stringhandler (approved):

PR #371, opened 2023-02-15
fix: allow web UI to use OS-assigned JRPC address

sdbondi:
Description
---
Sends correct listen address for JRPC to web UI, even when OS-assigned

Motivation and Context
---
When the JRPC falls back to an OS-assigned listen address, the web UI uses the configured address which points to some other socket (usually another running VN). This PR fixes this.

How Has This Been Tested?
---
Manually, going to the webui for another VN and clicking register an checking the correct VN registers


stringhandler (approved):
Nice. This was annoying

PR #372, opened 2023-02-15
fix: dont initially register automatically

sdbondi:
Description
---
Validator will not register the first time automatically until user has initiated registration via the UI/JRPC

Motivation and Context
---
Potentially controversial: If auto_register = true, the validator would immediately try to register. If a user has this flag on (perhaps unintentionally) but does not intend to spend significant funds on registration, either because they are not ready, dont have funds, they only want the VN to observe the network, or for any other reason. They should have to explicitly initiate a registration via the UI/CLI and thereafter only let auto-registration take care of ensuring the VN stays registered.

How Has This Been Tested?
---
Manually, auto registration happens after initial has been done in web UI


stringhandler (approved):

stringhandler:
Feels much better. If needed we can create a cli arg or config `force-auto-register` or something

PR #375, opened 2023-02-16
feat: add cucumber for leader failure with two committes

jorgeantonio21:
Description
---
Adds a cucumber test for leader failure with 8 VNs total.

Motivation and Context
---

How Has This Been Tested?
---


sdbondi (dismissed):
Test looks good but I havent got it to pass yet locally. 

I suspect that multi-committee consensus is broken, and it passes when the transaction happens to only involve one committee at a time (just a guess). Still looking at it.

Perhaps we should ignore this one and aim to fix the issues in subsequent PRs

stringhandler:
let's keep it open for now

PR #377, opened 2023-02-16
WIP: schema encoding / return owned values from components

sdbondi:
Description
---

Motivation and Context
---

How Has This Been Tested?
---


sdbondi:
This has been largely solved by using CBOR

PR #385, opened 2023-02-21
refactor: upgrade to diesel 2

sdbondi:
Description
---
Upgrades sqlite implementations to use diesel 2

Motivation and Context
---
Tari base layer recently upgraded to diesel 2. The main change is that transactions need to be mutable. 

How Has This Been Tested?
---
Existing tests pass


applications/tari_dan_wallet_daemon/src/main.rs
```diff
@@ -67,12 +67,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
     let shutdown_signal = shutdown.to_signal();
 
     let store = SqliteWalletStore::try_open(cli.base_dir().join("data/wallet.sqlite"))?;
-    let mut migration_output = BufWriter::new(Vec::new());
-    store.run_migrations_with_output(&mut migration_output)?;
-    let migration_output = migration_output.into_inner()?;
-    if !migration_output.is_empty() {
-        log::info!(target: LOG_TARGET, "{}", String::from_utf8_lossy(&migration_output));
-    }
+    store.run_migrations()?;
```
jorgeantonio21, line 70: Don't we need to handle the migration outputs anymore ?

sdbondi: I don't think having output is necessary really. As long as we get an error if it fails.

jorgeantonio21 (approved):
Nice ! utAck

stringhandler (approved):

PR #388, opened 2023-02-22
refactor: buckets hold confidential assets

sdbondi:
Description
---
Use existing buckets to hold confidential resources

Motivation and Context
---
Buckets holds resources which can be fungible, non-fungible and confidential. This was true with/without the ConfidentialBucket type. Any specific confidential operations are handled on the resource level. 

Invariant: A resource can only hold one type of resource (fungible, non-fungible, confidential)

The reveal/extract operation should either fail for non-confidential or should be a no-op (I think the latter, `reveal(fungible)->fungible`, reveal(confidential) -> fungible|non-fungible - for confidential asset tags, we'll need to figure out what resource type it reveals to)).

Reveal/Extract would likely 1:1 swap to a non-confidential resource.

```rust
         let confidential_coins = ResourceBuilder::confidential(ConfidentialResourceType::Fungible)
                .with_token_symbol("CFD")
                .initial_supply(commitment, commitment_signature)
                .build_bucket();
```

We can have a confidential bucket, but I vote to get further before we decide we need it. Specifically, if there are many differences in the API level and in security, we should use the abstraction.
The API for dealing with confidential assets can be exactly the same as any other asset.

How Has This Been Tested?
---
claim burn cucumber passes


stringhandler (approved):
I was in two minds about reusing the existing bucket. I decided that you may have different APIs for bucket and confidential bucket, but happy to go with this approach

PR #389, opened 2023-02-23
feat!: confidential resource minting

sdbondi:
Description
---
feat: adds resource builder for confidential assets
refactor: move resource builders into their own modules
test: add unit test for confidential resource minting

Motivation and Context
---
Allows a user to create a confidential resource by providing proof of knowledge of a commitment and a bulletproof.

```rust
pub struct ConfidentialProof {
    /// The public mask of the commitment
    pub public_mask: PublicKey,
    /// The commitment being proven
    pub commitment: Commitment,
    /// Proof of knowledge of each element in the commitment (value, mask, asset tag). Currently just a 2-degree 
    /// commitment signature <R, u, v>.
    pub knowledge_proof: ComitmentSignature,
    /// Proof that no elements are negative
    pub range_proof: Vec<u8>,
    pub minimum_value_promise: u64,
}
```

Here are some thoughts on implementation of confidential transfers:

Given account1 owned by $P_1$ (sender) and account2 owned by $P_2$ (receiver)
Transfer transaction manifest:
```rust
// The sender wallet generates the withdraw_proof by first retrieving the receiver account public key 
let withdraw_proof = global!["withdraw_proof"];
let confidential_resource_addr = global!["confidential_resource_addr"];
let bucket = account1.withdraw(confidential_resource_addr, withdraw_proof);
account2.deposit(bucket); // No authorisation required to deposit
```

- A withdraw will require a proof object that shows a balanced withdraw
- $k_1.G + 100.H - (k_2.G + 10.H) - (k_3.G + 90.H) =? (0)$, an encrypted value (via stealth key), fields <`public_stealth_nonce`, `public_stealth_key`> ($K_s = H(r.P_2).G + P_2 = H(p_2.R).G + P_2$), $k_2 = H(r.P_2)$
- The first commitment term is the input, the second the withdraw, and the third the change.
- To facilitate the stealth key, we should consider adding a mandatory public key to the Account component
- Once a `ConfidentialProof` is validated, a bucket is created containing the withdrawn commitment
- The de/inflation invariant holds because a bucket cannot be created without a correct proof.
- The bucket commitment is added to the resource contained in a vault.
- The receiver wallet would need to come online to validate the received funds 
- The receiver wallet could submit a transaction that aggregates the commitments. They update their mask for the resource
- Alternatively, we could hold incoming commitments separately and have the receiver "claim" incoming commitments by aggregating them into the single liquid commitment using a `ConfidentialProof`.

```rust
pub struct WithdrawProof {
   // Output commitment put in bucket
   output_commitment: Commitment,
   // Change commitment returned to the sender vault
   change_commitment: ConfidentialProof, 
   // Proof of knowledge of input, output and change commitments and balance proof
   excess_signature: Signature,
   // Batch range proof for commitments 
   rangeproof: Vec<u8>,
   // Value contained in output commitment, encrypted using the stealth key K_s
   encrypted_value: Vec<u8>,
   // R
   public_stealth_nonce: PublicKey,
   // K_s
   public_stealth_key: PublicKey,
}
```

How Has This Been Tested?
---
Basic unit test for minting.


dan_layer/engine/src/runtime/validation.rs
```diff
@@ -0,0 +1,120 @@
...
+        fn it_is_valid_if_proof_is_valid() {
+            let proof = create_valid_proof(100, 0);
+            validate_confidential_proof(proof).unwrap();
+        }
+
+        #[test]
+        fn it_is_invalid_if_minimum_value_changed() {
+            let mut proof = create_valid_proof(100, 100);
+            proof.minimum_value_promise = 1000;
+            validate_confidential_proof(proof).unwrap_err();
```
hansieodendaal: Intuitively it should fail if `minimum_value_promise > value`, so maybe worthwhile to show `proof.minimum_value_promise = 99;` will also fail.

dan_layer/engine/src/crypto.rs
```diff
@@ -0,0 +1,103 @@
...
+    let public_mask = PublicKey::from_secret_key(mask);
+    let value_mask = PrivateKey::from(value);
+    let nonce_a = PrivateKey::random(&mut OsRng);
+    let nonce_x = PrivateKey::random(&mut OsRng);
+    let factory = CommitmentFactory::default();
+    let public_nonce_agg = factory.commit(&nonce_x, &nonce_a);
+    let commitment = factory.commit_value(mask, value);
+    let challenge = challenges::confidential_commitment_proof(&commitment);
+    let challenge =
+        challenges::commitment_signature_fiat_shamir(&public_mask, public_nonce_agg.as_public_key(), &challenge);
```
hansieodendaal: Why split this in two and not just calculate it once with all the parameters?

Do you need to call `challenges::confidential_commitment_proof(&commitment)` for something other than just calculating the final challenge hash?

dan_layer/engine/src/crypto.rs
```diff
@@ -0,0 +1,103 @@
...
+lazy_static! {
+    static ref RANGE_PROOF_SERVICE: RangeProofService =
+        RangeProofService::init(64, 1, CommitmentFactory::default()).unwrap();
+}
+
+pub fn range_proof_service() -> &'static RangeProofService {
+    RANGE_PROOF_SERVICE.borrow()
+}
+
+pub fn generate_confidential_proof(mask: &PrivateKey, value: u64, minimum_value_promise: u64) -> ConfidentialProof {
```
hansieodendaal, line 28: Is there any set rule or expectation about the `minimum_value_promise`, or is it whatever the owner wants it to be?

dan_layer/engine/src/crypto.rs
```diff
@@ -0,0 +1,103 @@
...
+    let public_mask = PublicKey::from_secret_key(mask);
+    let (signature, commitment) = generate_commitment_knowledge_proof(mask, value);
+    let range_proof = generate_extended_bullet_proof(mask, value, minimum_value_promise);
```
hansieodendaal: The range proof is a zero-knowledge proof about the commitment openings already, so maybe `knowledge_proof` must be optional? What is the use case for `knowledge_proof`?

hansieodendaal: One aspect that may be worthwhile is that `public_mask` (actually just the public excess) is baked into the commitment signature, but I do not know if it has value. The validator nodes would need to verify the balance equation, and it would require the sum of the public excesses anyway as part of the Schnorr signature.

dan_layer/engine/src/crypto.rs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+fn generate_extended_bullet_proof(mask: &PrivateKey, value: u64, minimum_value_promise: u64) -> BulletRangeProof {
+    let extended_mask = RistrettoExtendedMask::assign(ExtensionDegree::DefaultPedersen, vec![mask.clone()]).unwrap();
+    let witnesses = vec![RistrettoExtendedWitness {
+        mask: extended_mask,
+        value,
+        minimum_value_promise,
+    }];
+    let range_proof = range_proof_service().construct_extended_proof(witnesses, None).unwrap();
```
hansieodendaal, line 66: We could easily support rewinding here

hansieodendaal (commented):
Nice progress.

I did not review all of the code, but I made some observations, which may or may not be of value. No need to resolve any of this.

dan_layer/engine/src/crypto.rs
sdbondi: For the mint use case, it makes sense, knowledge proof is not necessary. These are going to change somewhat to account for transfers which involve the balance proof.

dan_layer/engine/src/crypto.rs
sdbondi: Great to hear! I'll leave that out for now until we have the main details in place.

dan_layer/engine/src/crypto.rs
sdbondi: It's what the owner wants, they may want to openly declare that they have minted at least X.

dan_layer/engine/src/crypto.rs
sdbondi: I've removed the proof of knowledge and the public mask/excess (which leaks the commitment value).

The balance proof in transfers only requires an excess sig for the full transfer equation.

CjS77:
> /// Proof that no elements are negative

So random thought.

I briefly thought? What if we wanted to do DeFi loans, wouldn't we need negative numbers?
And the answer is NO.

What you want in this case is **double-entry bookkeeping**.

A loan is an asset to one party and a liability to the other party. In both cases, the value is positive. So requiring BPs to prove positive values is not restrictive to this future use case at all.

Plus double-entry bookkeeping comes with tons of additional benefits and safety features

sdbondi:
@CjS77 Makes sense, could these take the form of aggregate commitments in and out within the resource?

Vault 1: in: (100), out: (40+50)
Vault 1: in: (90), out: (0)

dan_layer/engine/src/crypto.rs
```diff
@@ -0,0 +1,86 @@
...
+//         challenges::confidential_commitment_proof(&public_mask, public_nonce_agg.as_public_key(), &commitment);
+//
+//     let signature =
+//         RistrettoComSig::sign(&value_mask, mask, &nonce_a, &nonce_x, &challenge, &COMMITMENT_FACTORY).unwrap();
+//
+//     (signature, commitment)
+// }
+
+fn generate_extended_bullet_proof(mask: &PrivateKey, value: u64, minimum_value_promise: u64) -> BulletRangeProof {
+    let extended_mask = RistrettoExtendedMask::assign(ExtensionDegree::DefaultPedersen, vec![mask.clone()]).unwrap();
```
jorgeantonio21, line 60: Why do we extend the degree of the mask for the range_proof below? We are using default anyway.

jorgeantonio21 (commented):
Superbe !

dan_layer/engine/src/crypto.rs
sdbondi: We'll be using multi-degree in future for confidential NFTs, likely `<blinding_factor, value, asset tag, asset instance>`

dan_layer/engine/tests/templates/confidential/faucet/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+                .with_token_symbol("🪙")
+                .initial_supply(confidential_proof)
+                .build_bucket();
+
+            Self {
+                vault: Vault::from_bucket(coins),
+            }
+        }
+
+        pub fn take_free_coins(&mut self) -> Bucket {
```
stringhandler, line 45: This seems like it won't work?

dan_layer/engine_types/src/resource_container.rs
```diff
@@ -187,6 +188,20 @@ impl ResourceContainer {
             )),
         }
     }
+
+    pub fn withdraw_confidential(&mut self, _proofs: &[ConfidentialProof]) -> Result<ResourceContainer, ResourceError> {
```
stringhandler, line 192: I think you should only be allowed to withdraw the full commitment and use Split to get change and deposit back

stringhandler (approved):
Happy but see comments

PR #391, opened 2023-02-24
feat: initial implementation of confidential resource transfer

sdbondi:
Description
---
- Adds the ability to withdraw and deposit confidential funds into vaults
- A single aggregate comitment is used in confidential resources

Motivation and Context
---
- A withdraw will require a proof object that shows a balanced withdraw
- $k_1.G + 100.H - (k_2.G + 10.H) - (k_3.G + 90.H) =? (0)$
- TBD: an encrypted value (via stealth key), fields <`public_stealth_nonce`, `public_stealth_key`> ($K_s = H(r.P_2).G + P_2 = H(p_2.R).G + P_2$), $k_2 = H(r.P_2)$
- The first commitment term is the input, the second the withdraw, and the third the change.
- To facilitate the stealth key, we should consider adding a mandatory public key to the Account component
- Once a `ConfidentialWithdrawProof` is validated, a bucket is created containing the withdrawn commitment
- The de/inflation invariant holds because a bucket cannot be created without a correct proof.
- The bucket commitment is added to the resource contained in a vault.
- The receiver wallet would need to come online to validate the received funds 
- The receiver wallet could submit a transaction that aggregates the commitments. They update their mask for the resource
- Alternatively, we could hold incoming commitments separately and have the receiver "claim" incoming commitments by aggregating them into the single liquid commitment using a `ConfidentialWithdrawProof`.

For now we aggregate the commitments to make input selection trivial. This aspect needs to be discussed further and is out of scope for this PR.

How Has This Been Tested?
---
New engine unit tests


hansieodendaal (commented):
I just looked at the extended range proof stuff; it looks good.

Note that should Bulletproof rewinding be required in future, it will not work for aggregation factors > 1.

dan_layer/engine/src/runtime/impl.rs
```diff
@@ -529,7 +540,7 @@ impl RuntimeInterface for RuntimeInterfaceImpl {
             },
             WorkspaceAction::Take => {
                 let key: Vec<u8> = args.get(0)?;
-                let value = self.tracker.take_from_workspace(&key)?;
+                let value = self.tracker.get_from_workspace(&key)?;
```
stringhandler, line 543: This naming could be confusing in future

stringhandler: Perhaps the WorkspaceAction should be renamed to `Get` ?

stringhandler (approved):
Didn't dig too much into the crypto because
1. It will be changed to non aggregated commitments in an upcoming PR
2. A deeper audit of all crypto at a later stage will be more thorough

A brief look over and it looks good.

Made some comments, but they are basically stylistic

dan_layer/engine/src/runtime/impl.rs
sdbondi: Oh yeah, good catch, I just ran into a problem with the way we do workspaces and changed that without changing the action - will update in PR #397

PR #392, opened 2023-02-24
feat: remove wait_for_result option on cli, thus running commands always waits

jorgeantonio21:
Description
---
Removes `wait_for_result` option in the dan wallet cli. Therefore, every execution of it will always for result to complete.

Motivation and Context
---

How Has This Been Tested?
---
Manual testing on the cli


sdbondi (approved):
Thanks, perfect

PR #394, opened 2023-02-24
fix(tari_indexer): replace template manager channel sink with a full mock

mrnaveira:
Description
---
* Created a new mock template manager system for the `base_layer_scanner` that handles all needed requests.
* Added some more descriptive error logging for indexers.

Motivation and Context
---
Recent changes in `tari_dan_app_utilities`, mainly related to channel management, broke the indexer `EpochManager` when detecting a new template in the base layer.

How Has This Been Tested?
---
The cucumber scenario for the indexer is working again

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6521,6 +6521,7 @@ dependencies = [
  "tari_template_lib",
  "tari_transaction",
  "tari_validator_node",
+ "tari_validator_node_client",
  "thiserror",
  "tokio",
  "tokio-stream",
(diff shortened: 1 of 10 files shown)


sdbondi (approved):
Thanks for fixing!

I think the coupling of all these services can be removed from the scanner. If the base layer scanner were architected to tell other services when something important on the base layer happens (new vn reg, new block, new template, new burn utxo) rather than doing the work on them inside the scanner itself. 

But let's discuss that as it's not super critical at this stage.

PR #395, opened 2023-02-24
feat: add indexer ui

Cifko:
Description
---
Add indexer UI. It supports info, monitored substate (add/remove), list of connections.
The indexer UI shows address with version number. For additional info only json of substate is shown as returned from jrpc call.

Motivation and Context
---
The index had only jrpc interface, which is not really user friendly.

How Has This Been Tested?
---
Manually.


sdbondi (dismissed):
LGTM utACK

sdbondi (approved):

PR #396, opened 2023-02-24
feat: new AddressList type that allows substate linking

mrnaveira:
Description
---
**Changes**:
* Created definitions for engine and templates to handle address lists and items, including new address variants.
* New `Address` type In `template_lib` to encapsulate any type of address. I could have opted for moving the `SubstateAddress` definition from engine to `template_lib` but I didn’t want to expose the concept of _substate_ to the template developer.
* New engine call for address lists, with two actions: create new list and push item to an existing list.
* Updated the tracker and runtime to generate the new substates related to lists.
* Updated the transaction builder and CLI clients to allow inputs for new items.

**Safety measures**:
* Lists and list items are append-only so the substates **cannot be mutated**: there cannot be any output with version > 0. We specifically check this situation.
* List items are prevented to reference other list items or list roots, to prevent cycles.

**Caveats**:
* Template developers need to specify the index that they are inserting into the list. They can choose any arbitrary index that is not used already (the substate UP will fail otherwise) so there could be gaps in the list. Ideally the developer should not need to specify the index and it should be automatically determined by the network, but this is a bigger issue outside of the scope of this PR.

Motivation and Context
---
In the context of DAN network scanning and indexing, currently there is no direct way of linking related substates together, you can only query individual substate addresses.

For example you cannot list all individual NFTs of a non fungible resource. The resource address of each individual NFT is identified in a way (random or user-provided) that is not known when scanning. This means that given an NFT resource address, we cannot know the ShardId of all the individual NFT's that have been minted (nor the order).

A naive approach would be to store the whole list of the individual NFT addresses inside the resource substate, but that would not scale well with collections that may have up to millions of minted NFTs. We need a way to create linked lists or maps of substates, that leverages the DAN layer sharding for scalability.

This PR implements a new struct `AddressList` accessible from template code, that provides an **append-only** list of substate addresses:
* There is a new substate for the address list **root**, with address of type `AddressListId`.
* A new substate is created for **each item** added to the list, with address composed of `AddressListId` + `index`. The index starts at 0 and is incremented by 1 for each new item. Each item substate contains the address of the referenced subsatate (e.g. an individual NFT). With this we have a predictable (and ordered) way of scanning all items regardless of the data they contain.

This approach decouples the list structure to the actual data that the list holds:
* It can be reused for any kind of addresable substate (not only NFTs)
* Provides a record of minting/issuing order, which is otherwise very difficult to get in the DAN layer due to it's decentralized sharded implementation.
* Allows for a substate (e.g. a NFT) to be included in many different lists, without modifying the actual data of the substate.

An **important open question** is deciding if the `AddressList` is a type that should be exposed to the template developer (as implemented in this PR) or something purely NFT-specific that should be transparent and automatically done for each minted NFT.

How Has This Been Tested?
---
* New test template (`nfts/nft_list`) for engine unit tests. The associated tests asserts the substates created for both the list and the list items.
* Updated the `basic_nft` template in cucumber integration tests, to hold an address list of minted NFTs.


sdbondi (approved):
LGTM

Will be good to see, in future, NFTs listed in the indexer UI

stringhandler (approved):
Nice. In a future PR I would like to see a restricted push that only can be done at the end of the list

PR #397, opened 2023-02-27
feat: confidential vault/bucket reveal

sdbondi:
Description
---
- feat: confidential vault/bucket reveal
- feat: adds revealed amount to confidential resource
- fix: bug in template macro that prevented private functions from returning certain types e.g. tuples

Motivation and Context
---
Allows a user to provide a withdraw proof that reveals a portion of the commitment in their bucket/vault. This
adds liquidity to a revealed "pool" of funds within the resource which then can be used in the same way as a non-confidential fungible resource. 

```rust
        // reveal_proof is a withdraw proof that specifies X funds should be withdrawn and revealed
        let revealed_funds = account1.reveal_confidential(resource, reveal_proof);
       // The account contains an additional X amount of non-blinded funds 
        account.deposit(revealed_funds); 
```

Balance proof: put 10 revealed and 0 blinded in bucket

$k_{input}.G + 100.H - (k_2.G + 0.H + 10.H) - (k_c.G + 90.H) \stackrel{?}{=} (0)$ 

Note the output commitment can include a non-zero value as long as the change $k_c.G + v_c.H$ balances.

How Has This Been Tested?
---
New unit tests that call vault and bucket reveal_confidential


stringhandler (approved):
Still using the aggregated commitment in the vault, which is already addressed in another pr. 

Otherwise, happy to merge it

PR #398, opened 2023-02-28
fix: port typo

Cifko:
Description
---
Fix port typo. We want the default port for UI to be on port 5000.


sdbondi (approved):

stringhandler:
Lol I preferred it on 15000

sdbondi:
lol :) I agree with that - I was helping someone with weird HTTP denied errors which I thought might be because of port 5000 being in use on their machine.

Cifko:
should we change the VN as well?

stringhandler:
probably, but it should be a different port to the indexer

PR #399, opened 2023-02-28
fix: add additional checks hotstuff checks

Cifko:
Description
---
Add additional checks for hotstuff.
- track who we asked from foreign committee.
- track how many times they sent election in progress.
Added two tests to confirm that hotstuff is currently broken.

How Has This Been Tested?
---
Cargo test.


dan_layer/integration_tests/src/test_consensus.rs
```diff
@@ -644,6 +644,34 @@ async fn test_leader_fails_only_foreignly() {
     instance1.assert_shuts_down_safely().await;
 }
 
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn test_if_hostuff_is_broken() {
```
stringhandler: I lolled at the name, but I think we'll struggle to remember which part of hotstuff this is testing in a few months. Maybe `test_that_hotstuff_reaches_decide`

stringhandler: A very useful and strangely simple test

Cifko: I've changed it to better reflects what's wrong with it.

dan_layer/integration_tests/src/test_consensus.rs
```diff
@@ -644,6 +644,34 @@ async fn test_leader_fails_only_foreignly() {
     instance1.assert_shuts_down_safely().await;
 }
 
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
```
stringhandler, line 638:
```suggestion
#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
#[should_panic]
```

dan_layer/core/src/workers/hotstuff_waiter.rs
```diff
@@ -531,6 +530,10 @@ where
             (node.height() == last_vote_height && node.leader_round() > last_leader_round)) &&
             (*node.parent() == locked_node || node.height() > locked_height)
         {
+            // We did receive something from the leader, so he was elected.
+            self.election_in_progress.remove(&(node.shard(), payload_id));
+            // If this proposal was a result of foreing leader failure, we need to delete the data we no longer need.
```
sdbondi:
```suggestion
            // If this proposal was a result of foreign leader failure, we need to delete the data we no longer need.
```

dan_layer/core/src/workers/hotstuff_waiter.rs
```diff
@@ -675,6 +677,13 @@ where
             .choose_multiple(&mut rand::thread_rng(), how_many_to_ask)
             .cloned()
             .collect();
+        self.asked_foreign_members.insert(
+            (shard_id, payload_id),
+            selected_foreing_committee_members
```
sdbondi:
```suggestion
            selected_foreign_committee_members
```

dan_layer/integration_tests/src/test_consensus.rs
Cifko: Done. Reworked the test little bit to panic.

Cifko:
Added a new broken scenario.

dan_layer/integration_tests/src/test_consensus.rs
```diff
@@ -644,22 +644,97 @@ async fn test_leader_fails_only_foreignly() {
...
+    ] {
+        let proposals = test.get_proposals(0).await;
+        test.send_all_proposals_to_all_committees(proposals, 0).await;
+        if phase != HotstuffPhase::Decide {
+            let votes = test.get_all_votes(0).await;
+            test.all_leaders_receive_votes(votes, 0).await;
+        }
+    }
+    test.send_tx_new_to_committee(0, 0).await;
+    // When the VNs receive the same transacation, which is already done. The leader gets the last proposal (decide
```
sdbondi:
```suggestion
    // When the VNs receive the same transaction, which is already done. The leader gets the last proposal (decide
```

dan_layer/integration_tests/src/test_consensus.rs
```diff
@@ -796,6 +883,21 @@ impl Committee {
...
+        let hs_message = join_all(
+            self.instances
+                .iter_mut()
+                .map(|instance| timeout(*TEN_SECONDS, instance.rx_leader.recv()))
+                .collect::<Vec<_>>(),
+        )
+        .await;
+        assert!(
+            hs_message.iter().all(Result::is_err),
+            "No message should be send!\n{:?}",
```
sdbondi:
```suggestion
            "No leader messages should have been sent!\n{:?}",
```

sdbondi (commented):
Might be naive, but would we not be able to simply check that the high QC for the duplicate transaction is already committed and not send a NEWVIEW?

```rust
// in on_next_sync_view
if new_view.high_qc().unwrap().payload_height() == NodeHeight(3) {
            warn!("🔥 Transaction {} has already been committed. Ignoring", payload_id);
            return Ok(());
        }
```

Cifko:
> Might be naive, but would we not be able to simply check that the high QC for the duplicate transaction is already committed and not send a NEWVIEW?
> 
> ```rust
> // in on_next_sync_view
> if new_view.high_qc().unwrap().payload_height() == NodeHeight(3) {
>             warn!("🔥 Transaction {} has already been committed. Ignoring", payload_id);
>             return Ok(());
>         }
> ```

It could work, but I didn't change it yet because, I don't understand why do we read the qc from the DB on `on_next_sync_view`. IMHO a single payload_id should never trigger `on_next_sync_view` twice.  And I don't even  know what will happen if someone send the first transaction more than one time. So the `on_next_sync_view` will trigger before we have anything in the DB. (This can be tested). But IMHO we should have an in-memory hashset of currently processed transacation, so we can ignore this double transacation.

sdbondi:
I've manually tested this case, well sort of, when sending a transaction with the same pledges. One or both transactions get rejected (race condition). For the exact same transaction i.e. same hash, the mempool already handles this by not sending through to consensus. Obviously, in the unit tests we dont have a mempool.

EDIT: it is still possible for this case to occur because the transaction could be broadcast after it's finalized. We could add a "recent transactions" cache in the mempool to mitigate transactions being passed through, however unless that cache is persistent and unlimited in size it's not guaranteed to prevent this. 

In the end, the authority is the consensus database and so I think we should handle this within the consensus handlers.

PR #403, opened 2023-02-28
feat: add rpc.discover methods to indexer and wallet daemon

stringhandler:
Description
---
Adds an `rpc.discover` endpoint which returns the openrpc.json specification of the jsonrpc

Motivation and Context
---
This is the standard approach for providing JSONRPC via discovery. I haven't found a working client that utilizes it, but I was trying to get it to work with https://playground.open-rpc.org/ but no luck. According to the openrpc spec, I believe this is the correct way of doing it.

As a side note, we also use the examples section in the openrpc.json file for the validator node to create mock servers.

How Has This Been Tested?
---
Tested via postman


sdbondi (dismissed):
Nice

applications/tari_dan_wallet_daemon/src/jrpc_server.rs
```diff
@@ -46,6 +46,17 @@ pub async fn listen(
 
 async fn handler(Extension(context): Extension<Arc<HandlerContext>>, value: JsonRpcExtractor) -> JrpcResult {
     info!(target: LOG_TARGET, "🌐 JSON-RPC request: {}", value.method);
+    if value.method == "rpc.discover" {
+        return Ok(JsonRpcResponse::success(
+            value.id,
+            serde_json::from_str::<HashMap<String, Value>>(include_str!("../openrpc.json")).map_err(|e| {
```
sdbondi, line 52: Think this can just be a Value
```suggestion
            serde_json::from_str::<Value>(include_str!("../openrpc.json")).map_err(|e| {
```

stringhandler: Nah, I tried that, but it automatically chooses an int or something and fails on deserialize

stringhandler: Basically got this error https://stackoverflow.com/questions/48595735/invalid-type-map-expected-a-sequence-when-deserializing-a-nested-json-struct

sdbondi: Very strange

sdbondi (approved):

applications/tari_dan_wallet_daemon/src/jrpc_server.rs
```diff
@@ -46,6 +46,17 @@ pub async fn listen(
 
 async fn handler(Extension(context): Extension<Arc<HandlerContext>>, value: JsonRpcExtractor) -> JrpcResult {
     info!(target: LOG_TARGET, "🌐 JSON-RPC request: {}", value.method);
+    if value.method == "rpc.discover" {
```
sdbondi, line 49: We use the `.` now so you could add this to the handlers below - I dont mind either way

PR #406, opened 2023-02-28
fix: consistent log targets, validation node output in cucumber

sdbondi:
and wait for block height in cucumber tests

Description
---
- Uses `tari::` prefix for all log targets
- Uses `tari::{app_name}` for application logs
- Uses `tari::dan` for dan layer crate logs
- Updates log4rs samples
- Updates cucumber log4rs to include validator node output
- Updates cucumber log4rs to reduce the hotstuff waiter output to WARN, ERROR 
- Adds a check for validator node reaching block height
- get epoch stats returns last scanned block height
- used correctly namespaced keys for epoch manager metadata entries

Motivation and Context
---
Validator node logs were not emitted in CI for cucumber. This PR reduces the base layer logs from debug to warn and includes all validator node logs. Hotstuff waiter logs are reduced to warn and up because they are very noisy. We should probably reduce a lot of hotstuff waiter logs to debug.

This allows us to see any errors in CI more clearly.

Also added a step that checks that the validator nodes have scanned up to the expected block height.
`    Then VN has scanned to height 18 within 10 seconds`

How Has This Been Tested?
---
Cucumbers pass


stringhandler (approved):

sdbondi:
Woah, hold on, looks like the tests didnt run lol

PR #409, opened 2023-03-01
fix: only copy compiled builtin templates to dest if changed

sdbondi:
Description
---
Skips copying builtin templates if their contents are the same

Motivation and Context
---
Often account.wasm is marked as changed in git unnecessarily because it was written even though contents were exactly the same. This leads to a lot of conflicts. It seems that deterministic builds were in fact working and all we needed to do was not overwrite the file if unchanged.

How Has This Been Tested?
---
Manually: making changes to account template that don't change the generated code (e.g. add a newline) and checking that the output file is not written.


jorgeantonio21 (approved):
Thanks, much appreciated !

PR #410, opened 2023-03-01
feat: template functions on new page

NovaT82:
Description
---
Moved template functions to a separate page.
Also added some general UI improvements, like on the connections "Add peer" buttons.
Changed the copy to clipboard component, so the hover state and confirmation message both display in the tooltip.

Motivation and Context
---
The template functions are currently displaying in a hover, making them difficult to read.

How Has This Been Tested?
---
Manually


sdbondi (approved):
ACK

Thanks for fixing the add peers form :)

PR #411, opened 2023-03-01
refactor: address lists are now non fungible indexes

mrnaveira:
Description
---
* The runtime tracker now automatically creates a new substate for NFT indexes every time a new NFT is minted
* Removing list functionality from the `template_lib` crate.
* Removing engine calls for list management.
* Renaming "address list" to "non fungible indexes" in all related definitions

Motivation and Context
---
On https://github.com/tari-project/tari-dan/pull/396 we introduced a new type of substate to keep an ordered list of other substates. It was mainly used to allow NFT collection scanning, as well as keep a record of minting order.

The list management was left to the user, but now we want it to be automatically done on every NFT mint. Also, the list is going to be specific for NFTs, as other types of substates can theoretically be tracked if the address emission is predictable.

So the goal of this PR is to refactor the address list functionality to:
* Make it part of the protocol, so every new NFT mint will also create a index substate, without the user (template developer) knowing.
* Make it specific to NFTs, instead of a general list of addresses, as it's very unlikely that we are going to use it for other purposes.

How Has This Been Tested?
---
The existing unit and integration tests for NFT indexes pass


applications/tari_validator_node/tests/features/indexer.feature
```diff
@@ -44,7 +44,6 @@ Feature: Indexer node
     When I submit a transaction manifest on VN with inputs "NFT, ACC1" and 4 outputs named "TX2"
         ```
             // $mint NFT/resources/0 1
-            // $list_item NFT/addresslists/0 0
```
sdbondi, line 47: Looks like a missing nft index output here.

`// $nft_index NFT/resources/0 0`

sdbondi (commented):
Looks good!

stringhandler (approved):
Tested but it's really difficult to manually call this from the cli. Nevertheless, I'm happy with the code. We just need to find a way to make all transactions easier to call manually

PR #413, opened 2023-03-02
feat: adds public nonce and encrypted value data to confidential resources

sdbondi:
Description
---
- adds `public nonce` and `encrypted value` data to confidential resources
- adds initial outputs table for SQLite implementation of wallet storage
- the bulletproof has been removed from the committed state (resource container)

Motivation and Context
---
In order for a sender to generate a commitment where the mask is only known by the receiver, we plan to implement a stealth address scheme. This requires the sender to generate a nonce used in a Diffie-Helman exchange as well as provide the value encrypted using another Diffie-Helman key.

Mask and encrypted value key derivation are identical except for the hash domain:
Sender: $K = H(r.P)G$, Output commitment: $K + v.H$, Reciever: $k = H(R.p)$
Note: the sender has to know the private mask to generate the withdrawal proof, however once spent into the receiver's account the sender cannot spend the receiver's funds as this is protected by account access rules.

The bulletproof is currently aggregated for the output and change commitments. I don't think that it can be verified given only the output or change commitment, so for now, once it is verified it is discarded from the actual substate. This may change in future if this is deemed incorrect.

How Has This Been Tested?
---
Existing tests to some degree, this is mostly a structural change.


dan_layer/template_lib/src/models/confidential_proof.rs
```diff
@@ -3,26 +3,36 @@
...
     pub revealed_amount: Amount,
 }
 
-#[derive(Debug, Clone, Default, Encode, Decode)]
+#[derive(Debug, Clone, Encode, Decode)]
 pub struct ConfidentialStatement {
     pub commitment: [u8; 32],
+    /// Public nonce (R) that was used to generate the commitment mask
+    pub sender_public_nonce: RistrettoPublicKeyBytes,
+    pub encrypted_value: EncryptedValue,
```
jorgeantonio21, line 26: A doc explaining the use of `encrypted_value` for the future, would be great

dan_layer/template_lib/src/models/component.rs
```diff
@@ -96,6 +96,7 @@ pub struct ComponentHeader {
     pub access_rules: AccessRules,
     // TODO: Split the state from the header
     pub state: ComponentBody,
+    // pub associated_public_key: Option<RistrettoPublicKeyBytes>,
```
jorgeantonio21: Will we need this, or can it be removed ?

dan_layer/template_lib/src/models/component.rs
```diff
@@ -106,6 +107,10 @@ impl ComponentHeader {
     pub fn state(&self) -> &[u8] {
         &self.state.state
     }
+
+    // pub fn associated_public_key(&self) -> Option<&RistrettoPublicKeyBytes> {
```
jorgeantonio21: Could it be removed ?

dan_layer/template_builtin/templates/account/src/lib.rs
```diff
@@ -93,17 +93,6 @@ mod account_template {
             vault_mut.deposit(bucket);
         }
 
-        // #[access_rules(allow_all)]
-        // pub fn deposit_confidential(&mut self, bucket: Bucket, proof: ConfidentialProof) {
```
jorgeantonio21, line 97: Currently, we don't have `deposit_confidential` functionality (which I presume we want in the future). This means that every transfer of a confidential asset, using `withdraw_confidential`, will make disclose some information about the asset value/amount on `deposit`. Also, it seems that confidential buckets can be only created but not transferred (really, i.e., maintaining confidentiality). But again, I am presuming we will add this functionality in the future ?

jorgeantonio21 (commented):
utAck ! 

Left some minor comments.

dan_layer/template_builtin/templates/account/src/lib.rs
sdbondi: You can deposit a bucket with`deposit` whether it contains confidential resources or not. No need for a separate function.

dan_layer/template_lib/src/models/component.rs
sdbondi: Not sure, still have to think about how to associate a public key used for shard DH key derivation with an account component. Will remove for now.

jorgeantonio21 (approved):

PR #416, opened 2023-03-02
feat: add brief explanation in readme to build the web ui on first build

jorgeantonio21:
Description
---
Update readme for first time one builds of the library (which depends on building the web ui, first).

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,10 @@ Network, or DAN.
 
 You can read about the technical specifications of the DAN in the [RFCs](https://rfc.tari.com).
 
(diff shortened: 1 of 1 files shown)


Cifko:
This is not necessary. On the fresh new compile it will run the `npm install` and `npm run build` for you. That's why we have the `build.rs`. The problem is that if you delete the build directory in the ui. The cargo:rerun will not pick that up.

PR #420, opened 2023-03-06
feat(indexer): track all NFT's of a resource

mrnaveira:
Description
---
* Added a new SQLite table to keep track of NFT indexes. The substates of the NFTs themselves are added to the existing `substates` table, so the last version is automatically updated on each poll.
* The DAN layer scanner now exposes a method to get all the NFTs of a resource. It queries from index 0 onwards until a index does not exist, and for each item it gets the latest substate of the NFT.
* The substate manager has some specific handling logic when a NFT resource address is added for tracking, so it also stores in database all the indexes and the substates of the individual NFTs. It also exposes methods for retrieving indexed NFTs from the database.
* The JSON-RPC API has new methods `get_non_fungible_count` and `get_non_fungibles`.

Motivation and Context
---
On https://github.com/tari-project/tari-dan/pull/411 we introduced a new substate type for keeping track of minted NFT indexes in the network. Continuing that work, the purpose of this PR is to update the `tari_indexer` application to expose new JSON-RPC methods for retrieving NFT lists.

Note that we don't need a new method to specify that the Indexer needs to track all NFTs of a resource, the existing `add_address` method will simply do it automatically when the requested address is a NFT resource address.

More specifically, we want **two new JSON-RPC methods** in the indexer that allow us to build a paginated UI listing of NFTs:

1. Get **total number of NFTs** in a resource:
```
{
    "jsonrpc": "2.0",
    "method": "get_non_fungible_count",
    "params": {
        "address": "resource_00e6c4a9e4b31fec0d0f98cb0af8e435b803b248ca159072a2e809b9d3c4bf41"
    },
    "id": 1
}
```
The result will be something like:
```
{
    "jsonrpc": "2.0",
    "result": 2,
    "id": 1
}
```

2. Get the **NFT substates between two indexes** of a resource:
```
{
    "jsonrpc": "2.0",
    "method": "get_non_fungibles",
    "params": { 
        "address": "resource_00e6c4a9e4b31fec0d0f98cb0af8e435b803b248ca159072a2e809b9d3c4bf41",
        "start_index": 0,
        "end_index": 9
    },
    "id": 1
}
```
The result will be something like:
```
{
    "jsonrpc": "2.0",
    "result": [
        {
            "address": { // substate address of the NFT},
            "index": 0,  // absolute position of the NFT in the resource
            "substate": { // substate object of the NFT, similar to the get_substate method response}
        },
        // ...rest of NFTs
    ],
    "id": 1
}
```

How Has This Been Tested?
---
The cucumber scenario for the Indexer application has been updated, so we assert that it returns all NFTs of a resource.

---
Manually:
* Run an indexer connected to a VN
* Deploy a template that creates a new `NonFungible` resource and allows to mint NFTs (e.g. the `basic_nft` test template)
* Initialize a new component of the NFT template, write down the new resource address.
* Add the resource address of the previous step to the list of tracked address in the Indexer (calling the JSON-RPC `add_address` method)
* Mint some NFTs (`CallMethod` to the NFT template)
* Wait a few seconds for the indexer to scan and update the NFT indexes and then call the new JSON-RPC methods `get_non_fungible_count` and `get_non_fungibles`. The newly minted NFTs should be in the result

Or by running and inspecting the `indexer.feature` cucumber scenario.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


applications/tari_indexer/src/dan_layer_scanner.rs
```diff
@@ -50,15 +51,74 @@ impl DanLayerScanner {
...
+
+        loop {
+            // build the address of the nft index substate
+            let index_address = NonFungibleIndexAddress::new(*resource_address, index);
+            let index_substate_address = SubstateAddress::NonFungibleIndex(index_address);
+
+            // get the nft index substate from the network
+            // nft index substates are immutable, so they are always on version 0
+            let index_substate_result = self
+                .get_specific_substate_from_commitee(&index_substate_address, 0, epoch)
```
sdbondi:
```suggestion
                .get_specific_substate_from_committee(&index_substate_address, 0, epoch)
```

applications/tari_indexer/src/dan_layer_scanner.rs
```diff
@@ -50,15 +51,74 @@ impl DanLayerScanner {
...
+
+            // now that we have the index substate, we need the latest substate of the referenced nft
+            let nft_address = match index_substate.into_non_fungible_index() {
+                Some(idx) => idx.referenced_address().clone(),
+                // the protocol should never produce this scenario, we stop querying for more indexes if it happens
+                None => break,
+            };
+            let nft_substate_address = SubstateAddress::NonFungible(nft_address);
+            match self
+                .get_latest_substate_from_commitee(&nft_substate_address, epoch)
```
sdbondi:
```suggestion
                .get_latest_substate_from_committee(&nft_substate_address, epoch)
```

applications/tari_indexer/src/dan_layer_scanner.rs
```diff
@@ -50,15 +51,74 @@ impl DanLayerScanner {
...
-        let epoch = match self.epoch_manager.current_epoch().await {
-            Ok(epoch) => epoch,
-            Err(e) => {
-                error!(target: LOG_TARGET, "Could not retrieve the current epoch: {} ", e);
-                return None;
-            },
+        let epoch = match self.get_current_epoch().await {
+            Some(epoch) => epoch,
+            None => return None,
         };
```
sdbondi: nit: can return None this way
```suggestion
        let epoch = self.get_current_epoch().await?;
```

sdbondi (dismissed):
utACK

sdbondi (approved):

PR #421, opened 2023-03-07
fix: fix claim commsig, allow JSON to be pasted on cli

sdbondi:
Description
---
- update claim proof protobuf representation to use CommitmentSig proto type
- allow user to paste json into claim-burn cli
- fix crash if user given malformed proofs
- updated cucumber

Motivation and Context
---
Claim burn works

How Has This Been Tested?
---
Manually

---
Burn funds on L1 using POSTman
Copy JSON from result
run: `cargo run --bin tari_dan_wallet_cli --  accounts get-balances --account-name primary`
Paste JSON

Breaking Changes
---

- [ ] None
- [x] Other - instruction changed. Transaction message not compatible with previous VNs


sdbondi:
Decided to lock tari to 0.47.0-pre-0 , otherwise I'd need to implement the BalancedMerkleTree in this PR

jorgeantonio21 (approved):
LGTM

PR #427, opened 2023-03-10
feat(claim_burn)!: adds encrypted value to UnclaimedConfidentialOutput substate

sdbondi:
Description
---
- renames LayerOneCommitment substate to UnclaimedConfidentialOutput
- adds the L1 wallet reciprocal burn public key to the claim burn instruction and includes that in the claimed resource
- improves wallet cli claim-burn instruction output
- fixes wallet cli claim-burn instruction by adding correct inputs and outputs

Motivation and Context
---
Depends on https://github.com/tari-project/tari/pull/5238
Ref https://github.com/tari-project/tari-dan/pull/425 - this PR will need to be updated to use the correct hasher (can be merged in any order though)

How Has This Been Tested?
---
Claim burn cucumber

What process can a PR reviewer use to test or verify this change?
---
Use POSTMan to burn funds (using wallet with https://github.com/tari-project/tari/pull/5238) 
Run vn and wallet daemon
```shell
$ cargo run --bin tari_dan_wallet_cli -- accounts create --name primary
$ cargo run --bin tari_dan_wallet_cli -- accounts claim-burn --name primary
<paste output from POSTman>
cargo run --bin tari_dan_wallet_cli -- accounts get-balances  --name primary
Checking balances for account 'primary'...
Account component_d74cd546e92985fd0ba5355233d53e963230e2470062155e13ab842a6d3bb991 balances:

Resource                                                                  | Balance
------------------------------------------------------------------------- | -------
resource_0101010101010101010101010101010101010101010101010101010101010101 | 0
resource_69e77a7104fcc9b8daf5a1125699edee08fe435565692c488134e53d594be57f | 6000

2 row(s)
```

Breaking Changes
---

- [ ] None
- [x] Requires data directory to be deleted
- [ ] Other - Please specify

PR #425, opened 2023-03-08
feat(wallet)!: monitor accounts for received funds

sdbondi:
Description
---
- adds monitor that periodically monitors accounts, validates commitments (if required) and updates balances
- adds `get_substate` call to validator node that returns the SubstateStatus::{Up, Down, NotExist}

Motivation and Context
---
When a wallet starts up or receives funds it needs to refresh its account balance.

How Has This Been Tested?
---
Manually with non-confidential assets. Checked that the balances of the vaults update.

```
18:55 INFO  Transaction 2496d255542e9a0a57db0ff92418a27a6092e9699d24c55dd57d319c933d8cfb has been finalized: Accepted
18:55 INFO  👁️‍🗨️ vault vault_e59cdccbd3f5d23e7490731f99699d907f1493a593d9a1d8aa2b7d0b05d17c79 in account component_dbabe540bc33e0a6b542f993a9f12fcc36c1a229a2a538990a6adef5a4e1406c has new balance 5790
18:55 INFO  👁️‍🗨️ vault vault_94257eb8b5116016bcbeb083e543bfb24fdbe8f1a36ab2c22370a9b53a5a4578 in account component_dbabe540bc33e0a6b542f993a9f12fcc36c1a229a2a538990a6adef5a4e1406c has new balance 210
```

What process can a PR reviewer use to test or verify this change?
---
Start a VN and Wallet daemon
```shell
cargo run --bin tari_dan_wallet_cli --  accounts create --name primary   
cargo run --bin tari_dan_wallet_cli --  accounts create --name alt
# Mint faucet
cargo run --bin tari_dan_wallet_cli -- transactions submit -n3 call-function c11278c1837a5cbaac07285ade9ef5c16b98a6cd127c0075ab073f8fdbab20c7 mint -a100000
# see manifest
cargo run --bin tari_dan_wallet_cli -- transactions submit-manifest -n1 ./manifest.rs -g account=$ACCOUNT -g faucet=$FAUCET   

# send from one account to another
cargo run --bin tari_dan_wallet_cli --  transactions send -n0 -i $RESX primary 10 $RESX component_ddea8abc10e75d67a132a35dd52fed5675cf93f7c52aee78d3ec9780c5a18d18
```

Manifest
```rust
     //take faucet tokens and deposit in account
    let faucet = var!["faucet"];
    let coins = faucet.take_free_coins();
    let account = var!["account"];
    account.deposit(coins);
```

Breaking Changes
---

- [ ] None
[...]

PR #429, opened 2023-03-13
feat(cli): generate output proof for minting, support file blob args

sdbondi:
Description
---
feat: add endpoint for generating output proofs for minting
feat: support base64 blob in cli args and manifest globals

Motivation and Context
---
When minting a confidential resource, an output commitment proof is required. This PR adds cli tools to support this.

The key used for the proof is from the key manager but we wont know which key to use for the transfer proof.
This is out of scope and requires PRs (#427 #425) to be merged before this can be worked on.

We may need the walletd to associate a key for all owned vaults regardless if they are part of an Account component so that transfer proofs can be generated. 

How Has This Been Tested?
---
Manually, `ConfidentialFaucet::mint` template call.

---
```shell
$ cargo run --bin tari_dan_wallet_cli --  proofs generate 10000 -o base64 > ./proof.txt   
$ cargo run --bin tari_dan_wallet_cli -- transactions submit-manifest -n3 ./manifest.rs -g proof=file://./proof.txt
```

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


applications/tari_dan_wallet_cli/src/command/transaction.rs
```diff
@@ -660,14 +660,20 @@ pub enum CliArg {
...
     NonFungibleId(NonFungibleId),
     SubstateAddress(SubstateAddress),
 }
 
 impl FromStr for CliArg {
-    type Err = String;
+    type Err = anyhow::Error;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if let Some(file) = s.strip_prefix('@') {
```
stringhandler, line 679: why not use a URL syntax instead:

`data:[<media type>][;base64],<data>`

and if it doesn't start with `data` then assume file path

stringhandler (dismissed):

applications/tari_dan_wallet_cli/src/command/transaction.rs
sdbondi: Yeah sounds good, was using cURL syntax but think this is better. I'll use `data:` `file:` urls and that lets us add support for http etc later (though doubt we'll ever really need that)

stringhandler (dismissed):

stringhandler (approved):

PR #430, opened 2023-03-13
refactor!: use CBOR as binary encoding

mrnaveira:
Description
---
- Replaced `borsh` with `ciborium` for binary encoding in the DAN layer. As `ciborium` depends on `serde`, added serde's `Serialize` and `Deserialize` derives to all types.
- We still use `borsh` on the `engine` crate, but only so we can match the base layer hashes in burn claims

Motivation and Context
---
Up until now we have been using `borsh` for binary encoding of DAN layer data (substates, etc.). The problem is that borsh is not a self-describing format, so there is no easy way to decode binary data from outside the network or outside of the WASM runtime.

We decided to switch to CBOR binary encoding, as it's a self-describing format. This PR aims to refactor all the binary encoding of the DAN layer to use CBOR based on the `ciborium` Rust crate. Ciborium is the easiest to implement CBOR crate, but has the downside of depending on `serde`, so the WASM sizes are increased (e.g. the account template grows from 280Kb to 460Kb). 

**TBD in future PRs:**
- Adding tags for template types that we want to recognise (VaultId, ResourceAddress, etc)
- (@sdbondi suggestion) Add a version prefix. The encode functions can prepend a pre-defined version, and the decode can check the version or error. The version should be useful if we decide to change to packed encoding later.
- Update the indexer to display NFT content by inspecting the binary encoding

How Has This Been Tested?
---
The existing unit and integration tests pass. The only exception is `test_dodgy_template` which was ignored because writing a low level invalid template is a lot harder after the change.

---
This change affects all use cases. To verify the change a reviewer can:
- manually spawn a new network, deploy some templates and call methods on them that update the state
- perform a burn claim

Breaking Changes
---

- [ ] None
- [x] Other - Breaks at all levels, all existing data must be deleted and VNs will not be able to sync with older versions prior to this change

BREAKING CHANGE: Breaks at all levels, all existing data must be deleted and VNs will not be able to sync with older versions prior to this change


sdbondi:
Looks good so far, we should be able to optimise template size later. 
We'll also need to define type tags for each type we're interested in but can be left for another PR.
I'd like to have a version prefix in this PR is possible. Think that the `encode` functions can prepend a pre-defined version, and the `decode` can check the version or error. Version should be useful if we decide to change to packed encoding later.

sdbondi (approved):

PR #424, opened 2023-03-08
feat: update to tari newest tari version

jorgeantonio21:
Description
---
Updates the Tari DAN repo with the latest Tari repo tag.

Motivation and Context
---

How Has This Been Tested?
---
Run `cargo build`.

What process can a PR reviewer use to test or verify this change?
---

Breaking Changes
---

- [x] None
- [ ] Requires data directory to be deleted
- [ ] Other - Please specify

PR #438, opened 2023-03-21
feat: introduce identity signature peer addresses from another node

jorgeantonio21:
Description
---
Refactors `Peer` validation using multiple addresses `identity_signature`.

Motivation and Context
---
Due to the recent changes in the Tari repo, network `Peer`'s allow multiple addresses, each having its own `identity_signature`. This implies that peer validation (and ultimately, its source) had to be refactored. On a first approach we set it to `Config`, everywhere, in #424. This PR tries to make the process more robust and accept `DanPeer` validation/source from other `DanPeer`s, instead.

How Has This Been Tested?
---

---
Run at least two validator nodes via the cli and check communication between these.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (approved):
utACK

PR #439, opened 2023-03-22
feat!: validator node fees

sdbondi:
Description
---
- feat: adds `FeeModule` to engine to calculate fees for a transaction
- feat: fees account for wasm size per kb, number of system invocations and storage 
- feat: adds fee instruction set that will execute and commit (if possible) even if the main transaction fails 
- feat: adds ability to reveal straight from `ClaimBurn`
- feat: pay fees via confidential/revealed from a vault
- feat: unused fees are refunded to vaults
- feat: adds small fee leeway for creating accounts
- feat: adds simple fee transaction validator to mempool
- fix: correct structure for confidential transfer proofs to allow for output or change reveal.
- feat: implement confidential reveal daemon endpoint and command for wallet cli
- feat: changed return structure of payload processor to included information about the transaction finalization, main transaction failure, fee breakdown, and resource containing the fees (currently burnt)
- feat: update state db tables to include fee instructions
- feat: allow account to be created with a n existing bucket
- feat: minor error message improvements
- feat: allow fees to be disabled in validators
- feat: `args![Workspace("...")]` is now valid syntax
- feat: improves ergonomics of the transaction builder
- feat: update wallet daemon to account for fees
- fix: dont require serde dependency to use template macro

---

- tests: add faucet to test runner that allows fees to easily be paid
- tests: add extensive unit tests for fee functionality
- tests: disable fees in cucumber for now

Motivation and Context
---

This PR adds functionality to allow validator nodes to recover fees for running transactions. 

Caveats:
- Fees are currently burnt. A solution for VN fee claiming is left for another PR.
- Account creation is currently "free". The engine "loans" minimal fees to allow an account to be created without fee instructions. All other transactions require fee instructions (validated in the mempool). In future, there may be a way to claim into a new account.
- Wallet commands do not currently support paying fees from confidential reveal proofs. Fees have to be revealed before payment (either in the same transaction or beforehand).

How Has This Been Tested?
---
Extensive engine unit tests
Manually

---
```shell
cargo run --bin tari_dan_wallet_cli --  accounts create --name primary
# Burn, mine and claim
cargo run --bin tari_dan_wallet_cli --  accounts claim-burn -n primary
# Reveal some funds used for fees
cargo run --bin tari_dan_wallet_cli --  accounts reveal primary 5000 --pay-from-reveal
cargo run --bin tari_dan_wallet_cli --  accounts get-balances primary
cargo run --bin tari_dan_wallet_cli --  accounts create --name alt
cargo run --bin tari_dan_wallet_cli --  transactions confidential-transfer -n1 primary 1000 $dest_account  $dest_pk
cargo run --bin tari_dan_wallet_cli --  accounts get-balances alt
# invoke costs fees
cargo run --bin tari_dan_wallet_cli --  accounts invoke -n primary get_balances
cargo run --bin tari_dan_wallet_cli --  accounts get-balances primary
```

Breaking Changes
---

- [ ] None
- [ ] Other - Please specify

BREAKING CHANGE: Requires data directory to be deleted. Also transaction JSONRPC schema has changed


applications/tari_dan_app_grpc/proto/transaction.proto
```diff
@@ -67,10 +67,10 @@ message MerkleProof {
 message Transaction {
   repeated ThaumInput inputs = 1;
```
stringhandler: These can probably be removed

applications/tari_dan_app_grpc/proto/transaction.proto
```diff
@@ -67,10 +67,10 @@ message MerkleProof {
...
   repeated ThaumOutput outputs = 2;
-  repeated Instruction instructions = 3;
-  bytes balance_proof = 4;
-  tari.dan.common.Signature signature = 5;
-  uint64 fee = 6;
+  repeated Instruction fee_instructions = 3;
+  repeated Instruction instructions = 4;
+  bytes balance_proof = 5;
```
stringhandler: I'm assuming this is a remnant of a previous attempt at fees, and can probably be removed

applications/tari_dan_app_utilities/src/base_layer_scanner.rs
```diff
@@ -290,11 +290,10 @@ impl BaseLayerScanner {
 
             for output in utxos.outputs {
                 let output_hash = output.hash();
-                let sidechain_feature = output.features.sidechain_feature.as_ref().ok_or_else(|| {
-                    BaseLayerScannerError::InvalidSideChainUtxoResponse(
-                        "Validator node registration output must have a sidechain features".to_string(),
-                    )
-                })?;
+                let Some(sidechain_feature) = output.features.sidechain_feature.as_ref() else {
+                    warn!(target: LOG_TARGET, "Validator node registration output must have sidechain features");
```
stringhandler, line 294: Which sidechain outputs is the base node returning that have this problem?

applications/tari_dan_wallet_cli/src/command/account.rs
```diff
@@ -90,8 +96,23 @@ pub struct ClaimBurnArgs {
...
-    #[clap(long, short = 'f')]
-    fee: Option<u64>,
+    #[clap(long, short = 'f', default_value_t = 1000)]
+    fee: u64,
+}
+
+#[derive(Debug, Args, Clone)]
+pub struct RevealFundsArgs {
+    /// The account name where the funds will be revealed
+    account_name: String,
```
stringhandler, line 106:
```suggestion
   #[clap(long, short='n', alias='name')]
    account_name: String,
```

applications/tari_dan_wallet_cli/src/command/account.rs
```diff
@@ -90,8 +96,23 @@ pub struct ClaimBurnArgs {
...
+    /// Amount of funds to reveal
+    reveal_amount: u64,
+    /// The fee to pay for the reveal transaction
+    #[clap(long, short = 'f', default_value_t = 1000)]
+    fee: u64,
+    /// If set, the fee will be paid from the revealed funds instead of from the account resulting in less revealed
+    /// funds than requested.
+    #[clap(long)]
+    pay_from_reveal: bool,
```
stringhandler, line 115: This is a difficult thing to name, suggesting this, but I don't think it's any better:
```suggestion
   use_as_fees: bool,
```

applications/tari_validator_node/src/p2p/services/mempool/validator.rs
```diff
@@ -50,3 +52,28 @@ impl Validator<Transaction> for MempoolTransactionValidator {
...
 }
+
+#[derive(Debug)]
+pub struct FeeTransactionValidator;
+
+#[async_trait]
+impl Validator<Transaction> for FeeTransactionValidator {
+    type Error = MempoolError;
+
+    async fn validate(&self, transaction: &Transaction) -> Result<(), MempoolError> {
```
stringhandler, line 63: This will probably change in future, but fine to have it now

applications/tari_validator_node/src/payload_processor.rs
```diff
@@ -85,26 +89,35 @@ where TTemplateProvider: TemplateProvider<Template = LoadedTemplate>
             initial_ownership_proofs: vec![owner_token],
         };
 
-        let modules = vec![]; // No modules for now, currently used in tests. Also will be useful for more advanced use-cases like fees, etc.
-
-        let processor = TransactionProcessor::new(package, state_store, auth_params, consensus, modules);
+        // 1 per byte
+        // Divide by 2 to account for the cost of CBOR
+        let initial_cost = self.fee_table.per_kb_wasm_size() * package.total_code_byte_size() as u64 / 1024 / 2;
```
stringhandler, line 94: nit: this logic is probably better in a separate struct or service, but fine to leave it here until more rules emerge

applications/tari_validator_node/tests/cucumber.rs
```diff
@@ -77,25 +77,25 @@ use crate::utils::{
...
-    current_scenario_name: Option<String>,
-    commitments: IndexMap<String, Vec<u8>>,
-    commitment_ownership_proofs: IndexMap<String, RistrettoComSig>,
-    rangeproofs: IndexMap<String, Vec<u8>>,
-    addresses: IndexMap<String, String>,
-    num_databases_saved: usize,
-    account_public_keys: IndexMap<String, (RistrettoSecretKey, PublicKey)>,
-    claim_public_keys: IndexMap<String, PublicKey>,
-    wallet_daemons: IndexMap<String, DanWalletDaemonProcess>,
+    pub base_nodes: IndexMap<String, BaseNodeProcess>,
```
stringhandler, line 80: TariWorld is OPEN TO THE PUBLIC lol

stringhandler (commented):
I like the module approach

applications/tari_dan_wallet_cli/src/command/account.rs
sdbondi: It's a mandatory field for now, so feels better to me to have it as an arg. `reveal <from_acc_name> <amount>`. If we maybe added a default account concept (maybe default account to pay fees), we could use that if not specified. Then I'd change it to a flag. 

Happy to change it if you disagree.

applications/tari_dan_app_utilities/src/base_layer_scanner.rs
sdbondi: I have a chain that has old burnt outputs without the sidechain feature set. This prevents the VN from scanning. 

There may be a use case for a burn without the Confidential Output sidechain feature (not sure) otherwise we should add a base layer validation.

applications/tari_dan_wallet_cli/src/command/account.rs
sdbondi: yeah `--pay-fees-from-revealed-funds` is probably better but a bit long. Perhaps we should always just pay from revealed funds since it amounts to the same final amount and it works even when no funds are previously revealed

applications/tari_validator_node/src/payload_processor.rs
sdbondi: yeah agree, might want to charge for args size etc so possibly just a cost function that takes in a Transaction will be sufficient.

applications/tari_validator_node/tests/cucumber.rs
sdbondi: lol, yeah think that is why guys put all the steps in one file - because they use the fields on TariWorld. #callthecodejanitor :P

applications/tari_dan_app_utilities/src/base_layer_scanner.rs
stringhandler: Yeah I think let's leave this as is for now, but I'll get an issue created for the base layer validation

applications/tari_dan_wallet_cli/src/command/account.rs
stringhandler: nah, leave it as is for now

PR #440, opened 2023-03-23
feat: add dan wallet ui

Cifko:
Description
---
As it is now, it supports only accounts (create, list) and keys(list, create, activate).

---
`npm install` and `npm run dev` in the `applications/tari_dan_wallet_web_ui` (currently it uses port 9000)

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (dismissed):
utACK - some files need a license

PR #441, opened 2023-03-23
feat: add flow templates

stringhandler:
Description
---
Adds flow template definitions for composing multiple templates into a single template. 

Other changes (sorry):
- Resume any downloads on templates on startup
- Templates are loaded dynamically instead of being loaded before execution starts (this is necessary for composability)
- Templates can now be published using a URL instead of a local folder path. 
   - this is for WASM and flow templates: `cargo run --bin tari_validator_node_cli -- templates publish -u HTTP://localhost:9000/file.wasm`
- Removed fees for per-kb for wasm since the templates are loaded dynamically, it can't be determined. The size of the template is also not really relevant IMO for fees
- Add names of parameters to the ABI. This is so that they can be correctly wired in flows, but also helps readability of the ABI.
- Added a max recursion to prevent calling a flow from a flow to infinity

Motivation and Context
---
Flow templates are JSON definitions that wire together other templates similar to how instructions are joined. 

How Has This Been Tested?
---
Tested manually with this example flow: 
https://raw.githubusercontent.com/tari-project/test-templates/72284e73678099cbc92a5318035542f5ee76935a/src/flows/faucet_to_account.json

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


dan_layer/engine/src/flow/workers/call_method_worker.rs
```diff
@@ -0,0 +1,98 @@
...
+            context.auth_scope.clone(),
+            &component_address,
+            method_name,
+            // TODO: put in rest of args
+            args,
+            context.recursion_depth + 1,
+            context.max_recursion_depth,
+        )?;
+
+        let result = exec_result.raw.clone();
```
stringhandler: Note:
Not sure if this is a bucket or a bucket_ref. 

If it is a bucket:
 We currently allow returning a bucket and passing a bucket to the next method. This opens the door to users creating binary buckets and passing them to valid method that expect the bucket to only be created by the run time

CjS77:
utACK

PR #450, opened 2023-03-30
feat!: make resource address deterministic

stringhandler:
Description
---
Make resource addresses deterministic, specifically: `Hash(template_address, token_symbol)`

Motivation and Context
---
Currently, the resource address is unpredictable, making some contracts difficult. For example, you can't create a resource and mint an NFT in the same transaction because the wallet does not know which shards to include in it.

How Has This Been Tested?
---
Unit tests, and tests with NFT contracts

---
Create an NFT or fungible token and check the resource address created is the same.

Breaking Changes
---

- [ ] None
- [x] Other - Resources will be created in a new place. Existing templates must be recompiled.

BREAKING CHANGE: Resources will be created in a new place. Existing templates must be recompiled.


applications/tari_validator_node/src/bootstrap.rs
```diff
@@ -356,13 +359,16 @@ where
...
+        // metadata.insert(TOKEN_SYMBOL, "tXTR2".to_string());
 
         tx.insert_substates(SubstateShardData::new(
             shard_id,
             address,
             0,
-            Substate::new(0, Resource::new(ResourceType::Confidential, metadata)),
+            Substate::new(
+                0,
+                Resource::new(ResourceType::Confidential, "tXTR2".to_string(), metadata),
```
jorgeantonio21, line 369: What does `tXTR2` stand for ?

jorgeantonio21 (commented):
Are we dropping the use of `ObjectClaim` for all ?

applications/tari_validator_node/src/bootstrap.rs
stringhandler: It's the testnet layer 2 tari. No official name yet. 

ObjectClaim was an idea I had in an earlier design. It was to prove ownership on a per-shard level. It turns out it's easier to use the whole transaction's ownership, so it's better to just use that

mrnaveira (dismissed):

PR #451, opened 2023-03-31
feat: stores claimable fees for VNs

jorgeantonio21:
Description
---

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


applications/tari_dan_app_grpc/proto/consensus.proto
```diff
@@ -39,9 +39,10 @@ message QuorumCertificate {
   uint64 local_node_height = 4;
   bytes shard = 5;
   uint64 epoch = 6;
-  QuorumDecision decision = 7;
-  repeated ShardPledge all_shard_pledges = 8;
-  repeated ValidatorMetadata validators_metadata = 9;
+  bytes proposed_by = 7;
```
Cifko, line 42: you shouldn't change the ids, you can add and delete, but not reuse or change, I think that's the ideology behind proto files

PR #458, opened 2023-04-05
fix(tari_indexer): invalid CBOR decoding of confidential vaults

mrnaveira:
Description
---
Used the existing `fix_invalid_object_keys` function right after decoding from CBOR, so the Json value is always correctly formed.

Motivation and Context
---
When adding an `Account` component address to be tracked by the `tari_indexer` application, if the account has a confidential resource then a `Error adding address: key must be a string` was returned.

How Has This Been Tested?
---
New unit test in the `substate_decoder` crate that reproduced the error.

---
Run an indexer and track an `Account` component that holds a confidential resource

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


stringhandler (approved):
Nice

PR #459, opened 2023-04-11
feat!: merge the mmr proofs into one

Cifko:
Description
---
Merge proof into one. This moves the mmr proof from the `ValidatorMetadata` to the `QuorumCertificate` / `VoteMessage`.

Motivation and Context
---
This will make the proof much smaller. Also the time to prove will be faster.

How Has This Been Tested?
---
Using script that creates many VN nodes and running TXs

---
Create more than one node and run some TXs

Breaking Changes
---

- [ ] None
- [x] Other - Please specify
  - This will change the Votes and QCs. So the second layer needs to be restarted.

### DIFF
--- a/dan_layer/common_types/src/quorum_certificate.rs
+++ b/dan_layer/common_types/src/quorum_certificate.rs
@@ -3,12 +3,13 @@
 
 use std::{borrow::Borrow, io};
 
-use digest::Digest;
 use serde::{Deserialize, Serialize};
 use tari_common_types::types::FixedHash;
 use tari_core::ValidatorNodeBmtHasherBlake256;
-use tari_crypto::hash::blake2::Blake256;
-use tari_engine_types::commit_result::RejectReason;
+use tari_engine_types::{
+    commit_result::RejectReason,
+    hashing::{hasher, EngineHashDomainLabel},
+};
 use tari_mmr::{Hash, MergedBalancedBinaryMerkleProof};
 
 use crate::{
@@ -217,20 +218,23 @@ impl<TAddr: NodeAddressable> QuorumCertificate<TAddr> {
     }
 
     pub fn to_hash(&self) -> FixedHash {
-        let result = Blake256::new()
-            .chain(self.local_node_hash.as_bytes())
-            .chain(self.local_node_height.to_le_bytes())
-            .chain(self.shard.as_bytes())
-            .chain((self.validators_metadata.len() as u64).to_le_bytes())
-            .chain(bincode::serialize(&self.merged_proof).unwrap())
-            .chain(bincode::serialize(&self.leaves_hashes).unwrap());
+        hasher(EngineHashDomainLabel::QuorumCertificate)
+            .chain(&self.local_node_hash)
+            .chain(&self.local_node_height)
+            .chain(&self.shard)
+            .chain(&(self.validators_metadata.len() as u64))
+            .chain(&self.merged_proof)
+            .chain(&self.leaves_hashes)
+            .result()
+            .into_array()
+            .into()
+
         // TODO: add all fields
 
         // result = result.chain((self.involved_shards.len() as u32).to_le_bytes());
         // for shard in &self.involved_shards {
         //     result = result.chain((*shard).to_le_bytes());
         // }
-        result.finalize().into()
     }
 
     pub fn payload_id(&self) -> PayloadId {
--- a/dan_layer/common_types/src/quorum_certificate.rs
+++ b/dan_layer/common_types/src/quorum_certificate.rs
@@ -1,13 +1,15 @@
 //   Copyright 2022 The Tari Project
 //   SPDX-License-Identifier: BSD-3-Clause
 
-use std::borrow::Borrow;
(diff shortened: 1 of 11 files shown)


dan_layer/common_types/src/quorum_certificate.rs
```diff
@@ -186,16 +217,15 @@ impl<TAddr: NodeAddressable> QuorumCertificate<TAddr> {
     }
 
     pub fn to_hash(&self) -> FixedHash {
-        let mut result = Blake256::new()
+        let result = Blake256::new()
```
sdbondi: We should use a domain-separated hasher 

```rust

hasher(EngineHashDomainLabel::QuorumCertificate)
            .chain(&self.local_node_hash)
            .chain(&self.local_node_height)
            .chain(&self.shard)
            .chain(&self.validators_metadata.len() as u64);
            .chain(&self.validators_metadata.len() as u64)
            .chain(&self.merged_proof)
            .chain(&self.leaves_hashes)
            .result();
```

dan_layer/core/src/models/vote_message.rs
```diff
@@ -159,4 +165,25 @@ impl VoteMessage {
     pub fn all_shard_pledges(&self) -> &ShardPledgeCollection {
         &self.all_shard_pledges
     }
+
+    pub fn merkle_proof(&self) -> Option<BalancedBinaryMerkleProof<ValidatorNodeBmtHasherBlake256>> {
+        self.merkle_proof.clone()
+    }
+
+    // TODO: impl CBOR for merged merkle proof
```
sdbondi: CBOR does not need to be implemented directly. `serde` is implemented so CBOR is. 

You could use 

```rust
tari_bor::encode(&self.merkle_proof)
```

dan_layer/core/src/models/vote_message.rs
```diff
@@ -47,6 +49,8 @@ pub struct VoteMessage {
     decision: QuorumDecision,
     all_shard_pledges: ShardPledgeCollection,
     validator_metadata: Option<ValidatorMetadata>,
+    merkle_proof: Option<BalancedBinaryMerkleProof<ValidatorNodeBmtHasherBlake256>>,
+    node_hash: Vec<u8>,
```
sdbondi: All hashes in DAN core should use `FixedHash`. I know that the MMR uses a variable length heap allocated hash type (`Vec<u8>`) but we can use `node_hash.to_vec()`  to get around that.

dan_layer/validator_node_rpc/src/conversions/consensus.rs
```diff
@@ -69,6 +71,8 @@ impl TryFrom<proto::consensus::VoteMessage> for VoteMessage {
                 .map(|n| n.try_into())
                 .collect::<Result<_, _>>()?,
             metadata.try_into()?,
+            VoteMessage::decode_merkle_proof(&value.merkle_proof)?,
+            value.node_hash,
```
sdbondi:
```suggestion
            FixedHash::try_from(value.node_hash)?,
```

dan_layer/core/src/workers/hotstuff_waiter.rs
```diff
@@ -1495,13 +1500,14 @@ where
...
-        for md in qc.validators_metadata() {
-            verify_all &= md
-                .merkle_proof
-                .verify(&validator_node_root, md.get_node_hash().to_vec());
-            if !verify_all {
+        if !qc.is_genesis() {
+            // Check the proof only for non-genesis blocks
+            let res = qc
+                .merged_proof()
+                .unwrap()
```
sdbondi:
```suggestion
                .ok_or(HotstuffError::...)?
```

dan_layer/core/src/workers/hotstuff_waiter.rs
```diff
@@ -1495,13 +1500,14 @@ where
...
+                .verify_consume(&validator_node_root, qc.leave_hashes());
+            let res = res.unwrap();
```
sdbondi: Can make a `BalancedBinaryMerkleProofError` variant in Hotstuff error
```suggestion
                .verify_consume(&validator_node_root, qc.leave_hashes())?;
```

sdbondi (commented):
LGTM - some minor comments. In casual testing, I saw an increase in message size. Will look into it a little more as that is probably more to do with testing with only 2 VNs but the important bit is that the proofs are valid and consensus still works - so not too concerned about that.

dan_layer/core/src/models/vote_message.rs
Cifko: It was a leftover comment, deleted, thanks.

sdbondi: Thanks - I meant change the encoding to use CBOR (uses serde so doesnt have to be specifically implemented for the Merkle proof) instead of bincode because we dont use this anywhere else and it adds a dependency. But CBOR will result in a larger encoding so let's leave to and we can decide how we want to do this later

sdbondi (dismissed):
ACK

PR #462, opened 2023-04-13
fix: compare addresses properly

Cifko:
Description
---
Compare the peer addresses differently. Because comparing everything lead to infinity network announcement going around.
Now I compare everything but the unverified data. Because that is filled during conversion [here](https://github.com/tari-project/tari/blob/development/comms/core/src/peer_manager/peer_identity_claim.rs#L105)

How Has This Been Tested?
---
Manually, starting a network with 4 VNs that works.

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


stringhandler:
Closing in favour of #464

PR #465, opened 2023-04-14
refactor!: component addresses are indexed

mrnaveira:
Description
---
- Changed the component addresses to be composed of template addresses + index
- Added a new field `new_component_outputs` in transactions, with the addresses for the new components that are going to be created:
    - If multiple components are going to be created for the same template in the same transaction, only the first new output is required to be specified.
    - Updated all clients and CLI applications to include this new field
- The `IdProvider` scans the incoming transaction for the `new_component_outputs` field, so the engine calls that create new components will use those addresses.

Motivation and Context
---
We want the component addresses to be deterministic and also be able to list all components of an specific template. For that, the new component address schema needs to be `H(template_address || index)`, were `index` is an incremental `u64`

How Has This Been Tested?
---
The existing unit and integration tests pass

---
Submit a transaction that creates a new component, specifying the address in the `new_component_outputs` field

Breaking Changes
---

- [ ] None
- [x] Other - Needs full Tari network reset, as the component addresses are now completely different


mrnaveira:
Abandoned as the new approach for component addresses is to be composed of template_address + user defined id (https://github.com/tari-project/tari-dan/pull/489)

PR #467, opened 2023-04-14
feat: jwt token to the wallet daemon jrpc

Cifko:
Description
---
Add jwt token to wallet daemon.
When starting the webrtc, you need to get the token, the webrtc calls then use this token for the jrpc server calls.
Currently the token is free to get (no real auth is going on). The expiration and secret key are set in config.

How Has This Been Tested?
---
Not yet.

---
Can be tested by setting up the extension (including the signaling server).

Breaking Changes
---
This will require dan wallet web ui to be updated.

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7521,6 +7521,7 @@ dependencies = [
  "config",
  "dirs",
  "futures 0.3.27",
+ "jsonwebtoken",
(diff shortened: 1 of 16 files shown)


stringhandler (approved):
ACK

Nice one

PR #475, opened 2023-04-19
feat(wallet_ui): add connector flow

NovaT82:
Description
---
Adds a button that can be embedded to initialise connector flow.
It'll check if a user already has a tari:// link in their clipboard and show different content on the first page depending on the info it gets back.
It then goes to a page where the user can allow / deny permissions.
Then goes to success page.

Note that the auto detection function works differently in different browsers:
**Chrome**: Everything should work
**Firefox**: Auto detection doesn’t work, because the Firefox API doesn’t support navigator.clipboard.readText()
**Safari**: Auto detection only works when the user clicks on the button, so it won’t automatically detect when they enter the page. This is because the read event has to be initiated by the user and is not allowed when the site tries to do it automatically.

Needs backend PR to function fully.

Also fixes some layout issues on the global theme.

Motivation and Context
---
Allow users to connect to the wallet.

How Has This Been Tested?
---
Manually

---
Click on the "Connect" button top right and follow the prompts.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


stringhandler (approved):
ACK

PR #478, opened 2023-04-20
feat: set simple graphql server for tari indexer

jorgeantonio21:
Description
---
Adds a mock GraphQL client/server, integrated with tari indexer application.

Motivation and Context
---

How Has This Been Tested?
---

---
A suitable cucumber test was added.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


applications/tari_indexer/src/cli.rs
```diff
@@ -36,7 +36,7 @@ pub struct Cli {
 
     /// Supply a network (overrides existing configuration)
     #[clap(long, env = "TARI_NETWORK")]
-    pub network: Option<String>,
+    pub cli_network: Option<String>,
```
sdbondi: This can be removed since the setting moved into CommonCliArgs

jorgeantonio21: I thought about it as well, just wasn't sure as it was being kept. Thanks for the remark

sdbondi (approved):
utACK

PR #481, opened 2023-04-24
feat: extract tagged values from wasm call return value

sdbondi:
Description
---
Extracts values of known types from binary data returned from WASM calls.

Motivation and Context
---
`IndexedValue::from_raw` can be used on any CBOR binary data e.g. input args to determine their contents

This allows us to implement additional validations that prevent template authors from making up bucket IDs and addresses. 

This also allows us to determine child substates related to a component.

With this PR we could recognize the returned value as a bucket and move the actual resource it contains on the worktop. Buckets could then be taken from the worktop (remove the need for the `PutLastInstructionOnWorktop` instruction)

How Has This Been Tested?
---
Basic unit test, manually

---
Check the instruction result in the payload database table. 

Breaking Changes
---

- [ ] None
- [ ] Other - Please specify


stringhandler (dismissed):

sdbondi:
ACK

mrnaveira (approved):
Nice

PR #482, opened 2023-04-24
chore: add template type to command line args

Cifko:
Description
---
Add template type to command line arguments. So the validator_node_cli can be run without human interaction.

How Has This Been Tested?
---
Manually.

---
Run tari_validator_node_cli with e.g. `--template-type wasm` when publishing template

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_validator_node_cli/src/command/template.rs
+++ b/applications/tari_validator_node_cli/src/command/template.rs
@@ -55,6 +55,9 @@ pub struct PublishTemplateArgs {
     #[clap(long, alias = "template-version")]
     pub template_version: Option<u16>,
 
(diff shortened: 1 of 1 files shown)


sdbondi:
LGTM - Would be cool if it could figure that out from the file extension

sdbondi (dismissed):

Cifko:
> LGTM - Would be cool if it could figure that out from the file extension

Good idea. Done. If the extension is `flow` it will be `flow` everything else defaults to `wasm`. Will the extension be `flow` for the flow type?

stringhandler (approved):

PR #485, opened 2023-04-25
feat: jwt token to the wallet daemon jrpc

Cifko:
Description
---

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


applications/tari_dan_wallet_daemon/src/handlers/accounts.rs
```diff
@@ -215,6 +219,7 @@ pub async fn handle_invoke(
 
 pub async fn handle_get_balances(
     context: &HandlerContext,
+    token: Option<String>,
```
stringhandler, line 226: I would add the token to the context, since it will almost always be used

Cifko: Yeah, I'm working on adding jwt to the wallet_sdk so I can leave the context immutable

Cifko: The tokens are now stored in db. But I still need to send it to each command, because the token is per request.

applications/tari_dan_wallet_cli/src/cli.rs
```diff
@@ -31,6 +31,8 @@ use crate::command::Command;
 pub struct Cli {
     #[clap(long, alias = "endpoint", env = "JRPC_ENDPOINT")]
     pub daemon_jrpc_endpoint: Option<Multiaddr>,
+    #[clap(long, alias = "token")]
```
sdbondi: nit:
```suggestion
    #[clap(long)]
```

applications/tari_dan_wallet_cli/src/main.rs
```diff
@@ -38,21 +38,23 @@ async fn main() -> Result<(), anyhow::Error> {
     log::info!("🌍️ Connecting to {}", endpoint);
     let client = WalletDaemonClient::connect(endpoint)?;
```
sdbondi: Perhaps we should set the token in the client to avoid having to pass it around everywhere.

```suggestion
    let mut client = WalletDaemonClient::connect(endpoint)?;
    if let Some(token) = token {
       client.set_jwt(token);
    }
```

applications/tari_dan_wallet_cli/src/cli.rs
sdbondi: I think we should allow it to be set via the env so that we can set it and forget about it for the session
e.g `#[clap(long, env = "TARI_WALLET_CLI_JWT")]` (naming up to you)

dan_layer/wallet/sdk/src/apis/jwt.rs
```diff
@@ -0,0 +1,205 @@
...
+            auth_token,
+            &DecodingKey::from_secret(self.auth_secret_key.as_ref()),
+            &Validation::default(),
+        )?;
+        Ok(auth_token_data.claims)
+    }
+
+    pub fn get_permissions(&self, token: Option<String>) -> Result<JrpcPermissions, JwtApiError> {
+        let token_data = decode::<Claims>(
+            token.ok_or(JwtApiError::TokenMissing)?.as_str(),
```
sdbondi: nit: Rather make it the caller's responsibility to provide a token
```suggestion
    pub fn get_permissions(&self, token: &str) -> Result<JrpcPermissions, JwtApiError> {
        let token_data = decode::<Claims>(
            token
```

applications/tari_dan_wallet_daemon/src/handlers/accounts.rs
```diff
@@ -69,9 +69,15 @@ const LOG_TARGET: &str = "tari::dan_wallet_daemon::handlers::transaction";
 
 pub async fn handle_create(
     context: &HandlerContext,
+    token: Option<String>,
     req: AccountsCreateRequest,
 ) -> Result<AccountsCreateResponse, anyhow::Error> {
     let sdk = context.wallet_sdk();
+    let jwt = sdk.jwt_api();
+    let permissions = jwt.get_permissions(token)?;
```
sdbondi:
```suggestion
    check_auth(&sdk, token, &[JrpcPermission::Admin])?;
    
    // Elsewhere
    fn check_auth(wallet_sdk: &WalletSdk, token: Option<String>, required_perms: &[JrpcPermission]) -> Result<JrpcPermissions, JsonRpcError> {
        let token = token.ok_or_else(|| unauthorized("no JWT token provided"))?;
        let permissions = jwt.get_permissions(&token).map_err(|e| unauthorized(&format!("malformed  JWT: {e}")))?;
        for p in required_perms {
        if !permissions.has_permission(p) {
￼        return Err( unauthorized(&format!("Insufficient permissions. Required {p}"))));
￼    }
}
    }
    
    fn unauthorized(details: &str) -> JsonRpcError {
      JsonRpcError::new(
        JsonRpcErrorReason::ApplicationError(401),
            format!("Not authorized: {details}")
        serde_json::Value::Null,
    )
     }
    
```

applications/tari_dan_wallet_daemon/src/webrtc.rs
```diff
@@ -75,22 +76,31 @@ pub async fn webrtc_start_session(
...
+                    } else {
+                        let result = handle_data(address, Some(request.token), request.method, request.params)
+                            .await
+                            .unwrap();
+                        response = Response {
+                            payload: result.to_string(),
+                            id: request.id,
+                        };
+                    }
+                    println!("response {response:?}");
```
sdbondi:
```suggestion

```

dan_layer/wallet/sdk/src/apis/jwt.rs
```diff
@@ -0,0 +1,205 @@
...
+                .unwrap()
+                .as_secs() as usize,
+        };
+        let permissions_token = encode(
+            &Header::default(),
+            &my_claims,
+            &EncodingKey::from_secret(self.jwt_secret_key.as_ref()),
+        )?;
+        let mut tx = self.store.create_write_tx()?;
+        println!("Storing ID {}", auth_claims.id);
```
sdbondi, line 188:
```suggestion
```

dan_layer/wallet/storage_sqlite/src/writer.rs
```diff
@@ -70,6 +76,35 @@ impl WalletStoreWriter for WriteTransaction<'_> {
...
+                .get_result(self.connection())
+                .unwrap();
+        Ok(last_inserted_id as u64)
+    }
+
+    fn jwt_store_decision(&mut self, id: u64, permissions_token: Option<String>) -> Result<(), WalletStorageError> {
+        // let values = match token {
+        //     Some(token) => (auth_status::user_decided.eq(true),auth_status::granted.eq(true),auth_status::token)
+        // }
+        println!("Storing {id} {permissions_token:?}");
```
sdbondi:
```suggestion

```

sdbondi (commented):
Looks like a good start - minor comments

dan_layer/wallet/sdk/src/apis/jwt.rs
Cifko: I wanted to, but then I will have to do the check somewhere else (in the jrpc_server handler function).

One solution (I don't know if it's possible) is to have two handlers (one with token and one without), then I could also get rid of the axum::middleware for the token.

applications/tari_dan_wallet_daemon/src/handlers/accounts.rs
Cifko: I did it slightly differently, now I get all JwtApiErrors as 401 (e.g. expiration of the token)

dan_layer/wallet/sdk/src/apis/jwt.rs
sdbondi: Yeah I think this should be checked elsewhere - like in the `check_auth` function

sdbondi: I also want the correct jsonRPC error code to be returned (see `check_auth`). 

You could have the `sdk.jwt_api().check_auth` (Think that check_auth should be in the JWT api) and map to a JRPC error. But there will be no `map_err` cruft if we make a `check_auth` that returns the error in the handlers.

```rust
sdk.jwt_api().check_auth(..).map_err(|e| {
 JsonRpcError::new(
        JsonRpcErrorReason::ApplicationError(401),
            format!("Not authorized: {e}")
        serde_json::Value::Null,
    )
})?;
```

Note: This can still be returned from handler as a `anyhow::Error` and the server will still return the correct JSONRPC error result

sdbondi: Ah now I understand what you did. Ok that works though I prefer less extra type handling code in the generic error resolver, but happy with the approach in any case. I think that the `check_auth` function should then live in jwt_api and should return `Result<(), JwtApiError>`

applications/tari_dan_wallet_daemon/src/jrpc_server.rs
```diff
@@ -82,84 +84,89 @@ async fn handler(
...
+                            JsonRpcErrorReason::ApplicationError(401),
+                            format!("Not authorized: {e}"),
+                            serde_json::Value::Null,
+                        ),
+                    )
+                })?;
+            let webrtc_start_request = value.parse_params::<WebRtcStartRequest>()?;
             let shutdown_signal = (*shutdown_signal).clone();
-            let (preferred_address, signaling_server_address) = *addresses.clone();
+            let (preffered_address, signaling_server_address) = *addresses.clone();
```
sdbondi: typo:
```suggestion
            let (preferred_address, signaling_server_address) = *addresses.clone();
```

sdbondi: Would prefer this code to be in a handler in `handlers/webrtc.rs`

sdbondi (approved):

PR #486, opened 2023-04-25
fix: return os error codes on error

stringhandler:
Description
---
Return error codes from applications

Motivation and Context
---
When scripting these applications, they would exit with a successful exit code, even though it had errored. This makes it more testable

How Has This Been Tested?
---
Manually

---
Run an application with bad input. Previously, it would exit without showing the error

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (approved):
utACK

PR #489, opened 2023-04-26
feat!: component addresses include the template

mrnaveira:
Description
---
* The `IdProvider` in the engine:
    * Calculates the `ComponentAddress` to be the hash of the template address and the component id.
* On the validator node mempool:
    * Perform a preemptive dry run processing of the transaction after receiving it, so it can add all the missing shard ids that the execution indicates.
* On dry run transaction processing:
    * Add utility functions for calculating missing shard ids for new components
    * Improve the robustness of remote shard querying, up until now it would fail if even one committee member was unresponsive. 
* Removed unused `new_component_address` field in the transaction builder.

Motivation and Context
---
The aim of this PR is to lay the foundation for user-specified component addresses.

For that the shard IDs for components will have the format:
```
H(template_address || component_id)
```

Where `component_id` is of Hash type and may be user defined. By default the engine will put a random value on it. 

So the specific goals for this PR are:
* Refactor the generation of component addresses so they are composed of `template_address` and (random for now) `component_id`
* Adapt the transaction processing in the validator nodes to infer missing outputs for new components, so the user does not need to specify them on client side.

For now the `component_id` field is going to be set as random every time. The capabilities for user defined `component_id` on template constructors will be left for future PRs.

How Has This Been Tested?
---
All existing unit and integration tests pass.

---
Submit a transaction that creates a new component

Breaking Changes
---

- [ ] None
- [x] Other - Needs full Tari network reset, as the component addresses are now completely different


dan_layer/transaction/src/id_provider.rs
```diff
@@ -71,8 +71,14 @@ impl IdProvider {
...
 
-    pub fn new_component_address(&self) -> Result<ComponentAddress, IdProviderError> {
-        Ok(self.new_id()?.into())
+    pub fn new_component_address(
+        &self,
+        template_address: TemplateAddress,
+        component_id: Option<Hash>,
+    ) -> Result<ComponentAddress, IdProviderError> {
+        // if the component_id is not specified by the caller, then it will be random
+        let component_id = component_id.unwrap_or(self.new_id()?);
```
sdbondi: Surprised clippy didn't catch this https://rust-lang.github.io/rust-clippy/v0.0.212/index.html#or_fun_call
This will increment the counter in every case.

mrnaveira: Nice catch, I forgot that `unwrap_or` is eagerly evaluated

dan_layer/transaction/src/id_provider.rs
```diff
@@ -77,7 +77,10 @@ impl IdProvider {
         component_id: Option<Hash>,
     ) -> Result<ComponentAddress, IdProviderError> {
         // if the component_id is not specified by the caller, then it will be random
-        let component_id = component_id.unwrap_or(self.new_id()?);
```
sdbondi: can also use `component_id.unwrap_or_else(|| self.new_id()?);`

sdbondi (approved):
Nice. Think this can go in as is. Some points that need to be addressed in future:

- security: involved shards should not be malleable
- perf: transaction executing twice
- potential problem: if we naively allow a user to specify bytes, a user may squat on any public key

PR #490, opened 2023-04-27
feat: add event templating

jorgeantonio21:
Description
---
Adds events to the Tari Runtime. It also corrects a minor bug on the runtime tracker `finalize` method, that would ignore logging.

Motivation and Context
---

How Has This Been Tested?
---
Add test `basic_emit_event` on the engine tests.

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


applications/tari_dan_app_grpc/proto/transaction.proto
```diff
@@ -46,6 +46,7 @@ message Instruction {
     PUT_OUTPUT_IN_WORKSPACE = 2;
     EMIT_LOG = 3;
     CLAIM_BURN = 4;
+    EMIT_EVENT = 5;
```
mrnaveira: I am not convinced that we need a new Instruction type for events, as I understand events should be emitted only by components and not by the user. That would also mean that we don't need the extra fields in transaction/instruction types.

dan_layer/engine/src/runtime/impl.rs
```diff
@@ -167,6 +175,7 @@ impl<TTemplateProvider: TemplateProvider<Template = LoadedTemplate>> RuntimeInte
             LogLevel::Warn => log::Level::Warn,
             LogLevel::Info => log::Level::Info,
             LogLevel::Debug => log::Level::Debug,
+            LogLevel::Store => todo!("Implement me later"),
```
mrnaveira: As we are using a separated event function, we probably need to remove the `Store` variant here

dan_layer/engine/src/runtime/impl.rs
```diff
@@ -821,11 +830,13 @@ impl<TTemplateProvider: TemplateProvider<Template = LoadedTemplate>> RuntimeInte
         let substates_to_persist = self.tracker.take_substates_to_persist();
         self.invoke_modules_on_before_finalize(&substates_to_persist)?;
 
-        let (result, fee_receipt) = self.tracker.finalize(substates_to_persist)?;
+        let events = self.tracker.take_events();
```
mrnaveira: At this point we may need a new struct to define the tracker finalization, that will include `result`, `fee_receipt` and the `events` as fields.

dan_layer/engine_types/src/events.rs
```diff
@@ -0,0 +1,53 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct Event {
+    pub timestamp: u64,
+    pub message: String,
+}
+
+impl Event {
+    pub fn new(message: String) -> Self {
+        let now = SystemTime::now()
```
mrnaveira: I am pretty sure that this way of getting the timestamp will not work in a real network (or even in a integration test), as each VN in the committee will get a different value thus the result of the transaction will be different as well.

We need a timestamp value that is calculated exactly the same in all VNs in the committee. That is probably non-trivial, so for now I would leave timestamps outside of the scope.

mrnaveira (dismissed):
The approach is good in general, but some changes are necessary IMO

dan_layer/engine/src/runtime/impl.rs
jorgeantonio21: Great catch, remove it

dan_layer/engine_types/src/events.rs
jorgeantonio21: Very good observation, agreed ! Removed the field

applications/tari_dan_app_grpc/proto/transaction.proto
jorgeantonio21: Removed it

dan_layer/transaction_manifest/src/parser.rs
```diff
@@ -324,6 +324,7 @@ fn macro_call(mac: &Ident, tokens: TokenStream) -> Result<ManifestIntent, syn::E
             level: LogLevel::Error,
             message: parse2::<LitStr>(tokens)?.value(),
         })),
+        "store" => todo!("Implement me later"),
```
sdbondi: Is this needed? What is the intention?

sdbondi (dismissed):
Looks like a good start

dan_layer/transaction_manifest/src/parser.rs
jorgeantonio21: It was supposed to be removed, but missed it. Thanks for catching it :)

sdbondi (approved):

PR #496, opened 2023-05-02
feat: hash built-in template content

Cifko:
Description
---
Hash the content of a built-in template to prevent recompile when there was no change. The wasm build is not stable.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (commented):
We already compare the contents of the target wasm to the committed wasm. What benefit`*` is there to this approach?

`*` other than hashing + 32 byte comparison _might_ outperform byte for byte comparison

EDIT: Oh I see this is hashing the source code. cargo only checks the modified date of the files when using the `rerun-if-modified` AFAIK. The problem with this comes in when pulling new code/switching branches in git, the system then rebuilds and the resulting binary often isn't identical. Your approach could work, it would need to take all of the dependencies of Account into account. In addition to the lib file in Account, hashing the Account Cargo.lock file and template_lib source and the template_abi source should work in almost all cases (hopefully not too slow).

PR #497, opened 2023-05-02
feat: add jwt expiration from cli

Cifko:
Description
---
Now you can specify the validity duration of the token via cli. Also now the `permissions token` that will be granted will have the same expiration datetime. This way the generated permissions token will not rely on when the user accepts. And if we somehow accepts twice the same request, it will generated the same token. If no value is provided then the default value from daemon is used (currently that's 5 minutes).

Motivation and Context
---
Sometimes there is a need for specifying the time.

How Has This Been Tested?
---
Manually.

---
`cargo run --bin tari_dan_wallet_cli -- auth request Admin 1` - this means one second validity.
You can generate a token with validity one second and then try to use it (keep in mind that JWT library allows for 1 minute threshold. So if you have validity one sec, in reality it's 61 seconds.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (dismissed):
Similar to https://github.com/tari-project/tari-dan/pull/492 but for the `auth request` command

utACK

sdbondi:
utACK

sdbondi (approved):

PR #500, opened 2023-05-03
feat: event integration with graphql 

jorgeantonio21:
Description
---
Adds `saveEvent` and `getEvent` graphQL calls

Motivation and Context
---
Allow consumers to query the indexer for events by transaction hash and template address.

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


applications/tari_indexer/src/substate_manager.rs
```diff
@@ -251,6 +255,23 @@ impl SubstateManager {
         Ok(nfts)
     }
 
+    pub async fn get_event_from_db(
+        &self,
+        template_address: String,
+        tx_hash: String,
```
sdbondi: nit: What are these strings? They can technically be anything. In general, we want to use the compiler to ensure that the database layer can only have well-formed data. How that data is serialized is not a concern of callers. Not doing this results in: 
- malformed data being entered into the database
- Changes to data types may only break things at the highest level (e.g. something in the call to the jrpc broke vs the decoding in this specific db function is broken)
- conversion code littering the codebase (see `EventQuery::get_event`)
- get and update/insert operations can have inconsistencies 
- tight coupling with a specific database and/or ORM (e.g. below `NewEvent`  is a diesel helper struct, we may want to use a non-relational database, we may want to use postgres etc)
- changing how something is serialized (e.g. JSON, hex etc) results in many changes to the code

```suggestion
        template_address: &TemplateAddress,
        tx_hash: PayloadId, // This is usually a bit muddy,  Hash and FixedHash are also used across the codebase
```

applications/tari_indexer/src/substate_manager.rs
```diff
@@ -251,6 +255,23 @@ impl SubstateManager {
...
+    ) -> Result<Vec<EventData>, anyhow::Error> {
+        let mut tx = self.substate_store.create_read_tx()?;
+        let events = tx.get_events(template_address, tx_hash)?;
+        Ok(events)
+    }
+
+    pub async fn save_event_to_db(&self, new_event: NewEvent) -> Result<(), anyhow::Error> {
```
sdbondi: Same for `NewEvent` (diesel struct). For the payload, we dont want the caller to choose how to serialize something. If we're defining payload as a HashMap<String, String> then use that type. 

```suggestion
    pub async fn save_event_to_db(&self, template_address: TemplateAddress, tx_hash: PayloadId, topic: String, payload: &HashMap<String, String>) -> Result<(), anyhow::Error> {
// or even
    pub async fn save_event_to_db<T: Serialize>(&self, template_address: TemplateAddress, tx_hash: PayloadId, topic: String, payload: &T) -> Result<(), anyhow::Error> {
// where in the current implementation T happens to be serialized to JSON
```

applications/tari_indexer/src/substate_manager.rs
jorgeantonio21: Great suggestion ! I did refactor the code accordingly, using the types `TemplateAddress` and `PayloadId` for the signature types. Then made use of the `.to_string()` method to convert to string, for query build in the `fn get_event` method on the substate manager reader.

applications/tari_indexer/src/substate_manager.rs
jorgeantonio21: That's slightly more subtle, as I think GraphQL doesn't support, at the moment of writing and if I am not mistaken, queries parametrized by `HashMap` directly.

sdbondi (dismissed):
Thanks for the changes - ACK

sdbondi (approved):

applications/tari_indexer/src/substate_manager.rs
sdbondi: Was more referring to the database layer, not graphQL, but I see you came right with the changes :) There is no PR description, but I'm curious what the use case is for saveEvent in the graphQL interface. Happy to merge as is though

PR #503, opened 2023-05-04
fix(consenus): vote reject and release pledges if remote pledges invalid

sdbondi:
Description
---
fix(consenus): vote reject and release pledges if remote pledges invalid
fix: remove extraneous new output shards from most of the walletd transactions

Motivation and Context
---
InvalidPledges error variant was not handled correctly from `validate_pledges` leading to a hard error rather than a REJECT vote. With the change in this PR, InvalidPledge results in the transaction being rejected and the pledges released. 

Also reduced the number of involved shards by removing some `new_output`s. Another PR will remove this entirely.

How Has This Been Tested?
---
Tested by reusing the same keypair in walletd for more than one account. The pledge included in this case was for a previously down'd account shard.

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


dan_layer/core/src/workers/hotstuff_waiter.rs
```diff
@@ -1229,7 +1229,7 @@ where
                             })?;
                             Ok(finalize)
                         }),
-                        Err(e @ HotStuffError::MissingPledges(_)) => {
```
sdbondi, line 1238: This is the consensus ~~change~~ fix

applications/tari_validator_node/src/dry_run_transaction_processor.rs
```diff
@@ -56,10 +56,11 @@ use tari_validator_node_rpc::{
     proto::rpc::VnStateSyncResponse,
 };
 use thiserror::Error;
+use tokio::task;
 
 use crate::{p2p::services::template_manager::TemplateManager, payload_processor::TariDanPayloadProcessor};
 
-const LOG_TARGET: &str = "tari::validator_node::dry_run_transaction_processor";
+const LOG_TARGET: &str = "tari::dan::validator_node::dry_run_transaction_processor";
```
jorgeantonio21, line 63: Should we make it the default across codebase ?

dan_layer/core/src/workers/hotstuff_waiter.rs
jorgeantonio21: I don't have enough context to understand this comment

sdbondi: 1) fix(consenus): vote reject and release pledges if remote pledges invalid 
2) fix: remove extraneous new output shards from most of the walletd transactions

This change relates to 1, the rest of the changes relate to 2

applications/tari_validator_node/src/dry_run_transaction_processor.rs
sdbondi: This is how it is currently set up in the `log4rs.yaml`s so `tari::xxxx` will not emit logs (other than specific configs for the L1 tari repo, tari::comms, etc). The current standard for this repo is `tari::dan::xxxx`.

PR #505, opened 2023-05-05
feat: adds test for leader failure with multiple committees, corrects previous bug

jorgeantonio21:
Description
---
In this PR, we add a leader failure with multiple committees cucumber test. We also correct an existing bug in the transaction builder logic. Namely, we were adding shard ids for possibly non-existing substates (this happened when creating an account with no funds, in which only a component address of the account is created as a substate, and no vault). 

To resolve this issue, we deleted the code

```rs
transaction
            .meta_mut()
            .involved_objects_mut()
            .extend((0..max_outputs).map(|_| {
                let new_hash = id_provider
                    .new_address_hash()
                    .expect("id provider provides num_outputs IDs");
                (ShardId::from_hash(&new_hash, 0), SubstateChange::Create)
            }));
```

from 

```rs
pub async fn submit_transaction(
    instructions: Vec<Instruction>,
    common: CommonSubmitArgs,
    base_dir: impl AsRef<Path>,
    client: &mut ValidatorNodeClient,
) -> Result<SubmitTransactionResponse, anyhow::Error>
```

in the `applications/tari_validator_node_cli/src/command/transaction.rs`. 

Motivation and Context
---
Improves test suite on leader failure.

How Has This Been Tested?
---
Added cucumber test and extensive state checks on manual testing,

---
Run 

```
cargo test --test cucumber -- -n "Leader failure with multiple committees"
```

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (dismissed):

applications/tari_validator_node_cli/src/command/transaction.rs
```diff
@@ -289,7 +289,7 @@ pub async fn submit_transaction(
         .sign(&key.secret_key)
         .build();
 
-    if transaction.meta().involved_shards().is_empty() {
+    if transaction.meta().involved_shards().is_empty() && transaction.meta().max_outputs() == 0 {
```
sdbondi, line 292: Since the VNs attempt to determine outputs, we should still be able to submit a transaction with 0 outputs (max_outputs should be named num_new_outputs)

sdbondi (approved):

PR #508, opened 2023-05-05
feat: add pid to validator node

stringhandler:
Description
---
1. Update the indexer to allow nodes to query locally, or ask the network
2. Update tari-crypto to use a minimum compilable version
3. Allow the indexer to run without a public address
4. Create a pid file on the indexer once it starts
5. Add an error when the indexer does not have a committee

Motivation and Context
---
1. So that you can query a non-stored substate from a client easily
2. To reduce the reliance on tags on the base layer
3. Indexers may run in "Listen only mode", and generally won't be running as public nodes
4. This allows tests to wait until the indexer has properly started
5. The indexer may not have scanned to a point on the base layer where a committee is present. This usually happens when the indexer starts, especially in cucumber tests

How Has This Been Tested?
---
locally, using spawn_vn_tests

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


applications/tari_indexer/src/json_rpc/handlers.rs
```diff
@@ -128,7 +128,12 @@ impl JsonRpcHandlers {
         let response = GetIdentityResponse {
             node_id: self.node_identity.node_id().to_hex(),
             public_key: self.node_identity.public_key().to_hex(),
-            public_address: self.node_identity.public_addresses().first().unwrap().to_string(),
+            public_address: self
+                .node_identity
+                .public_addresses()
+                .first()
+                .map(|s| s.to_string())
+                .unwrap_or_else(|| "No public address".to_string()),
```
sdbondi, line 136: We should change this field to a `Vec<MultiAddr>`. Doesn't have to be in this PR though the change should be a minor one.

applications/tari_indexer/src/main.rs
```diff
@@ -56,7 +56,8 @@ async fn main_inner() -> Result<(), ExitError> {
     let cfg = load_configuration(config_path, true, &cli)?;
     let config = ApplicationConfig::load_from(&cfg)?;
     println!("Starting validator node on network {}", config.network);
-
+    // Remove the file if it was left behind by a previous run
+    let _file = fs::remove_file(config.common.base_path.join("pid"));
```
sdbondi, line 60: Could also be if another instance is already started. Probably fine for our use cases though since we're really just using this for readiness.

dan_layer/template_lib/src/hash.rs
```diff
@@ -25,21 +25,23 @@ use std::{
...
+    str::FromStr,
 };
 
+use generic_array::{typenum::U32, GenericArray};
 use serde::{Deserialize, Serialize};
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Ord, PartialOrd, Hash, Default, Serialize, Deserialize)]
 #[serde(transparent)]
-pub struct Hash([u8; 32]);
+pub struct Hash(GenericArray<u8, U32>);
```
sdbondi, line 36: Interesting - what is the reason for this change?

sdbondi (approved):
utACK

applications/tari_indexer/src/main.rs
stringhandler: I suppose we could create a lock and make sure it can't delete it, but if the process previously crashed, it should also delete it

PR #509, opened 2023-05-05
feat: account transfer creates destination account component

mrnaveira:
Description
---
* `tari_dan_wallet_daemon`:
    * New `transfer` method that encapsulates the logic of transfers, including the destination account creation if needed.
* `tari_dan_wallet_cli`:
    * Update the existing `send` command to invoke the new `transfer` method in the daemon
* `engine_types`:
    *  New utility function `new_component_address_from_parts` that is used both in the `wallet_daemon` and in the engine's `id_provider`. It could have been created as an additional construction in the `ComponentAddress` type, but there are some extra dependencies that we would have to add to the `template_lib` for that.
* New` account.feature` cucumber scenario to test transfers, plus all the new required utility to support it. 

Motivation and Context
---
When sending funds to other accounts, right now the destination account component must have been created previously. This requires the receiver of funds to create the account beforehand, which will eventually cost fees. This is very inconvenient when we want to bring new people to try the Tari network, as they would need to burn and claim Minotari.

We want the sender to transparently create the destination account component if it does not exits.

This PR aims to implement that by:
* Creating a new method in the wallet daemon API that implements the transfer (and destination account creation) logic
* Update the wallet `send` command to use the new method and only require the destination public key instead of the component address.

How Has This Been Tested?
---
New cucumber scenario to test account transfers

---
Use the updated `send` command in the wallet cli

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (dismissed):
Thanks, looks great! utACK

sdbondi (approved):

PR #510, opened 2023-05-08
refactor: remove fee loans

mrnaveira:
Description
---
Removed fee loans and all related logic from the project

Motivation and Context
---
Up until know we had the concept of _free loan_, which was a way to enable account initialisation without having received the funds yet.

But now, after https://github.com/tari-project/tari-dan/pull/509, accounts can be created when other accounts send funds for the first time, so the fee loans are no longer needed.

So the goal of this PR is to remove fee loans and all related logic from the project.

How Has This Been Tested?
---
Existing unit and integration tests pass

---
Existing use cases must still work. For example, use the `send` command in the wallet cli to create an account 

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (dismissed):
utACK

With #512 we'll be able to create accounts from test tokens

stringhandler (approved):

PR #513, opened 2023-05-09
test(cucumber): fix cucumbers

sdbondi:
Description
---
fix(cucumber): Uses separate directories for each scenario, application, node combination.
feat(cucumber): Allow signing key to be specified for manifest transactions
fix(cucumber): fail step if transaction fails
fix(cucumber): fix some failing transactions 
fix(walletd): wallet will detect new vaults when transaction is complete
fix(cucumber): only run one mock server at a time
fix(cucumber): manifest transactions panic on failure and fix tests which did not pass
fix(cucumber): ad-hoc improvements to error output

Motivation and Context
---
Uses directories per scenario per application, per node name e.g
`/temp/cucumber_894581/Mint--mutate-and-burn-non-fungible-tokens/base_node/BASE`

Validator node cucumber already had this, did the same for the wallet daemon, indexer, base node and console wallet.

How Has This Been Tested?
---
Cucumbers pass

---
CI

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


stringhandler (dismissed):
Nice

stringhandler (approved):

PR #514, opened 2023-05-10
feat: add selecting permissions

Cifko:
Description
---
Add jwt permissions, the web3 developer can set optional and required permission. In the dan wallet you can deselect the optional permissions and grant the token. Then the web3 developer gets the token, which is then sent with every request. The wallet store all tokens. And in the DB there is a flag for revoking the token (no ui yet).

How Has This Been Tested?
---
Manually.

---
You can setup danwallet, and create a simple page using [tari-connector](https://github.com/tari-project/tari-connector). This will connect to the signaling server and create an url/qrcode (there is a signaling server jwt and permissions). Pass this to the dan wallet. Authorize that. Next the web3 page can reguest the token by sending "get.token" message. After that you can start using the token for other jrpc (over webrtc) requests.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_dan_wallet_daemon/src/webrtc.rs
+++ b/applications/tari_dan_wallet_daemon/src/webrtc.rs
@@ -1,7 +1,7 @@
 //   Copyright 2023 The Tari Project
 //   SPDX-License-Identifier: BSD-3-Clause
 
-use std::{net::SocketAddr, sync::Arc};
+use std::{net::SocketAddr, pin::Pin, sync::Arc};
 
 use anyhow::Result;
 use axum_jrpc::{JsonRpcAnswer, JsonRpcRequest, JsonRpcResponse};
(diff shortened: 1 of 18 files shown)


applications/tari_dan_wallet_daemon/src/webrtc.rs
```diff
@@ -81,58 +152,19 @@ pub async fn webrtc_start_session(
...
-            });
-        }
-        Box::pin(async {})
+        on_ice_candidate(
+            ice_candidate,
+            signaling_server_token_clone.clone(),
+            signaling_server_address,
+        )
     }));
+    dbg!();
```
stringhandler: Some dbgs still present

applications/tari_dan_wallet_web_ui/src/utils/tari_permissions.tsx
```diff
@@ -0,0 +1,378 @@
...
+  }
+  toJSON() {
+    return this.tagged.toJSON();
+  }
+}
+
+export class VaultId {
+  private tagged: Tagged;
+  constructor(value: any) {
+    this.tagged = new Tagged(value["@@TAGGED@@"][0], new Hash(value["@@TAGGED@@"][1]));
```
sdbondi, line 193: This wnt be needed after PR #515 :)

sdbondi (dismissed):
utACK 

Is there an example site in the repo that uses the connector I could use to test?

stringhandler (approved):

PR #515, opened 2023-05-10
fix!: simplify JSON encoding for all JSON-RPC calls

sdbondi:
Description
---
feat!: add BorTag wrapper that does not emit tags in HRFs e.g. JSON
fix!: implement simple arg parser
fix!: encode addresses as string and bytes as hex
chore: clean up deps

Motivation and Context
---
Previously, very difficult to create a correct JSON-RPC request.
String and hex representations are used where possible. Args can now be passed in as strings.
```json
{
    "id": 0,
    "jsonrpc": "2.0",
    "method": "transactions.submit",
    "params": {
        "fee_instructions": [
            {
                "CallMethod": {
                    "component_address": "component_6476a7dc6ee72290c6e667f2d12caa061489109bc0e5f15d479a30ea0ec13988",
                    "method": "pay_fee",
                    "args": [
                        "Amount(1000)"
                    ]
                }
            }
        ],
        "instructions": [{
            "CallMethod": {
                    "component_address": "component_6476a7dc6ee72290c6e667f2d12caa061489109bc0e5f15d479a30ea0ec13988",
                    "method": "balance",
                    "args": [
                        "resource_0101010101010101010101010101010101010101010101010101010101010101"
                    ]
                }
        }],
        "inputs": [],
        "override_inputs": false,
        "new_outputs": 0,
        "specific_non_fungible_outputs": [],
        "new_resources": [],
        "new_non_fungible_outputs": [],
        "new_non_fungible_index_outputs": [],
        "is_dry_run": false,
        "proof_ids": []
    }
}
```

How Has This Been Tested?
---
Manually with POSTMAN

---
With an account with tokens in, call `transactions.submit` (above request)
call `transactions.get_result`

Breaking Changes
---

- [ ] None
- [x] Other - most JSONRPC requests are not backward compatible.


stringhandler:
Looks great!

dan_layer/engine_types/src/lib.rs
```diff
@@ -17,6 +18,7 @@ pub mod non_fungible;
 pub mod non_fungible_index;
 pub mod resource;
 pub mod resource_container;
+pub mod serde_with;
```
stringhandler, line 21: Perhaps in a different PR, but we should avoid the term `serde_with` because it's a commonly used crate

stringhandler (approved):

dan_layer/engine_types/src/lib.rs
sdbondi: yeah good point could be confusing - will change in another PR

PR #518, opened 2023-05-15
feat: committee chart and inspect in new page

NovaT82:
Description
---
Installs React E-charts.
Adds a chart to committees that displays the number of committees and amount of members per committee.
Adds a details page to inspect committee members.
Adds the committees page to the side menu.
I moved where some of the state gets generated from ValidatorNode.tsx to App.tsx and exported it as a context object. This way the current epoch, shardKey and identity are all easily accessible from anywhere in the app.

Motivation and Context
---
Adds the ability to inspect committees and show them visually.
We can use E-charts for other charts inside the app.
Data like current epoch, identity, etc was being passed down as props, making it difficult to access from pages that were not being called from ValidatorNode.tsx, so it's now in a context object that you can import into any page.

How Has This Been Tested?
---
Manually

---
When running the app, there should now be a chart above committees where you can click to inspect the committee members.
You can also click on the committees menu item and it should now display the committee info.

Breaking Changes
---
x

- [x] None
- [ ] Other - Please specify


stringhandler (dismissed):

sdbondi:
In one of the PRs (#521) I had to push the package-lock.json changes and then it built. Looks like the same issue here

sdbondi (approved):

PR #516, opened 2023-05-11
feat: adds tx result to network substates

jorgeantonio21:
Description
---
Adds a `TransactionReceipt` as a substate to the network, for every successfully finalized transaction. The structure is as follows:

```rs
pub struct TransactionReceipt {
  pub transaction_hash: Hash,
  pub events: Vec<Event>,
  pub logs: Vec<LogEntry>,
  pub fee_receipt: FeeReceipt,
}
```
On every transaction submission, after the transaction has been accepted by the network, each involved substate is stored in the network ledger. Additionally, we are now storing an additional substate, which contains metadata about the transaction.

Notice that, at the moment, we decided to leave the involved `SubstateDiff` from the definition of `TransactionReceipt` as it would involve recursive calls on substates to be persisted, on the tracker storage.

Motivation and Context
---
The current goal is to index events from substates of the network.

How Has This Been Tested?
---
Checked count of substates in tests.

What process can a PR reviewer use to test or verify this change?
---

Breaking Changes
---

- [x] None
- [ ] Requires data directory to be deleted
- [ ] Other - Please specify

PR #527, opened 2023-05-18
feat: adds event indexing over component addresses

jorgeantonio21:
Description
---

In previous work (#516, #500, #490, #478) we developed the tools for both a tari indexer graphql service, templating events and adding transaction receipts to the network substates (upon acceptance). The goal of this PR is to bring forth the previous efforts together. Namely, this PR aims at:

1. Add an event scanning interface, integrated with `SubstateManager` in the tari indexer, that scans for events on the network and stores these into the local indexer db for cache (whenever needed). 
2. The above logic is integrated with the GraphQL interface, so that we are to request events from a GraphQL service.
3. Refactored the `Event` interface (both on tari template lib and tari indexer), so that component addresses are optional and we require both `template_address` and `transaction_hash` fields.
4. We updated the tracker event formation, to automatically set the both `transaction_hash` and `template_address` from its current state.
5. Refactored the enum `SubstateValue` so that deleted substates keep track of which transaction hash they have been destroyed. In this way, given a `ComponentAddress` and a `version` we can query directly the network committee responsible for storing the substate containing the `TransactionReceipt` associated with that same transaction. This greatly improves the efficiency of the event scanning on the tari indexer, since we don't have to query multiple committees to get the actual event state.
6. Added `ConsensusConstants` to the `EpochManagerHandle`. This was done in an initial version, to get the committee size at the level of `SubstateManager` committee data to randomnize the process of querying VNs within a committee. After a suggestion of @sdbondi, we can't do this process in a more direct way. Therefore, we might as well remove this step as well.

Motivation and Context
---

Updates the Tari Indexer service to be able to query the network to get events associated with components (account creation, transfer, etc). GraphQL integration is made.

How Has This Been Tested?
---
Added cucumber tests for querying events on the network, upon component creation.

---
Run cucumber, create accounts and use the Tari Indexer to scan the network for the corresponding events stored in the Indexer local db. Manifests using the `emit_event` could also be used to further testing.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


applications/tari_indexer/src/substate_manager.rs
```diff
@@ -260,35 +263,135 @@ impl SubstateManager {
...
+            .iter()
+            .map(|e| Event::try_from(e.clone()))
+            .collect::<Result<Vec<Event>, anyhow::Error>>()?;
+
+        if events.is_empty() {
+            let network_events = self.substate_scanner.get_events_for_transaction(tx_hash).await?;
+            events.extend(network_events);
+        }
+
+        Ok(events)
```
sdbondi: nit: simplified and removes allocations
```suggestion
        let events = {
            let mut tx = self.substate_store.create_read_tx()?;
            tx.get_events_for_transaction(PayloadId::from_array(tx_hash.into_array()))?
        };

        let mut events = events
            .into_iter()
            .map(|e| Event::try_from(e))
            .collect::<Result<Vec<Event>, anyhow::Error>>()?;

        // If we have no events locally, fetch from the network if possible. 
        if events.is_empty() {
            let events = self.substate_scanner.get_events_for_transaction(tx_hash).await?;
            return Ok(events)
        }

        Ok(events)
```

applications/tari_indexer/src/substate_manager.rs
```diff
@@ -260,35 +263,135 @@ impl SubstateManager {
...
-        &self,
-        template_address: TemplateAddress,
-        tx_hash: PayloadId,
-    ) -> Result<Vec<EventData>, anyhow::Error> {
-        let mut tx = self.substate_store.create_read_tx()?;
-        let events = tx.get_events(template_address, tx_hash)?;
-        Ok(events)
-    }
-
     pub async fn save_event_to_db(
```
sdbondi: Not an async function
```suggestion
    pub fn save_event_to_db(
```

dan_layer/engine/src/runtime/impl.rs
```diff
@@ -167,14 +165,16 @@ impl<TTemplateProvider: TemplateProvider<Template = LoadedTemplate>> RuntimeInte
 
     fn emit_event(
         &self,
-        template_address: TemplateAddress,
-        tx_hash: Hash,
+        component_address: Option<ComponentAddress>,
```
sdbondi: The component should not come from the template since a template/user could send anything they want there.  It is known by the execution context so we should use the component address from there.

Suggest: In the RuntimeState struct you can add:

```rust
#[derive(Debug, Clone)]
pub struct RuntimeState {
    pub template_address: TemplateAddress,
// Add this
    pub component_address: Option<ComponentAddress>,
}
```

The for call function `component_address: None` and call method `component_address: Some(*component_address)`

stringhandler (commented):
Tested locally, but got stuck on passing byte arrays to graphql. I suggest we use hex (or "component_xxxxx") for inputs, but also return bytearrays as strings a well

applications/tari_indexer/src/substate_storage_sqlite/sqlite_substate_store_factory.rs
```diff
@@ -293,23 +301,89 @@ impl SubstateStoreReadTransaction for SqliteSubstateStoreReadTransaction<'_> {
...
+            reason: format!("get_events_by_version: {}", e),
         })?;
 
         Ok(res)
     }
+
+    fn get_all_events(&mut self, component_address: &ComponentAddress) -> Result<Vec<EventData>, StorageError> {
+        let res =
+            sql_query("SELECT component_address, tx_hash, topic, payload FROM events WHERE component_address = ?")
+                .bind::<Text, _>(component_address.hash().to_string())
```
sdbondi:
```suggestion
                .bind::<Text, _>(component_address.to_string())
```

applications/tari_indexer/src/substate_storage_sqlite/sqlite_substate_store_factory.rs
```diff
@@ -293,23 +301,89 @@ impl SubstateStoreReadTransaction for SqliteSubstateStoreReadTransaction<'_> {
...
+            version
+        );
         let res = sql_query(
-            "SELECT template_address, tx_hash, topic, payload FROM events WHERE template_address = ? AND tx_hash = ?",
+            "SELECT component_address, template_address, tx_hash, topic, payload FROM events WHERE component_address \
+             = ? AND version = ?",
         )
-        .bind::<Text, _>(template_address.to_string())
-        .bind::<Text, _>(tx_hash.to_string())
+        .bind::<Nullable<Text>, _>(Some(component_address.hash().to_string()))
```
sdbondi:
```suggestion
        .bind::<Nullable<Text>, _>(Some(component_address.to_string()))
```

applications/tari_indexer/src/substate_storage_sqlite/sqlite_substate_store_factory.rs
```diff
@@ -293,23 +301,89 @@ impl SubstateStoreReadTransaction for SqliteSubstateStoreReadTransaction<'_> {
...
+            "Querying substate scanner database: get_stored_versions_of_events with component_address = {} and \
+             start_version = {}",
+            component_address,
+            start_version
+        );
+        use crate::substate_storage_sqlite::schema::events;
+        let res: Vec<i32> = events::table
+            .filter(
+                events::component_address
+                    .eq(&component_address.hash().to_string())
```
sdbondi:
```suggestion
                    .eq(&component_address.to_string())
```

sdbondi (commented):
Nice, LGTM. Only thing I picked up was that the component address is set by the WASM code template when it should come from the execution context

applications/tari_indexer/src/substate_manager.rs
jorgeantonio21: Awesome, thanks !

applications/tari_indexer/src/substate_storage_sqlite/models/events.rs
```diff
@@ -58,28 +63,62 @@ pub struct EventData {
...
-        tx_hash.copy_from_slice(&tx_hash_buffer);
+        let template_address = Hash::from_hex(&event_data.template_address)
+            .map_err(|e| anyhow::anyhow!(e.to_string()))?
+            .into_array();
+
+        let tx_hash = Hash::from_hex(&event_data.tx_hash)
+            .map_err(|e| anyhow::anyhow!(e.to_string()))?
+            .into_array();
 
         let payload = serde_json::from_str(event_data.payload.as_str()).map_err(|e| anyhow::anyhow!(e.to_string()))?;
```
sdbondi: Anyhow implements `From` for any `Error` and `?` desugars to
```rust
match stmt {
   Ok(v) => v,
   Err(e) => return Err(e.into()), // Note the .into()
}
```

```suggestion
        let payload = serde_json::from_str(event_data.payload.as_str())?;
```

applications/tari_indexer/src/substate_storage_sqlite/models/events.rs
```diff
@@ -58,28 +63,62 @@ pub struct EventData {
...
+    fn try_from(event_data: EventData) -> Result<Self, Self::Error> {
+        let component_address = event_data
+            .component_address
+            .clone()
+            .map(|comp_addr| ComponentAddress::from_str(comp_addr.as_str()))
+            .transpose()?;
+        let template_address =
+            Hash::from_hex(&event_data.template_address).map_err(|e| anyhow::anyhow!(e.to_string()))?;
+        let tx_hash = Hash::from_hex(&event_data.tx_hash).map_err(|e| anyhow::anyhow!(e.to_string()))?;
+        let payload = serde_json::from_str(event_data.payload.as_str()).map_err(|e| anyhow::anyhow!(e.to_string()))?;
```
sdbondi: Anyhow implements `From` for any `Error` and `?` desugars to
```rust
match stmt {
   Ok(v) => v,
   Err(e) => return Err(e.into()), // Note the .into()
}
```

```suggestion
        let template_address =
            Hash::from_hex(&event_data.template_address)?;
        let tx_hash = Hash::from_hex(&event_data.tx_hash)?;
        let payload = serde_json::from_str(event_data.payload.as_str())?;
```

applications/tari_indexer/src/substate_storage_sqlite/models/events.rs
jorgeantonio21: Thanks ! Addressed !

sdbondi (dismissed):

applications/tari_validator_node/tests/cucumber.rs
```diff
@@ -885,24 +885,73 @@ async fn works_indexer_graphql(world: &mut TariWorld, indexer_name: String) {
...
+        .map(|(_, v)| {
+            v.address
+                .as_component_address()
+                .expect("Failed to parse `ComponentAddress`")
+        })
+        .collect::<Vec<_>>();
+
+    let component_address = *component_addresses
+        .first()
+        .unwrap_or_else(|| panic!("Did not find component addresses"));
```
sdbondi: nit: If you are only interested in finding a single item, you can make use of `Iterator::find` without having to collect all matches first

```suggestion
    let component_address = accounts_component_addresses
        .into_iter()
        .find(|(k, _)| k.contains("components/Account"))
        .map(|(_, v)| {
            v.address
                .as_component_address()
                .expect("Failed to parse `ComponentAddress`")
        })
       .expect("Did not find component address");

```

applications/tari_validator_node/tests/cucumber.rs
```diff
@@ -885,24 +885,73 @@ async fn works_indexer_graphql(world: &mut TariWorld, indexer_name: String) {
...
+    let mut graphql_client = indexer.get_graphql_indexer_client().await;
+    let query = format!(
+        "{{ getEventsForComponent(componentAddress: {:?}, version: 0) {{ componentAddress, templateAddress, txHash, \
+         topic, payload }} }}",
+        component_address.to_string()
+    );
+    let res = graphql_client
+        .send_request::<HashMap<String, Vec<tari_indexer::graphql::model::events::Event>>>(&query, None, None)
+        .await
+        .unwrap_or_else(|_| panic!("Failed to obtain getEventsForComponent query result"));
```
sdbondi: You don't have to change anything, just to complete the discussion about this lint and increase understanding: A `&'static str` does not cost anything to create (it's a pointer to a hard-coded string in the program) so a closure is not necessary in this case. 

This is fine:
```suggestion
        .expect("Failed to obtain getEventsForComponent query result");
```

If you had to say allocate on the heap or perform any other call, e.g  `.expect("whatever".to_string().as_str())` `to_string()` (heap allocation) will always be called regardless of the expect triggering. So `unwrap_or_else(|err| panic!("whatever {}", err))` is needed, then the closure is only called if the unwrap fails.

sdbondi (approved):

sdbondi:
utACK

Went over the code in some depth ~~but did not test it~~

PR #528, opened 2023-05-18
fix: reject transaction if component address is duplicate

sdbondi:
Description
---
1. Fixes bug in tracker which can cause the component state to be overwritten if a template creates a component using an existing component_id
2. Adds test for this bug
3. replaces `transaction_receipt_` substate address prefix with `txreceipt_`
4. use default account to pay for fees when creating a new account
5. always use a new key to create a new account
6. fixes failing cucumbers due to 4

Motivation and Context
---
3. address convention should have a single delimiter
4. account creation is no longer zero-rated (#510) so fees must be paid by an existing account
5. a new key is required when creating a new account 

How Has This Been Tested?
---
Manually, and new unit test

---
```
cargo run --bin tari_dan_wallet_cli --  accounts create-free-test-coins acc1
cargo run --bin tari_dan_wallet_cli --  accounts create --name acc2
```

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


mrnaveira (approved):
LGTM

PR #529, opened 2023-05-19
fix(engine)!: use resource addresses derived from tx hash

sdbondi:
Description
---
- uses resource address = `H(tx_hash || n)`
- reverts some of the changes in #450 
- removes the ability to add output shards in the transaction builder

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_dan_wallet_cli/src/command/transaction.rs
+++ b/applications/tari_dan_wallet_cli/src/command/transaction.rs
@@ -425,7 +425,7 @@ pub async fn submit_transaction(
     println!("✅ Transaction {} submitted.", resp.hash);
     println!();
     // TODO: Would be great if we could display the Substate addresses as well as shard ids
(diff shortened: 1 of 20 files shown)


sdbondi:
Pausing this for now to focus on more important items. TODO:
- add a cucumber that mints some initial NFTs in the same transaction as resource creation
- (possibly required) update code and cucumbers to remove all places where outputs are explicitly declared

stringhandler:
The main problem that the previous PR was trying to fix was that you couldn't create a Resource (specifically an NFT) and mint an NFT in the same transaction. If you can find a way to still achieve that I'm open to it, but I'm not keen on merging anything that doesn't fill that requirement

sdbondi:
Closing in favour of #743

PR #530, opened 2023-05-22
fix: handle (confidential) transfer

Cifko:
Description
---
If anyone wants to use this right now here it is. But this will eventually not be committed. Because there will be a different PR from @mrnaveira that solve this and more.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


applications/tari_dan_wallet_daemon/src/handlers/accounts.rs
```diff
@@ -913,6 +913,23 @@ pub async fn handle_transfer(
     let destination_account_address =
         get_or_create_account_address(&sdk, &req.destination_public_key, &mut inputs, &mut instructions).await?;
 
+    let related = context
+        .wallet_sdk()
+        .substate_api()
+        .get_related_substates(&destination_account_address.into(), None)
```
stringhandler, line 924: Does this return all versions or only the latest version?

Cifko: This returns only the address, the latest version is then retrieved just after this (I do it only for the vault right now).
It could be bundled together to return the versions directly. But @mrnaveira is working on a PR where all of this will be done on the indexer side. So we don't have to do it here.

PR #532, opened 2023-05-24
fix: remove unnecessary checked_sub and expect

stringhandler:
Description
---
removes a checked_sub and expect which caused a panic

Motivation and Context
---
The Validator Node would panic when this checked_sub returns a negative. I'm not sure exactly why it mismatched, but the base layer height was 39, so I assume some kind of edge case or perhaps a sync was happening. Either way, the next line does a checked_sub and it doesn't look necessary to prevent a negative number here.

How Has This Been Tested?
---
Validator Node would crash without this change when using dan_testing

---
Not really easy to reproduce

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_validator_node/src/p2p/services/epoch_manager/base_layer_epoch_manager.rs
+++ b/applications/tari_validator_node/src/p2p/services/epoch_manager/base_layer_epoch_manager.rs
@@ -500,10 +500,8 @@ impl BaseLayerEpochManager {
         let constants = self.get_base_layer_consensus_constants().await?;
         let expiry = constants.validator_node_registration_expiry();
 
-        let num_blocks_since_last_reg = self
-            .current_epoch
-            .checked_sub(last_registration_epoch)
-            .expect("current epoch was less than the epoch we registered"); // Reorgs are not supported
+        // Note this can be negative in some cases
+        let num_blocks_since_last_reg = self.current_epoch - last_registration_epoch;
 
         // None indicates that we are not registered, or a previous registration has expired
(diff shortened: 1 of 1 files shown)


applications/tari_validator_node/src/p2p/services/epoch_manager/base_layer_epoch_manager.rs
```diff
@@ -500,10 +500,8 @@ impl BaseLayerEpochManager {
         let constants = self.get_base_layer_consensus_constants().await?;
         let expiry = constants.validator_node_registration_expiry();
 
-        let num_blocks_since_last_reg = self
-            .current_epoch
-            .checked_sub(last_registration_epoch)
-            .expect("current epoch was less than the epoch we registered"); // Reorgs are not supported
+        // Note this can be negative in some cases
+        let num_blocks_since_last_reg = self.current_epoch - last_registration_epoch;
```
sdbondi: These are unsigned (u64) so cannot be negative and will panic/underflow. Ideally we should find out why this expectation failed, something like current epoch is not updated before the last_registration. 

Otherwise we can return an error for now (which will just log/not crash the node)

stringhandler: my bad, I didn't check the types. Switched it to a saturating sub which should give a similar result

stringhandler: Returning 0 or an error would incorrectly get it to re-register

sdbondi (approved):

PR #535, opened 2023-05-29
feat(walletui): transaction confirmation popup

NovaT82:
Description
---
Adds a popup with a UI to confirm a transaction in the wallet.
Currently checks for the value: 'transaction://', but this can be changed once we have the correct info.

Motivation and Context
---

How Has This Been Tested?
---
Manually

---
The popup will trigger if you have 'transaction://' in your clipboard (in Chrome / Safari).

Breaking Changes
---
x

- [x] None
- [ ] Other - Please specify


sdbondi (approved):
utACK

PR #536, opened 2023-05-29
feat: indexer auto-fills transaction inputs

mrnaveira:
Description
---
* Created a new `required_inputs` in the transaction meta, to hold the inputs that the user/clients specifies.
* Clients and CLI populate the new `required_inputs` field.
* The indexer uses the autofill functionality in the `submit_transaction` method (used by the wallet daemon) so all related inputs are included.
* Moved substate decoding and transaction autofill functionality into the `indexer_lib` to allow them to be reused in other applications in the future.

Motivation and Context
---
Up until now, the client submitting a transaction into the network must specify all inputs that are needed for consensus. This makes it impractical to submit transactions as it can be quite difficult to know all the inputs involved, specially on more complex templates.

So the goal of this PR is to **implement a input auto-filler in the Indexer**, so clients only need to specify the higher level components involved (i.e. accounts and the components they want to call) and the indexer will automatically scan the network to include all related substates (as well as the latests version of those).

For security reasons, we are creating a separated field in transactions (`required_inputs`) which contains the inputs that the client requires (typically the account address component with a specific version) and it's read-only (i.e. included in the transaction hash). The indexer just adds all related inputs into the existing `involved_objects` field that is not included in the hash.

Note that there may be other auto-fill actions to be developed in the future to have a seamlessly experience in transaction creation. 

How Has This Been Tested?
---
- New cucumber scenario for transfers between accounts that exist
- Manually submitting a transaction

---
Submit a transaction thought the indexer, only specifying the component addresses as inputs, and let the indexer figure out all other dependent inputs

Breaking Changes
---

- [ ] None
- [x] Requires data directory to be deleted (new `required_inputs` field in transactions)
- [ ] Other - Please specify


dan_layer/indexer_lib/src/transaction_autofiller.rs
```diff
@@ -0,0 +1,110 @@
...
+        let mut autofilled_transaction = original_transaction.clone();
+
+        // scan the network to fetch all the substates for each required input
+        // TODO: perform this loop concurrently by spawning a tokio task for each scan
+        let mut input_addresses: Vec<(SubstateAddress, u32)> = vec![];
+        let mut input_substates: Vec<Substate> = vec![];
+        for r in autofilled_transaction.meta().required_inputs() {
+            let scan_res = self.substate_scanner.get_substate(r.address(), r.version()).await?;
+
+            // TODO: should we return an error if some of the inputs are not "Up"?
```
stringhandler: Maybe a warning? It might be useful to allow this method to not fail during testing, or if we pass a flag `allow_downed_inputs`

stringhandler (dismissed):
Runs fine when creating accounts, but I wasn't able to test the autofill, so more of an untested ACK

Code looks good.

sdbondi:
Fixed the broken test unrelated to this PR here: https://github.com/tari-project/tari-dan/pull/538

dan_layer/indexer_lib/src/transaction_autofiller.rs
mrnaveira: Added a warning and also refactored the input substate fetching (to differentiate if a specific version was required). I think we can skip returning errors and just log a warning, as if the input is actually needed it will not be included anyway and consensus will fail. If in the future we see that we want to fail sooner we can add more complex logic with a parameter here.

dan_layer/indexer_lib/src/substate_decoder.rs
```diff
@@ -136,26 +145,33 @@ fn fix_invalid_object_keys(value: &CborValue) -> CborValue {
...
-    Ok(ciborium::de::from_reader::<CborValue, _>(bytes)?)
+fn decode_into_cbor(bytes: &[u8]) -> Result<CborValue, SubstateDecoderError> {
+    let value = ciborium::de::from_reader::<CborValue, _>(bytes)
+        .map_err(|e| SubstateDecoderError::UnexpectedError(e.to_string()))?;
+    Ok(value)
 }
 
 // Recursively scan a substate for references to other substates
-pub fn find_related_substates(substate: &Substate) -> Result<Vec<SubstateAddress>, anyhow::Error> {
+pub fn find_related_substates(substate: &Substate) -> Result<Vec<SubstateAddress>, SubstateDecoderError> {
```
sdbondi, line 174: At some point in a future PR, should we use IndexedValue to extract referenced substate addresses? 
```rust
let state_blob = match substate.substate_value() { 
 //...
};
let value = IndexedValue::from_raw(state_blob)?;
let related_substates = value.owned_substates();
```

sdbondi (dismissed):
utACK

Looks good!

dan_layer/wallet/sdk/src/models/substate.rs
```diff
@@ -54,3 +55,9 @@ impl Display for VersionedSubstateAddress {
 #[derive(Debug, thiserror::Error)]
 #[error("Failed to parse versioned substate address {0}")]
 pub struct VersionedSubstateAddressParseError(String);
+
+impl From<VersionedSubstateAddress> for SubstateRequirement {
```
sdbondi, line 59: We can probably remove the VersionedSubstateAddress struct in future

dan_layer/indexer_lib/src/transaction_autofiller.rs
```diff
@@ -0,0 +1,132 @@
...
+        // add all substates related to the inputs
+        // TODO: perform this loop concurrently by spawning a tokio task for each scan
+        // TODO: we are going to only check the first level of recursion, for composability we may want to do it
+        // recursively (with a recursion limit)
+        let mut autofilled_inputs: Vec<(SubstateAddress, u32)> = vec![];
+        let related_addresses: Vec<Vec<SubstateAddress>> = input_substates
+            .iter()
+            .map(find_related_substates)
+            .collect::<Result<_, _>>()?;
+        let related_addresses = related_addresses.into_iter().flatten();
```
sdbondi: Should probably not scan for related substates if the tx creator has already provided them.

```suggestion
        let related_addresses: Vec<Vec<SubstateAddress>> = input_substates
            .iter()
            .map(find_related_substates)           
            .collect::<Result<_, _>>()?;
        let related_addresses = related_addresses.into_iter().flatten()
                    // Something like
            .filter(|s| !transactions.meta().has_substate(s))
```

dan_layer/indexer_lib/src/substate_decoder.rs
mrnaveira: That could make it easier to interpret transaction results from clients, specially when we have transactions with many instructions

dan_layer/transaction/src/transaction.rs
```diff
@@ -146,18 +153,36 @@ impl Eq for Transaction {}
 
 #[derive(Debug, Clone, Default, Deserialize, Serialize, Eq, PartialEq)]
 pub struct TransactionMeta {
+    required_inputs: Vec<SubstateRequirement>,
     involved_objects: HashMap<ShardId, SubstateChange>,
```
sdbondi, line 157: Think we can remove involved_objects in future and just have explicit `<Substate, Option<Version>>` inputs. VNs can translate to shard ids in the mempool

mrnaveira: I agree, the `involved_objects` use ShardIds which is a bit inconvenient (makes it hard to interpret), we may need a new field with a new type for auto-filled inputs or outputs without the optional version (something like `<Substate, u32>`)

sdbondi: :+1: Makes sense, the required_inputs are in the sig challenge and the autofilled_inputs are included by some notary (e.g. maybe indexer and/or maybe a VN), which signs the additional inputs/outputs and these are "validated" by VNs - that is, the VN can quickly validate that these should have been included without running the template.

sdbondi (dismissed):

sdbondi (approved):

PR #543, opened 2023-06-02
refactor: represent tariswap fee as a per-mil amount

mrnaveira:
Description
---
* Changed the `fee` parameter in the `tariswap` template constructor to be an integer (per-mil amount) instead of a float
* Updated the corresponding unit test accordingly

Motivation and Context
---
On https://github.com/tari-project/tari-dan/pull/517 we introduced a new template example `tariswap`, that implements a decentralised exchange that allows adding/removing liquidity and token swaps.

The template works on unit tests, but presents a problem when trying to initialize it (calling the `new` function) via wallet/VN CLI's or through a `tari-connector` web page. The `fee` parameter in the constructor is defined as a float number which is currently not supported in clients, so this PR refactors it to be a integer (a per-mil amount) instead.

The goal is to enable `tariswap` to work manually as soon as possible. We might consider adding float number support for clients in the future.

How Has This Been Tested?
---
* The existing unit test has been updated and passes
* Manually we can initialize now components of the template via wallet/VN CLI's or through a `tari-connector` web page.

---
Run the unit test or manually initialize a new component of the Tariswap template (calling the `new` function)

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/dan_layer/engine/tests/tariswap.rs
+++ b/dan_layer/engine/tests/tariswap.rs
@@ -19,7 +19,7 @@ struct TariSwapTest {
     account_proof: NonFungibleAddress,
 }
 
-fn setup(fee: f64) -> TariSwapTest {
+fn setup(fee: u16) -> TariSwapTest {
     let mut template_test = TemplateTest::new(vec!["tests/templates/tariswap", "tests/templates/faucet"]);
 
     // create the token pair for the swap pool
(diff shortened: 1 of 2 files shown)


sdbondi (approved):
LGTM - Floats are troublesome and imprecise / precision is lost for division etc and so are explicitly disallowed. All operations should use integers.

PR #544, opened 2023-06-02
feat(tari_validator_node_cli): extract each type of substate address

mrnaveira:
Description
---
Extract each type of substate address when passing them as arguments on template functions or method calls

Motivation and Context
---
The `tari_dan_wallet_cli` does extract each type of substate address. That functionality should also be present on the `tari_validator_node_cli`.

How Has This Been Tested?
---
Manually invoking a template function with a resource address as parameter

---
Manually invoke a template function (or component method) with a substate address as an input argument

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (approved):
Thanks

PR #547, opened 2023-06-05
feat(walletd): add naming to the JWT tokens

Cifko:
Description
---
Add names to the jwt (the user needs to see which token is which).
Add jwt method for listing of the (still valid) tokens.

Motivation and Context
---
Adding a name to the token makes it much easier for user to revoke it. Currently the name comes from user when it calls the `auth.accept`. So we do need a change in the UI for this. Or we can provide the name from the `tari-connector`. But I suggest we can provide the name and still be replaced by the user in the acceptance dialog in the dan wallet ui.

How Has This Been Tested?
---

1. Request a token
`cargo run --bin tari_dan_wallet_cli -- auth request Admin`
Outputs `Auth token <auth_token>`
2. Grant the permissions
`cargo run --bin tari_dan_wallet_cli -- auth grant <auth_token> <name>`
Outputs `Access granted. Your JRPC token : <permission_token>`
3. List the tokens:
`cargo run --bin tari_dan_wallet_cli -- --token <permissions_token> auth list`
Outputs `Id 1 name <name>`

---
Use the same step I used for testing.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/dan_layer/wallet/storage_sqlite/src/reader.rs
+++ b/dan_layer/wallet/storage_sqlite/src/reader.rs
@@ -26,11 +26,7 @@ use tari_engine_types::substate::{InvalidSubstateAddressFormat, SubstateAddress}
 use tari_template_lib::models::ResourceAddress;
 use tari_utilities::hex::Hex;
 
-use crate::{
-    diesel::ExpressionMethods,
-    models,
-    serialization::deserialize_json,
-};
+use crate::{diesel::ExpressionMethods, models, serialization::deserialize_json};
 
 const LOG_TARGET: &str = "tari::dan::wallet_sdk::storage_sqlite::reader";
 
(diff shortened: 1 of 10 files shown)


sdbondi (approved):
utACK

dan_layer/wallet/storage_sqlite/src/reader.rs
```diff
@@ -150,6 +150,17 @@ impl WalletStoreReader for ReadTransaction<'_> {
         })
     }
 
+    // -------------------------------- JWT -------------------------------- //
+    fn jwt_get_all(&mut self) -> Result<Vec<(i32, String)>, WalletStorageError> {
+        use crate::schema::auth_status;
+        let res = auth_status::table
+            .select((auth_status::id, auth_status::token))
+            .filter(auth_status::granted.eq(true))
```
sdbondi, line 158: This will also return revoked tokens - intentional?

PR #548, opened 2023-06-05
feat: wallet daemon cli nfts

jorgeantonio21:
Description
---
1. Adds mint account NFT logic to the wallet daemon, so a user can mint NFTs using its own account, via the daemon. 
2. Adds a wallet sdk non fungible tokens api.
3. Add daemon client request for minting NFTs.
4. Adds cucumber scenario to test correctness of the logic.

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


stringhandler (approved):
utACK

PR #549, opened 2023-06-06
fix: jwt permission issues between wallet daemon and signaling server

mrnaveira:
Description
---
* Define new permission types for transactions and get default account, so those basic operations do not require an `Admin` access
* The `TransactionSend` permission accepts optional addresses, so if it's `None` it allows to send transaction sto any address. This is needed for now to unblock development of applications, but we would want to switch back to specific addresses in the future, probably instead of `TransactionSend` it would be more convenient a `ResourceInput` permission.
* Fixed mismatch of the JWT secret keys between the wallet daemon and the signaling server. Anyway they are hardcoded so in the future we would need a safer solution.
* Fixed the `auth.login` method in the signaling server, the permissions were not being included in the JWT so the tokens were empty when used on the dan wallet.

Motivation and Context
---
Up until now on web applications that use the `tari-connector` project to interact with a wallet daemon, the tokens being generated did not work. This PR fixes all issues preventing it to work on the `tari-dan` repo. There is a corresponding PR (https://github.com/tari-project/tari-connector/pull/4) on the `tari-connector` side.

How Has This Been Tested?
---
Manually using a web application that uses the `tari-connector` to interact with the wallet daemon

---
Manually use a web application that uses the `tari-connector` to interact with the wallet daemon

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (approved):
utACK

jorgeantonio21 (approved):
utAck

Question: Should we from now on, default to use `JrpcPermission::AccountInfo` instead of `JrpcPermission::Admin`, whenever necesary ?

stringhandler (approved):

mrnaveira:
> utAck
> 
> Question: Should we from now on, default to use `JrpcPermission::AccountInfo` instead of `JrpcPermission::Admin`, whenever necesary ?

The `JrpcPermission::Admin` is the highest level of permissions and allows to manage the wallet daemon completely. It's used on the wallet daemon UI and should not be used/required for web3 apps IMO. We should always require the bare minimum privileges possible for doing operations.

That's why I created the `JrpcPermission::AccountInfo` for getting the information on the current default account in the wallet. It's intended for getting read-only and non-sensible information, for example to have the component address to craft actual transactions that involve the current account.

PR #557, opened 2023-06-08
feat(consensus): initial structure for batched consensus

sdbondi:
Description
---
feat(consensus): adds basic code structure for batched consensus
feat(consensus): implement on new transaction
feat(consensus): adds interface traits for services external to consensus
feat(consensus): adds initial storage interface
feat(consensus): adds sqlite storage implementation
feat(consensus): initial implementations of on receive new view, on beat and on propose
feat(consensus): partially implements on propose received

Motivation and Context
---
Initial structure is not currently used so that we can continue work on this without breaking existing code. 

Structure is as follows:

- tari_consensus crate 
  - worker - handles and dispatches all hotstuff messages
  - on propose - triggered when a node is the leader and there progress should be made on batches
  - on receive new view - primarily for new leader election for leader failure
  - on proposal received - handle a block from a local/foreign leader
  - on vote received - handle a vote received for a local block

- state_store - generic trait interfaces for batched store
- sqlite_state_store - SQLite implementation of the state store

How Has This Been Tested?
---
Tests need to be implemented

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


stringhandler (approved):
happy with this for now, as it's building into a full process later

PR #566, opened 2023-06-13
fix: gitignore wasm binaries

sdbondi:
Description
---
Ignore builtin wasm binaries

Motivation and Context
---
These often cause conflicts due to non-deterministic builds. Until we can build deterministically, having these committed is unnecessary.

How Has This Been Tested?
---
Code still compiles 

---
Check that wasms are still built

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi:
image


:hocho: :hocho:

mrnaveira (approved):

stringhandler (approved):

PR #567, opened 2023-06-13
fix(wasm)!: avoid using hashmaps in templates

sdbondi:
Description
---
fix(wasm)!: replace hashmap with Metadata type for emit_event
fix(wasm)!: replace hashmap with btree in account builtin
fix(wasm)!: improve ergonomics of emit_event api

Motivation and Context
---
Previously, when an account is deserialized and re-serialized, the contents will always change regardless of there actually being any changes. This is because a HashMap uses random state (by default) so when the deserializer builds it, the hash ordering is completely different.

emit_event previously required a hashmap to be built, but that is replaced by the Metadata type in prelude. The emit event api generalizes this to `Into<Metadata>` which allows:

```rust
emit_event("my_topic", [("hello", "world")]);
```

The use of a hashmap in events causes hash mismatches when pledging tx_recipt substates that contain more than one event

How Has This Been Tested?
---
Engine tests pass

---
Covered in CI

Breaking Changes
---

- [ ] None
- [x] Other - change to template lib, some templates using emit_event will no longer work


CjS77:
@sdbondi -- it's possible to change the hashing algo used by `HashMap` fwiw

sdbondi:
@CjS77 Yup, though that does potentially open up hashDoS attacks. I figured it's better to use a data type that has ordering as an invariant for events and other metadata. For a general use map substate in templates, we probably need to look into a lazy map substate that only requires the keys being mutated/referenced as inputs.

stringhandler (approved):

PR #568, opened 2023-06-14
feat(consensus): add on vote receved and transaction executor

sdbondi:
Description
---
feat(consensus): implements on vote received
feat(consensus): adds transaction executor
fix(consensus): correct safenode predicate in OnReceiveProposal

Motivation and Context
---
*OnReceiveVote*
Collects votes from replicas and generates a new high QC
On beat will trigger later and propose the next block containing the QC for the previous block. 

*TransactionExecutor*
Responsible for executing transactions in parallel when received from the network and when missing transactions are requested in response to a proposed block with unknown local transactions. 

This component is not currently used in this PR as it will require batched consensus integration with the validator node. This work is left for a subsequent PR.

*safeNode predicate*
@stringhandler pointed out that I had the incorrect boolean logic in the safeNode predicate. This has been corrected.

How Has This Been Tested?
---
Tests required

---
Tests to be created in subsequent PR

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


dan_layer/common_types/src/committee.rs
```diff
@@ -71,25 +71,41 @@ impl<TAddr: NodeAddressable> FromIterator<Committee<TAddr>> for Committee<TAddr>
 #[derive(Debug, Clone, Copy)]
 pub struct CommitteeShard {
     num_committees: u64,
+    num_local_members: u64,
```
stringhandler: u32 would probably be fine here

dan_layer/consensus/src/hotstuff/on_receive_vote.rs
```diff
@@ -1,109 +1,142 @@
...
+        let committee = self.epoch_manager.get_local_committee(message.epoch).await?;
+        if !committee.contains(&addr) {
+            return Err(HotStuffError::ReceivedMessageFromNonCommitteeMember {
+                epoch: message.epoch,
+                sender: from.to_string(),
+                context: "OnVoteReceived".to_string(),
+            });
+        }
+
+        if !self.leader_strategy.is_leader(&addr, &committee, &message.block_id, 0) {
```
stringhandler, line 64: There are actually two leaders involved here. The leader that proposes a block, and then the leader who proposes the next block will be the one who receives the vote. Given that, it's not clear whether this logic is correct. It should maybe use a different method `is_leader_for_receiving_votes()` or `is_leader_for_next_block`

dan_layer/consensus/src/hotstuff/on_receive_vote.rs
```diff
@@ -1,109 +1,142 @@
...
+            return Err(HotStuffError::NotTheLeader {
+                details: format!("Not this leader for block {}, vote sent by {}", message.block_id, addr),
+            });
+        }
+
+        let local_commmittee_shard = self.epoch_manager.get_local_committee_shard(message.epoch).await?;
+
+        // Get the sender shard, and check that they are in the local committee
+        let sender_shard = self
```
stringhandler: Seems to be a mismatch of committees and shards. Is shard the shard space or the individual state address?

dan_layer/consensus/src/hotstuff/on_receive_vote.rs
```diff
@@ -1,109 +1,142 @@
...
+            .epoch_manager
+            .get_validator_shard(message.epoch, from.clone())
+            .await?;
+        if !local_commmittee_shard.includes_shard(&sender_shard) {
```
stringhandler: wouldn't the local_committee_shard only be 1 shard?

dan_layer/consensus/src/hotstuff/on_receive_vote.rs
```diff
@@ -1,109 +1,142 @@
...
+            decision: message.decision,
+            sender: sender_shard,
+            signature: message.signature,
+            merkle_proof: message.merkle_proof,
+        }
+        .save(&mut tx)?;
+
+        let count = Vote::count_for_block(tx.deref_mut(), &message.block_id)?;
+
+        if count < local_commmittee_shard.quorum_threshold() {
```
stringhandler: This handles the case where you have less than the required votes, but you also need a check for whether you have constructed a QC already and distributed it. 

In other words, when a vote comes in over the threshold, what happens. 
For example, in a 4 member committee, the 3rd vote comes in, we continue the logic below. When the 4th vote comes in, we will repeat this process, when (I believe) we shouldn't

dan_layer/consensus/src/hotstuff/on_receive_vote.rs
sdbondi: This returns a ShardId, it is the ShardId/Key of the sender. The `CommitteeShard` struct is all the info we need to determine if a shard is in the shard space. I think we need to find some ubiquitous language for these concepts. 

Currently I have:
CommitteeShard = a "slice" of the 256-bit shard space. 
ShardId = a single shard id, a single u256 number. This ID applies to substates and validators
Bucket = (ShardBucket?) is the index of the CommitteeShard in the 256-bit shard space.

dan_layer/consensus/src/hotstuff/on_receive_vote.rs
sdbondi: A `CommitteeShard` represents the shard space for a committee. So includes_shard will return true for any ShardId that falls within a start and end range (represented by the bucket and num_committees)

```
get_shard_range(bucket: int) -> Range<ShardId> {
  committee_shard_size = u256::MAX / num_committees
  start = bucket * committee_shard_size
  end = (bucket + 1) * committee_shard_size

  return start..end
}
```

dan_layer/consensus/src/hotstuff/on_receive_vote.rs
sdbondi: ah yup you are right - good catch

dan_layer/consensus/src/hotstuff/on_receive_vote.rs
sdbondi: The votes are sent to the leader for the block which was originally proposed by another leader. So here we check that we're the leader for the block being voted for. `leader_for_next_block(previous) == is_leader(current)`

dan_layer/consensus/src/hotstuff/on_receive_vote.rs
sdbondi: Improved the names and added some comments

dan_layer/consensus/src/hotstuff/on_receive_vote.rs
stringhandler: ok, yeah can be a bit clearer, but I think this is fine for now. If it's wrong it'll be very apparent in early testing

dan_layer/consensus/src/hotstuff/on_receive_vote.rs
stringhandler: alright, we should standardize on terminology but doesn't need to be corrected in this pr

PR #569, opened 2023-06-15
feat: template composability

mrnaveira:
Description
---
* New `template_lib` utilities available to template developers for creating composable templates:
    * `TemplateManager` to allow template function calls
    * The existing `ComponentManager` now allows to get other components and call methods on them
    * `invoke_args!` macro for CBOR argument encoding. It already existed but was only used internally
* New operation code `EngineOp::CallInvoke` in the `template_abi` for composable calls 
* The engine runtime implements the new `CallInvoke` operation (with two actions `call_method` and `call_function`) reusing the `TransactionManager` to execute the calls

Motivation and Context
---
One of the key features we want to provide is the ability to compose templates together, so that one function or method can internally call another function or method in other template/component. This allows developers to create a much richer ecosystem of templates that can reuse other templates/components as building blocks.

Template developers can use composability like this: 
```
impl MyTemplate {
        // typical use case of a constructor initializing and storing other component
        pub fn new() -> Self {
            let other_component_address = TemplateManager::get(NESTED_TEMPLATE_ADDRESS)
                .call("new".to_string(), vec![]);
            Self { other_component_address }
        }

        // at any moment the component can call the other component
       pub fn composed_message(&self) -> String {
            let message: String = ComponentManager::get(self.other_component_address);
                  .call("message".to_string(), invoke_args![self.some_value]);
            format!("composed: {}", message)
        }
 }
```

The `invoke_args!` macro, already present in the `template_lib`, is now public and it's used for conveniently encode the array of arguments for calls.

For safety reasons, **authorizations (i.e. owner tokens) are only permitted at the first level** and not passed down to nested components. That way composable calls can only be done to functions/methods that are not protected. We want to avoid the possibility of malicious behaviour without the knowledge of the user/developer (e.g. maliciously account withdrawals).

How Has This Been Tested?
---
New engine test template (`Composability`) used for:
* All possible combinations of calls between template functions and component methods
* Permissions cannot be abused in malicious methods (e.g. account withdrawals in nested components)
* Call depth limits

---
Create a new template (or use the new test one) to created composed components that call each other

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


stringhandler (dismissed):
Untested but looks great

sdbondi (dismissed):
utACK

Looks great!

Minor: I don't think that `max_call_recursion_depth`  doesn't need to be a consensus constant as we'll probably not change it between networks. It would probably only change with template engine versions.

Really nice tests and good call to not propagate the auth scope, think we can work through that once we work on auth overall

mrnaveira:
> utACK
> 
> Looks great!
> 
> Minor: I don't think that `max_call_recursion_depth` doesn't need to be a consensus constant as we'll probably not change it between networks. It would probably only change with template engine versions.
> 
> Really nice tests and good call to not propagate the auth scope, think we can work through that once we work on auth overall

Thanks for your comments. On point with the `max_call_recursion_depth` thing. I refactored it to be a simple (public) constant `MAX_CALL_DEPTH` on the engine's `TransactionProcessor`, so it's now outside of the consensus constants

sdbondi (approved):

PR #574, opened 2023-06-16
fix(consensus): batched consensus fixes, test infrastructure and tests

sdbondi:
Description
---
fix(consensus): on beat triggered conditionally rather than every 1 second
fix(consensus): mark transactions in local-only blocks as ready immediately
fix(consensus): don't double-propose the same block
fix(consensus): dont move  transactions into pools when constructing a block (happens in on recv proposal)
fix(consensus): correctly sign and validate vote signatures
feat(sqlite_state_store): separate table for QCs 
feat(consensus): QCs are referenced in blocks by their ID
tests(consensus): adds consensus test infrastructure 
tests(consensus): adds single committee tests

Motivation and Context
---
Testing revealed a number of issues. Multi-shard blocks are not supported yet so tests were not added.

How Has This Been Tested?
---
Tests pass

---
Run new tests

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


dan_layer/consensus/src/hotstuff/on_propose.rs
```diff
@@ -64,26 +61,55 @@ where
...
+                    parent_block.height(),
+                    child.id(),
+                    child.height(),
+                );
+                tx.rollback()?;
+                return Ok(());
+            }
+
+            next_block = self.build_next_block(&mut *tx, epoch, &parent_block, high_qc, validator_id)?;
+            if next_block.exists(&mut *tx)? {
```
stringhandler, line 89: this check is probably not necessary since it should be caught by the above statement, but probably doesn't hurt

dan_layer/consensus/src/hotstuff/on_propose.rs
```diff
@@ -144,11 +171,14 @@ where
         const TARGET_BLOCK_SIZE: usize = 1000;
         let mut remaining_txs = TARGET_BLOCK_SIZE;
 
-        let commit_txs = PrecommitTransactionPool::move_many_to_committed(tx, remaining_txs)?;
+        // TODO: Decide if we should lock the txs to the next block somehow
```
stringhandler, line 174: This should be fine (tm) since the replica will receive this block via loopback and it should in almost all cases be locked anyway

stringhandler (approved):
always difficult to assess whether a big PR like this should go in or wait for all the functionality, but I'm happy with it as is.

dan_layer/consensus/src/hotstuff/on_propose.rs
sdbondi: yeah you're right, i put that in at first but then thought of the above check after and forgot to remove this one. will leave it in for now, as you say, doesnt hurt

PR #575, opened 2023-06-21
feat: add account nft get and list methods on the daemon

jorgeantonio21:
Description
---

Adds two account nfts methods for the wallet daemon:

1. `get`: Given a `NonFungibleId` it gets the corresponding account NFT metadata;
2. `list`: Given `limit` and `offset` `u64` values, it gets all the NFT metadata available for the given account, with `limit` and `offset`.

Motivation and Context
---

How Has This Been Tested?
---

---

Mint some account NFTs using the wallet daemon/wallet dan cli and then query these, using:

1. `cargo run --bin tari_dan_wallet_cli -- nfts get --nft_id <NON_FUNGIBLE_ID>`

2. `cargo run --bin tari_dan_wallet_cli -- nfts list --limit <LIMIT> --offset <OFFSET>`,

where `--limit` and `--offset` are optional in 2.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


applications/tari_dan_wallet_cli/src/command/nfts.rs
```diff
@@ -56,12 +64,28 @@ pub struct MintAccountNftArgs {
...
 
+#[derive(Debug, Args, Clone)]
+pub struct GetAccountNftArgs {
+    #[clap(long, short = 'i', alias = "nft_id")]
+    pub nft_id: String,
+}
+
+#[derive(Debug, Args, Clone)]
+pub struct ListAccountNftArgs {
+    #[clap(long, short = 'l', alias = "limit")]
```
sdbondi: nit: alias to the same name as the field is redundant
```suggestion
    #[clap(long, short = 'l')]
```

dan_layer/template_lib/Cargo.toml
```diff
@@ -15,6 +15,7 @@ tari_crypto = "0.16"
 
 newtype-ops = "0.1.4"
 serde = { version = "1.0.143", default-features = false, features = ["derive", "alloc"] }
+serde_json = { version = "1.0.64" }
```
sdbondi: The template lib must have minimal dependencies as they will be compiled into every WASM.

dan_layer/template_lib/src/models/metadata.rs
```diff
@@ -45,6 +46,14 @@ impl Metadata {
...
         self.0.get(key).map(|v| v.as_str())
     }
+
+    pub fn into_json(self) -> Value {
+        serde_json::to_value(self.0).expect("Failed to parse to JSON")
+    }
+
+    pub fn as_json(&self) -> Value {
+        serde_json::to_value(&self.0).expect("Failed to parse JSON")
+    }
```
sdbondi: Remove these and serialize outside of the template_lib crate

applications/tari_dan_wallet_daemon/src/handlers/nfts.rs
```diff
@@ -37,6 +45,56 @@ use crate::{
...
+    sdk.jwt_api().check_auth(token.clone(), &[JrpcPermission::Admin])?;
+
+    let non_fungible_api = sdk.non_fungible_api();
+
+    let non_fungible = non_fungible_api
+        .non_fungible_token_get_by_nft_id(req.nft_id)
+        .map_err(|e| anyhow!("Failed to get non fungible token, with error: {}", e))?;
+    let token_symbol = non_fungible.token_symbol.clone();
+    let is_burned = non_fungible.is_burned;
+    let metadata = non_fungible.metadata.into_json();
```
sdbondi: To avoid adding serde_json as a dependency to every wasm 
```suggestion
    let metadata = serde_json::to_value(&non_fungible.metadata)?;
```

dan_layer/wallet/storage_sqlite/src/reader.rs
```diff
@@ -665,29 +668,68 @@ impl WalletStoreReader for ReadTransaction<'_> {
...
+            .offset(offset as i64)
+            .load::<models::NonFungibleToken>(self.connection())
+            .map_err(|e| WalletStorageError::general("non_fungible_token_get_all", e))?;
+
+        let vault_ids = non_fungibles.iter().map(|n| n.vault_id).collect::<Vec<_>>();
+        let vault_addresses = vaults::table
+            .select(vaults::address)
+            .filter(vaults::id.eq_any(vault_ids))
+            .load::<String>(self.connection())
+            .map_err(|e| WalletStorageError::general("accounts_get_by_vault", e))?;
```
sdbondi: neat trick: eq_any also takes an iterator so no need to collect it into a buffer first
```suggestion
        let vault_ids = non_fungibles.iter().map(|n| n.vault_id);
        let vault_addresses = vaults::table
            .select(vaults::address)
            .filter(vaults::id.eq_any(vault_ids))
            .load::<String>(self.connection())
            .map_err(|e| WalletStorageError::general("accounts_get_by_vault", e))?;
```

sdbondi (commented):
Nice LGTM, only barrier is adding an unncessary dependency to the template lib

applications/tari_dan_wallet_cli/src/command/nfts.rs
jorgeantonio21: Nice !

dan_layer/wallet/storage_sqlite/src/reader.rs
jorgeantonio21: Cool ! Thanks for the hint :)

stringhandler (approved):
Nice. In future, we should add the ability to create (and verify) a proof of ownership and existence, but this is useful, thanks!

sdbondi (approved):

PR #576, opened 2023-06-22
feat: dry run transactions in the indexer through the wallet

mrnaveira:
Description
---
* Created new dry run processor in the indexer
    * Immediately returns the result to the caller on `submit_transaction` if the `is_dry_run` option is enabled
    * Simulates fees if there are any instructions in the `fee_instructions` field
* Indexer now scans and stores templates (needed for executing the transactions)
* Wallet daemon (and CLI) now properly handle the **dry run** option (up until now it was ignored)
* The utility crate `tari_dan_app_utilities` now contains the `TemplateProvider` as well as the `PayloadProcessor`. They are reused in both `tari_validator_node` and `tari_indexer`
* The wallet daemon now does not store a transaction in the database if it's a dry run
* Renamed in the wallet daemon `substate_provider.rs` to `network.rs`, to better represent the logic of that module

Motivation and Context
---
**Dry run** transactions are transactions that are simulated, that means the wasm is executed like a regular transaction but they are **not committed into the network** (no consensus, no side effects, etc.). This is useful for testing purposes and to call read-only methods without paying fees.

Currently only the validator node supports dry run transactions. We need a way to submit dry run transactions in the wallet daemon. A very easy solution could be to simply make the indexer proxy the dry run request to the validator node, but there is value on **executing the dry run transaction in the indexer**:
* Read-only transactions will be very frequent. If they are executed in the indexer we reduce the load on validator nodes (which are the ones responsible for consensus).
* Simulating a transaction in the indexer will allow us in the future to calculate a more precise input/output autofill

**TBD**:
* We may consider removing the dry run processor from the validator node, as it's a user feature and VN's should only focus on consensus. But this is debatable so I left it out of this PR.
* There is further improvement on the fee simulation, right now it requires valid `fee_instruction` to simulate fees. It would be better if the fees are always simulated in each transaction regardless.

How Has This Been Tested?
---
Manually submitted dry run transactions on the wallet CLI. For example:
```
cargo run --bin tari_dan_wallet_cli -- transactions submit --dry-run --num-outputs 2 call-function -a 100000 fe6b6a4ada71fd292d21235ee33289d3ed2a23f679334e65e1d7f9b314e790f2 mint
```
Also tested dry run executions (including fee estimation) using the tariswap web UI.

---
Manually submit dry run transactions using the wallet CLI or the JSON RPC in the wallet daemon

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


applications/tari_dan_wallet_cli/src/command/transaction.rs
```diff
@@ -260,13 +260,19 @@ pub async fn handle_submit(args: SubmitArgs, client: &mut WalletDaemonClient) ->
         });
     }
 
-    let request = TransactionSubmitRequest {
-        signing_key_index: None,
-        fee_instructions: vec![Instruction::CallMethod {
+    let fee_instructions = if common.dry_run {
+        vec![]
```
sdbondi: Any reason why we remove fee instructions here? Dry Run should be able to simulate fees too. Could be useful for determining the exact fee to pay.

dan_layer/wallet/sdk/src/network.rs
```diff
@@ -35,6 +39,7 @@ pub struct SubstateQueryResult {
 }
 
 #[derive(Debug, Clone, Deserialize, Serialize)]
-pub struct TransactionQueryResult {
+pub struct TransactionResult {
```
sdbondi: Happy to rename, but the main reason I named this TransactionQueryResult is because there is already a TransactionResult type.

sdbondi (commented):
Looks good so far, still need to go over all the code and run some tests

dan_layer/wallet/sdk/src/network.rs
mrnaveira: Fair point. To avoid any confusion with the already existing `TransactionResult`  type, I just reverted the renaming back to `TransactionQueryResult`

sdbondi (dismissed):
utACK

LGTM, we can decide if we want fees in dry run, maybe optional or maybe always required - one main use case is calculating exact fees.

applications/tari_dan_wallet_cli/src/command/transaction.rs
mrnaveira: Updated the dry run to automatically simulate fees if there is any `fee_instructions` in the transaction, which should include the required amount as well. 

IMO for now it's good enough. It would be better to simulate it always regardless of `fee_instructions`, but it would require a lot of refactor inside the transaction processor and tracker.

sdbondi (dismissed):

applications/tari_dan_wallet_cli/src/command/transaction.rs
sdbondi: Thanks LGTM

stringhandler (dismissed):
utACK

stringhandler (approved):

PR #580, opened 2023-06-29
chore!: update hash domain consistency

brianp:
Description
---
Update the hash domains to be
- more consistent
- align with core

Motivation and Context
---
Consistency is key

How Has This Been Tested?
---
It has not been

Breaking Changes
---

- [ ] None
- [x] Other - Please specify

This is a breaking change, old nodes may not be able to communicate with new nodes as the message signatures will be different.


brianp:
Ideally some of these should be imported from core, and not re-defined.

sdbondi:
Thanks, I'm in favour of a tari repo crate that exposes the shared hash domains, tari-core comes with a lot of dependencies and a lot of code we don't need in the DAN, I suspect it's already a dependency in places just for 1 or 2 hash domains.

sdbondi (approved):

stringhandler (approved):

PR #581, opened 2023-06-30
feat: jrpc for get all txs

Cifko:
Description
---
Add jrpc call the get all transactions, either by status or all at once.

Motivation and Context
---
We need this for the UI.

How Has This Been Tested?
---
Manually.

---
Call the new jrpc method on a wallet with some transactions.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (dismissed):

applications/tari_dan_wallet_daemon/src/handlers/transaction.rs
```diff
@@ -184,6 +186,27 @@ pub async fn handle_get(
     })
 }
 
+pub async fn handle_get_all_by_status(
```
jorgeantonio21, line 189: I would suggest rename it to `handle_get_all_txs_by_status`

applications/tari_dan_wallet_daemon/src/jrpc_server.rs
```diff
@@ -104,6 +104,7 @@ async fn handler(
             "get" => call_handler(context, value, token, transaction::handle_get).await,
             "get_result" => call_handler(context, value, token, transaction::handle_get_result).await,
             "wait_result" => call_handler(context, value, token, transaction::handle_wait_result).await,
+            "get_all_by_status" => call_handler(context, value, token, transaction::handle_get_all_by_status).await,
```
jorgeantonio21, line 107: Could we rename it to `get_all_txs_by_status`? Or `get_all_transacations_by_status`.

clients/wallet_daemon_client/src/types.rs
```diff
@@ -108,6 +108,11 @@ pub struct TransactionGetRequest {
     pub hash: FixedHash,
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct TransactionGetAllRequest {
```
jorgeantonio21, line 112: Can we rename the struct to `GetAllTransactionsRequest` or `GetAllTransactionsByStatusRequest` ?  I think we use `Get` at the beginning in the previous examples.

clients/wallet_daemon_client/src/types.rs
```diff
@@ -124,6 +129,16 @@ pub struct TransactionGetResponse {
     pub transaction_failure: Option<RejectReason>,
 }
 
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct TransactionGetAllResponse {
```
jorgeantonio21, line 133: Can we rename it to `GetAllTransactionsResponse` or to `GetAllTransactionsByStatusResponse` ?

jorgeantonio21 (approved):
LGTM !

nit: I proposed to do some renaming, that I think it helps a lot on the reading.

applications/tari_dan_wallet_daemon/src/jrpc_server.rs
Cifko: it's a `transacation.get_all_by_status` command, that sounds redundant to me and also not consistent with other transactions commands.

applications/tari_dan_wallet_daemon/src/handlers/transaction.rs
Cifko: Redundant and not consistent with rest of the file. e.g. just after this is `handle_get_result`

clients/wallet_daemon_client/src/types.rs
Cifko: I want to keep it consistent with `TransactionGetRequest` `TransactionSubmitRequest` etc.

clients/wallet_daemon_client/src/types.rs
Cifko: Same :-)

sdbondi (approved):

PR #584, opened 2023-07-05
fix(ci): use nightly only for cargo fmt

sdbondi:
Description
---

Use nightly for cargo fmt and stable for everything else.
Remove tari_crypto dependency from template lib

Motivation and Context
---
We use unstable cargo fmt options that require nightly, for testing and clippy we can use stable.
Clippy currently doesnt compile with nightly e.g. https://github.com/tari-project/tari-dan/actions/runs/5455992887/jobs/9928182920?pr=577 

Tari crypto adds a [bunch](https://github.com/tari-project/tari-dan/pull/584/files#diff-d9d5a2564a9e608c65be8c61f26887fa7953149a9c3a50b375b840c9d95b77d4) of dependencies to all WASM templates. There is a zero dependency type for public keys that should be used. Tari crypto could still be used directly in templates but template lib dependencies should be kept to an absolute minimum.

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


stringhandler (dismissed):
Looks good. Will have to see CI if it works

sdbondi:
No idea why/how the linker is set to mold and if it should be, but linker fails without it and it seems to work

stringhandler (approved):
Happy. Yeah, we only need the bytes version of the public key for now. I think we might need a stripped-down version of tari-crypto for other use cases, but I think almost all crypto operations should happen in rust in the engine runtime. Running them in wasm would create really slow operating times and could slow down consensus

PR #577, opened 2023-06-23
feat(consensus): implement simpler batching co-ordination

sdbondi:
Description
---
- feat: use Prepare, LocalPrepared, and Accept commands in block
- refactor: use single transaction pool with New, LocalPrepared, AllPrepared, SomePrepared stages
- feat: BlockCommitted event
- feat: replaced pledges with substate multi-read/single-write lock 
- test: implement multi-shard consensus tests
- test: implement virtual network that connects all consensus instances 

Motivation and Context
---
Implement ideas from the [chainspace paper](https://arxiv.org/abs/1708.03778) - thanks @stringhandler for conceptualising this.

How Has This Been Tested?
---
Implemented multi-shard functional tests in the `consensus_tests` crate.

What process can a PR reviewer use to test or verify this change?
---
Only by running the tests as consensus has not been integrated with the validator node yet

Breaking Changes
---

- [x] None
- [ ] Requires data directory to be deleted
- [ ] Other - Please specify

PR #590, opened 2023-07-06
ci: fix testing jobs

stringhandler:
Description
---
Separate the unit tests from the cucumber tests and publish results for both. 
Fixes the path of the cucumber junit result file

Motivation and Context
---
The cucumber result file had moved in a previous PR where the integration tests were moved from the validator_node crate to their own crate. This move also makes it easier to call `cargo nextest` which provides us with unit test results in a junit format.

How Has This Been Tested?
---
Ran the commands for nextest and cucumber manually, but will have to see how it runs on CI to see if the GHA files are correct. 

---
N/A

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -269,7 +269,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --all-features --release --
+          args: --all-features --release --package integration_tests
       - name: upload artifact
         uses: actions/upload-artifact@v3  # upload test results as artifact
         if: success() || failure()
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -261,7 +261,7 @@ jobs:
         run: cargo nextest run --all-features --release -E "not package(integration_tests)" --profile ci
       - name: upload artifact
         uses: actions/upload-artifact@v3  # upload test results as artifact
-        if: always()
+        if: success() || failure()
         with:
           name: test-results
           path: ${{ github.workspace }}/target/nextest/ci/junit.xml
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -249,22 +249,33 @@ jobs:
       - uses: rui314/setup-mold@v1
       - name: wasm target install
         run: rustup target add wasm32-unknown-unknown
+      - name: Install cargo-nextest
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -249,22 +249,33 @@ jobs:
...
+        if: always()
+        with:
+          name: test-results
+          path: ${{ github.workspace }}/target/nextest/ci/junit.xml
+      - name: cargo test cucumber
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: -v --all-features --release
+          args: --all-features --release --
```
sdbondi: is '--' intentional?

sdbondi (dismissed):
Nice!

sdbondi:
This test `multi_validator_propose_blocks_with_new_transactions_until_all_committed` is flaky, I did pick that up and it should be fixed in https://github.com/tari-project/tari-dan/pull/577. Restarted the test.

EDIT: lol guess it's just broken, at least confirms this is the right change to CI. It always passes locally so very strange.

I've ignored it for now since #577 should fix it

sdbondi (dismissed):

.github/workflows/ci.yml
stringhandler: ah wait

.github/workflows/ci.yml
```diff
@@ -249,22 +249,33 @@ jobs:
...
         uses: actions-rs/cargo@v1
         with:
           command: test
           args: --no-run --locked --all-features --release
+
       - name: cargo test
+        run: cargo nextest run --all-features --release -E "not package(integration_tests)" --profile ci
+      - name: upload artifact
+        uses: actions/upload-artifact@v3  # upload test results as artifact
+        if: always()
```
CjS77: You want to upload on cancels as well I gather?

CjS77 (dismissed):

.github/workflows/ci.yml
stringhandler: ah no, this should be success() || failure() as well, thanks

stringhandler: Not 100% sure, but `always()` seems to change the outcome to success if the previous step failed

sdbondi (approved):

PR #592, opened 2023-07-10
ci: new artifacts

stringhandler:
Description
---
Creates new artifacts for releases

Motivation and Context
---
Similar to the base layer, this job will create executables that can be downloaded without compiling from source.

How Has This Been Tested?
---
It's CI files, so will be testing on github CI

---
N/A

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


jorgeantonio21 (approved):
utAck

PR #593, opened 2023-07-10
feat: request missing TXs

Cifko:
Description
---
Request TXs in a proposal that the VN doesn't know of. It sends all the ids of the missing transactions to the leader. And store the information into two tables. One is to know which transaction belong to which block, and the other one is to know which block is missing which transaction. This way the look ups are fast.

How Has This Been Tested?
---
We have a test `consensus::node_requests_missing_transaction_from_local_leader`. There are two nodes, but only one receives the transactions. The other one receives the block and asks for the transactions.

---
You can run the same test.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/dan_layer/consensus/src/hotstuff/on_receive_proposal.rs
+++ b/dan_layer/consensus/src/hotstuff/on_receive_proposal.rs
@@ -12,6 +12,7 @@ use tari_dan_common_types::{
 use tari_dan_storage::{
     consensus_models::{
         Block,
+        BlockId,
         Command,
         Decision,
         ExecutedTransaction,
@@ -26,6 +27,7 @@ use tari_dan_storage::{
     },
     StateStore,
     StateStoreReadTransaction,
+    StateStoreWriteTransaction,
 };
 use tari_epoch_manager::EpochManagerReader;
 use tari_transaction::Transaction;
@@ -116,14 +118,45 @@ where TConsensusSpec: ConsensusSpec
         }
     }
 
+    async fn block_has_missing_transaction(
+        &self,
+        local_committee: &Committee<TConsensusSpec::Addr>,
+        block: &Block,
+    ) -> Result<bool, HotStuffError> {
+        let mut missing_tx_ids = Vec::new();
+        self.store.with_read_tx(|tx| {
+            for tx_id in block.all_transaction_ids() {
+                if tx.transactions_get(tx_id).is_err() {
+                    missing_tx_ids.push(*tx_id);
+                }
+            }
+            Ok::<_, HotStuffError>(())
+        })?;
+        if missing_tx_ids.is_empty() {
+            Ok(false)
+        } else {
+            self.store
+                .with_write_tx(|tx| tx.insert_missing_transactions(block.id(), missing_tx_ids.clone()))?;
+            self.send_to_leader(
+                local_committee,
+                block.id(),
+                HotstuffMessage::RequestMissingTx(RequestMissingTransactionsMessage {
+                    block_id: *block.id(),
+                    epoch: block.epoch(),
+                    transactions: missing_tx_ids,
+                }),
+            )
+            .await?;
+            Ok(true)
+        }
+    }
+
     async fn handle_local_proposal(
         &self,
         from: TConsensusSpec::Addr,
         local_committee: Committee<TConsensusSpec::Addr>,
         block: Block,
     ) -> Result<(), HotStuffError> {
-        let local_committee_shard = self.epoch_manager.get_local_committee_shard(block.epoch()).await?;
-
         // First save the block in one db transaction
         self.store.with_write_tx(|tx| {
             self.validate_local_proposed_block(&mut *tx, &from, &block)?;
@@ -133,12 +166,31 @@ where TConsensusSpec: ConsensusSpec
             Ok::<_, HotStuffError>(())
         })?;
 
+        if self.block_has_missing_transaction(&local_committee, &block).await? {
+            Ok(())
+        } else {
+            self.process_block(&local_committee, &block).await
+        }
+    }
+
+    pub async fn reprocess_block(&self, block_id: &BlockId) -> Result<(), HotStuffError> {
+        let block = self.store.with_read_tx(|tx| tx.blocks_get(block_id))?;
+        let local_committee = self.epoch_manager.get_local_committee(block.epoch()).await?;
+        self.process_block(&local_committee, &block).await
+    }
+
+    async fn process_block(
+        &self,
+        local_committee: &Committee<<TConsensusSpec as ConsensusSpec>::Addr>,
+        block: &Block,
+    ) -> Result<(), HotStuffError> {
+        let local_committee_shard = self.epoch_manager.get_local_committee_shard(block.epoch()).await?;
         let maybe_decision = self.store.with_write_tx(|tx| {
-            let should_vote = self.should_vote(&mut *tx, &block)?;
+            let should_vote = self.should_vote(&mut *tx, block)?;
 
             let mut maybe_decision = None;
             if should_vote {
-                maybe_decision = self.decide_what_to_vote(tx, &block, &local_committee_shard)?;
+                maybe_decision = self.decide_what_to_vote(tx, block, &local_committee_shard)?;
             }
 
             self.update_nodes(tx, &block, &local_committee_shard)?;
@@ -146,17 +198,16 @@ where TConsensusSpec: ConsensusSpec
         })?;
 
         if let Some(decision) = maybe_decision {
-            let vote = self.generate_vote_message(&block, decision).await?;
+            let vote = self.generate_vote_message(block, decision).await?;
             debug!(
                 target: LOG_TARGET,
-                "🔥 Send {:?} VOTE for block {}, parent {}, height {} to {}",
+                "🔥 Send {:?} VOTE for block {}, parent {}, height {}",
                 decision,
                 block.id(),
                 block.parent(),
                 block.height(),
-                from,
             );
-            self.send_to_leader(&local_committee, vote).await?;
+            self.send_vote_to_leader(local_committee, vote).await?;
         }
 
         Ok(())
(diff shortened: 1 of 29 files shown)


dan_layer/consensus_tests/src/support/harness.rs
```diff
@@ -88,6 +93,17 @@ impl Test {
         self.validators.values().for_each(f);
     }
 
+    pub async fn wait_until_new_pool_count_for_vn(&self, count: usize, vn: TestAddress) {
+        self.wait_all_for_predicate(format!("new pool count to be {}", count), |v| {
+            v.address != vn ||
+                v.state_store
+                    .with_read_tx(|tx| tx.transaction_pool_count(None, None))
```
sdbondi: Did you mean this?
```suggestion
                    .with_read_tx(|tx| tx.transaction_pool_count(Some(TransactionPoolStage::New), None))
```

dan_layer/consensus/src/hotstuff/on_receive_proposal.rs
```diff
@@ -133,30 +166,48 @@ where
...
 
+        if self.block_has_missing_transaction(&local_committee, &block).await? {
+            Ok(())
+        } else {
+            self.process_block(&local_committee, &block).await
+        }
+    }
+
+    pub async fn reprocess_block(&self, block_id: &BlockId) -> Result<(), HotStuffError> {
+        let block = self.store.with_read_tx(|tx| tx.blocks_get(block_id))?;
```
sdbondi: Should use the record types in the consensus layer
```suggestion
        let block = self.store.with_read_tx(|tx| Block::get(tx, block_id))?;
```

dan_layer/consensus/src/hotstuff/on_receive_request_missing_transactions.rs
```diff
@@ -0,0 +1,58 @@
...
+    pub async fn handle(
+        &mut self,
+        from: TConsensusSpec::Addr,
+        msg: RequestMissingTransactionsMessage,
+    ) -> Result<(), HotStuffError> {
+        info!(target: LOG_TARGET, "{:?} is requesting missing transactions from block {} with ids {:?}", from,msg.block_id, msg.transactions);
+        let txs = self
+            .store
+            .with_read_tx(|tx| tx.transactions_get_many(&msg.transactions));
+        let txs = txs?;
```
sdbondi:
```suggestion
        debug!(target: LOG_TARGET, "{:?} is requesting missing transactions from block {} with ids {:?}", from,msg.block_id, msg.transactions);
        let txs = self
            .store
            .with_read_tx(|tx| ExecutedTransaction::get_many(tx, &msg.transactions))?;
```

dan_layer/consensus/src/messages/requested_transaction.rs
```diff
@@ -0,0 +1,13 @@
...
+
+use serde::Serialize;
+use tari_dan_common_types::Epoch;
+use tari_dan_storage::consensus_models::{BlockId, ExecutedTransaction};
+
+#[derive(Debug, Clone, Serialize)]
+pub struct RequestedTransactionMessage {
+    pub epoch: Epoch,
+    pub block_id: BlockId,
+    pub transactions: Vec<ExecutedTransaction>,
```
sdbondi: Should be sending the transaction without the execution result
```suggestion
    pub transactions: Vec<Transaction>,
```

dan_layer/consensus/src/hotstuff/on_receive_request_missing_transactions.rs
```diff
@@ -0,0 +1,58 @@
...
+        self.tx_leader
+            .send((
+                from,
+                HotstuffMessage::RequestedTransaction(RequestedTransactionMessage {
+                    epoch: msg.epoch,
+                    block_id: msg.block_id,
+                    transactions: txs,
```
sdbondi:
```suggestion
                    transactions: txs.into_iter().map(|tx|tx.into_transaction()).collect(),
```

dan_layer/consensus/src/hotstuff/on_receive_proposal.rs
```diff
@@ -116,14 +118,45 @@ where
...
 
+    async fn block_has_missing_transaction(
+        &self,
+        local_committee: &Committee<TConsensusSpec::Addr>,
+        block: &Block,
+    ) -> Result<bool, HotStuffError> {
+        let mut missing_tx_ids = Vec::new();
+        self.store.with_read_tx(|tx| {
+            for tx_id in block.all_transaction_ids() {
+                if tx.transactions_get(tx_id).is_err() {
```
sdbondi:
```suggestion
                if ExecutedTransaction::exists(tx, tx_id)? {
```

dan_layer/consensus/src/hotstuff/worker.rs
```diff
@@ -232,17 +249,23 @@ where
         from: TConsensusSpec::Addr,
         msg: HotstuffMessage,
     ) -> Result<(), HotStuffError> {
+        dbg!(&self.validator_addr);
```
sdbondi: remove `dbg!`s

dan_layer/consensus/src/hotstuff/on_receive_requested_transactions.rs
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    pub async fn handle(
+        &mut self,
+        from: TConsensusSpec::Addr,
+        msg: RequestedTransactionMessage,
+    ) -> Result<(), HotStuffError> {
+        info!(target: LOG_TARGET, "{:?} sent the requested transactions for block {}", from,msg.block_id);
+        for tx in msg.transactions {
+            self.tx_new_transaction
```
sdbondi: Not sure if the intention is to start consensus on transactions that have not been executed locally, but we'll need to send the transactions to the mempool to be executed - this channel can take a `Transaction` and we can hook that up.

dan_layer/consensus/src/hotstuff/on_receive_requested_transactions.rs
```diff
@@ -0,0 +1,49 @@
...
+    messages::RequestedTransactionMessage,
+    traits::{ConsensusSpec, EpochManager},
+};
+
+const LOG_TARGET: &str = "tari::dan::consensus::hotstuff::on_receive_requested_transactions";
+
+pub struct OnReceiveRequestedTransactions<TConsensusSpec: ConsensusSpec> {
+    tx_new_transaction: mpsc::Sender<ExecutedTransaction>,
+    _phantom: std::marker::PhantomData<TConsensusSpec>,
+}
```
sdbondi: This generic is unused
```suggestion
pub struct OnReceiveRequestedTransactions {
    tx_new_transaction: mpsc::Sender<ExecutedTransaction>,
}
```

sdbondi (commented):
Ok let's try this approach, it looks like it'll work, if we run into code bloat and lots of edge cases we can revisit.

We should see a lot of requests for transactions which are we have but are still executing. This is partially a symptom of the fact that we propagate transactions after executing them (to determine outputs, can be made cleverer) so there is always a close race between proposal and transaction (but the transaction has to execute so always takes longer) and partially because we dont have a check that checks this case in the DB. Both are very fixable.

Fair amount of work adding the GRPC conversions and hooking up channels etc to get this working. Let me know if you run into any trouble with these!

dan_layer/consensus_tests/src/support/harness.rs
Cifko: Maybe :-) I copy pasted the function bellow, in that case I will fix both.

dan_layer/consensus/src/hotstuff/on_receive_requested_transactions.rs
Cifko: I change this to send it to local mempool. I've added new trait function to outbound services to send_to_self, so I don't have to send my address around in the channels.

dan_layer/consensus/src/hotstuff/on_receive_requested_transactions.rs
Cifko: The `TConsensusSpec` is needed for the `from` field in handle.

Cifko:
I had to delete the test. Because now the transactions are send via mempool. But the test is not executing the transaction so they don't go via mempool. So there is no way for me to check that. We can talk about how to test this (maybe an integration tests will cover this).

dan_layer/consensus/src/hotstuff/on_receive_proposal.rs
```diff
@@ -133,30 +166,48 @@ where TConsensusSpec: ConsensusSpec
             Ok::<_, HotStuffError>(())
         })?;
 
+        if self.block_has_missing_transaction(&local_committee, &block).await? {
+            Ok(())
```
stringhandler, line 170: It might be useful for testing to return an enum saying which decision we took

```suggestion
            Ok(LocalProposalResult::MissingTransactions)
```

dan_layer/consensus/src/hotstuff/on_receive_request_missing_transactions.rs
```diff
@@ -0,0 +1,55 @@
...
+    hotstuff::error::HotStuffError,
+    messages::{HotstuffMessage, RequestMissingTransactionsMessage, RequestedTransactionMessage},
+    traits::ConsensusSpec,
+};
+
+const LOG_TARGET: &str = "tari::dan::consensus::hotstuff::on_receive_request_missing_transactions";
+
+pub struct OnReceiveRequestMissingTransactions<TConsensusSpec: ConsensusSpec> {
+    store: TConsensusSpec::StateStore,
+    tx_leader: mpsc::Sender<(TConsensusSpec::Addr, HotstuffMessage)>,
```
stringhandler: This might not only be the leader who processes this, but anyone who has it
```suggestion
    tx_request_missing_tx: mpsc::Sender<(TConsensusSpec::Addr, HotstuffMessage)>,
```

dan_layer/consensus/src/hotstuff/on_receive_requested_transactions.rs
```diff
@@ -0,0 +1,43 @@
...
+        }
+    }
+
+    pub async fn handle(
+        &mut self,
+        from: TConsensusSpec::Addr,
+        msg: RequestedTransactionMessage,
+    ) -> Result<(), HotStuffError> {
+        info!(target: LOG_TARGET, "{:?} sent the requested transactions for block {}", from,msg.block_id);
+        for tx in msg.transactions {
```
stringhandler, line 33: Might be useful to send these as a single message in batches, but can be optimized later

dan_layer/consensus/src/hotstuff/on_receive_proposal.rs
Cifko: In the test we don't call the function directly, it's inside a loop over tokio::select, so the return value doesn't bring any value for testing. But if the missing transactions is triggered we can check if the node sent the `RequestMissingTransactions`. That should be enough for testing.

dan_layer/consensus/src/hotstuff/on_receive_request_missing_transactions.rs
Cifko: yeah, the name was just an copy paste
renamed

dan_layer/consensus/src/hotstuff/on_receive_requested_transactions.rs
Cifko: The request from other node uses batch, but inserting into mempool is one by one. But at that's done locally, so it's just a propagation over some channels, not using comms, so it should be fast.

stringhandler (approved):
tested locally and working, nice one

sdbondi (approved):

PR #599, opened 2023-07-12
wip: feat: block propagation for non-leader members

Cifko:
Description
---
If a node receives a new local block, it checks if it needs to resend it to foreign committees. If the node receives a foreign block and it's not the leader, it sends it to the leader. The local leader is responsible for distributing the foreign block within its committee members.

Motivation and Context
---
We need to make sure that at least one honest node sends the block to at least one honest node in the foreign committee.

How Has This Been Tested?
---
There is a new test `foreign_block_distribution` where non of the leaders can send messages to the foreign committee. So other nodes are responsible to get the message across and to the leaders.

---
Read and run the `foreign_block_distribution`.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/dan_layer/consensus/src/hotstuff/on_receive_proposal.rs
+++ b/dan_layer/consensus/src/hotstuff/on_receive_proposal.rs
@@ -5,7 +5,7 @@
 // ----[foreign:LocalPrepared]--->(LocalPrepared, true) ----cmd:AllPrepare ---> (AllPrepared, true) ---cmd:Accept --->
 // Complete
 
-use std::ops::DerefMut;
+use std::{collections::HashSet, ops::DerefMut};
 
 use log::*;
 use tari_dan_common_types::{
@@ -22,6 +22,7 @@ use tari_dan_storage::{
         ExecutedTransaction,
         LastExecuted,
         LastVoted,
+        LeafBlock,
         LockedBlock,
         QuorumDecision,
         SubstateLockFlag,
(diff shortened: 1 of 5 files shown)


dan_layer/consensus/src/hotstuff/on_receive_proposal.rs
```diff
@@ -218,16 +348,60 @@ where TConsensusSpec: ConsensusSpec
...
+        // If we received the foreign proposal, we send it to the leader (if we are not the leader), the leader then
+        // redistributes the block to all other nodes. This way if the leader is not faulty O(n) messages will be send
+        // around. If the leader doesn't have the message it will take 2 delta (if the delta time is the maximum latency
+        // between any two nodes) to have it everywhere. If the leader has the message already it will be just 1 delta.
+        // Worst case scenario is when we have f faulty nodes, and 2f honest nodes have the message and 1 node doesnt,
+        // but he is the (f+1)th leader. In this case we send exactly 2f*f+2f+3f messages around. 2f*f to the
+        // faulty leaders, 2f to the honest leader, and 3f from the leader.
+        if self
+            .leader_strategy
+            .is_leader(&self.validator_addr, local_committee, block.id(), block.height())
```
stringhandler: This block will not be in our chain, so this call will not return the correct leader. You might have to get the `last_voted` block or highest block in our local chain to determine who our current leader is

Cifko: Done. I had to also change other methods so I can send the leader around when I want to send the foreign block.

PR #600, opened 2023-07-12
feat: new builtin template for atomic swaps

mrnaveira:
Description
---
* New builtin template `atomic_swap`, with the Tari side of a HTLC implementation
* New engine unit test for the `atomic_swap` template

Motivation and Context
---
We want to provide a method for users to buy or sell Tari resources in exchange from other network (Ethereum, Minotari, Monero, Bitcoin, etc.) tokens. The most decentralized approach of implementing this is through atomic swaps, where transactions in different networks can be intertwined in a way that either both execute or both are reverted.

This PR is the first step to have a functional atomic swap implementation in the Tari network. It provides a **builtin template for the Tari side of an atomic swap**, following a Hashed Time Locked Contract (HTLC) approach. HTLC is the most widespread and easier to implement method, so it's the ideal for a first iteration.

TBD in future PRs:
* Ethereum side of the atomic swap
    * This may force use us to use `kecckak256` or `sha3` instead of the `Blake256` that we use now in the Tari template
* User application to find and execute desired atomic swaps
* Include the block number in the consensus context, so we don't need to use epochs
* Switch to a more sophisticated implementation, like Point Time Locked Contracts (PTLCs).

How Has This Been Tested?
---
New unit test for all possible scenarios of the swap process of the new template

---
Create a couple of atomic swap components and simulate a swap

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


stringhandler (approved):
Nice. 

I think it's subjective whether it should be a builtin template, be a simple test or live in the cargo-generate template repo, but I'm fine with it living here for now

mrnaveira:
Closed as for now this template is going to be part of a separated atomic swap application, not a builtin template.

PR #620, opened 2023-07-14
feat!: integrate batched consensus

sdbondi:
Description
---
- feat!: integrate batched consensus
- fix(consensus): add last proposed bookkeeping variable to prevent duplicate proposal
- fix(consensus): unlock substates when committed/aborted
- fix(consensus): handle substate already locked case (TODO: change tx decision)
- fix(consensus): mark transaction as finalized once accepted 
- fix(consensus): remove erroneous check for votes being sent to proposer (sent to "next leader")
- fix(consensus): count majority (2f + 1) votes as QC decision
- fix(consensus): create new genesis block on epoch change
- feat(mempool): resolve foreign substates and execute transaction before consensus
- feat(mempool): fill transaction outputs
- fix(mempool): remove zero fee check for accounts in memool validator
- feat(epoch_manager): implement required interface for consensus
- refactor(autofiller): update for new transaction struct
- refactor(wallet-daemon): update for new transaction struct
- refactor(indexer): update for new consensus/transaction struct

Motivation and Context
---
Integrate batched consensus changes into the validator node and other applications.

Limitations:
1. Multi-VN can fail due to the known race condition partially fixed in #593 
2. Multi-shard is not well tested and will likely fail due to above race condition
3. Some failure cases are not handled correctly and result in a transaction being stuck
4. RPC calls for transaction result dont seem correct, wallet unable to determine a tx failure
5. Autofilling is very slow, though this isn't new, we just use it in more places

How Has This Been Tested?
---
Manually, tested account creation and transfer.
Existing cucumbers 

---
Run. a validator node, indexer, walletd and submit transactions

Breaking Changes
---

- [ ] None
- [x] Other - various database, network and protocol breaking changes


applications/tari_dan_app_utilities/src/base_layer_scanner.rs
```diff
@@ -356,23 +361,39 @@ impl BaseLayerScanner {
...
-        );
-        let shard_id = ShardId::from_address(&address, 0);
-        self.shard_store
-            .with_write_tx(|tx| tx.save_burnt_utxo(&substate, address, shard_id))
+        let substate = SubstateValue::UnclaimedConfidentialOutput(UnclaimedConfidentialOutput {
+            commitment: output.commitment.clone(),
+            encrypted_data: EncryptedData(output.encrypted_data.to_bytes()),
+        });
+        let epoch = self.epoch_manager.current_epoch().await?;
+        self.state_store
```
stringhandler, line 376: I wonder if these should be proposed by the leader of the local committee as well?

stringhandler (commented):
Obviously a huge change. I tried to skip the stuff like the indexer and dry run changes. 

Looks fine from what I see. I'm going to test it locally and see what the database looks like and how it progresses through stages

applications/tari_dan_app_utilities/src/base_layer_scanner.rs
sdbondi: yeah I think that makes sense, at some point we need to fix this. If only the shard(UnclaimedOutpt) stores it, it could be enough however on the claim transaction, every validator involved would need to validate that they have seen the output on the base layer. Maybe we can have "validator transactions" or "system transactions" so that validators can propose things like a new unclaimed output, but not sure if that is necessary in this case

stringhandler (approved):
Nice. Ran locally, some changes I'd like for a future PR:

1. I'm not sure about having a genesis block for every epoch. I suppose this is needed for the burns and claiming? We probably need a proper design for start-up (where the epoch on the base layer has already progressed quite far), and also for epoch changes, where a committee splits or joins
2. When I tested it, the transaction was completed, but not in consecutive blocks. Probably just an optimization missing there
3. There were no shard/committee ids on the blocks or QCs. Probably will naturally find them

sdbondi:
> Nice. Ran locally, some changes I'd like for a future PR:
> 
> 1. I'm not sure about having a genesis block for every epoch. I suppose this is needed for the burns and claiming? We probably need a proper design for start-up (where the epoch on the base layer has already progressed quite far), and also for epoch changes, where a committee splits or joins
> 2. When I tested it, the transaction was completed, but not in consecutive blocks. Probably just an optimization missing there
> 3. There were no shard/committee ids on the blocks or QCs. Probably will naturally find them

Thanks 

1. Yeah this is more of a placeholder for something better, this was just very simple. Nothing really to do with burns/claiming. FYI The epoch manager doesnt emit these events for each epoch while starting up, only once sync has finished. Perhaps we can reference parent blocks in the new genesis blocks and perhaps continue from the max height of any of the previous tips. 
2. Interesting, will look into that. EDIT: ok yeah this would happen if the block has foreign committees (we need a 3 chain before we can send local prepared) but there is an optimisation for local only blocks (i.e. all transactions are only local) that I'll add.
3. The committee address is part of the validator node signature, not a separate field in the vote message/QCs

PR #621, opened 2023-07-18
fix: getting/revoking tokens

Cifko:
Description
---
Added more information exported by the `get_all_jwt` function. Fix the revoke function (takes id, instead of the token).

How Has This Been Tested?
---
Manually.

---
The UI changes are in separate PR by @NovaT82 .

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (approved):
utACK

stringhandler (approved):

PR #622, opened 2023-07-19
fix(consensus)!: handle SomePrepared case, use less blocks

sdbondi:
Description
---
fix(consensus): change decision to ABORT if any foreign shard decides to abort
fix(consensus): remove extraneous involved shards list in TransactionAtom, these already exist in the evidence struct 
test(consensus): add test where one shard decides to COMMIT and the other ABORTS, check that all ABORT
fix(rpc/jrpc): ensure that final consensus decision is returned to clients when finalized rather than the execution decision  
chore(consensus): improve debug and info log output
fix: use single task for messaging pipelines to ensure message ordering is maintained

Motivation and Context
---

Extra blocks were produced to get a transaction to finalization. 

When any foreign committee decides ABORT, we must set SomePrepared(tx, ABORT). This PR implements this decision change.

Prepare stage was added to differentiate between the stage where transactions are prepared but not yet included as LocalPrepared in a local block, this allows us to wait for LocalPrepared before we proceed to AllPrepared regardless of whether our local proposal arrives first or the remote proposal does.  

How Has This Been Tested?
---
Started with failing test, which now passes. The fix that necessitated the addition of the Prepare stage is difficult to test directly in integration tests however multishard tests expose the race condition quite nicely by bein flaky before the fix. 

---
Manual tests by submitting transactions

Breaking Changes
---

- [ ] None
- [ ] Other - Please specify


dan_layer/consensus/src/hotstuff/on_receive_proposal.rs
```diff
@@ -229,6 +229,57 @@ where TConsensusSpec: ConsensusSpec
...
+
+            tx_rec.update_evidence(tx, foreign_committee_shard, *block.justify().id())?;
+            let change_to_abort = cmd.decision().is_abort() && tx_rec.original_decision().is_commit();
+            if change_to_abort {
+                info!(
+                    target: LOG_TARGET,
+                    "⚠️ Foreign shard ABORT {}. Update decision to ABORT",
+                    tx_rec.transaction.id
+                );
+                tx_rec.update_decision(tx, Decision::Abort)?;
```
stringhandler: I don't think this is 100% correct. We should update the __pending__ decision to abort, but we only actually update the decision when the local leader proposes to us to abort.

In other words, we change the decision that we would make if we would become the leader and propose to change this transaction.

Interestingly, if we receive this foreign proposal and it looks like it's an Abort, but the leader proposes an Accept, we would have to check the evidence that they have provided that we should Accept, and if it checks out, it should override our local abort. This is only possible in the threat model of an entire committee being malicious so we should throw up a big warning flag and potentially even record this somewhere for reconciliation. It might even be worth stalling consensus at this point so that we don't mangle the data

dan_layer/consensus/src/hotstuff/on_receive_proposal.rs
```diff
@@ -229,6 +229,57 @@ where TConsensusSpec: ConsensusSpec
...
+            }
+
+            // If we've received we know that all locals have prepared and have all the evidence from all shards, we can
+            // transition to All/SomePrepared
+            if tx_rec.stage().is_local_prepared() && tx_rec.transaction.evidence.all_shards_complete() {
```
stringhandler, line 278: Only transistion when our local leader proposes, so that all local nodes have the same data

stringhandler (approved):
utACK

Will test locally though

stringhandler:
Tested locally and working great

PR #623, opened 2023-07-19
chore(tari_dan_storage): set exact `tari_utilities` version

mrnaveira:
Description
---
Set the exact version for the `tari_utilities` dependency on the `tari_dan_storage` crate

Motivation and Context
---
When importing (directly or indirectly) the `tari_dan_storage` package from external Rust applications, the project does not compile due to version incompatibilities with the latest `tari_utilities` dependency. The same happens after running `cargo clean` on the `tari-dan` project.

How Has This Been Tested?
---
External projects that have the `tari_dan_storage` dependency do compile now

---
Compile an external project that depends on `tari_dan_storage`

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (approved):
lol whoops

PR #624, opened 2023-07-19
fix: use round robin leader strategy

stringhandler:
Description
---
Change from random leaders (determined by blockid) to round robin through the committee using the block height

Motivation and Context
---
This is how hotstuff is usually done, but for more motivation it means that every leader has a fair opportunity to get fees, and leaders cannot be censored in any way.

It also is a step towards implementing leader failure, which will be in another PR

How Has This Been Tested?
---
Tested manually with dan testing

---
Run a chain and see that the blocks are proposed in a round robin fashion

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (commented):
LGTM - just clippy

sdbondi (approved):

PR #625, opened 2023-07-20
fix(mempool): add transaction to state store before execution

sdbondi:
Description
---
fix(mempool): add transaction to state store before execution
feat(sqlite state store): add executed_time_ms to transactions

Motivation and Context
---
Add the transaction to the state store before execution to allow missing transactions to check if a transaction exists but is still executing.

How Has This Been Tested?
---
Existing cucumber tests, manually

---
Check that missing transaction does not request transactions that are busy executing

Breaking Changes
---

- [ ] None
- [ ] Other - Please specify


stringhandler (dismissed):
For a future PR it might be worth adding the Execution start time in the database so that you we can see when it was queued, but not necessary

stringhandler (approved):
reapproving

PR #626, opened 2023-07-20
fix: better errors and create free coins button

stringhandler:
Description
---
Propagates error from the indexer to the caller.
Adds a Create Free Test coins button in the DAN wallet UI so that you can create a transaction easily.

Motivation and Context
---
Errors were being swallowed in JSONRPC
Also, it helps with testing to be able to press a single button to create a transaction on the network, as well as adding  test coins to play with

How Has This Been Tested?
---
Using DAN testing

---
Start a DAN wallet and press the button. Go to the account page and see the balance increase

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (dismissed):
utACK

PR #627, opened 2023-07-21
test: add debugging tracking trigger to transaction_pool

stringhandler:
Description
---
Creates a trigger that saves every update to the transaction_pool to a new table transaction_pool_history

Motivation and Context
---
This is purely for debugging and can be used for tracking changes and when foreign proposals came in. This is typically quite hard to track from logs and the transaction_pool table is empty after the transaction has run.

How Has This Been Tested?
---
With dan testing, checking if the table is populated

---
Run a transaction, open the transaction_pool_history table and see what changed at which time

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (dismissed):
utACK

sdbondi (dismissed):

PR #632, opened 2023-07-24
feat!: add basic leader failure recovery

stringhandler:
Description
---
Implements leader failure, according to the hotstuff paper. Also added a timeout after which an empty block is proposed. Previously the nodes would stop proposing blocks when there was no work to be done. Since a node cannot distinguish between a node having no work and a failure, this was changed to keep the chain going. This also keeps the nodes in better sync since they can keep their timeouts in the same order of magnitude.

Some other changes:
1. Removed the per-epoch genesis block.
2. Removed BlockId from leader strategy
3. Deleted some old TIP files from ages ago.
4. Removed epoch from most housekeeping variables
5. The safe_node predicate was moved to **before** a block is added. Previously it would add the block and then not vote, but this could lead to some bugs in future where we think we have validated a block
6. NB: Added a rule that we will not propose a node unless we have received our own vote for it. This prevents a case where we have not applied all the data to our state database and we lazily propose a node
7. Fixed a bug (or weird behavior) in `blocks_is_ancestor` 

Motivation and Context
---
Leader failure and recovery is a big part of hotstuff.
The per-epoch genesis blocks made it difficult to determine if the parent was a genesis block, mainly because `BlockId::is_genesis()` did not take it into account. IMO it's simpler to have a single chain from a single genesis block. This may have an impact on how burnt funds are imported, but they seemed to work in my manual testing.

Removing the BlockId from the leader strategy was done to allow simpler adding of dummy blocks. The BlockID will probably never be used in a valid production LeaderStrategy, so I feel like this is fine, but if we need it in future, we can always add it back.

In a future PR I'd like to also change the genesis QC to have an ID of 000000..000. It currently has a value of 28,,, When a leader has failed, the nodes will populate dummy blocks, which have a genesis QC to mark them as dummy. Perhaps it's better to explicitly mark this in the database, but for now I just used a genesis QC. 

Note: This is implemented as closely to the hotstuff paper, according to my understanding. A potentially unclear consequence is that the validator node just before the leader who fails never gets their proposal accepted. This is because it uses `b_leaf` which is the node before the highest QC and the votes for the node that proposed it are lost. (They would have gone to the failed leader). So if you test this manually, you'll see the proposal by the VN before the failed leader get rolled back and replaced by a dummy node, as well as the dummy node of the failed leader. 

A possible future improvement would be to resend the votes to the next leader as part of the NewView message. The new leader could assemble the QC and only fill in the single dummy node. I'm not sure if this solution is secure.

This also means that with 4 nodes and a single failure, a 3 chain is never reached, so the minimum number of nodes in each committee before any can fail is 5. 

How Has This Been Tested?
---
Tested locally with DAN testing. Will put tests in a new PR

---
Run 5 vns and kill one. Note: this technically could work with 4 VNs, but a 3 chain is never completed, so transactions will stay in pending.

Breaking Changes
---

- [ ] None
- [ ] Other - Please specify


dan_layer/consensus/src/hotstuff/pacemaker.rs
```diff
@@ -0,0 +1,98 @@
...
+                    }
+                    // if let Err(e) = self.on_beat().await {
+                    //     error!(target: LOG_TARGET, "Error (on_beat): {}", e);
+                    // }
+
+                },
+                () = &mut sleep => {
+                    // tx_on_leader_timeout.send(()).map_err(|e| HotStuffError::PacemakerChannelDropped{ details: e.to_string()})?;
+                    on_leader_timeout.leader_timed_out();
+                    self.current_delta *= 2;
```
Cifko, line 114: This is never reset.

dan_layer/consensus/src/hotstuff/on_receive_proposal.rs
```diff
@@ -204,6 +204,7 @@ where TConsensusSpec: ConsensusSpec
         })?;
 
         if let Some(decision) = maybe_decision {
+            self.pacemaker.reset_leader_timeout();
```
Cifko: This will cause leader rotation even if the foreign committee fails, right?

dan_layer/consensus/src/hotstuff/pacemaker.rs
stringhandler: Correct. I have a commit that brings it down, but haven't pushed it yet. Will continue the discussion when it's pushed

sdbondi (approved):
Very nice - scanned over all the code and took a closer look at the important bits, no tests but happy for that to be in a new PR - utACK :rocket:

PR #640, opened 2023-08-02
feat: add epoch manager stats to indexer

Cifko:
Description
---
Add jrpc call to get epoch manager stats.

Motivation and Context
---
In the dan-testing there is no way to wait for indexers to sync to the current epoch.

How Has This Been Tested?
---
I'm now syncing the dan-testing indexers to the current epoch.

---
Call the new jrpc method to get the stats.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_indexer/src/json_rpc/handlers.rs
+++ b/applications/tari_indexer/src/json_rpc/handlers.rs
@@ -553,7 +553,7 @@ impl JsonRpcHandlers {
             JsonRpcResponse::error(
                 answer_id,
                 JsonRpcError::new(
-                    JsonRpcErrorReason::InvalidParams,
+                    JsonRpcErrorReason::InternalError,
                     format!("Could not get current epoch: {}", e),
                     json::Value::Null,
                 ),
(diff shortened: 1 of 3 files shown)


applications/tari_indexer/src/json_rpc/handlers.rs
```diff
@@ -540,6 +547,47 @@ impl JsonRpcHandlers {
         }
     }
 
+    pub async fn get_epoch_manager_stats(&self, value: JsonRpcExtractor) -> JrpcResult {
+        let answer_id = value.get_answer_id();
+        let current_epoch = self.epoch_manager.current_epoch().await.map_err(|e| {
+            JsonRpcResponse::error(
+                answer_id,
+                JsonRpcError::new(
+                    JsonRpcErrorReason::InvalidParams,
```
sdbondi: InvalidParams is used to indicate that the user passed in something invalid. JsonRpcErrorReason::InternalError is  more appropriate. You can use the `Self::internal_error` helper. Same for the others below

Cifko: It was copy-pasted, so I changed the original as well.

sdbondi (approved):

PR #641, opened 2023-08-02
fix: transfer inputs (refs)

Cifko:
Description
---
Fix inputs for transfer, template is now as input refs, so transfers don't collide.

How Has This Been Tested?
---
Running multiple transaction in parallel.

---
Run dan-testing with multiple transaction in parallel.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (approved):
utACK - probably needed in a few other places too

PR #649, opened 2023-08-07
feat(engine): add decoded json field on instruction results

mrnaveira:
Description
---
* Added a new `json` field in the `InstructionResult` struct, alongside the existing raw CBOR value. 
* Some of the sanity checks of the CBOR to JSON conversion is common with the indexer operations, so it was carried over to the engine instead.
* The wasm instruction processor fills the new `json` field in the result after a successful execution.
* A few unrelated code was reformatted by `rust fmt` 

Motivation and Context
---
Up until now, the instruction results of a transaction included only the raw CBOR value and the decoded substates. From user web applications this is insufficient as many template methods/functions can return any type of data, and CBOR decoding in the web may not be the best approach.

This PR aims to solve this problem by adding a new field (`json`) in the `InstructionResult` struct in the engine, alongside the existing raw CBOR value. 

How Has This Been Tested?
---
* Manually by inspecting the execution results of a template method that returns values
* The logic for JSON decoding was already present in the indexer with a unit test that was carried over

---
Call a template method/function and inspecting the new field value in the result

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_indexer/src/json_rpc/special_substate_encoding.rs
+++ b/applications/tari_indexer/src/json_rpc/special_substate_encoding.rs
@@ -2,6 +2,7 @@
 //   SPDX-License-Identifier: BSD-3-Clause
 
 use serde_json as json;
+use tari_dan_engine::json_encoder::fix_invalid_object_keys;
 use tari_engine_types::{
     component::ComponentHeader,
     non_fungible::NonFungibleContainer,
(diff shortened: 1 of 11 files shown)


sdbondi:
I don't agree with this change. The engine should not have to care about display concerns for web clients. This also adds a  lot of redundant data for the validator node to store. 

We can add this code to a common crate at the application level (e.g app utils). The JSON mangling can perhaps be done in the JRPC handlers if needed since there are squarely concerned with web clients.

mrnaveira:
> I don't agree with this change. The engine should not have to care about display concerns for web clients. This also adds a lot of redundant data for the validator node to store.
> 
> We can add this code to a common crate at the application level (e.g app utils). The JSON mangling can perhaps be done in the JRPC handlers if needed since there are squarely concerned with web clients.

Fair enough. I will close this PR and approach the problem from a better angle

PR #651, opened 2023-08-08
feat: block sync

Cifko:
Description
---
Sync blocks.

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_validator_node/src/p2p/rpc/service_impl.rs
+++ b/applications/tari_validator_node/src/p2p/rpc/service_impl.rs
@@ -28,7 +28,7 @@ use tari_comms::protocol::rpc::{Request, Response, RpcStatus, Streaming};
 use tari_dan_common_types::{optional::Optional, NodeAddressable, ShardId};
 use tari_dan_p2p::PeerProvider;
 use tari_dan_storage::{
-    consensus_models::{SubstateRecord, TransactionRecord},
+    consensus_models::{Block, BlockId, SubstateRecord, TransactionRecord},
     StateStore,
 };
 use tari_engine_types::virtual_substate::VirtualSubstateAddress;
@@ -39,6 +39,8 @@ use tari_transaction::{Transaction, TransactionId};
 use tari_validator_node_rpc::{
     proto,
     proto::rpc::{
+        BlockSyncRequest,
+        BlockSyncResponse,
         GetSubstateRequest,
         GetSubstateResponse,
         GetTransactionResultRequest,
@@ -287,6 +289,50 @@ where TPeerProvider: PeerProvider + Clone + Send + Sync + 'static
         };
 
         Ok(Response::new(resp))
+    }
+
+    async fn sync_blocks(&self, req: Request<BlockSyncRequest>) -> Result<Streaming<BlockSyncResponse>, RpcStatus> {
+        let (sender, receiver) = mpsc::channel(100);
+        let req = req.into_message();
+
+        let block_id = BlockId::try_from(req.block_id)
+            .map_err(|e| RpcStatus::bad_request(&format!("Invalid encoded block id: {}", e)))?;
+        let epoch = req.epoch;
+        let shard_db = self.shard_state_store.clone();
+        task::spawn(async move {
+            let block = shard_db.with_read_tx(|tx| Block::get_tip(tx, tari_dan_common_types::Epoch(epoch)));
+            match block {
+                Ok(block) => {
+                    let mut block = block;
+                    while *block.id() != block_id {
+                        if sender
+                            .send(Ok(BlockSyncResponse {
+                                block: Some(block.clone().into()),
+                            }))
+                            .await
+                            .is_err()
+                        {
+                            debug!(
+                                target: LOG_TARGET,
+                                "Peer stream closed by client before completing. Aborting"
+                            );
+                        }
+                        let x = shard_db.with_read_tx(|tx| block.get_parent(tx));
+                        block = match x {
+                            Ok(block) => block,
+                            Err(e) => {
+                                let _ignore = sender.send(Err(RpcStatus::general(&e))).await;
+                                return;
+                            },
+                        };
+                    }
+                },
+                Err(e) => {
+                    let _ignore = sender.send(Err(RpcStatus::general(&e))).await;
+                },
+            }
+        });
+        Ok(Streaming::new(receiver))
     }
 
     async fn get_transaction_result(
--- a/applications/tari_validator_node/src/p2p/rpc/service_impl.rs
+++ b/applications/tari_validator_node/src/p2p/rpc/service_impl.rs
@@ -301,7 +301,7 @@ where TPeerProvider: PeerProvider + Clone + Send + Sync + 'static
         // Check if we have such block
         shard_db
             .with_read_tx(|tx| Block::get(tx, &block_id))
-            .map_err(|e| RpcStatus::general(&e))?;
+            .map_err(|e| RpcStatus::bad_request(&e))?;
         let epoch = req.epoch;
         task::spawn(async move {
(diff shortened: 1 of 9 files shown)


applications/tari_validator_node/src/p2p/rpc/service_impl.rs
```diff
@@ -260,6 +262,51 @@ where TPeerProvider: PeerProvider + Clone + Send + Sync + 'static
...
     }
 
+    async fn vn_blocks_sync(&self, req: Request<BlockSyncRequest>) -> Result<Streaming<BlockSyncResponse>, RpcStatus> {
+        let (sender, receiver) = mpsc::channel(100);
+        let req = req.into_message();
+
+        let block_id = BlockId::try_from(req.block_id)
+            .map_err(|e| RpcStatus::bad_request(&format!("Invalid encoded block id: {}", e)))?;
+        let epoch = req.epoch;
+        let shard_db = self.shard_state_store.clone();
```
sdbondi: Need to check the requested block exists, otherwise return bad request.

applications/tari_validator_node/src/p2p/rpc/service_impl.rs
```diff
@@ -260,6 +262,51 @@ where TPeerProvider: PeerProvider + Clone + Send + Sync + 'static
         Ok(Response::new(resp))
     }
 
+    async fn vn_blocks_sync(&self, req: Request<BlockSyncRequest>) -> Result<Streaming<BlockSyncResponse>, RpcStatus> {
```
sdbondi: nit(naming): I prefer sync_blocks - I know this name was taken from a state sync :)

sdbondi (commented):
Looks good so far

applications/tari_validator_node/src/p2p/services/committee_state_sync.rs
```diff
@@ -158,16 +180,63 @@ impl CommitteeStateSync {
...
+                    let msg = resp?;
+                    let block: Block<CommsPublicKey> = msg
+                        .block
+                        .ok_or(CommitteeStateSyncError::InvalidStateSyncData(anyhow::anyhow!(
+                            "No block"
+                        )))?
+                        .try_into()
+                        .map_err(CommitteeStateSyncError::InvalidStateSyncData)?;
+                    self.shard_store.with_write_tx(|tx| block.justify().save(tx))?;
+                    self.shard_store.with_write_tx(|tx| block.insert(tx))?;
```
sdbondi: These should run in a single transaction
```suggestion
                    self.shard_store.with_write_tx(|tx| {
                       block.justify().save(tx)?;
                       block.insert(tx)
                    })?;
```

applications/tari_validator_node/src/p2p/services/committee_state_sync.rs
```diff
@@ -158,16 +180,63 @@ impl CommitteeStateSync {
...
+                    last_synced_block = Some(*block.id());
+                    // TODO: When we start splitting chain this will be very diffirent.
+                    block_count += 1;
+                }
+                if let Some(last_synced_block) = last_synced_block {
+                    if self
+                        .shard_store
+                        .with_read_tx(|tx| Block::get(tx, &last_synced_block)?.get_parent(tx))
+                        .is_err()
```
sdbondi: is_err will be true for any error - not just the not found case

```suggestion
                    if self
                        .shard_store
                        .with_read_tx(|tx| Block::get(tx, &last_synced_block)?.get_parent(tx))
                        .optional()?.is_none()
```

dan_layer/consensus/src/hotstuff/event.rs
```diff
@@ -1,10 +1,12 @@
 //    Copyright 2023 The Tari Project
 //    SPDX-License-Identifier: BSD-3-Clause
 
+use tari_dan_common_types::Epoch;
 use tari_dan_storage::consensus_models::BlockId;
 
 #[derive(Debug, Clone)]
 pub enum HotstuffEvent {
     BlockCommitted { block_id: BlockId },
+    BlockSyncRequest { block_id: BlockId, epoch: Epoch },
```
sdbondi: Events are usually named for something (an event) that happened

`BlockchainBehind` or `BlocksOutOfSync` or (insert the best name here)

dan_layer/consensus/src/hotstuff/on_receive_proposal.rs
```diff
@@ -133,6 +133,15 @@ where TConsensusSpec: ConsensusSpec
         local_committee: Committee<TConsensusSpec::Addr>,
         block: Block<TConsensusSpec::Addr>,
     ) -> Result<(), HotStuffError> {
+        if self.store.with_read_tx(|tx| block.get_parent(tx)).is_err() {
```
sdbondi: Same as above

```suggestion
        if self.store.with_read_tx(|tx| block.get_parent(tx)).optional().is_none() {
```

We should make Block::exists a non-instance function rather:

```rust
    pub fn exists<TTx: StateStoreReadTransaction<Addr = TAddr> + ?Sized>(
     
        tx: &mut TTx,
block_id: &BlockId    
) -> Result<bool, StorageError> {
        tx.blocks_exists(block_id)
    }
```

applications/tari_validator_node/src/p2p/rpc/service_impl.rs
```diff
@@ -290,6 +292,57 @@ where TPeerProvider: PeerProvider + Clone + Send + Sync + 'static
...
+        let start_block_id = BlockId::try_from(req.start_block_id)
+            .map_err(|e| RpcStatus::bad_request(&format!("Invalid encoded block id: {}", e)))?;
+        // Check if we have the blocks
+        shard_db
+            .with_read_tx(|tx| Block::get(tx, &start_block_id))
+            .map_err(|e| RpcStatus::bad_request(&e))?;
+        let epoch = req.epoch;
+        // Sync from tip, or from the block that the requester has.
+        let mut block = match req.end_block_id.len() {
+            0 => shard_db.with_read_tx(|tx| Block::get_tip(tx, tari_dan_common_types::Epoch(epoch))),
```
sdbondi: You'll have to change get_tip to return the MAX height and to not take in an epoch anymore. It changed with the leader failure PR not to have a chain per epoch but I don't think this function was changed.

applications/tari_validator_node/src/p2p/services/committee_state_sync.rs
```diff
@@ -113,7 +118,24 @@ impl CommitteeStateSync {
...
+        let tip = self.shard_store.with_read_tx(|tx| Block::get_tip(tx, epoch))?;
+        // Missing children is an array (from, to), where we have `from` and `to` blocks, but nothing in between. In
+        // case the `to` is `None`, then we want everything up to the tip.
+
+        let mut missing_blocks = {
+            let mut tx = self.global_db.create_transaction()?;
+            match self
+                .global_db
+                .metadata(&mut tx)
+                .get_metadata::<Vec<(BlockId, BlockId)>>(MetadataKey::MissingBlocks)?
```
sdbondi, line 137: I don't really understand why we need to persist this data.

Cifko: This is for the case when you sync fails in the middle. So you can end up with a ranges of blocks.
E.g. you have tip 10, and now blocks 20 comes in. You start syncing (the blocks are coming from the tip). You will receive 20,19,18 (and it fails). So now you have 1-10,18-20. You start syncing again and now the tip is 30, so you get 30down to 25 (and fail again). This way you can end up with many block ranges missing. And we probably don't want to sync all the time from the tip to our last know tip. In this example you don't want again all blocks from 10 to 30. But I'm open to any other solution.

stringhandler:
Closing in favour of @sdbondi's block sync pr

PR #653, opened 2023-08-09
feat(template-lib)!: impl human readable (de)serializer for address newtypes

sdbondi:
Description
---
Implement a serializer and deserializer for address newtypes that de/serialize from/to string for human-readable formats.

Remove generic_array dependency from template_lib

Motivation and Context
---
Some structs use address type but do not or cannot add the serde with annotations. This PR implements better (more compact and readable) de/serialization for JSON internally to the address type. This does not add much extra code from what already exists in the template lib because Serde's Serialize and Deserialize derive macros generate similar code anyway.

I did not remove the existing annotations though now they are unnecessary.

generic_array is likely eventually to be replaced with native const generics in many crates. Removing this also allows us to remove lazy_static too.  Size reductions are fairly modest e.g. in account template size is reduced by 3998 bytes (previous size: 499531 bytes  ~ 0.8% reduction).

How Has This Been Tested?
---
Unit test for ResourceAddress, cucumbers

---
A few ways e.g. use a JRPC call that includes a NonFungibleAddress type

Breaking Changes
---

- [ ] None
- [ ] Other - Please specify


stringhandler (dismissed):

Cifko (approved):
Cool, this means the tari-connector sample will be broken again (because of the serialization of the permissions), If that's the case I will look at that after this gets merged.

stringhandler (approved):

PR #655, opened 2023-08-11
fix(engine): return engine error from instruction invocation 

sdbondi:
Description
---
fix(engine): return engine error from instruction invocation 
fix: only unlock on abort if inputs were originally locked
fix(wallet): propagate result error from vn to wallet cli
fix(wallet/cli): enable amounts bigger than u32::MAX to be input as an arg

Motivation and Context
---
Hook up errors returned by the runtime to the final transaction result. This takes precedence over a panic (which MUST always occur after an engine error). 

Hook up the result properly so that it is returned even when the transaction is aborted. This allows the client to get the actual error that occurred during execution.

How Has This Been Tested?
---
New engine unit test that performs an invalid engine call, the WASM panics and we check that the result includes the underlying engine error rather than the null ptr panic error.
Manually, send more funds than available.

---
Create a wallet and send more funds than available, the error is now output rather than a unknown error or WASM panic text

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_dan_wallet_cli/src/command/transaction.rs
+++ b/applications/tari_dan_wallet_cli/src/command/transaction.rs
@@ -137,7 +137,7 @@ pub struct SubmitManifestArgs {
 
 #[derive(Debug, Args, Clone)]
 pub struct SendArgs {
-    amount: u32,
+    amount: u64,
     resource_address: ResourceAddress,
     destination_public_key: FromHex<Vec<u8>>,
     #[clap(flatten)]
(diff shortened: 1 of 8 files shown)


stringhandler (approved):
Nice, much simpler than what I was trying to do

Cifko (approved):

Cifko (approved):
cool, the new error message `Fee transaction rejected: Shard was rejected: Validators decided to abort: Execution failure: Runtime error: Resource did not contain sufficient balance: Bucket contained insufficient revealed funds. Required: 1500, Available: 100` this will be super useful when testing

PR #657, opened 2023-08-14
fix!: peer sync

sdbondi:
Description
---
Fix peer sync and announce by correctly handling peer signature claims

Motivation and Context
---
Peer sync and announce worked only when a single address is used, when multiple addresses are added (this happens whenever an address is changed), the identity signature validation fails. 

How Has This Been Tested?
---
100 validator nodes which local address changes on each startup, able to peer sync and receiving valid announces

---
Change the validator node address and check that the network can receive the updated address

Breaking Changes
---

- [ ] None
- [x] Other - peer sync and announce messages have breaking changes


Cifko (approved):
tested, 5 VNs, stop and start with new address, process TX, works
nice

stringhandler (approved):

PR #661, opened 2023-08-17
fix(consensus): commit the block that forms a 3-chain

sdbondi:
Description
---
Only commit node once it forms a 3-chain

Motivation and Context
---
The proposed block is committed instead of the `prepare_node` block. This is incorrect as per https://arxiv.org/pdf/1804.00399.pdf

```
b* is the proposed block
b'' ← b∗.justify.node; 
b' ← b''.justify.node
b ← b'.justify.node

if b forms a 3-chain {
   // Bug was here - b* was committed previously 
    commit(b); 
    execute(b.cmd)
 }
```

How Has This Been Tested?
---
Existing consensus tests pass - fast block production stops after a 3 chain

---
Submit a transaction and check that 3 empty blocks are created after the transaction has been accepted

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_validator_node_cli/src/command/transaction.rs
+++ b/applications/tari_validator_node_cli/src/command/transaction.rs
@@ -309,9 +309,10 @@ async fn wait_for_transaction_result(
 
         if let Some(resp) = resp {
             if resp.is_finalized {
-                if let Some(result) = resp.result {
-                    return Ok(result);
-                }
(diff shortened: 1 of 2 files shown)


stringhandler (dismissed):
lol woops

stringhandler (approved):

PR #666, opened 2023-08-21
fix(consensus)!: fixes for leader failure w/ locked substates

sdbondi:
Description
---
fix(consensus): change local decision when locking fails
fix(consensus): propose local decision for Prepare/LocalPrepared phases
fix(consensus)!: track local and remote decisions for transactions
fix(consensus)!: add block_id to quorum_certificate to allow fetching of qc by block id
fix(consensus): fail to lock DOWNed substates
fix(consensus): set the last proposed block to account for dummy blocks
fix(consensus): only update nodes and last voted if we are voting on a proposal
fix(consensus): start leader failure timer even if a proposal was not voted on
fix(consensus): dont vote if we do not have a transaction in the pool
fix(consensus): various logging improvements
feat(vn/cli): allow input_refs to be specified
feat(consensus_tests): add `debug_sql` to easily create a sqlite database file in consensus tests
feat(consensus_tests): add timeout for hotstuff events 
tests(consensus): add leader_failure_output_confict test
tests(cucumber): fix mint test by removing DOWNed inputs from being used in transactions
fix(consensus/mempool): add transaction result and atom in single db transaction to prevent having to handle an edge case 
fix(consensus): missing transactions does not request transactions that are awaiting execution
fix(consensus): add pending stage change to account for stage changes that are reverted (TODO: undo stage change when creating dummy blocks for already processed blocks)
fix(consensus): use unbounded channel between consensus and mempool, bounded causes lockups due to cyclic service dependency (mempool -> consensus ---(now unbounded)---> mempool)
fix(consensus): add correct check for local vn registration for epoch (TODO: correctly define valid epoch range)
feat(consensus): fixed block time of 10s
fix(consensus): stop/start pacemaker when node is un/registered
fix(consensus): exponentially increasing leader timeout increasing by height difference between current height and high QC
fix(clippy): fix clippies unrelated to this PR
tests(consensus): add fern logger

Motivation and Context
---
Many of these fixes were discovered by stress tests and when testing transaction failure due to conflicting outputs.

Currently, conflicting inputs/ outputs cause a leader failure as nodes will not vote on the block. This is bad because it allows users to trigger a leader failure. A validator should propose ABORT for conflicting objects, however, this complicates the on_propose quite a bit so is left for another PR so that we can think about how to simplify it more.

Other changes are pretty self-explanatory and allow stress tests to work.

How Has This Been Tested?
---
New functional test `leader_failure_output_confict`
Updated cucumber tests

---
Submit a transaction with a DOWNed input or duplicate outputs in multiple transactions

Breaking Changes
---

- [ ] None
- [ ] Other - Please specify


stringhandler (approved):
Difficult to review all these changes, but tested it locally and it runs smoothly

PR #667, opened 2023-08-22
fix(template-manager): prevent unncessary wasm loading in high concurrency

sdbondi:
Description
---
- Protect a critical area for template loading by
1. preventing more than 100 concurrent accesses
2. preventing more than one load of the same template
      
- Preload the builtin templates

Motivation and Context
---
In stress testing, when submitting around 200+ transactions, the validator's stack overflows. This is because the template manager would load the (same) template 200+ times concurrently. This PR prevents concurrent loading of the same template, instead loading from the cache once loaded and more than CONCURRENT_ACCESS_LIMIT (100) concurrent template loads.

We validate that all templates used in transactions exist. So a user can not use up the CONCURRENT_ACCESS_LIMIT with non-existent templates.

In this PR, we preload builtin templates into the cache, as we assume that these are used more commonly. 

How Has This Been Tested?
---
Manually, stress testing without first preloading the builtin templates

---
Submit 1000+ transactions before any templates have been loaded does not stack overflow. 
Note since the stress tester currently only submits ACCOUNT template instructions and since ACCOUNT is now preloaded, the stack overflow would not occur even without the locks.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_dan_app_utilities/src/template_manager/implementation/cmap_semaphore.rs
+++ b/applications/tari_dan_app_utilities/src/template_manager/implementation/cmap_semaphore.rs
@@ -1,0 +1,67 @@
+//   Copyright 2023 The Tari Project
+//   SPDX-License-Identifier: BSD-3-Clause
+
+use std::{
+    fmt::{Debug, Formatter},
+    hash::Hash,
+    sync::{Arc, Mutex, MutexGuard},
+};
+
+#[derive(Clone)]
+pub struct CMapSemaphore<K: Hash + Eq> {
+    map: Arc<dashmap::DashMap<K, Arc<Mutex<()>>>>,
+    global: Arc<std_semaphore::Semaphore>,
+}
+
+impl<K: Hash + Eq + Clone> CMapSemaphore<K> {
+    pub fn new(max_global_access: isize) -> Self {
+        Self {
+            map: Arc::new(dashmap::DashMap::new()),
+            global: Arc::new(std_semaphore::Semaphore::new(max_global_access)),
+        }
+    }
+
+    pub fn acquire(&self, key: K) -> CMapSemaphoreGuard<'_, K> {
+        let global_access = self.global.access();
+        let map_mutex = self
+            .map
+            .entry(key.clone())
+            .or_insert_with(|| Arc::new(Mutex::new(())))
+            .clone();
+        CMapSemaphoreGuard {
+            _global_access: global_access,
+            map: self.map.clone(),
+            map_mutex,
+            key,
+        }
+    }
+}
+
+impl<K: Hash + Eq> Debug for CMapSemaphore<K> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("CMapSemaphore")
+            .field("map", &self.map.len())
+            .field("global", &"...")
+            .finish()
+    }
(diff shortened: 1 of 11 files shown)


applications/tari_dan_app_utilities/src/template_manager/implementation/cmap_semaphore.rs
```diff
@@ -0,0 +1,67 @@
...
+//   SPDX-License-Identifier: BSD-3-Clause
+
+use std::{
+    fmt::{Debug, Formatter},
+    hash::Hash,
+    sync::{Arc, Mutex, MutexGuard},
+};
+
+#[derive(Clone)]
+pub struct CMapSemaphore<K: Hash + Eq> {
```
stringhandler:
```suggestion
pub struct ConcurrentMapSemaphore<K: Hash + Eq> {
```

stringhandler: I would prefer a full name here to avoid confusion with C********* types that represent native objects. E.g. CString

applications/tari_dan_app_utilities/src/template_manager/implementation/cmap_semaphore.rs
```diff
@@ -0,0 +1,67 @@
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("CMapSemaphore")
+            .field("map", &self.map.len())
+            .field("global", &"...")
+            .finish()
+    }
+}
+
+pub struct CMapSemaphoreGuard<'a, K: Hash + Eq> {
+    _global_access: std_semaphore::SemaphoreGuard<'a>,
```
stringhandler, line 51: Perhaps add a comment here to prevent people from deleting this thinking it is unused

applications/tari_dan_app_utilities/src/template_manager/implementation/cmap_semaphore.rs
```diff
@@ -0,0 +1,67 @@
...
+    map: Arc<dashmap::DashMap<K, Arc<Mutex<()>>>>,
+    map_mutex: Arc<Mutex<()>>,
+    key: K,
+}
+
+impl<'a, K: Hash + Eq> CMapSemaphoreGuard<'a, K> {
+    pub fn access(&self) -> MutexGuard<'_, ()> {
+        // Unwrap: only errors if the mutex is poisoned, which is a bug
+        self.map_mutex.lock().unwrap()
```
stringhandler, line 60: why not use tokio?

stringhandler (commented):
Happy, just some suggestions

applications/tari_dan_app_utilities/src/template_manager/implementation/cmap_semaphore.rs
sdbondi: :100:

applications/tari_dan_app_utilities/src/template_manager/implementation/cmap_semaphore.rs
sdbondi: We're not in a async context here, transaction execution happened in a spawn_blocking

stringhandler (approved):

PR #673, opened 2023-09-05
feat(wallet/indexer): add decoded json field on transaction results

mrnaveira:
Description
---
* Added a new `json_result` field in multiple JSON RPC result types (both in the indexer as well as the wallet daemon) 
* The execution results of a transaction are decoded as JSON in the indexer, reusing some of the substate decoding logic already existing in the indexer
* Added a new column for the `json_result` field of a transaction in In the wallet sqlite storage
* A few unrelated code was reformatted by rust fmt

Motivation and Context
---
Up until now, the instruction results of a transaction included only the raw CBOR value and the decoded substates. From user web applications this is insufficient as many template methods/functions can return any type of data, and CBOR decoding in the web may not be the best approach.

This PR aims to solve this problem by adding a new field (`json_result`) in the indexer and wallet JSON RPC responses

How Has This Been Tested?
---
Manually by inspecting the execution results of a template method that returns values

---
Call a template method/function and inspecting the new field value in the result

Breaking Changes
---
- [ ] None
- [x] Requires data directory to be deleted (wallet daemon sqlite)
- [ ] Other - Please specify

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7827,6 +7827,7 @@ dependencies = [
  "log",
  "rand 0.7.3",
  "serde",
+ "serde_json",
  "tari_common_types",
  "tari_crypto",
  "tari_dan_common_types",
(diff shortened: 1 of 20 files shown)


stringhandler (approved):

sdbondi (commented):
LGTM :+1:

PR #677, opened 2023-09-08
fix(consensus)!: undo pending transaction stage transitions on leader fail

sdbondi:
Description
---
undo pending transaction stage transitions on leader fail
Add leader failure test where one node in 5 goes offline
Use RoundRobin leader strategy in consensus_tests

Motivation and Context
---
Each head block affects stage changes and object locks which need to be undone when the block(s) fail.

This PR doesn't have this quite right. An alternative is to not do any state changes when processing the block and 
rather do stage transition processing when proposing by processing blocks from the locked block up to and excluding the next proposed block (typically a single block). This would avoid the need for any "reorg" logic.

How Has This Been Tested?
---
Manually by inducing leader failure by spamming nodes with transactions.

---
As above

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_validator_node/src/dan_node.rs
+++ b/applications/tari_validator_node/src/dan_node.rs
@@ -115,7 +115,7 @@ impl DanNode {
     }
 
     async fn handle_hotstuff_event(&self, event: HotstuffEvent) -> Result<(), anyhow::Error> {
-        let HotstuffEvent::BlockCommitted { block_id } = event else {
+        let HotstuffEvent::BlockCommitted { block_id, .. } = event else {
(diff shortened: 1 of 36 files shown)


sdbondi:
I've commented out the "undo"ing and stage transitions on locked block for now. This PR adds a new leader failure test, cutting off messages in the test harness, aborting of all conflicting transactions as well as some other logic improvements to the way proposals are handled.

stringhandler (dismissed):
tested locally. I don't think we're 100% there yet, but this can go in

PR #678, opened 2023-09-10
feat: count votes as newview

Cifko:
Description
---

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


dan_layer/consensus/src/hotstuff/on_receive_proposal.rs
```diff
@@ -261,9 +261,14 @@ where TConsensusSpec: ConsensusSpec
...
             let vote = self.generate_vote_message(block, decision).await?;
 
-            self.send_vote_to_leader(local_committee, vote, block.height()).await?;
+            self.send_vote_to_leader(local_committee, vote.clone(), block.height())
+                .await?;
 
-            self.store.with_write_tx(|tx| block.as_last_voted().set(tx))?;
+            self.store.with_write_tx(|tx| {
+                block
+                    .as_last_voted(vote.epoch, vote.decision, vote.signature, vote.merkle_proof)
```
sdbondi: I think Block::as_last_voted doesnt make sense anymore. What about `VoteMessage::to_last_voted` ? 

```suggestion
                vote.to_last_voted().set(tx)
```

sdbondi (commented):
LGTM so far

dan_layer/consensus/src/hotstuff/on_receive_proposal.rs
Cifko: hmm, last voted is now getting destroyed with dummy blocks on leader failure trigger. I need to store it somewhere else.

sdbondi: Yeah that part of the code is in flux at the moment. The latest code is in #651.

sdbondi:
Question: how do you propose that we handle the fact that votes for a block are bound to the block height?

e.g. votes for block 5 are collected, block 6 fails, dummy block 6' and next block 7 are produced. Votes for block 5 are not valid for block 7.

sdbondi:
Figured out what causes the leader failure in the failing `New validator node registers and syncs` cucumber. Basically this inadvertently tests epoch change with new validators which is not well tested.

What happens is that at epoch 2 we have $vn_1$ and it proposes and accepts block 28. It would then propose 29 justifying 28. But then the epoch changes to 3, and suddenly $vn_2$ is registered and is the leader for block 29, however it cannot propose 29 because it does not have the justify/qc for 28 (its leaf block is 27). 

Few ways to solve this but what comes to mind: 
1. on epoch change, all nodes send votes to leader for epoch 3, block 28, or
2. the leader for epoch2 block 29  should send an EpochChangeSync message to all nodes on epoch change that contains their high qc.

Not related to this PR just mentioning it here

dan_layer/consensus/src/hotstuff/on_next_sync_view.rs
```diff
@@ -41,10 +44,23 @@ impl<TConsensusSpec: ConsensusSpec> OnNextSyncViewHandler<TConsensusSpec> {
         info!(target: LOG_TARGET, "⚠️ Leader failure: NEXTSYNCVIEW for epoch {} and node height {}", epoch, new_height);
         let local_committee = self.epoch_manager.get_local_committee(epoch).await?;
         let current_epoch = self.epoch_manager.current_epoch().await?;
-
+        let last_sent_vote = self.store.with_read_tx(|tx| LastSentVote::get(tx));
         let high_qc = self
             .store
             .with_read_tx(|tx| HighQc::get(tx)?.get_quorum_certificate(tx))?;
+        let mut vote = None;
+        if let Ok(last_sent_vote) = last_sent_vote {
```
sdbondi, line 52:
```suggestion
        let last_sent_vote = self.store.with_read_tx(|tx| LastSentVote::get(tx)).optional()?;
        let high_qc = self
            .store
            .with_read_tx(|tx| HighQc::get(tx)?.get_quorum_certificate(tx))?;
        let mut vote = None;
        if let Some(last_sent_vote) = last_sent_vote {
```

dan_layer/consensus/src/hotstuff/on_next_sync_view.rs
```diff
@@ -41,10 +44,23 @@ impl<TConsensusSpec: ConsensusSpec> OnNextSyncViewHandler<TConsensusSpec> {
...
+                vote = Some(VoteMessage {
+                    epoch: last_sent_vote.epoch,
+                    block_id: last_sent_vote.block_id,
+                    block_height: last_sent_vote.block_height,
+                    decision: last_sent_vote.decision,
+                    signature: last_sent_vote.signature,
+                    merkle_proof: last_sent_vote.merkle_proof,
+                });
+            }
+        }
```
sdbondi: Use `optional` combinator to  return Option on NotFound and still return other errors
```suggestion
        let last_sent_vote = self.store.with_read_tx(|tx| LastSentVote::get(tx)).optional()?;
        let high_qc = self
            .store
            .with_read_tx(|tx| HighQc::get(tx)?.get_quorum_certificate(tx))?;
        let vote = last_sent_vote
            .filter(|vote| high_qc.block_height() < vote.block_height)
            .map(|vote| VoteMessage {
                epoch: vote.epoch,
                block_id: vote.block_id,
                block_height: vote.block_height,
                decision: vote.decision,
                signature: vote.signature,
                merkle_proof: vote.merkle_proof,
            });
```

sdbondi (approved):
utACK - didnt test the specific case but in general consensus works.

I added some code to send the last vote from the previous epoch on epoch change to the leader in the next epoch to ensure they have the votes required to justify blocks from a previous epoch. I also had to increase the timeout on the sync test because there are still cases where leader failure can delay things esp on CI

PR #681, opened 2023-09-12
fix(vn): add dedicated messaging substream for consensus msgs

sdbondi:
Description
---
Adds a dedicated substream for consensus messages

Motivation and Context
---
When many transaction messages are being propagated, consensus messages have to wait in line for their turn to be received. This can cause excessive delays in receiving proposals which leads to leader failure on well-bahaving nodes. This PR allows consensus messages to be sent/received in a separate "queue" from transaction messages avoiding the backed up queue of transaction messages.

Observed a decrease in leader failures in stress test conditions and overall improved recovery to GST.

The current transaction gossip strategy needs to be improved. Message complexity is $O(n^2)$ for a single transaction within the local shard 

How Has This Been Tested?
---
Manually, spamming validators with transactions and checking that consensus messages are still sent/received, reducing leader failures

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


stringhandler (approved):

mrnaveira (commented):
LGTM

PR #682, opened 2023-09-12
feat: random JWT secrets on wallet daemon and signaling server

mrnaveira:
Description
---
* Add a new common utility function on `tari_common_types` for secret generation
* Update wallet daemon and signaling server JWT secrets to be randomized at startup

Motivation and Context
---
Right now JWT secrets are hardcoded both in the wallet daemon as well as in the signaling server. This is not acceptable from a security standpoint if we want to create a public test network. So the purpose of this PR is to change both the wallet daemon and the signaling server to haver random JWT secrets each time they start.

Please do not merge this PR until https://github.com/tari-project/tari-connector/pull/9 is merged 

How Has This Been Tested?
---
Manually launching both applications and inspecting the JWTs

---
Add log instructions to see the JWTs and launch both wallet daemon and signaling server

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8152,6 +8152,7 @@ dependencies = [
  "serde",
  "serde_json",
  "tari_common",
+ "tari_dan_common_types",
  "tari_dan_wallet_sdk",
  "tari_shutdown",
  "tokio",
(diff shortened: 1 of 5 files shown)


sdbondi (approved):
Tested wallet daemon. An ephemeral JWT secret makes sense to me. Can override with ` -pdan_wallet_daemon.jwt_secret_key=secret` to avoid having to auth on each restart when deving on the walletd and using e.g. the cli.

Obtaining a JWT for the signaling server is probably not necessary so not too worried about invalidating all of them on restart.

PR #691, opened 2023-09-13
fix(template-lib): re-add buggy template test + fix nostd support

sdbondi:
Description
---
Unignore test_buggy_template test
Fix nostd support in tari-bor
Use ordered map (BTreeMap) directly in AccessRules type, rather than internally for serialization
Remove wee_alloc

Motivation and Context
---
test_buggy_template is currently the only test that uses nostd support. nostd works for templates that choose to use it.

wee_alloc is unmaintained. Replaced with lol_alloc which is actively maintained. Currently used in web extensions and for a single no_std template test

How Has This Been Tested?
---
Engine tests, test_buggy_template works

---
CI

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


stringhandler (dismissed):
I think there is an advisory on weealloc anyway, so this helps

PR #693, opened 2023-09-14
feat(mempool): efficient transaction gossip

sdbondi:
Description
---
Adds efficient gossip procedure to mempool

Motivation and Context
---
The procedure is as follows:
- as a non-involved validator node, gossip to a random set of $f + 1$ nodes in each input shard

If the transaction is not already known:
- as an input shard node: 
   - gossip to $f + 1$ local replicas after the node's position in the local VN set
   - gossip to a small portion𖭫 of other input shard replicas according to their relative position in their respective VN sets
   - do not gossip back to the shard that sent you the transaction
   - after execution, gossip to a small portion𖭫 according to their relative position in their respective VN sets of output shard replicas that are not already part of the input shards
- as an output shard node:
  - verify that the sender is an input shard replica for the transaction, if not discard the message
  - gossip to $f + 1$ local replicas after the node's position in the local VN set
  - do not gossip to any foreign shards (exception: since CreateFreeTestCoins have no inputs we make an exception for 0 input transactions and propagate as if belonging to an input shard)

𖭫 depending on the relative number of committee members in the shard. For example, if the sending shard $V_{a}$ is 3 members and the destination shard $V_{b}$ is 6, each node sends $6/3=2$ messages. If the sending shard size is 6, and the remote 3, each node sends 1 message to the respective node (i.e. $V_{a_1}$ and $V_{a_2}$ send to $V_{b_1}$ etc)

How Has This Been Tested?
---
Manually, 3 committee network sending CreateFreeTestCoin transactions.

---
Test that we do not hit missing transactions in consensus very often

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


stringhandler (dismissed):

stringhandler:
there was a failing test, so rerunning to check if it's related to this pr

mrnaveira (approved):

PR #702, opened 2023-09-15
feat(consensus): consensus state machine

sdbondi:
Description
---
Adds a state machine for consensus worker to allow consensus to switch to/from sync and running modes.

Motivation and Context
---
This PR adds the state machine, the sync implementation will be in a follow-up PR. 
Consensus should not be processing messages during sync. This pattern also fits well with consensus switching to idle when not registered for an epoch.

Diagram [here](https://github.com/tari-project/tari-dan/blob/development/docs/consensus_state_machine.md). This diagram needs to be updated once the transitions and events are closer to final.

The current sync state immediately returns a SyncComplete event, so behaviour is almost identical to the current consensus behaviour.

How Has This Been Tested?
---
Existing test harness updated to use state machine.

---
VN works as before

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_validator_node/src/consensus/mod.rs
+++ b/applications/tari_validator_node/src/consensus/mod.rs
@@ -6,7 +6,7 @@ use std::sync::Arc;
 use tari_common_types::types::PublicKey;
 use tari_comms::{types::CommsPublicKey, NodeIdentity};
 use tari_consensus::{
-    hotstuff::{HotstuffEvent, HotstuffWorker},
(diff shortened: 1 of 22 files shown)


stringhandler (commented):
Happy but seems like there are a lot of CI failures

sdbondi:
Thanks - will check it out

stringhandler (dismissed):

stringhandler (approved):

PR #707, opened 2023-09-29
feat: cache balance merkle tree

Cifko:
Description
---
Cache the BMT per epoch.

Motivation and Context
---
It was computed everytime we needed it, but it's a constant for the whole epoch.

How Has This Been Tested?
---
Manually.

---
Just run the VN with some transaction and you will see that the BMT is stored in the DB.

Breaking Changes
---

- [x] None

### DIFF
--- a/dan_layer/epoch_manager/src/base_layer/base_layer_epoch_manager.rs
+++ b/dan_layer/epoch_manager/src/base_layer/base_layer_epoch_manager.rs
@@ -461,16 +461,24 @@ impl BaseLayerEpochManager<SqliteGlobalDbAdapter, GrpcBaseNodeClient> {
         &self,
         epoch: Epoch,
     ) -> Result<ValidatorNodeBalancedMerkleTree, EpochManagerError> {
+        let db = {
+            let mut tx = self.global_db.create_transaction()?;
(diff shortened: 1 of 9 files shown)


sdbondi (approved):
utACK

dan_layer/epoch_manager/src/base_layer/base_layer_epoch_manager.rs
```diff
@@ -461,16 +461,24 @@ impl BaseLayerEpochManager<SqliteGlobalDbAdapter, GrpcBaseNodeClient> {
...
 
-        // TODO: the MMR struct should be serializable to store it only once and avoid recalculating it every time per
-        // epoch
         let mut vn_bmt_vec = Vec::with_capacity(vns.len());
         for vn in vns {
             vn_bmt_vec.push(vn.node_hash().to_vec())
         }
 
-        let vn_bmt = ValidatorNodeBalancedMerkleTree::create(vn_bmt_vec);
+        let vn_bmt = ValidatorNodeBalancedMerkleTree::create(vn_bmt_vec.clone());
```
sdbondi, line 478: Dont think this clone is necessary
```suggestion
        let vn_bmt = ValidatorNodeBalancedMerkleTree::create(vn_bmt_vec);
```

PR #714, opened 2023-10-04
fix(transactions)!: remove transaction output

Cifko:
Description
---
Remove explicit outputs from transaction.

Motivation and Context
---
Outputs are always determined for the transaction by the validator. A client application will never need to specify these This also enables a vulnerability allowing any user to involve any number of validators in their transaction.

How Has This Been Tested?
---
Dan testing still works.

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/dan_layer/wallet/storage_sqlite/src/models/transaction.rs
+++ b/dan_layer/wallet/storage_sqlite/src/models/transaction.rs
@@ -44,7 +44,6 @@ pub struct Transaction {
 pub struct InputsAndOutputs {
     pub inputs: Vec<ShardId>,
     pub input_refs: Vec<ShardId>,
-    pub outputs: Vec<ShardId>,
 }
 
 impl Transaction {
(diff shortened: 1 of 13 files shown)


dan_layer/wallet/storage_sqlite/src/models/transaction.rs
```diff
@@ -55,11 +54,7 @@ impl Transaction {
                 details: e.to_string(),
             })?;
         let signature = TransactionSignature::new(sender_public_key, signature);
-        let InputsAndOutputs {
-            inputs,
-            input_refs,
-            outputs,
-        } = deserialize_json(&self.meta)?;
+        let InputsAndOutputs { inputs, input_refs } = deserialize_json(&self.meta)?;
```
sdbondi: Maybe:
```suggestion
        let TransactionInputs { inputs, input_refs } = deserialize_json(&self.meta)?;
```

sdbondi (commented):
utACk - one nit

sdbondi (dismissed):

sdbondi (approved):

PR #717, opened 2023-10-05
feat(walletui): add onboarding and settings page

NovaT82:
Description
---
Added an onboarding page.
Added a settings page and moved accounts, keys and access tokens there.
Improved the way adding and switching accounts works.
Also made various UI improvements and restructured the AssetVault component so the various elements are in their own components.

Motivation and Context
---
The onboarding page is necessary for when the wallet is run the first time. I'll work on a nicer design / flow for that page, but the functionality is working so long.
General UI improvements.

How Has This Been Tested?
---
Manually

---
Run the wallet and click on "Create Account" on the onboarding page.
If the onboarding flow isn't showing, clear your localStorage - Zustand saves the current account name there, so there could still be an account name from a previous build.

Breaking Changes
---
x

- [x] None
- [ ] Other - Please specify


sdbondi (dismissed):
utACK

sdbondi (approved):

PR #718, opened 2023-10-06
fix: confidential transfer

Cifko:
Description
---
Fix confidential  transfers.

How Has This Been Tested?
---
By running multiple confidential transfers on dan-testing.

---
Run multiple confidential transfers between different wallets.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (dismissed):
Nice - right so `secret_key_index` refers to the key used for encryption which is $k_a*R$ not $k_{change}*R$. Maybe should rename that to `encryption_secret_key_index` or something, but in any case LGTM!

utACK

sdbondi (dismissed):

sdbondi:
@Cifko Looks like some related tests are failing

sdbondi (approved):

PR #719, opened 2023-10-06
feat: add cli app scaffolder

stringhandler:
Description
---
Adds tari_scaffolder which creates a CLI rust project using clap for calling commands on a contract.

Motivation and Context
---
Usually when you make a template, there will be some methods that are called from a frontend and some that are called from by an admin or superuser. This helps with testing, but also allows you to simply call admin functions without having to create a web front end or use a jsonrpc client like postman.

At some point in the future it would be good to add more output types, other than cli, for example rust or node based websites.

Also, it might be nice to add the ability to download a template from the VN, but for now a local folder must be used.

How Has This Been Tested?
---
Tested locally

---
I run the app using `cargo run --bin tari_scaffolder -- -c  -o "d:\temp\test2" --crates-root "D:/projects/tari-dan/"  scaffold D:\projects\dan-testing\Data\templates\fungible\package\` 

It's important to use `--crates-root` at this time, because the crates are not published on crates.io and also there are changes to the dan wallet client that must be merged with this PR.

Note: it's also better to use an output folder outside of your tari-dan folder so that you don't get that nested workspace issue.

Once that runs, you can run `cargo run -- --help` in the output folder to see the available options. 

I usually copy the dan wallet auth token from the dan wallet frontend, at some point I'll add a flow to get that automatically.

Using the fungible wasm-template. 

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (dismissed):
Nice! utACK

sdbondi (approved):

PR #724, opened 2023-10-11
fix: foreign vaults/accounts

Cifko:
Description
---
Don't add foreign vaults/accounts to your DB.

Motivation and Context
---

How Has This Been Tested?
---
Run dan-testing with some transaction and look into the DB to see there are no foreign vaults.

---
Same as above.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_dan_wallet_daemon/src/services/account_monitor.rs
+++ b/applications/tari_dan_wallet_daemon/src/services/account_monitor.rs
@@ -150,6 +150,11 @@ where
         let substate_api = self.wallet_sdk.substate_api();
         let accounts_api = self.wallet_sdk.accounts_api();
 
+        if !accounts_api.exists_by_address(account_address)? {
+            // This is not our account
+            return Ok(false);
+        }
+
         let mut is_updated = false;
         let account_substate = substate_api.get_substate(account_address)?;
         let ValidatorScanResult {
@@ -227,6 +232,10 @@ where
 
         let balance = vault.balance();
         let vault_addr = SubstateAddress::Vault(*vault.vault_id());
+        if !accounts_api.exists_by_address(account_addr)? {
+            // This is not our account
+            return Ok(());
+        }
         if !accounts_api.has_vault(&vault_addr)? {
             info!(
                 target: LOG_TARGET,
@@ -434,6 +443,10 @@ where
     ) -> Result<(), AccountMonitorError> {
         let vault_addr = SubstateAddress::Vault(*vault.vault_id());
         let accounts_api = self.wallet_sdk.accounts_api();
+        if !accounts_api.exists_by_address(account_addr)? {
+            // This is not our account
+            return Ok(());
+        }
         if accounts_api.has_vault(&vault_addr)? {
             return Ok(());
         }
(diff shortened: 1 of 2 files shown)


dan_layer/wallet/storage_sqlite/src/writer.rs
```diff
@@ -455,21 +455,23 @@ impl WalletStoreWriter for WriteTransaction<'_> {
...
-            vaults::resource_address.eq(vault.resource_address.to_string()),
-            vaults::resource_type.eq(format!("{:?}", vault.resource_type)),
-            vaults::token_symbol.eq(vault.token_symbol),
-        );
-        diesel::insert_into(vaults::table)
-            .values(values)
-            .execute(self.connection())
-            .map_err(|e| WalletStorageError::general("vaults_insert", e))?;
-
+        if let Some(account_id) = account_id {
```
sdbondi: I think that we should return an error here if account_id is None because it should be invalid to call `vault_insert` without an associated account for the vault - rather than silently not inserting the vault. Would need to check that the account exists somewhere before calling `vault_insert`.

applications/tari_dan_wallet_daemon/src/services/account_monitor.rs
```diff
@@ -244,6 +244,10 @@ where
             )?;
         }
 
+        if !accounts_api.has_vault(&vault_addr)? {
```
sdbondi: Shouldn't we add this check before adding the vault? around line 228

Cifko: yes

applications/tari_dan_wallet_daemon/src/services/account_monitor.rs
```diff
@@ -227,6 +227,10 @@ where
 
         let balance = vault.balance();
         let vault_addr = SubstateAddress::Vault(*vault.vault_id());
+        if accounts_api.get_account_by_address(account_addr).is_err() {
```
sdbondi:
```suggestion
        if accounts_api.get_account_by_address(account_addr).optional()?.is_none() {
```

sdbondi: Happy to merge with above

Bonus: create `accounts_api.exists_by_address(account_addr)?` and internally either create the `tx.accounts_exists_by_address` call or use `accounts_api.get_account_by_address(account_addr).optional()?.is_none()` inside the function.

This is so that if we want to change the backend for these calls we don't have to change usage everywhere.

Cifko: done

Cifko:
Btw we still endup with substates that don't belong to our vallet. Because we just add everything that is up/down in a transaction.

sdbondi:
> Btw we still endup with substates that don't belong to our vallet. Because we just add everything that is up/down in a transaction.

Yeah thats ok, we just take note of substate addresses we encounter in the wallet to help us figure out which substates are related. We dont keep the actual state

sdbondi (approved):

PR #727, opened 2023-10-12
feat!: min/max epoch for transaction

Cifko:
Description
---
Add min/max epoch for transaction. If the current epoch is not in between (included) the epochs, it's discarded.

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [ ] None
- [ ] Other - Please specify


sdbondi (commented):
LGTM, the description says it discards the message if outside the epoch range, however I don't see the code for that. I'm fine to merge this if either that code is added, or the description is changed and we do another PR to add that validation to the mempool and consensus.

Cifko:
> LGTM, the description says it discards the message if outside the epoch range, however I don't see the code for that. I'm fine to merge this if either that code is added, or the description is changed and we do another PR to add that validation to the mempool and consensus.

It got lost in rebase :-) now added in last commit

sdbondi (approved):
utACK

PR #729, opened 2023-10-16
feat: add transaction failure reason

Cifko:
Description
---
Fill the transaction failure field in transaction. Add the reason to ui. Also add reason when the transaction is committed but only fee is accepted. Change transfer to still pay fees if owner tries to send too much tari.

Motivation and Context
---

How Has This Been Tested?
---
Running dan-testing. With 3 transaction (accepted, only fee is accepted, all failed). To see the ui.

---
Same as me.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


applications/tari_dan_wallet_daemon/src/handlers/accounts.rs
```diff
@@ -680,11 +681,13 @@ pub async fn handle_claim_burn(
         new_account: None,
     });
 
-    let finalized = wait_for_result(&mut events, tx_id).await?;
+    let mut finalized = wait_for_result(&mut events, tx_id).await?;
     if let Some(reject) = finalized.finalize.result.reject() {
         return Err(anyhow::anyhow!("Fee transaction rejected: {}", reject));
     }
     if let Some(reason) = finalized.transaction_failure {
+        finalized.finalize.result =
```
sdbondi: The transaction result types are a bit of a mess since fee instructions were added (my bad). I think that the execution engine should return this enum variant (maybe `FeeAccept` and `FullAccept` idk) rather than the wallet modifying things. Then we should be able to remove the "weird" `transaction_result` field entirely.

stringhandler: Yeah, I agree with that. I don't mind `FeeAccept` and `FullAccept`, but here are some other suggestions:

`FailedButCharged`
`FailedButFeesDeducted`
`OnlyFeesAccept`
`AcceptWithFailures`

dan_layer/engine/tests/fees.rs
```diff
@@ -238,8 +238,12 @@ fn fail_partial_paid_fees() {
 
     test.disable_fees();
 
+    println!("HERE?");
```
sdbondi: :broom:

dan_layer/engine/src/transaction/processor.rs
```diff
@@ -154,6 +154,14 @@ impl<TTemplateProvider: TemplateProvider<Template = LoadedTemplate> + 'static> T
                 } = runtime.interface().finalize()?;
 
                 if !fee_receipt.is_paid_in_full() {
+                    finalized.result = TransactionResult::AcceptFeeRejectRest(
+                        finalized.result.accept().unwrap().clone(),
```
sdbondi: finalized is not guaranteed to be accept

integration_tests/src/wallet_daemon_cli.rs
```diff
@@ -514,7 +514,7 @@ pub async fn submit_manifest_with_signing_keys(
         timeout_secs: Some(120),
     };
     let wait_resp = client.wait_transaction_result(wait_req).await.unwrap();
-    if let Some(reason) = wait_resp.transaction_failure {
+    if let Some(reason) = wait_resp.result.clone().and_then(|result| result.reject().cloned()) {
```
sdbondi: No need to double clone
```suggestion
    if let Some(reason) = wait_resp.result.as_ref().and_then(|result| result.reject().cloned()) {
```

dan_layer/wallet/sdk/src/apis/transaction.rs
```diff
@@ -148,11 +146,20 @@ where
             TransactionFinalizedResult::Finalized {
                 final_decision,
                 execution_result,
-                abort_details,
+                abort_details: _,
                 json_results,
             } => {
                 let new_status = if final_decision.is_commit() {
-                    TransactionStatus::Accepted
+                    match execution_result.clone() {
```
sdbondi:
```suggestion
                    match execution_result.as_ref() {
```

applications/tari_dan_wallet_daemon/src/handlers/accounts.rs
```diff
@@ -902,19 +904,21 @@ pub async fn handle_transfer(
...
             args: args![Workspace("bucket")],
         },
-        Instruction::CallMethod {
-            component_address: source_account_address,
-            method: "pay_fee".to_string(),
-            args: args![max_fee],
-        },
     ]);
 
+    fee_instructions.append(&mut vec![Instruction::CallMethod {
```
sdbondi, line 909: Why this change?

Suggest adding the instruction to the end and using `Vec::extend` rather than `Vec::append`.
```rust
// vec![....] allocates on the heap, then append takes each allocated item and the vec is dropped
    fee_instructions.append(&mut vec![Instruction::CallMethod {
```

`Vec::extend`  takes any iterator, in this case, a static array of instructions 
```suggestion
    fee_instructions.extend([Instruction::CallMethod {
```

integration_tests/src/wallet_daemon_cli.rs
```diff
@@ -540,8 +540,8 @@ pub async fn submit_manifest(
     };
     let wait_resp = client.wait_transaction_result(wait_req).await.unwrap();
 
-    if let Some(reason) = wait_resp.transaction_failure {
-        panic!("Transaction failed: {}", reason);
+    if let Some(reason) = wait_resp.result.clone().and_then(|finalize| finalize.reject().cloned()) {
```
sdbondi, line 543:
```suggestion
    if let Some(reason) = wait_resp.result.as_ref().and_then(|finalize| finalize.reject()) {
```

integration_tests/src/wallet_daemon_cli.rs
```diff
@@ -466,7 +466,7 @@ pub async fn submit_manifest_with_signing_keys(
         timeout_secs: Some(120),
     };
     let wait_resp = client.wait_transaction_result(wait_req).await.unwrap();
-    if let Some(reason) = wait_resp.transaction_failure {
+    if let Some(reason) = wait_resp.result.as_ref().and_then(|result| result.reject().cloned()) {
```
sdbondi, line 469:
```suggestion
    if let Some(reason) = wait_resp.result.as_ref().and_then(|result| result.reject()) {
```

integration_tests/src/wallet_daemon_cli.rs
```diff
@@ -637,7 +637,7 @@ pub async fn create_component(
     };
     let wait_resp = client.wait_transaction_result(wait_req).await.unwrap();
 
-    if let Some(reason) = wait_resp.transaction_failure {
+    if let Some(reason) = wait_resp.result.clone().and_then(|finalize| finalize.reject().cloned()) {
```
sdbondi, line 640:
```suggestion
    if let Some(reason) = wait_resp.result.as_ref().and_then(|finalize| finalize.reject()) {
```

sdbondi (approved):

PR #738, opened 2023-10-17
feat: add jrpc address for webui to connect to

Cifko:
Description
---

Motivation and Context
---

How Has This Been Tested?
---
I created a new [PR](https://github.com/tari-project/dan-testing/pull/43) in dan-testing that set the listening address to `0.0.0.0:<port>` and the webui jrpc address to `127.0.0.1:<port>`. 

---
Either do the same as me or test my other 

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_indexer/src/config.rs
+++ b/applications/tari_indexer/src/config.rs
@@ -83,6 +83,8 @@ pub struct IndexerConfig {
     pub graphql_address: Option<SocketAddr>,
     /// The address of the HTTP UI
     pub http_ui_address: Option<SocketAddr>,
+    /// The jrpc address where the UI should connect
+    pub jrpc_for_ui_address: Option<String>,
     /// Substate addresses to keep watching
     pub address_watchlist: Vec<SubstateAddress>,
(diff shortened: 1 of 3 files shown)


applications/tari_indexer/src/config.rs
```diff
@@ -83,6 +83,8 @@ pub struct IndexerConfig {
     pub graphql_address: Option<SocketAddr>,
     /// The address of the HTTP UI
     pub http_ui_address: Option<SocketAddr>,
+    /// The jrpc address where the UI should connect
+    pub jrpc_for_ui_address: Option<String>,
```
sdbondi, line 87: You can use 
```suggestion
    pub jrpc_for_ui_address: Option<SocketAddr>,
```

sdbondi (approved):

applications/tari_indexer/src/config.rs
Cifko: I can't because there will be no real ip in some case, it will be a domain name.

PR #741, opened 2023-10-19
feat: change base node grpc address to string

Cifko:
Description
---
Change the base node grpc address to string so we can use domain names instead of ips.

Motivation and Context
---

How Has This Been Tested?
---
Running `dan-testing`

---
Set the `base_node_grpc_address` in the config.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi:
Looks like the cucumbers need updating to use an URL

integration_tests/src/miner.rs
```diff
@@ -63,7 +63,11 @@ pub fn register_miner_process(world: &mut TariWorld, miner_name: String, base_no
 
 pub async fn mine_blocks(world: &mut TariWorld, miner_name: String, num_blocks: u64) {
     let miner = world.get_miner(&miner_name);
+    println!("a");
```
sdbondi: :broom:

sdbondi (approved):

PR #744, opened 2023-10-20
feat: adds settings JSON-RPC call to allow changes to indexer URL

Cifko:
Description
---
Add new jrpc calls that will allow change of the indexer jrpc url in the wallet daemon.

Motivation and Context
---

How Has This Been Tested?
---
as below

---
Run dan-testing with 2 indexers. And call the jrpc method to change it. And e.g. send money using the new indexer. Then you should clearly see in the logs which one was used.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


applications/tari_dan_wallet_daemon/src/jrpc_server.rs
```diff
@@ -101,6 +102,11 @@ async fn handler(
             "get_all_jwt" => call_handler(context, value, token, rpc::handle_get_all_jwt).await,
             _ => Ok(value.method_not_found(&value.method)),
         },
+        Some(("settings", method)) => match method {
+            "get_indexer_url" => call_handler(context, value, token, settings::handle_get_indexer_url).await,
```
sdbondi: We may want to have `"get"` and `"set"` operations for settings. Then for now we only have the `indexer_url` in the Request/Response objects but later we can add more, rather than having get/set handlers for each setting.

Cifko: done

sdbondi: Thanks

sdbondi (dismissed):
utACK

sdbondi (dismissed):

sdbondi (approved):

PR #746, opened 2023-10-23
fix(template)!: make token_symbol part of the metadata again in resources

mrnaveira:
Description
---
Integrated back the `token_symbol` field into the `metadata` field in resources

Motivation and Context
---
After https://github.com/tari-project/tari-dan/pull/743, the `token_symbol` field in resources is not used anymore for address calculation. This means that the original reason the field was made mandatory thought the whole project (and extracted out of the `metadata` field) is no longer present.

This PR makes the token `token_symbol` field in resources optional again, by integrating it into the `metadata` field as originally implemented.

How Has This Been Tested?
---
* All existing unit and integration tests pass
* Manually started a fresh new network, created an account with funds and inspected the balances in the wallet daemon 

---
Launch the network and perform operations on resources

Breaking Changes
---

- [ ] None
- [x] Requires a full network reset, as the template engine API to create resources changes, as well as the account_nfts builtin template


sdbondi (approved):
utACK

The wallet could still index the token symbol in the database, but we can add that if needed

PR #762, opened 2023-11-08
chore: bump up the tari-crypto and tari-utilities

Cifko:
Description
---
Bump up the versions of `tari-crypto` and `tari-utilities`

Motivation and Context
---

How Has This Been Tested?
---
I run the dan-testing with stress test enabled.

---
The same as above.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -140,31 +140,32 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.7.6"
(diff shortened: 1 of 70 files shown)


stringhandler (dismissed):

stringhandler (approved):
There's probably a way to do a generic for the U32 and U64 but this is fine for now

PR #766, opened 2023-11-10
fix(mempool): correct shard id for transaction receipt

sdbondi:
Description
---
Calculate the correct shard id for the transaction receipt in the mempool and indexer
fix(indexer): indexer submits the auto-filled transaction instead of the original transaction
feat(indexer): indexer attempts to submit the transaction to any member of all input shards

Motivation and Context
---
The transaction receipt shard id was calculated incorrectly causing the incorrect output shard to be selected for propagation in some cases. Thanks to @Cifko for discovering the issue.

How Has This Been Tested?
---
This case is not explicitly tested :/

---
Check that the transaction is propagated to the correct output shards in a multi-shard network

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_indexer/src/transaction_manager/mod.rs
+++ b/applications/tari_indexer/src/transaction_manager/mod.rs
@@ -22,10 +22,9 @@
 
 mod error;
 
-use std::{fmt::Display, future::Future, sync::Arc};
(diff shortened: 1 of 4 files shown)


Cifko (approved):
tested
LGTM

PR #768, opened 2023-11-14
fix: multi committee transaction

Cifko:
Description
---
This fixes two bugs:
- every VN now distributes the transaction after execution (we can maybe limit this to the VNs that have the TxReceipt, because that's the shard where is it always distributed).
- VNs is no longer trying to delete substates that are in foreign committees.

Motivation and Context
---

How Has This Been Tested?
---
Running dan-testing with 2 committees and several transactions.

---
Same as above.

Breaking Changes
---

- [ ] None
- [x] Other - Older VNs are currently not working 100% properly and they need to be recompiled.

### DIFF
--- a/applications/tari_validator_node/src/consensus/state_manager.rs
+++ b/applications/tari_validator_node/src/consensus/state_manager.rs
@@ -2,7 +2,7 @@
 //    SPDX-License-Identifier: BSD-3-Clause
 
 use tari_consensus::traits::StateManager;
-use tari_dan_common_types::ShardId;
+use tari_dan_common_types::{committee::CommitteeShard, ShardId};
 use tari_dan_storage::{
     consensus_models::{Block, ExecutedTransaction, SubstateRecord},
     StateStore,
@@ -25,6 +25,7 @@ impl<TStateStore: StateStore> StateManager<TStateStore> for TariStateManager {
         tx: &mut TStateStore::WriteTransaction<'_>,
         block: &Block<TStateStore::Addr>,
         transaction: &ExecutedTransaction,
+        local_committee_shard: &CommitteeShard,
     ) -> Result<(), Self::Error> {
         let Some(diff) = transaction.result().finalize.result.accept() else {
             // We should only commit accepted transactions, might want to change this API to reflect that
@@ -33,7 +34,8 @@ impl<TStateStore: StateStore> StateManager<TStateStore> for TariStateManager {
 
         let down_shards = diff
             .down_iter()
-            .map(|(addr, version)| ShardId::from_address(addr, *version));
+            .map(|(addr, version)| ShardId::from_address(addr, *version))
+            .filter(|shard| local_committee_shard.includes_shard(shard));
         SubstateRecord::destroy_many(
             tx,
             down_shards,
@@ -44,17 +46,22 @@ impl<TStateStore: StateStore> StateManager<TStateStore> for TariStateManager {
             true,
         )?;
 
-        let to_up = diff.up_iter().map(|(addr, substate)| {
-            SubstateRecord::new(
-                addr.clone(),
-                substate.version(),
-                substate.substate_value().clone(),
-                block.epoch(),
-                block.height(),
-                *block.id(),
-                *transaction.id(),
-                *block.justify().id(),
-            )
+        let to_up = diff.up_iter().filter_map(|(addr, substate)| {
+            let shard_id = ShardId::from_address(addr, substate.version());
+            if local_committee_shard.includes_shard(&shard_id) {
+                Some(SubstateRecord::new(
+                    addr.clone(),
+                    substate.version(),
+                    substate.substate_value().clone(),
+                    block.epoch(),
+                    block.height(),
+                    *block.id(),
+                    *transaction.id(),
+                    *block.justify().id(),
+                ))
+            } else {
+                None
+            }
         });
 
         for up in to_up {
--- a/applications/tari_validator_node/src/p2p/services/mempool/service.rs
+++ b/applications/tari_validator_node/src/p2p/services/mempool/service.rs
@@ -344,7 +344,7 @@ where
                     );
                 }
 
-                // Only input shards propagate to foreign shards
+                // Only tx receipt shards propagate to foreign shards
                 if is_input_shard {
                     // Forward to foreign replicas.
                     // We assume that at least f other local replicas receive this transaction and also forward to their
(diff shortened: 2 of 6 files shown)


applications/tari_validator_node/src/consensus/state_manager.rs
```diff
@@ -33,7 +34,8 @@ impl<TStateStore: StateStore> StateManager<TStateStore> for TariStateManager {
 
         let down_shards = diff
             .down_iter()
-            .map(|(addr, version)| ShardId::from_address(addr, *version));
+            .map(|(addr, version)| ShardId::from_address(addr, *version))
+            .filter(|shard| local_committee_shard.includes_shard(shard));
```
sdbondi, line 38: We should also do the same for the UP substates no?

applications/tari_validator_node/src/p2p/services/mempool/service.rs
```diff
@@ -586,21 +586,14 @@ where
...
-            // Forward the transaction to any output shards that are not part of the input shard set as these have
-            // already been forwarded
-            let num_committees = self.epoch_manager.get_num_committees(current_epoch).await?;
-            let input_buckets = executed
-                .involved_shards_iter()
-                .map(|s| s.to_committee_bucket(num_committees))
-                .collect::<HashSet<_>>();
+
+        // TODO: Should we always propagate? The TX is initially send only to single output shard (TxReceipt).
+        if should_propagate {
```
sdbondi: Please explain this further. The rule is that only input shards can send to output shards i.e. VNs which are not involved in the transaction will not execute it unless an input shard VN tells them that they are an output shard. This is done to limit who can cause VNs to execute that shouldn't, to input shard VNs (and you'll know that the input shard VN was acting incorrectly). Happy to change that rule but there should be a reason. If there is a problem/bug/edge-case with the implementation then that should be fixed.

sdbondi (commented):
+1 to only DOWNing/UPing local substates. Have some questions on the other change.

applications/tari_validator_node/src/p2p/services/mempool/service.rs
sdbondi: We propagate to all input shards before execution. This propagation happens after execution, and that should only be sent to output shards since input shard propagation has already happened.

sdbondi: The TxReceipt is a special case, so we could/should propagate before execution to the txreceipt shard

applications/tari_validator_node/src/consensus/state_manager.rs
Cifko: added (they were just not critical)

applications/tari_validator_node/src/p2p/services/mempool/service.rs
Cifko: Added back the filtering. Working on the distribution via TxReceipt only.

dan_layer/consensus/src/hotstuff/on_propose.rs
```diff
@@ -196,26 +196,33 @@ where TConsensusSpec: ConsensusSpec
...
-                    let involved = NonZeroU64::new(involved as u64).ok_or_else(|| {
-                        HotStuffError::InvariantError(format!(
-                            "Number of involved shards is zero for transaction {}",
-                            t.transaction_id(),
-                        ))
-                    })?;
-                    let leader_fee = t.calculate_leader_fee(involved, EXHAUST_DIVISOR);
-                    total_leader_fee += leader_fee;
-                    Ok(Command::Accept(t.get_final_transaction_atom(leader_fee)))
+                    if t.transaction().evidence.all_shards_complete() {
```
sdbondi: The transaction is never marked as is_ready == true unless we have `evidence.all_shards_complete()`. Only is_ready==true transactions are returned `get_batch_for_next_block`. So this shouldn't be necessary to check again (this change should have no effect). Are you seeing Accept commands before evidence is complete?

sdbondi: If there is a bug where is_ready is true, this will create a block with less transactions and hide the bug. There may be a bug in the way we work out each transaction atom's state (stage & is_ready) from the current chain from the locked block to the leaf block.

sdbondi: Or it probably is a race condition, where some validators have seen the foreign local prepared, and some haven't. In the case where we dont have all the evidence, I think we currently abstain from voting which is probably correct as long as we're able to recover once the block gets to locked block. This is where we may have to request the foreign `LocalPrepared`.

applications/tari_validator_node/src/p2p/services/mempool/service.rs
sdbondi: I think we can readd the is_input_shard check now (since it will result in messages being sent by output shards to the output shards again). And have the input shards propagate before execution to the TxReciept output shard. We already check that we're an [output shard from the TxReceipt](https://github.com/tari-project/tari-dan/blob/9c2f369f69c27b570fce96b5316e1f62e694a032/applications/tari_validator_node/src/p2p/services/mempool/service.rs#L316). We already handle the [special case when receiving a transaction for output shards](https://github.com/tari-project/tari-dan/blob/9c2f369f69c27b570fce96b5316e1f62e694a032/applications/tari_validator_node/src/p2p/services/mempool/service.rs#L254) and the transaction has no inputs.

sdbondi: So I think the only change that is needed is to include the tx receipt in propagation [here](https://github.com/tari-project/tari-dan/blob/9c2f369f69c27b570fce96b5316e1f62e694a032/applications/tari_validator_node/src/p2p/services/mempool/service.rs#L356)

```rust
   transaction.involved_shards_iter().copied().chain(Some(tx_shard_id)).collect(),
```

sdbondi: Ah ok, so we would have other output shards too if the account is created from free coins, not just the TxReciept.

In this case, we also need to propagate to non-TxReceipt output shards after execution. So maybe we should add another clause to the if for this case rather than always propagating? So that "normal" transactions don't result in output shards repropagating to other output shards needlessly.

```rust
// If we're an input shard, or in the special case of no inputs, we propagate to the output shards
       if should_propagate && (is_input_shard || executed.transaction().num_involved_shards() == 0) {
```

Cifko: What I do now is that indexer sends the TX only to transaction id shard. That propagates to known (including filled) input shards. After the execution each if we are the transaction id shard we propagate it to any shard that was not in the inputs (i.e. output shards). So there is much less propagation going around, only from the transaction id shard. And only once to every other bucket.

stringhandler (approved):

sdbondi (approved):

PR #785, opened 2023-11-24
feat: better tuple support in templates

mrnaveira:
Description
---
* Added a new variant (`Type::Tuple`) for tuples in the template ABI definition
* Updated the template macros to generate correct ABI for tuples and support tuple arguments
* Updated the existing engine test (and associated template) regarding tuples
* Updated CLI display for tuples in `validator_node_cli` and `tari_dan_wallet_cli` 

Motivation and Context
---
Up until now we had support for tuples in templates, but only for return types (https://github.com/tari-project/tari-dan/pull/10)

Recently we have stumbled upon some limitations with tuples:
* The ABI definition is not descriptive, as we lack a specific variant to handle tuples. We want the display to look like `Tuple<String,U32>`
* Tuples are not supported as arguments to functions or methods

This PR aims to fix those limitations.

How Has This Been Tested?
---
Improved existing tuple template test in the engine:
* It checks more scenarios (tuples as arguments as well as return types)
* It checks ABI for both tuple arguments and return types

---
Deploy a template that deals with tuples and inspect ABI or do calls

Breaking Changes
---

- [ ] None
- [x] Requires network reset. Template ABI has changed, so new templates may be incompatible with previous versions of the network.

### DIFF
--- a/applications/tari_dan_wallet_cli/src/command/transaction.rs
+++ b/applications/tari_dan_wallet_cli/src/command/transaction.rs
@@ -668,7 +668,7 @@ pub fn print_execution_results(results: &[InstructionResult]) {
                     },
                 }
             },
-            Type::Tuple(subtypes)=> {
+            Type::Tuple(subtypes) => {
                 let str = format_tuple(subtypes, result);
                 println!("{}", str);
             },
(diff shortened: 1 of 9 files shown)


sdbondi (approved):
utACK

Nice!

PR #791, opened 2023-11-27
feat: proper transaction signature and validation

mrnaveira:
Description
---
* Added all relevant fields to the transaction signature
* Renamed the corresponding domain separated hash from `InstructionSignature` to `TransactionSignature`
* Created a new validation in the validator node's mempool for signatures, with a new `InvalidSignature` error variant

Motivation and Context
---
There are some missing logic regarding transaction signatures:
* It only includes the `instructions` field in the challenge
* It's not validated on validator nodes when arriving to the mempool

How Has This Been Tested?
---
* Existing unit and integration test pass
* Manually submit valid and invalid transactions using `dan-testing`

---
Use the network to submit transactions, both valid and invalid

Breaking Changes
---
- [ ] None
- [x] Breaking change. Transaction signatures are now validated and include new fields, so VNs will reject transactions built on previous versions


sdbondi (approved):
utACK

Awesome thanks - LGTM

PR #757, opened 2023-10-31
feat!: foreign broadcast reliability counter

Cifko:
Description
---
Add indexes/counters to blocks. So nodes in foreign committees know if they missed some foreign proposals. The VN ignores blocks with invalid foreign counter :
- receive foreign proposal with invalid index
- foreign proposal is missing index for our bucket
- local proposal has wrong foreign indexes

Motivation and Context
---

How Has This Been Tested?
---

What process can a PR reviewer use to test or verify this change?
---

Breaking Changes
---

- [ ] None
- [ ] Requires data directory to be deleted
- [x] Other - Please specify

BREAKING CHANGE: New consensus message structure

PR #792, opened 2023-11-28
feat: foreign proposal command

Cifko:
Description
---
It's build on top of #757 
In current state everyone propose the foreign proposal command all the time.
If I build on top of block where the foreign proposal is, then I can not propose it and mark it as proposed.
But everyone else needs to check if there is a block on top of the block where the command is to mark it as well. But the command is not in the locked block. So it can be rollbacked, right?
For the transactions we have something similar where we keep track of the changes until they are locked, in the `transaction_pool_state_updates` table.

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


stringhandler (dismissed):
tested locally with 20 vns of 3 committee size

stringhandler (approved):

PR #802, opened 2023-12-01
feat: foreign proposal transaction timeout

Cifko:
Description
---
Transaction (from foreign proposals) timeout.
For every Foreign proposal we store new column with all the transactions id. Once we receive local block where the transaction should be resolved and it's not (it's not finalized, or the current state is not at least LocalPrepared) we update the local decision to abort.

Motivation and Context
---

How Has This Been Tested?
---
N/A

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/dan_layer/consensus/src/hotstuff/on_receive_local_proposal.rs
+++ b/dan_layer/consensus/src/hotstuff/on_receive_local_proposal.rs
@@ -126,9 +126,7 @@ impl<TConsensusSpec: ConsensusSpec> OnReceiveLocalProposalHandler<TConsensusSpec
         let local_committee_shard = self.epoch_manager.get_local_committee_shard(block.epoch()).await?;
 
         let maybe_high_qc_and_block = self.store.with_write_tx(|tx| {
-            let Some(valid_block) =
-                self.validate_block(tx.deref_mut(), block, &local_committee, &local_committee_shard)?
-            else {
+            let Some(valid_block) = self.validate_block(tx, block, &local_committee, &local_committee_shard)? else {
                 return Ok(None);
             };
 
@@ -164,7 +162,7 @@ impl<TConsensusSpec: ConsensusSpec> OnReceiveLocalProposalHandler<TConsensusSpec
 
     fn validate_block(
         &self,
-        tx: &mut <TConsensusSpec::StateStore as StateStore>::ReadTransaction<'_>,
+        tx: &mut <TConsensusSpec::StateStore as StateStore>::WriteTransaction<'_>,
         block: Block,
         local_committee: &Committee<TConsensusSpec::Addr>,
         local_committee_shard: &CommitteeShard,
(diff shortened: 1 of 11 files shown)


stringhandler (dismissed):
utACK, seems correct

dan_layer/consensus/src/hotstuff/on_receive_local_proposal.rs
```diff
@@ -346,14 +358,40 @@ impl<TConsensusSpec: ConsensusSpec> OnReceiveLocalProposalHandler<TConsensusSpec
...
             return Err(ProposalValidationError::NotSafeBlock {
                 proposed_by: candidate_block.proposed_by().to_string(),
                 hash: *candidate_block.id(),
             }
             .into());
         }
 
+        // TODO: Move this to consensus constants
+        const TIMEOUT: u64 = 1000;
+        let all_mined_proposals = ForeignProposal::get_all_mined(
```
sdbondi: We should avoid the 'mined' terminology on L2. maybe `get_all_proposed`

dan_layer/p2p/proto/consensus.proto
```diff
@@ -74,6 +74,7 @@ message ForeignProposal {
   bytes block_id = 2;
   ForeignProposalState state = 3;
   uint64 mined_at = 4;
```
sdbondi, line 77: Suggest `proposed_height`. This has already changed in development but I missed this in the proto file

dan_layer/state_store_sqlite/migrations/2023-06-08-091819_create_state_store/up.sql
```diff
@@ -249,11 +249,12 @@ CREATE TABLE missing_transactions
...
-    id         integer   not NULL primary key AUTOINCREMENT,
-    bucket     bigint    not NULL,
-    block_id   text      not NULL,
-    state      text      not NULL,
-    mined_at   bigint    NULL,
+    id           integer not NULL primary key AUTOINCREMENT,
+    bucket       bigint  not NULL,
+    block_id     text    not NULL,
+    state        text    not NULL,
+    mined_at     bigint  NULL,
```
sdbondi: suggest: `proposed_height`

dan_layer/state_store_sqlite/src/reader.rs
```diff
@@ -438,6 +439,21 @@ impl<TAddr: NodeAddressable + Serialize + DeserializeOwned> StateStoreReadTransa
             .collect::<Vec<ForeignProposal>>())
     }
 
+    fn foreign_proposal_get_all_mined(&mut self, to_height: NodeHeight) -> Result<Vec<ForeignProposal>, StorageError> {
+        use crate::schema::foreign_proposals;
+
+        let foreign_proposals = foreign_proposals::table
+            .filter(foreign_proposals::state.eq("Mined"))
```
sdbondi: This will result in a bug since ForeignProposalState::Mined no longer exists
```suggestion
            .filter(foreign_proposals::state.eq(ForeignProposalState::Proposed.to_string()))
```

sdbondi (commented):
LGTM
I don't think we should be using 'mined` when talking about L2 blocks.

sdbondi (approved):
utACK

Tested but only with a single shard

PR #825, opened 2023-12-07
feat: add version to template WASMs

mrnaveira:
Description
---
* The `template_lib` crate exports a constant `VERSION` with its Cargo package version at compile time.
* On template WASM compilation, a new function `_GET_TEMPLATE_LIB_VERSION` will be exported that will reexport the `template_lib` version used when compiling the template.
* Validator nodes call this new function when loading a new template WASM for execution, and check the version against the engine's own version. For now, no errors are returned even if they don't match, only logging is performed.

Motivation and Context
---
We want to know the `template_lib` version that a template was compiled with, as this determines the engine calls that it may request during execution. The goal is to be able to detect version incompatibilities when executing.

For now this PR only adds the version to the template WASMs and logs version match/mismatch, no error is returned. In the future this setup will allow us to load the appropriate `template_lib` version for each template and avoid incompatibilities.

How Has This Been Tested?
---
Launching a local network and inspecting the VN logs when executing a transaction for the `template_lib` version messages

---
Inspect the VN logs when executing a transaction for the `template_lib` version messages

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


dan_layer/template_macros/src/template/template_lib_version.rs
```diff
@@ -0,0 +1,41 @@
...
+    let version_function_name = format_ident!("{TEMPLATE_LIB_VERSION_FUNCTION_GLOBAL_NAME}");
+
+    quote! {
+        #[no_mangle]
+        pub unsafe extern "C" fn #version_function_name() -> *mut u8 {
+            use ::tari_template_lib::VERSION;
+            use ::tari_template_lib::template_dependencies::{wrap_ptr, encode_with_len};
+            let result = encode_with_len(&VERSION);
+            wrap_ptr(result)
+        }
```
sdbondi, line 39: I think we should add a version field in the template def. See `dan_layer/template_macros/src/template/abi.rs`

A malicious template developer could add e.g. an infinite loop in a function call, which is handled by metering in the dispatcher call but not here. It's simpler to add it as a static value and we may as well add it to the template def

PR #826, opened 2023-12-08
feat: do fee estimation in UI transfer

Cifko:
Description
---
Add fee estimation to transfer in the UI.
The way it works, the fee field is readonly, instead of "send tari" there is a "Estimate Fee", once you click it, it will calculate the fees and populate the fee field and the button changes to "Send", if you change anything, it will go back to "Estimate Fee".
The problem is that we have to enter the maximum fee for estimation. If user don't have enough money (the amount for transfer+the maximum fee for calculation) that's problem. In current state the maximum fee is set to 1000. So if it's going to be more, the estimation.

Motivation and Context
---

How Has This Been Tested?
---
Manually.

---
Go to the send dialog in the wallet ui.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_dan_wallet_daemon/src/handlers/accounts.rs
+++ b/applications/tari_dan_wallet_daemon/src/handlers/accounts.rs
@@ -933,9 +933,21 @@ pub async fn handle_transfer(
         let execute_result = result.result.into_execute_result().unwrap();
         return Ok(TransferResponse {
             transaction_id: result.transaction_id,
-            fee: execute_result.fee_receipt.clone().unwrap().total_fees_paid,
-            fee_refunded: execute_result.fee_receipt.clone().unwrap().total_fee_payment -
-                execute_result.fee_receipt.clone().unwrap().total_fees_paid,
+            fee: execute_result
+                .fee_receipt
+                .clone()
+                .map(|fee_receipt| fee_receipt.total_fees_paid)
+                .unwrap_or_default(),
+            fee_refunded: execute_result
+                .fee_receipt
+                .clone()
+                .map(|fee_receipt| fee_receipt.total_fee_payment)
+                .unwrap_or_default() -
+                execute_result
+                    .fee_receipt
+                    .clone()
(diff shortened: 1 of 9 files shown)


sdbondi (approved):

applications/tari_dan_wallet_daemon/src/handlers/accounts.rs
```diff
@@ -925,6 +925,32 @@ pub async fn handle_transfer(
...
+        let result = sdk
+            .transaction_api()
+            .submit_dry_run_transaction(transaction, required_inputs)
+            .await?;
+        let execute_result = result.result.into_execute_result().unwrap();
+        return Ok(TransferResponse {
+            transaction_id: result.transaction_id,
+            fee: execute_result
+                .fee_receipt
+                .clone()
```
sdbondi, line 938:
```suggestion
                .as_ref()
```

applications/tari_dan_wallet_daemon/src/handlers/accounts.rs
```diff
@@ -925,6 +925,32 @@ pub async fn handle_transfer(
...
+                .map(|fee_receipt| fee_receipt.total_fees_paid)
+                .unwrap_or_default(),
+            fee_refunded: execute_result
+                .fee_receipt
+                .clone()
```
sdbondi, line 943:
```suggestion
                .as_ref()
```

applications/tari_dan_wallet_daemon/src/handlers/accounts.rs
```diff
@@ -925,6 +925,32 @@ pub async fn handle_transfer(
...
+                .map(|fee_receipt| fee_receipt.total_fee_payment)
+                .unwrap_or_default() -
+                execute_result
+                    .fee_receipt
+                    .clone()
```
sdbondi, line 948:
```suggestion
                    .as_ref()
```

PR #828, opened 2023-12-11
feat: propagation of foreign proposal by non leader

Cifko:
Description
---
If a node locks a new local block, it checks if it needs to resend it to foreign committees. If a node receives a foreign block and it's not the leader, it sends it to the leader. The local leader is responsible for distributing the foreign block within its committee members.

Motivation and Context
---
We need to make sure that at least one honest node sends the block to at least one honest node in the foreign committee.

How Has This Been Tested?
---
There is a new test `foreign_block_distribution` where non of the leaders can send messages to the foreign committee. So other nodes are responsible to get the message across and to the leaders. Also there are new tests for the computation of the indexes for sending (who to whom).

---
Read and run the `foreign_block_distribution`.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/dan_layer/consensus/src/hotstuff/on_ready_to_vote_on_local_block.rs
+++ b/dan_layer/consensus/src/hotstuff/on_ready_to_vote_on_local_block.rs
@@ -937,15 +937,8 @@ where TConsensusSpec: ConsensusSpec
 
     async fn propose_newly_locked_blocks(&self, blocks: Vec<Block<TConsensusSpec::Addr>>) -> Result<(), HotStuffError> {
         for block in blocks {
-            let local_committee = self.epoch_manager.get_local_committee(block.epoch()).await?;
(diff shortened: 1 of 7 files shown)


sdbondi (approved):
utACK

Haven't tested but code looks great, thanks for the explanations in the comments

Cifko:
I'm closing this to go with easier approach.
Approach in this PR minimizes the number of messages between committees.  But creates new issues with distribution in the foreign committee. Because we send it to the leader, but when the leader fails, we have to handle that, or we have to force the leader to send the foreign proposal before it proposes block with it.

I'm going to prepare a new simplistic PR, where `f+1` nodes (so at least one is hones) send the proposal to the whole foreign committee. This way we know that the whole committee will have it and we don't have to do any distribution anymore.

PR #838, opened 2023-12-12
chore: unify cargo workspace

mrnaveira:
Description
---
* All crates in the workspace reuse the package properties of the parent `Cargo.toml`: `version`, `edition`, `authors`, `repository` and `license`. This means that from now on we only need to change them on one place
* All the dependencies (path, git and third-party crates) are defined only once on the parent `Cargo.toml`, so all versions are centralised. There are a couple of exceptions (`clap` and `base64`) which have incompatibilities so some crates have their specific versions.
* Added the crate `tari_comms_logging` to the workspace members.
* Removed the script `update_crate_metadata.sh` as it will not longer be necessary.

Motivation and Context
---
Up until now, every crate in the workspace defined it's own values in their Cargo.toml:
* Package properties: `version`, `edition`, `authors`, `repository` and `license`
* Dependency location and versions for all types (path, git and third-party crates) 

This means we had a lot of duplicity, so doing changes was very tedious and prone to error. This PR aims to solve it by refactoring the Cargo workspace to unify (as much as possible) properties into the parent Cargo definition, reusing them in each crate. This will have multiple benefits:
* Easier to consistently update the version of the project (no need for a custom script)
* Easier to update versions of dependencies across all crates (e.g. to fix vulnerabilities)
* We can see all the dependencies of the whole project in one place, making it easier to spot potential problems.
* Reduce possibilities of runtime or compilation conflicts

How Has This Been Tested?
---
Not applicable

---
Not applicable

Breaking Changes
---
- [x] None
- [ ] Other - Please specify


sdbondi (approved):
LGTM

PR #839, opened 2023-12-13
feat: reduce messages sent when broadcasting to foreign committee

Cifko:
Description
---
This is simple propagation of foreign proposal. We pick `f+1` nodes that will send it to the whole foreign committee. This way at least one node is honest.

Motivation and Context
---

How Has This Been Tested?
---
There is a new test `foreign_block_distribution`.

---
Run the test, it's very hard to manually check this.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (commented):
LGTM 

The title of this PR mentions "reliable broadcast" which implies the implementation of a known reliable broadcast protocol, however, this appears to simply be an optimisation where a proposal is forwarded by f + 1 local nodes instead of all local nodes. If that makes sense, could we update the PR title to reflect that?

sdbondi (approved):

PR #848, opened 2023-12-15
fix: claim burn to create account

Cifko:
Description
---
The claim burn was working only for account that were already created. I refactored the code for create-free-test-coins as well, because it has a lot in common with claiming burns.

Motivation and Context
---
Before this PR there was no way to create an account without having fees and claiming burn needed an account to work.

How Has This Been Tested?
---
Burned and claimed funds for a new account.

---
The claim public key for burning needs to be filled with key from the dan wallet. After that you can call the claim burn with new account name and key index for the respective key.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9293,6 +9293,7 @@ dependencies = [
  "tari_dan_wallet_storage_sqlite",
  "tari_engine_types",
  "tari_indexer_client",
+ "tari_key_manager",
  "tari_shutdown",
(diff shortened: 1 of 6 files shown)


sdbondi (approved):
ACK

Tested and working. I added `--key-id` to the walletd cli so that we could set the correct claim key from there when creating a new account. We can update the UI in another PR.

PR #857, opened 2023-12-21
feat: validate qc

Cifko:
Description
---
Add QC validation.

Motivation and Context
---

How Has This Been Tested?
---
I just ran a single committee 3 VNs to see if it works.

---
You can also try multicommittee test run.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


dan_layer/consensus/src/block_validations.rs
```diff
@@ -42,17 +46,54 @@ pub fn check_proposed_by_leader<TAddr: DerivableFromPublicKey, TLeaderStrategy:
...
+    if !proof.verify_consume(&merkle_root, vns.iter().map(|hash| hash.to_vec()).collect())? {
+        return Err(ProposalValidationError::QCisNotValid { qc: qc.clone() }.into());
+    }
+
+    let mut votes_for_decision = 0;
+    for (sign, leaf) in qc.signatures().iter().zip(vns.iter()) {
+        let challenge = vote_signing_service.create_challenge(leaf, qc.block_id(), &qc.decision());
+        // The verify can fail if the signature is not valid or if the challenge is for opposite decision. We
+        // don't care if it's not valid. We count the valid ones and they need to reach the
+        // quorum.
```
sdbondi: It feels wrong to have a QC that has invalid signatures in it. Perhaps we should only include the signatures of the votes that reached the same decision as the quorum

dan_layer/consensus/src/block_validations.rs
```diff
@@ -42,17 +46,50 @@ pub fn check_proposed_by_leader<TAddr: DerivableFromPublicKey, TLeaderStrategy:
...
+        let challenge = vote_signing_service.create_challenge(leaf, qc.block_id(), &qc.decision());
+        if !sign.verify(challenge) {
+            return Err(ProposalValidationError::QCInvalidSignature { qc: qc.clone() }.into());
+        }
+    }
+    let committee_shard = epoch_manager
+        .get_committee_shard_by_validator_public_key(candidate_block.epoch(), candidate_block.proposed_by())
+        .await?;
 
+    if committee_shard.quorum_threshold() > u32::try_from(qc.signatures().len()).unwrap() {
```
sdbondi: Avoids unwrap
```suggestion
    if usize::from(committee_shard.quorum_threshold()) > qc.signatures().len() {
```

Cifko: usize::from<u32> is not defined, I had to add a map_err

sdbondi (approved):

PR #858, opened 2023-12-29
feat: sign proposal

Cifko:
Description
---
Sing the proposal (the proposer signs the block).

Motivation and Context
---

How Has This Been Tested?
---
I just did sample run on dan-testing.

---
You can run the dan-testing and maybe do a debug logs on the signature check to make sure that the checks are enforced.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (approved):
utACK

Nice

dan_layer/consensus/src/block_validations.rs
```diff
@@ -42,6 +42,30 @@ pub fn check_proposed_by_leader<TAddr: DerivableFromPublicKey, TLeaderStrategy:
     Ok(())
 }
 
+pub fn check_signature(candidate_block: &Block) -> Result<(), ProposalValidationError> {
+    if candidate_block.is_dummy() {
```
sdbondi, line 46: I think we should change the sync to never send and receive dummy blocks, We could calculate dummy blocks instead.

PR #861, opened 2024-01-02
fix(engine): remove unnecessary allocations from GetNonFungibleIds

sdbondi:
Description
---
Remove unnecessary allocation when calling `VaultAction::GetNonFungibleIds` and `BucketAction::GetNonFungibleIds`

Motivation and Context
---

Previously, when we were using borsh encoding, `BTreeSet`s would not deserialize in the WASM (unfortunately, no specific details in my code comment). Since the change to use CBOR this is no longer the case, which allows us to encode a reference to the BTreeSet without the into_iter/collect conversion to a Vec. Since Sets and Vecs are represented in the same CBOR binary format, we can deserialize from a BTreeSet to a Vec. 

It is important to remove any unnecessary contract execution resource usage. 

How Has This Been Tested?
---
Modified existing test to call the bucket get_non_fungible_ids method. Vault get_non_fungible is already tested.

---
Engine tests

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


mrnaveira (approved):
LGTM

PR #863, opened 2024-01-02
fix(template-lib): minor cross-template call api improvement

sdbondi:
Description
---
Add "shortcut" for cross-template calls that return the unit type i.e `call::<_, ()>(...)`

Motivation and Context
---
It is a little unwieldy to use `manager.call::<_, ()>(...)`. This PR adds `manager.invoke(...)`

How Has This Been Tested?
---
`call` has been tested in "composability" tests

---
Use `manager.invoke`

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


mrnaveira (approved):
LGTM

PR #864, opened 2024-01-03
fix!: use configured network string in networking protocol

sdbondi:
Description
---
Use string from `Network::as_str` in the protocol identifier

Motivation and Context
---
This change uses the configured network in the `/tari/{network}/{version}` protocol string. This string is used to identify different networks. If peers are on a different network (by libp2p identify) the peers disconnect.

How Has This Been Tested?
---
Partially manually by checking that the protocol string is set per the network. Cucumbers also partially test this case.

---
Try connect two VNs, one with `--network localnet` and the other with `--network igor`

Breaking Changes
---

- [ ] None
- [x] Other - previous nodes will not stay connected to new nodes and vice versa


mrnaveira (approved):
LGTM

PR #865, opened 2024-01-03
fix: gossip

Cifko:
Description
---
Fix the gossiping duplicate problem.

Motivation and Context
---
The gossiping discard duplicates, it doesn't take into account that the same message is being send to different topic. So when a message was send to local committee and then to foreign committee (this was counted as a duplicated), the foreign committee never received it. The topic is now added to the message id.

How Has This Been Tested?
---
Running dan-testing with multiple committees and now I see `tx_<id>` in the `message_log.sqlite`.

---
Do the same as above.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (approved):
Ah interesting I didn't consider that deduplication is not scoped to the topic already. Nice find!

PR #866, opened 2024-01-03
test: break infinity loop in template

Cifko:
Description
---
Test that the infinity loop will eventually break because of the increasing costs.

Motivation and Context
---

How Has This Been Tested?
---
There is a new test `test_break_infinity_loop`

---
The cost for loop was increased to 1 (from 0), so we should confirm that nothing else is breaking.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/dan_layer/engine/tests/test.rs
+++ b/dan_layer/engine/tests/test.rs
@@ -293,6 +293,16 @@ fn test_random() {
     assert_ne!(value, vec![0; 300]);
 }
 
+#[test]
+#[should_panic(
+    expected = "called `Result::unwrap()` on an `Err` value: Transaction failed: Execution failure: RuntimeError: \
+                unreachable\n    at tari_free (<module>[327]:0x2390c)"
+)]
+fn test_break_infinity_loop() {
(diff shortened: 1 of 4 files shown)


dan_layer/engine/tests/test.rs
```diff
@@ -293,6 +293,16 @@ fn test_random() {
     assert_ne!(value, vec![0; 300]);
 }
 
+#[test]
+#[should_panic(
+    expected = "called `Result::unwrap()` on an `Err` value: Transaction failed: Execution failure: RuntimeError: \
+                unreachable\n    at tari_free (<module>[327]:0x2390c)"
```
Cifko, line 299: This message is very specific, if we change the template it will differ. I tested this on ubuntu and windows to make sure it passes on both.

sdbondi (approved):

dan_layer/engine/tests/test.rs
sdbondi: You could also use `test.execute_expect_failure` and inspect the error, though you would have to build the transaction.

PR #868, opened 2024-01-03
feat(template): allow resource actions to be restricted by component

sdbondi:
Description
---
Allows resource access rules  to restrict actions to one or more specific components/templates
Adds the ability to get pre-allocate component addresses in code before component creation

Motivation and Context
---
Thanks to @mrnaveira for this idea!

Some use cases may want a user to mint their own badge by executing a method on a specific template without explicit permission from the component owner (permissionless). This PR allows a template author to be permissable in who can perform resource actions, but force all users to interact with their component/template to control how those resource actions are used. 

To set component-restricted rules in code, the component address is required. Often access rules are set in the constructor of the component, where the address is not known. This PR adds the ability ask the runtime for a "pre-allocated" component address and later apply that to the component that is to be created.

How Has This Been Tested?
---
New unit tests

---

Template using the new access rules, e.g
```rust
  let allocation = CallerContext::allocate_component_address();
            let tokens = ResourceBuilder::fungible()
                .initial_supply(1000)
                .mintable(AccessRule::Restricted(RestrictedAccessRule::Require(
                    RequireRule::Require(allocation.address().clone().into()),
                )))
                .build_bucket();

  Component::new(Self {
                tokens: Vault::from_bucket(tokens),
            })
            .with_address_allocation(allocation)
            .create()
```

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/dan_layer/engine/src/runtime/error.rs
+++ b/dan_layer/engine/src/runtime/error.rs
@@ -220,6 +220,11 @@ pub enum RuntimeError {
     DuplicateBucket { bucket_id: BucketId },
     #[error("Duplicate proof {proof_id}")]
     DuplicateProof { proof_id: ProofId },
+
+    #[error("Address allocation not found with id {id}")]
+    AddressAllocationNotFound { id: u32 },
+    #[error("Address allocation type mismatch: {address}")]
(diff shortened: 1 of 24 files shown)


mrnaveira (approved):
LGTM

PR #879, opened 2024-01-05
feat: show keys in claim burn

Cifko:
Description
---
Show keys in the claim burn in wallet ui.
If the key is associated with an account you will see the account name.
The account name is editable only when you select key without an associated account.
Fix adding keys in the wallet ui.

Motivation and Context
---

How Has This Been Tested?
---
Burning funds into new account.

---
Same as above. Create a burn for a new key and then claim it for a new account for this key.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (approved):
utACK

PR #881, opened 2024-01-05
feat(template_lib): builtin template addresses

mrnaveira:
Description
---
* `template_lib` exports an enum `BuiltinTemplate` that enumerates all the builtin templates and allows to retrieve the addresses via a new `BuiltinTemplateInvoke` engine operation
* The engine implements the `EngineOp::BuiltinTemplateInvoke` by accessing the appropriate constants in the `template_builtin` crate

Motivation and Context
---
Template developers will want to perform cross-template calls with builtin templates (e.g. accounts). For convenience we want to expose the corresponding template addresses in the `template_lib` crate.

There are two main ways of doing this:

1. Export the template addresses directly as constants
2. Create a new engine call for getting the template address

This PR follows (2) for template WASM size optimisation reasons, as (1) would require to include the full address (32 bytes for each one of the builtin templates, and we may have more of them in the future) in every template WASM.

How Has This Been Tested?
---
New engine unit test for builtin template addresses

---
Use the new `BuiltinTemplate` enum in a template

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


dan_layer/engine/src/runtime/impl.rs
```diff
@@ -1694,6 +1697,17 @@ impl<TTemplateProvider: TemplateProvider<Template = LoadedTemplate>> RuntimeInte
         self.tracker.pop_call_frame()?;
         Ok(())
     }
+
+    fn builtin_template_invoke(&self, action: BuiltinTemplateAction) -> Result<InvokeResult, RuntimeError> {
+        let address = match action {
```
sdbondi, line 1704: Need to add the module call so that fees are tracked

```rust
        self.invoke_modules_on_runtime_call("builtin_template_invoke")?;
```

sdbondi (commented):
Looks good, could we also provide a method to fetch the currently executing template address in templates?

sdbondi (approved):

PR #885, opened 2024-01-10
feat: don't add substates that are already being processed

Cifko:
Description
---
Don't add transaction to a proposal if any of their substates is being processed.

Motivation and Context
---
I left one TODO in the code. Because if we don't switch to the "no version" transaction, them we can add aborts directly to the proposal.

How Has This Been Tested?
---
I tried 9 concurrent transaction 9 accounts to 1, and 1 accounts to 9.

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/dan_layer/state_store_sqlite/src/reader.rs
+++ b/dan_layer/state_store_sqlite/src/reader.rs
@@ -32,6 +32,8 @@ use tari_dan_storage::{
         Block,
         BlockId,
         Command,
+        Decision,
+        Evidence,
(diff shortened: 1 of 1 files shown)


sdbondi (approved):
utACK

Ah I didn't initially think this was happening on the database layer,  we should definitely write some docs on the expected behaviour of each method in the storage interface (#890). We may encounter other issues but this should fix the specific issue with incorrect proposals.

PR #886, opened 2024-01-11
fix(consensus): dont send dummy blocks on sync

sdbondi:
Description
---
Excludes dummy blocks from catch-up sync
Excludes dummy blocks from RPC sync
Rename `tari_comms_rpc_state_sync` to `tari_rpc_state_sync`
Removed commented-out and dead code

Motivation and Context
---
The is_dummy flag is not sent on catch-up, which leads to a "missing proposal signature" error, since summy blocks cannot be signed. Dummy blocks are calculated as needed by the proposal handler so do not need to be sent.

When testing manually this works, however, we sometimes run into #887 

How Has This Been Tested?
---
Manually, synced chain with dummy blocks and checked that dummy blocks created

---
4 VN committee, shut down one node and delete its data, wait for leader failures (to create some dummy blocks), restart node, and check that it syncs

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


Cifko (approved):
utACK

PR #889, opened 2024-01-12
feat: add foreign proposal validation

### LINKED ISSUE #887: bug(consensus): inbound message handler future can be aborted resulting in dropped messages and blocks
Discovered by @Cifko 

In the main select loop in the consensus worker

The `inbound_message_worker.next(..)` future performs missing transaction processing on proposals. 
If any other branch in this select loop is ready, this future is aborted. This manifests most commonly 
as blocks arriving in the logs but then no further logs appear.

In most cases, [this epoch manager call](https://github.com/tari-project/tari-dan/blob/b7ccff2084577985f5ff2c0ed53714ee9a77635a/dan_layer/consensus/src/hotstuff/on_inbound_message.rs#L164) is aborted. "Requester abandoned request" error in logs

```rust
   tokio::select! {

                msg_or_sync = self.inbound_message_worker.next(current_height) => {
                    if let Err(e) = self.on_new_hs_message(msg_or_sync).await {
                        self.on_failure("on_new_hs_message", &e).await;
                        return Err(e);
                    }
                },

                Ok(event) = epoch_manager_events.recv() => {
                    self.handle_epoch_manager_event(event).await?;
                },
// ...
}
```

Cifko:
Description
---
Do most of the validation for foreign proposals as their are done on the local proposals.
I had to move the tokio::select from on_inbound_message to the parent tokio select, so it will stop being cancelled. I tried to fuse it instead but then I got into mutability concurrency issue.

Motivation and Context
---
Fixes #887 

How Has This Been Tested?
---
Running dan-testing and looking at the logs. The `requester abandoned request` is gone.

---
Same as above.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/dan_layer/consensus/src/hotstuff/on_inbound_message.rs
+++ b/dan_layer/consensus/src/hotstuff/on_inbound_message.rs
@@ -35,11 +35,11 @@ pub struct OnInboundMessage<TConsensusSpec: ConsensusSpec> {
     leader_strategy: TConsensusSpec::LeaderStrategy,
     pacemaker: PaceMakerHandle,
     vote_signing_service: TConsensusSpec::SignatureService,
-    rx_hotstuff_message: mpsc::Receiver<(TConsensusSpec::Addr, HotstuffMessage)>,
+    pub rx_hotstuff_message: mpsc::Receiver<(TConsensusSpec::Addr, HotstuffMessage)>,
     tx_outbound_message: mpsc::Sender<(TConsensusSpec::Addr, HotstuffMessage)>,
     tx_msg_ready: mpsc::UnboundedSender<(TConsensusSpec::Addr, HotstuffMessage)>,
-    rx_new_transactions: mpsc::Receiver<TransactionId>,
-    message_buffer: MessageBuffer<TConsensusSpec::Addr>,
+    pub rx_new_transactions: mpsc::Receiver<TransactionId>,
+    pub message_buffer: MessageBuffer<TConsensusSpec::Addr>,
     shutdown: ShutdownSignal,
 }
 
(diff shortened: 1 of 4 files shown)


dan_layer/consensus/src/hotstuff/on_inbound_message.rs
```diff
@@ -35,11 +35,11 @@ pub struct OnInboundMessage<TConsensusSpec: ConsensusSpec> {
     leader_strategy: TConsensusSpec::LeaderStrategy,
     pacemaker: PaceMakerHandle,
     vote_signing_service: TConsensusSpec::SignatureService,
-    rx_hotstuff_message: mpsc::Receiver<(TConsensusSpec::Addr, HotstuffMessage)>,
+    pub rx_hotstuff_message: mpsc::Receiver<(TConsensusSpec::Addr, HotstuffMessage)>,
     tx_outbound_message: mpsc::Sender<(TConsensusSpec::Addr, HotstuffMessage)>,
     tx_msg_ready: mpsc::UnboundedSender<(TConsensusSpec::Addr, HotstuffMessage)>,
-    rx_new_transactions: mpsc::Receiver<TransactionId>,
-    message_buffer: MessageBuffer<TConsensusSpec::Addr>,
+    pub rx_new_transactions: mpsc::Receiver<TransactionId>,
```
sdbondi, line 41: nit: structs with a mix of pub and private fields is code smell (couldn't find a good write up but that is my two cents ;) - also https://users.rust-lang.org/t/structs-with-public-fields-are-brittle/989)
I see why you had to do that though due to multiple mutable references, but we should clean this up in future.

sdbondi (approved):
Thanks for fixing that issue - might have to refactor later but getting this issue fixed is higher priority

dan_layer/consensus/src/hotstuff/on_inbound_message.rs
Cifko: yeah, I tried to do it another way, but as you said, multiple mutable references was the problem

PR #892, opened 2024-01-12
refactor(consensus): clean up inbound message processing

sdbondi:
Description
---
Move HS message receivers into worker, pass messages to the OnInboundMessage handler. This checks proposals for missing transactions and parks them if necessary. New transaction notifications are sent to the OnInboundMessage handler to check if any proposals can be unparked. Once unparked, they are added to the "next message buffer". The 
message buffer is polled (cancel safe) in the consensus worker tokio select loop.

Motivation and Context
---
Clean up after fix in #889  

How Has This Been Tested?
---
Running stress test from PR #880 

---
Consensus should still work as before

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


mrnaveira (approved):
LGTM

PR #893, opened 2024-01-14
feat(indexer): quorum certificate validation

mrnaveira:
Description
---
* VNs now include QCs in the `GetSubstateResponse` type.
* Unified the QC validation function inside the `consensus` crate, and reuse it for block validations inside the VN as well as in the indexer.
* Created a new error type for QC validation errors
* Moved the `TariSignatureService` to `tari_dan_app_utilities`, so it can be reused from VNs and indexers.

Motivation and Context
---
Right now indexers and wallets trust the substate values received from VNs. Ideally we would want this process to be totally trustless by validating a QC sent by the VN alongside the substate value.

But right now this is not possible as the QC does not include the appropriate information for validating the substate value. Anyway, the goal of this PR to implement the boilerplate needed as the first step towards trustless VNs:
* Add the QC to the VN responses along the substate values
* Perform a initial QC validation: signatures, merkle root and minimum quorum. As mentioned earlier, in the future (when it's possible) we would like to check that a substate value is validated by the QC.

How Has This Been Tested?
---
Related integration tests pass

---
Not applicable

Breaking Changes
---
- [ ] None
- [x] Indexers now expect a valid QC from VNs, so older VNs are incompatible


sdbondi (dismissed):
utACK

sdbondi (approved):

sdbondi:
PR cleanup

PR #894, opened 2024-01-15
fix(consensus)!: allow multiple read-only shard references in proposals

sdbondi:
Description
---
Includes locks in evidence
Checks locks when proposing to allow multiple input refs

Motivation and Context
---
Evidence for each shard within a TransactionAtom now includes the lock type. 
Updated the proposal db query to check each lock type and ensure that conflicts either don't occur or if they do they are all read locks.

How Has This Been Tested?
---
Ran a stress test from #880, previously after #885 was merged, funding the tariswap components would take a very long time (I've never actually run it to completion, but ran for 30 mins without completing). With this PR funding took roughly a minute on my test. Swap batches are also reaching finalization within an acceptable timeframe.

---
Submit multiple transactions which use a single substate as an input ref and check that they can be added to the same block.

Breaking Changes
---

- [ ] None
- [x] Requires data directory to be deleted (Evidence struct changed)
- [ ] Other - Please specify

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4012,6 +4012,7 @@ checksum = "d530e1a18b1cb4c484e6e34556a0d948706958449fca0cab753d649f2bce3d1f"
 dependencies = [
  "equivalent",
  "hashbrown 0.14.3",
+ "serde",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


mrnaveira (approved):
LGTM

PR #896, opened 2024-01-15
fix: qc was validated based on current epoch, not the qc's epoch

Cifko:
Description
---
The qc validation was based on the epoch of the candidate block. So if the QC was from previous epoch the committee shard could be different. E.g. when you add a new VN into the committee, the qc was expected to have 2 signatures. But in the previous epoch there was only one VN so the QC never passed the validation.

Motivation and Context
---

How Has This Been Tested?
---
Start dan-testing with 1 VN, add a new VN and mine it into the committee. Without this change you would never see a new block coming in. You still have to wait for some leader failure on the epoch change but eventually new blocks starts coming in.

---
As above.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/dan_layer/consensus/src/block_validations.rs
+++ b/dan_layer/consensus/src/block_validations.rs
@@ -109,7 +109,13 @@ pub async fn check_quorum_certificate<TConsensusSpec: ConsensusSpec>(
         }
     }
     let committee_shard = epoch_manager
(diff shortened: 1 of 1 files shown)


sdbondi (approved):
LGTM

PR #897, opened 2024-01-15
feat(template_lib): improve API for extracting data from NFTs

mrnaveira:
Description
---
* Added new `get_non_fungibles` method to vaults and buckets in `template_lib`, that returns all the `NonFungible` objects contained within.
* Added new engine operations to support the new method in both vaults and buckets

Motivation and Context
---
We expect templates to often read NFT data from buckets or vaults. Currently the only way to do it is doing something like:
```
 let seller_badge_id = &seller_badge_bucket.get_non_fungible_ids()[0];
 let seller_badge = ResourceManager::get(self.seller_badge_resource).get_non_fungible(&seller_badge_id);
 let nft_metadata = seller_badge.get_data::<Metadata>();
```

This PR adds a new method `get_non_fungibles` to vaults and buckets to we can write the previous example as:
```
 let seller_badge = seller_badge_bucket.get_non_fungibles()[0];
 let nft_metadata = seller_badge.get_mutable_data::<Metadata>();
```

How Has This Been Tested?
---
New unit test in the `engine` crate for the new methods

---
Call the new methods inside a template

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/dan_layer/engine/src/runtime/impl.rs
+++ b/dan_layer/engine/src/runtime/impl.rs
@@ -1086,6 +1086,28 @@ impl<TTemplateProvider: TemplateProvider<Template = LoadedTemplate>> RuntimeInte
                 })
             },
             VaultAction::CreateProofByConfidentialResource => todo!("CreateProofByConfidentialResource"),
+            VaultAction::GetNonFungibles => {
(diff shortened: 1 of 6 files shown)


sdbondi (approved):
utACK

PR #899, opened 2024-01-16
feat(engine): check WASM version compatibility

mrnaveira:
Description
---
* Added `semver` dependency for version parsing
* Added new `version` module to encapsulate version compatibility logic
* The engine's WASM execution uses the new version compatibility logic and returns a new types of errors associated with versions.

Motivation and Context
---
Previously, on https://github.com/tari-project/tari-dan/pull/835 we implemented a version system in template WASMs. On WASM execution, the engine only logged if the WASM version is equal to the engine one, no error was returned.

We want a more sophisticated approach, following [Cargo's SemVer convention](https://doc.rust-lang.org/cargo/reference/resolver.html) where versions are considered compatible if their left-most non-zero major/minor/patch component is the same. We also want to fail the execution if the WASM version is incompatible with the engine one. 

How Has This Been Tested?
---
New unit tests for the SemVer compatibility check function

---
Inspect the logs on transaction executions for success/error in version checks

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (approved):
Nice and neat!

PR #905, opened 2024-01-18
fix: uncomment sync cucumber test

Cifko:
Description
---
It looks like the test is working. I had rerun it couple of times. 

Motivation and Context
---

How Has This Been Tested?
---
`cargo test --test cucumber -- -n "New validator node registers and syncs"`

---
`cargo test --test cucumber -- -n "New validator node registers and syncs"`

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/integration_tests/tests/features/block_sync.feature
+++ b/integration_tests/tests/features/block_sync.feature
@@ -3,57 +3,56 @@
 
 Feature: Block Sync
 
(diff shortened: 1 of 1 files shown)


sdbondi (approved):
Thanks - fixed in #886

PR #911, opened 2024-01-22
fix(consensus): minor foreign counters fixes

sdbondi:
Description
---

Rollback counter correctly for foreign proposals that error
Validate and save proposed block in single db transaction
Replaces hashmap in proposal with IndexMap
Load counters internally, remove from initialization
Validate correct sorting of counters in block
Validate counters without allocations
Dont add empty foreign counters to the database 

Motivation and Context
---
Hashmaps have random ordering (by default). After looking at it,  the previous implementation was correct because the PartialEq implementation of HashMap is independent of ordering. However, IMO there is a risk of introducing non-determinism bugs in the future as any checks that rely on iterating through the hashmap may break for size > 1. 

IndexMap seems like a good choice as it maintains order but does not have the memory overhead of BTreeMap.

This PR also removes the need to store empty counters for each local-only block.

How Has This Been Tested?
---

Partially tested by  some consensus tests (e.g. `leader_failure_output_conflict`). Manually, multi shard test using dan-testing

---
Should work as before. 

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


Cifko (approved):
utACK

sdbondi:
Thanks @Cifko  Did more multishard testing on this in the meanitme, working as before so going to merge

PR #914, opened 2024-01-24
feat(consensus): adds metrics to consensus

sdbondi:
Description
---
feat(consensus): adds metrics to consensus
Adds local setup for prometheus and grafana using docker compose 
Serves prometheus metrics at `{jrpc_url}/_metrics`
simplify update_node usage and collect finalized transactions
Adds a Grafana dashboard export that can be imported when testing
fix(consensus): get committee shard from block epoch and proposer, rather than local since this fails because the local node is not registered for the epoch. This caused cucumber `New validator node registers and syncs` to fail, this test passes now.

Motivation and Context
---
Collects consensus metrics to allow us to evaluate performance and detect errors.

Grafana and Prometheus can be run locally:

```shell
cd prometheus
docker-compose up -d
open http://localhost:3000
```

image

How Has This Been Tested?
---
Manually

---
Run a VN and prometheus+grafana dockers 

Update the scrape config as required to point to the JRPC if your validator

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


stringhandler (dismissed):

stringhandler (approved):
Nice

PR #915, opened 2024-01-24
feat: add ts-rs mapping to our jrpc

Cifko:
Description
---
The `ts-rs` crate is very helpful but it has some downsides (bugs). 
If you want to generate new mappings. You have to run test with feature ts without integration test : `cargo test --workspace --exclude integration_tests export_bindings --features ts`. This will run only the ts-rs exporting. But there is a catch, some of the structs are export to wrong directory (`dan_layer\bindigs`). You have to manually move them to the `/bindings` folder. Don't forget to run prettier afterwards.
Probably not all settings are exported to typescript correctly. We will fix once we start replacing the `any`s in the webui projects.

Motivation and Context
---
 It seems like you can change the `export_to` path to get it into the right directory. But then the `ts-rs` will panic (because the are expecting that each struct is on the same level.... Maybe in the future version it will work. 

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi:
Nice - perhaps we can add a node package for these bindings and add a command to build them?

Then we can include that package in the web UIs

stringhandler (dismissed):
utACK

sdbondi (approved):
utACK

PR #919, opened 2024-01-29
feat: transparent bindings and bigints

Cifko:
Description
---
Handle the `#[serde(transparent)]` structs, and also fix the bigint. i64 and u64 were exported as bigint, instead of number. The number can hold anything in range(inclusive) from -(2^53-1)...(2^53-1). Anything above or below can be hold but it's not safe to be precise.

Motivation and Context
---
Some of the structures are `#[serde(transparent)]` this is not recognized by the ts-rs crate. If the structure is "real struct" (e.g. struct A{b:u64} and not a struct A(u64), then we can call #[ts(flatten)] but you have to call it on every instance. And it will rewrite the variable e.g.
```
struct A{
  #[ts(flatten)]
  b:B
}

struct B {
  c:u8
}
```
Will be exported as
```
export interface A {
  c : number; // instead of b : number
}
```
So I have decided instead of flatten define the types by hand 

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


dan_layer/engine_types/src/fees.rs
```diff
@@ -14,10 +14,13 @@ use crate::resource_container::ResourceContainer;
 #[cfg_attr(feature = "ts", derive(TS), ts(export, export_to = "../../bindings/src/types/"))]
```
sdbondi, line 14: To fix where the bindings are output you need to add another `../`
```suggestion
#[cfg_attr(feature = "ts", derive(TS), ts(export, export_to = "../../../bindings/src/types/"))]
```

sdbondi (commented):
Loosely tested

Perhaps just change the paths to be correct inside dan_layer so that they go into  `bindings` in the git root. It might be better to package JRPC types for the VN, indexer and wallet in separate NPM packages but we can clean that stuff up at a later point

sdbondi (dismissed):
Approving because we can clean the paths up later, since build.sh moves them

btw probably want to add `npx` infront of `shx mv...`  I get a command not found error in build.sh but if the paths are fixed we can remove the mv commands anyway

dan_layer/engine_types/src/fees.rs
Cifko: if I do this the includes in the generated files will be wrong. That's why I move them, instead of correcting the paths in the generated files.

Cifko:
> Approving because we can clean the paths up later, since build.sh moves them
> 
> btw probably want to add `npx` infront of `shx mv...` I get a command not found error in build.sh but if the paths are fixed we can remove the mv commands anyway

done

sdbondi (approved):

PR #921, opened 2024-01-29
feat: adapt indexer to ts-rs exported interfaces

Cifko:
Description
---
The #919, #920 need to go in first. Then I will rebase this.
This PR removes all the any that could be replaced by our auto-generated interfaces/types.
Now the jsonrpc calls is expecting the proper request and gives proper response. So when the structure changes we will know.

Motivation and Context
---

How Has This Been Tested?
---
Running dan-testing with a wallet with some accounts and transactions.

---
Same as above, maybe there is something I haven't tested (looked at). But at the first glance it's working as expected.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (approved):
utACK

PR #922, opened 2024-01-29
feat!: optional versions in transaction inputs

mrnaveira:
Description
---
* Refactored transactions so inputs have optional versions. For that I reused the `SubstateRequirement` struct instead of the `SubstateAddress` we were using. This change cascaded through all the project.
* Mempool skips execution of transactions that have some inputs without versions. It stores a mock result in database so the transaction is picked up when creating the next HotStuff block. We may want to change this behaviour later (e.g. real execution in mempool, or separated database table for those types of transactions).
* Created a `BlockTransactionExecutor` used by the consensus when voting on local blocks (I leave remote consensus out of the scope for now). The goal of the executor is to allow the re-execution of transactions taking into account the outputs of previous transactions in the block. The executor will be initialised at the beginning of the block, and will keep a temporary substate memory during the block execution, so each consecutive transaction will read the most recent versions of the inputs from there. After a transaction finishes, the outputs will be used to update the memory.
* After a transaction is executed in consensus, the results are stored in the SQLite database.

Motivation and Context
---
We want to allow _concurrent_ transactions that affect the same substates. Think for example of a Uniswap-like pool, in which each transaction alters the pool ratio (price impact), in this case transactions must be executed in strict sequential order without conflicts.

The implementation of concurrent transactions is going to spawn across multiple PRs. The goal of this PR is to set the foundation needed for concurrency:
* Allow for clients to send transactions whose inputs don't have specific versions. This is necessary because clients do not know which is the last version of the input they need in case of a high demand component.
* Transaction execution at the HotStuff level instead of at mempool level for transactions without versions. The latest input version is selected for each transaction.

Note that transaction execution and consensus behaviour has not been changed for transactions with specific input versions (the default way of interacting with the network). To trigger the new behaviour the user has to manually craft and send to a VN a transaction that has one or more inputs without version.

<a name="todo"></a> 
Current Limitations and Future Work
---
* The scope of this PR is concurrent single-shard transactions. Concurrent multi-shard transactions, that would involve more that one committee, are left for future work.
* We may need a separated DB API for substates, with a memory implementation so each block can be processed in parallel without DB conflicts.
* The `TariDanBlockTransactionExecutor`, which implements the consensus execution of transactions, has some `TODO` comments for improvements to be made.
* In this PR concurrent transactions do not work (the second onwards fails) due to the behaviour of the pacemaker that we will need to address in future PRs:
    * Blocks seem to be generated very fast, at least in the cucumber scenarios (maybe due to having a single VN network), so it's not possible to force concurrent transactions into the same block. We may need a new network parameter to set up a minimum block time and use it in cucumber tests.
    * Also, there seems to be overlapping of blocks processing. Transactions of block N+x are executed in consensus before transactions in block N are stored into the DB. We will need strict sequential block execution so outputs of one transaction can be used as inputs in another transaction in a future block. 

How Has This Been Tested?
---
New cucumber scenario for concurrent transaction execution. Currently only one transaction without input versions is executed due to the behaviour of the consensus pacemaker (check [Current Limitations and Future Work](#todo))

---
Run the new cucumber scenario or try to manually send a transaction with no input versions to the VN. I recommend the `counter` template for testing as it's easy to craft transactions that affect only one substate.

Breaking Changes
---

- [ ] None
- [x] Network reset. Now transactions inputs contain the substate ID + version, where up until now they were just a substate address (a hash). **This change breaks both consensus and clients** (UIs, JavaScript, etc.).


dan_layer/consensus/src/hotstuff/block_transaction_executor.rs
```diff
@@ -0,0 +1,155 @@
...
+    // TODO: remove this variant when we have a remote substate implementation
+    #[error("Remote substates are now allowed")]
+    RemoteSubstatesNotAllowed,
+}
+
+// TODO: we should keep a "proxy" hashmap (or memory storage) of updated states from previous transactions in the same
+// block,       so each consecutive transaction gets the most updated version for their inputs.
+//       If no previous tx wrote on a input, just get it from the regular state store
+#[derive(Debug, Clone)]
+pub struct BlockTransactionExecutor<TConsensusSpec: ConsensusSpec> {
```
sdbondi: I think this should live in the validator node bin crate and be abstracted behind the TransactionExecutor trait. 
We should maintain a separation between the tari_engine and consensus.

The validator node bin can pass in the implementation.

dan_layer/consensus/Cargo.toml
```diff
@@ -13,6 +13,7 @@ tari_dan_storage = { workspace = true }
 tari_engine_types = { workspace = true }
 tari_transaction = { workspace = true }
 tari_epoch_manager = { workspace = true }
+# Used for PublicKey and Signature
```
sdbondi: Is this comment intentional?

dan_layer/consensus/src/hotstuff/on_ready_to_vote_on_local_block.rs
```diff
@@ -795,16 +864,31 @@ where TConsensusSpec: ConsensusSpec
...
+            local_committee_shard.filter(write_inputs.iter()),
             SubstateLockFlag::Write,
         )?;
+        // We ignore inputs without version
+        let read_inputs: Vec<SubstateAddress> = transaction
+            .input_refs()
+            .iter()
+            .filter(|i| i.version().is_some())
+            .map(|i| i.to_substate_address())
+            .collect();
```
sdbondi, line 887: nit: we can save several heap allocations by just passing in the iterator directly to the filter function instead of first allocating a new vec and then passing it in. But not super critical can clean up later.

applications/tari_validator_node/src/consensus/spec.rs
```diff
@@ -26,6 +31,10 @@ pub struct TariConsensusSpec;
 
 impl ConsensusSpec for TariConsensusSpec {
     type Addr = PeerAddress;
+    type BlockTransactionExecutorBuilder = TariDanBlockTransactionExecutorBuilder<
```
sdbondi, line 34: I suspect we won't need a build interface, seems to be just avoiding the borrow checker (immutable self, mutable executor). We can probably just pass in an executor, but not important at this stage.

sdbondi (dismissed):
ACK

Submitted some transactions and tested and seems to be working.
Nice, looks like a good first step - happy to merge

I suspect the builder interface on the spec can be removed in favour of passing in the executor directly, it seems to be a way to keep the borrow checker happy (immutable self, mutable execute) but this is very minor so no change required at this point.

We'll need to update typescript bindings (this is currently a manual process) by running `npm run build` in `bindings/`

dan_layer/consensus/Cargo.toml
mrnaveira: Nice catch, it's a byproduct of conflict resolution, I will remove it

applications/tari_validator_node/src/consensus/spec.rs
mrnaveira: The problem that the builder interface is trying to solve is that HotStuff needs to build a brand new executor for each incoming block, because each executor has an internal state so we cannot just pass a executor when initialising the HotStuff (well, maybe we could but we would need to call a `clear` function after each block). But I agree it does seem a bit convoluted, maybe we can simplify it in the future.

mrnaveira:
> We'll need to update typescript bindings (this is currently a manual process) by running `npm run build` in `bindings/`

Done

applications/tari_validator_node/src/consensus/spec.rs
sdbondi: Ah I see the output versions cache `output_versions`. Makes sense

sdbondi (approved):

PR #924, opened 2024-02-01
feat(consensus)!: implements state merkle root

sdbondi:
Description
---
Implements Merkle root committing to substate state within a shard

Motivation and Context
---
This PR adds a correct Merkle root within block proposals and implements validation against the state to be committed in the proposed block. The state Merkle has is also validated when syncing. The tree is an implementation of [Jellyfish Merkle tree](https://diem-developers-components.netlify.app/papers/jellyfish-merkle-tree/2021-01-14.pdf) adapted from Scrypto.

The tree is not currently pruned although each node is marked to be removed. A state tree diff is produced for each uncommitted block and applied and removed once the block is committed (3-chain).

How Has This Been Tested?
---
New unit tests, existing cucumbers, existing consensus tests, manually (swap bench, stress test, sync fresh node, sync after offline)

When running the swap bench, clearing and re-syncing a node, there was an MR mismatch for one very large transaction (creating hundreds of accounts in 1 tx). I have not been able to track down the cause, but I believe that this is a pre-existing determinism issue with sync/substate store. Other than that, this appears to be working as expected.

---
Nodes should function as before

Breaking Changes
---

- [ ] None
- [ ] Other - Please specify


stringhandler (dismissed):
I think we should eventually replace the substate table with just the tree and then move it to rocksdb or lmdb. I am still unsure whether this approach is more efficient than just storing the keys in rocksdb directly, but I will investigate and see if we can benchmark the two

stringhandler (approved):

PR #925, opened 2024-02-01
fix: use hash domains from tari_hash_domains

Cifko:
Description
---
Use hash domains from the `tari` repo.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/dan_layer/engine_types/src/hashing.rs
+++ b/dan_layer/engine_types/src/hashing.rs
@@ -29,10 +29,9 @@ use blake2::{
 use digest::Digest;
 use serde::Serialize;
 use tari_bor::encode_into;
-use tari_crypto::{hash_domain, hashing::DomainSeparation};
+use tari_crypto::hashing::DomainSeparation;
+use tari_hash_domains::TariEngineHashDomain;
 use tari_template_lib::Hash;
-
-hash_domain!(TariEngineHashDomain, "com.tari.dan.engine", 0);
 
 pub fn hasher64(label: EngineHashDomainLabel) -> TariHasher64 {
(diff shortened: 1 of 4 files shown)


dan_layer/engine_types/src/hashing.rs
```diff
@@ -29,11 +29,10 @@ use blake2::{
 use digest::Digest;
 use serde::Serialize;
 use tari_bor::encode_into;
-use tari_crypto::{hash_domain, hashing::DomainSeparation};
+use tari_crypto::hashing::DomainSeparation;
+use tari_hash_domains::TariEngineHashDomain;
 use tari_template_lib::Hash;
 
-hash_domain!(TariEngineHashDomain, "com.tari.dan.engine", 0);
```
sdbondi, line 35: Is there a reason that this DAN-level hash domain goes in the common hash domain crate?

IMO only minotari hash domains that are needed to interact with the base layer should go there. We can manage L2 hash domains in this repo

sdbondi (commented):
Thanks, missed the `TransactionSecureNonceKdfDomain` one

Not sure about the TariEngine hash domain

dan_layer/engine_types/src/hashing.rs
Cifko: It's because there is a template registration tab in the console
https://github.com/tari-project/tari/blob/development/applications/minotari_console_wallet/src/ui/components/register_template_tab.rs#L541

sdbondi: Ahh ok thanks

sdbondi (approved):

PR #942, opened 2024-02-21
fix(consensus)!: add exhaust burn per transaction, fix block fee malleability

sdbondi:
Description
---
~~fix(consensus)!: calculate fees per block~~
fix(consensus)!: immalleable block fee 
feat(consensus)!: enable fixed exhaust of 5%

Motivation and Context
---
~~Previously, fees and exhaust burns were calculated per transaction. This increases the deviation caused by rounding and isa little more complex implementation-wise. This PR sums up fees for a block and then computes the leader fee and exhaust burn.~~

The leader fee calculation is improved to better account for burnt fees due to the remainder when dividing the fees by the number of involved shards.

This PR enables the exhaust at a fixed rate of 5% - this number is not finalized and will change but is enabled to allow for better testing.

The fee field in the block was previously malleable, this PR includes the fees in the block hash, and therefore block signature.

How Has This Been Tested?
---
New unit test, manually checked that fee agreement is maintained

---
Check the expected fee and burn rate in VN database

Breaking Changes
---

- [ ] None
- [ ] Other - Please specify


sdbondi:
Hmmm, it may have to be calculated per transaction as different shards with a different tx set would calculate a different burn for their block, and therefore a different burn for those transactions. $(\sum{f_i})/X \neq \sum{f_i/X}$ for integer division

dan_layer/state_store_sqlite/src/schema.rs
```diff
@@ -224,7 +224,7 @@ diesel::table! {
         evidence -> Text,
         remote_evidence -> Nullable<Text>,
         transaction_fee -> BigInt,
-        leader_fee -> BigInt,
+        leader_fee -> Nullable<Text>,
```
stringhandler: I would prefer two separate fields, but this is fine as well

stringhandler (dismissed):

dan_layer/state_store_sqlite/src/schema.rs
sdbondi: Yeah definitely better for queriability. Since there's a conflict I'll make those changes too.

stringhandler (dismissed):

stringhandler (dismissed):

stringhandler (approved):

PR #945, opened 2024-02-23
feat!: block pagination and header timestamp

Cifko:
Description
---
Add pagination to get blocks. All the filtering/sorting was moved from UI to backend.

Motivation and Context
---

How Has This Been Tested?
---
Works as usual, but now you can see in chrome that the amount of data downloaded on `get_blocks` call.

---
Above.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


dan_layer/storage/src/consensus_models/block.rs
```diff
@@ -96,6 +96,9 @@ pub struct Block {
     /// Signature of block by the proposer.
     #[cfg_attr(feature = "ts", ts(type = "{public_nonce : string, signature: string} | null"))]
     signature: Option<ValidatorSchnorrSignature>,
+    block_time: Option<u64>,
```
sdbondi, line 99: block_time should not be in the block that is sent between peers, over the wire, though it can still be calculated when inserting the block in the database and sent via JRPC.

dan_layer/storage/src/consensus_models/block.rs
```diff
@@ -96,6 +96,9 @@ pub struct Block {
...
+    #[cfg_attr(feature = "ts", ts(type = "number"))]
+    timestamp: u64,
```
sdbondi, line 102: timestamp needs to be included in the signature challenge (block id). The block_time should not since it's just informational and not sent over the wire.

dan_layer/storage/src/consensus_models/block.rs
```diff
@@ -129,6 +132,8 @@ impl Block {
             foreign_indexes: sorted_foreign_indexes,
             stored_at: None,
             signature,
+            block_time: None,
+            timestamp: EpochTime::now().as_u64(),
```
sdbondi: You cant use epoch time here since this will typically be different for each node, which will break things when timestamp is added to the signature. You need to add the timestamp field to the protobuf proposal message and set it from that. 

You should use EpochTime::now() in `on_propose`

dan_layer/storage/src/consensus_models/block.rs
```diff
@@ -231,6 +245,7 @@ impl Block {
             0,
             IndexMap::new(),
             None,
+            EpochTime::now().as_u64(),
```
sdbondi: This will be a problem as hashes will not match. I think let's pass in the same timestamp as the parent block.

Cifko: done

sdbondi (dismissed):

sdbondi:
Tested and working

sdbondi (dismissed):

sdbondi (approved):

PR #946, opened 2024-02-23
feat(engine)!: allow revealed-only mint and transfers

sdbondi:
Description
---
feat(engine)!: allow revealed-only mint and transfers
feat(engine): add the ability to count confidential commitments in a bucket
feat(engine): allow fetching of badge data from proofs

Motivation and Context
---
There are use cases where you want to mint or transfer confidential assets using only revealed funds. This was previously possible if the wallet generated a new (0) output commitment and set the revealed funds. This PR allows the user to explicitly omit the confidential output and only specify the revealed output. A user may still create the (0) output ($k.G + 0.H$) as before if required. For mints, this change will mint only revealed funds, and for transfers this will consume the inputs and only produce revealed funds for the transfer output/change output or both.

This allows the template to create a confidential output proof that only creates revealed funds (like a fungible resource) without the wallet having to generate commitments/rangeproofs.

```rust
       pub fn mint_revealed(&mut self, amount: Amount) {
            let proof = ConfidentialOutputProof::mint_revealed(amount);
            let bucket = ResourceManager::get(self.vault.resource_address()).mint_confidential(proof);
            self.vault.deposit(bucket);
        }
```

By allowing a template to fetch the number of confidential commitments in a bucket, a template author can either prevent confidential commitments from being used, or assert that they are being used.

Allowing the template to fetch NFT data from proofs enables templates to implement more complex rules for badges.

Breaking changes are due to the ConfidentialOutputProof struct change

How Has This Been Tested?
---
New unit tests that check minting

---
Create a template with the code above.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_dan_wallet_daemon/src/handlers/accounts.rs
+++ b/applications/tari_dan_wallet_daemon/src/handlers/accounts.rs
@@ -397,9 +397,9 @@ pub async fn handle_reveal_funds(
             .confidential_outputs_api()
             .resolve_output_masks(inputs, key_manager::TRANSACTION_BRANCH)?;
 
-        let reveal_proof = sdk
-            .confidential_crypto_api()
-            .generate_withdraw_proof(&inputs, &output_statement, None)?;
+        let reveal_proof =
+            sdk.confidential_crypto_api()
+                .generate_withdraw_proof(&inputs, Amount::zero(), &output_statement, None)?;
 
         info!(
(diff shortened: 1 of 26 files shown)


stringhandler (approved):

sdbondi:
Tried out git butler, looks like it didn't sign my commits :/

stringhandler (approved):

PR #948, opened 2024-02-26
fix(validator_node_web_ui): infinite loop on transaction details

mrnaveira:
Description
---
Fixed a `useEffect` call on the VN's transaction page to avoid an infinite refreshing issue

Motivation and Context
---
The VN UI transaction details page is currently broken due to an infinite loop.

How Has This Been Tested?
---
Manually: Blocks -> More Info -> Expand "Accept" -> Click on one of the transactions

---
See previous section

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_validator_node_web_ui/src/routes/Transactions/TransactionDetails.tsx
+++ b/applications/tari_validator_node_web_ui/src/routes/Transactions/TransactionDetails.tsx
@@ -113,7 +113,7 @@ export default function TransactionDetails() {
 
   useEffect(() => {
     getTransactionByHash();
-  });
+  }, []);
 
   const handleChange = (panel: string) => (event: React.SyntheticEvent, isExpanded: boolean) => {
     setExpandedPanels((prevExpandedPanels) => {


applications/tari_validator_node_web_ui/src/routes/Transactions/TransactionDetails.tsx
```diff
@@ -113,7 +113,7 @@ export default function TransactionDetails() {
 
   useEffect(() => {
     getTransactionByHash();
```
sdbondi, line 115: Isnt `[]` the default (not sure)? I think the problem is that we dont check if we're already loading or done.

Maybe something like:
```typescript
  useEffect(() => {
    if (!loading && !fee && !logs && !error) {
       getTransactionByHash();
    }
  }, [loading, fee, logs, error]);
```

mrnaveira: Without the `[]` it will run on any update on the page (which makes it very easy to get into infinite loops), the `[]` makes it run only on mount/unmount of the page (which is what we want in the transaction page)

https://stackoverflow.com/questions/53070970/infinite-loop-in-useeffect

sdbondi (approved):

PR #950, opened 2024-02-28
fix(walletd): fix query for NonFungible

sdbondi:
Description
---
fix(walletd): fix query for NonFungible
fix(walletd): serialize NFT id as string
chore: update bindings 
fix(bindings): fix NonFungible parsing in stringToSubstateId helper

Motivation and Context
---
Fetching by substate type `NonFungible` never returned any results. This PR fixes this.

How Has This Been Tested?
---
Calling `substates.list` with `filter_by_type: "NonFungible"`

---
Calling `substates.list` with `filter_by_type: "NonFungible"`

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/dan_layer/wallet/storage_sqlite/src/reader.rs
+++ b/dan_layer/wallet/storage_sqlite/src/reader.rs
@@ -257,7 +257,15 @@ impl WalletStoreReader for ReadTransaction<'_> {
             query = query.filter(substates::template_address.eq(template_address.to_string()));
         }
         if let Some(substate_type) = by_type {
-            query = query.filter(substates::address.like(format!("{}_%", substate_type.as_prefix_str())));
+            match substate_type {
+                SubstateType::NonFungible => {
+                    query = query
+                        .filter(substates::address.like(format!("resource_% {}_%", substate_type.as_prefix_str())));
+                },
+                _ => {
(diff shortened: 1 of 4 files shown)


dan_layer/wallet/storage_sqlite/src/reader.rs
```diff
@@ -257,7 +257,15 @@ impl WalletStoreReader for ReadTransaction<'_> {
             query = query.filter(substates::template_address.eq(template_address.to_string()));
         }
         if let Some(substate_type) = by_type {
-            query = query.filter(substates::address.like(format!("{}_%", substate_type.as_prefix_str())));
+            match substate_type {
+                SubstateType::NonFungible => {
```
stringhandler, line 261: Possible sql injection here, but I think substate_type can't be a generic string, so looks ok

stringhandler (approved):

dan_layer/wallet/storage_sqlite/src/reader.rs
sdbondi: Yeah cant be too careful, as_prefix_str returns a static str so it cant be anything other than a hardcoded string, so will be ok - NFT addresses need improving

PR #951, opened 2024-03-01
refactor(engine)!: substate address based on entity id prefix

sdbondi:
Description
---
Prefix all substate addresses with an entity ID to group objects related to a particular component in the same shard
All shards commit transaction receipt
Adds CreateAccount command which uses the owner public key prefix as the entity ID
Adds OwnerRule::ByPublicKey

Motivation and Context
---
Substate addresses for Components, Resources, Vaults and NFTs are composed of the entity id (20 bytes), componentKey (8 bytes) and version (4 bytes) (4.29 billion versions).
CreateAccount is implemented the same as CallFunction but sets the entity ID to the owner public key for the call. This can be later changed to create the account without invoking WASM

These "versionable" substate IDs are now 28 bytes long (entity id  + component key). 

Transaction receipts are committed by all involved shards. 

How Has This Been Tested?
---
CI, Manually

---
Verify that all objects created by a template/component call share the same prefix.

Breaking Changes
---

- [ ] None
- [x] Other - Please specify

BREAKING CHANGE: all SubstateIds are shorter which breaks all wasm templates (need to recompile), new instruction is not recognised by older versions.


stringhandler (dismissed):
tested locally and it's working as expected

stringhandler (approved):

PR #954, opened 2024-03-05
fix(consensus): reject transactions that DOWN substates pledged as input refs

sdbondi:
Description
---
fix(consensus): reject substates that are DOWNed as input refs

Motivation and Context
---
When checking substate locked, we use the diff to determine which inputs were written and attempt to lock those.
We can determine which inputs should have been input refs from this. We should consider removing input_refs because (1) these are difficult to determine by the user, and (2) the validator node can determine which inputs are never written from the transaction result which it always has by the time we vote on proposals.

How Has This Been Tested?
---
Manually tested that create account and create nft still work. Other manual tests that check this specific case are difficult because the indexer (autofiller) will always fill the inputs (unless the version matches exactly). Since some of this code will change when we can submit transactions to validator nodes without autofilling, no fixes for this are in this PR.

---
Submit a transaction that has an input ref including the correct version that is not written to

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


stringhandler (approved):

sdbondi:
Closing as this issue has been addressed in #1023

PR #967, opened 2024-03-11
feat(wallet/webui): send any resource

sdbondi:
Description
---
Add send button to each row in asset list to allow sending of any resource
Fix use of default URL if URL is not available (used for development)
Add dropdown to select optional badge for transfer

Motivation and Context
---
Previously you could only send Tari. This PR allows any resource to be sent.

When using vite during development, to get hot reload, the JRPC calls would fetch an HTML page. This would break all calls. This PR checks that the returned URL is well formed, if not the default is used.

How Has This Been Tested?
---
Manually

---
Use the wallet web UI to send non-Tari resources.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


stringhandler (dismissed):
Finally, thanks!

stringhandler (approved):

PR #968, opened 2024-03-12
feat: store base layer block hashes

Cifko:
Description
---
Store the base layer block hashes (not just the ones we sync, but to the very tip of the chain).

Motivation and Context
---
We will need this for accepting/rejecting the proposed blocks. Currently we have base layer block hash in the blocks, but we do nothing with it. We need to validate if the hash is indeed correct. For that we need all the hashes from the base layer.

---
Look into the `global_storage.sqlite` table to see that the correct data is stored.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_dan_app_utilities/src/base_layer_scanner.rs
+++ b/applications/tari_dan_app_utilities/src/base_layer_scanner.rs
@@ -264,7 +264,22 @@ impl<TAddr: NodeAddressable + 'static> BaseLayerScanner<TAddr> {
             },
             Some(end_height) => end_height,
         };
-
+        let mut scan = tip.tip_hash;
+        loop {
+            let header = self.base_node_client.get_header_by_hash(scan).await?;
+            if let Some(last_tip) = self.last_scanned_tip {
+                if last_tip == scan {
+                    // This was processed on the previous call to this function.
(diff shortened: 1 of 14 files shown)


applications/tari_dan_app_utilities/src/base_layer_scanner.rs
```diff
@@ -264,7 +264,22 @@ impl<TAddr: NodeAddressable + 'static> BaseLayerScanner<TAddr> {
...
+                if last_tip == scan {
+                    // This was processed on the previous call to this function.
+                    break;
+                }
+            }
+            if header.height == end_height {
+                // This will be processed down below.
+                break;
+            }
+            self.epoch_manager.update_epoch(header.height, scan, false).await?;
```
sdbondi: Rather make two calls than pass in some opaque bool. 

First call (here) adds the block. Second call (below) activates/updates the epoch

Cifko: I did separate the calls, and also the message that is send around in the function. That way if we decide to delete this code it will be easier to do so.

sdbondi (approved):

PR #970, opened 2024-03-12
fix: allow multiple concurrent transactions into the same block

mrnaveira:
Description
---
* Avoid adding inputs without versions to the input evidence. Before we were adding always the version 0.
* The count for involved shards (used for fee calculation) is calculated in a special case for transactions without versions

Motivation and Context
---
Following the previous work on concurrency (https://github.com/tari-project/tari-dan/pull/922), this PR solves the problem of concurrent transactions not being batched together in the same block.

The problem happened on the new block calculation, the database function `transaction_pool_get_many_ready` uses the `evidence` field to detect conflicts and select transactions. The `evidence` field for inputs without version was always adding the version 0, which result on blocks only picking up one concurrent transaction per block.

This PR solves this problem by avoiding adding inputs without versions into the `evidence` field. 

How Has This Been Tested?
---
Using the `concurrency` cucumber scenario, setting a lot of concurrent transactions (e.g. 100) and checking logs and database for block commands. Transactions eventually fail but this is due to other problems in concurrency.

---
See previous section

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


dan_layer/consensus/src/hotstuff/on_propose.rs
```diff
@@ -234,7 +234,15 @@ where TConsensusSpec: ConsensusSpec
                 // prepared. We can now propose to Accept it. We also propose the decision change which everyone
                 // should agree with if they received the same foreign LocalPrepare.
                 TransactionPoolStage::LocalPrepared => {
-                    let involved = local_committee_shard.count_distinct_shards(t.transaction().evidence.shards_iter());
+                    // For now we need to treat transactions without versions in a special case
+                    // TODO: update the evidence after execution so all transactions are treated equally here
+                    let db_transaction = t.get_transaction(tx)?;
+                    let involved = if db_transaction.transaction().has_inputs_without_version() {
+                        db_transaction.transaction().all_inputs_iter().count()
```
sdbondi, line 241: The number of inputs is not necessarily (and usually not) the number of involved shards

sdbondi (commented):
This appears to be a workaround for the partial implementation of late execution in consensus. The evidence should always contain all involved shards, and working around it may cause unexpected bugs. I think this would revert to the previous code when we implement a more complete solution for "consensus-executed" transactions. Currently, we insert a "dummy" executed transaction in the mempool. 

Consensus can fetch unexecuted transaction using `TransactionRecord::get` (loads either executed or unexecuted) instead of ExecutedTransaction::get (only executed) in some places and take appropriate action if the transaction was not executed in the mempool. I'd rather work towards correcting this than merge a workaround if that makes sense.

sdbondi:
PR cleanup

PR #971, opened 2024-03-13
fix: broken webui builds and abort compilation when this happens

Cifko:
Description
---
Fix the broken web ui builds (some of the types in the webui did not exists).
The build scripts were update to catch these errors.you
Also apply `.prettierrc` to all files in the repo.

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_dan_wallet_daemon/build.rs
+++ b/applications/tari_dan_wallet_daemon/build.rs
@@ -20,13 +20,19 @@
 // WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-use std::{env, process::Command};
+use std::process::Command;
+
+fn exit_on_ci() {
+    if option_env!("CI").is_some() {
+        std::process::exit(1);
+    }
+}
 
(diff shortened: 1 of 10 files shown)


applications/tari_dan_wallet_daemon/build.rs
```diff
@@ -39,23 +39,25 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-            if !s.success() || s.code().unwrap_or(0) != 0 {
-                println!("cargo:warning='npm run build' failed!");
-                println!("cargo:warning=The web ui will not be included!");
-                panic!("npm run build failed");
-            }
+        Ok(output) if !output.status.success() => {
+            println!("cargo:warning='npm run build' exited with non-zero status code");
+            println!("cargo:warning=Output: {}", String::from_utf8_lossy(&output.stdout));
+            println!("cargo:warning=Error: {}", String::from_utf8_lossy(&output.stderr));
+            std::process::exit(1);
```
sdbondi: I seem to remember that we didn't want to exit because some shells don't have nodejs available. Also if something is broken on the webui, you may still want to run the wallet, VN and indexer when developing.

I suggest that you create a function called `exit_on_ci()` 

```rust
fn exit_on_ci() {
 if  option_env!("CI").is_some() {
    std::process::exit(1);
}
}
```

Cifko: done

sdbondi (approved):

PR #972, opened 2024-03-13
feat!: check the incoming base layer block hash in proposals

Cifko:
Description
---
We are adding the check for the base layer hash in the proposal to be within some +- from the current synced height. And it has to be known.
The check if the proposed block is too much behind is disabled. It's causing issues during sync (added TODO).

Motivation and Context
---

How Has This Been Tested?
---
I just did some logging to see if the height is being compared. But not the failing path.

---
Maybe run a couple vns, with running miner and see what's going on, will it fail? Make sure you run multiple committees to also test that foreignproposals will be validated.

Breaking Changes
---

- [ ] None
- [x] Other - The 2nd layer should be restarted, otherwise the blocks validation will fail. But wait for the next PR that also have new breaking changes.

### DIFF
--- a/applications/tari_dan_app_utilities/src/consensus_constants.rs
+++ b/applications/tari_dan_app_utilities/src/consensus_constants.rs
@@ -24,6 +24,8 @@
 pub struct ConsensusConstants {
     pub base_layer_confirmations: u64,
     pub committee_size: u32,
+    pub max_base_layer_blocks_ahead: u64,
+    pub max_base_layer_blocks_behind: u64,
 }
 
 impl ConsensusConstants {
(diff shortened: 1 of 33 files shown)


sdbondi (dismissed):
utACK

sdbondi:
Looks like some tests are failing

```
        FAIL [  16.111s] consensus_tests consensus::leader_failure_node_goes_down
273
        FAIL [  39.565s] consensus_tests consensus::leader_failure_output_conflict
```

`Consensus failure: Storage error: Not found in operation blocks_get: Record not found`

sdbondi (approved):

PR #975, opened 2024-03-14
feat(engine): builtin vault events for deposits and withdrawals

mrnaveira:
Description
---
* Updated engine vault actions for deposit and withdrawal to emit standard events
* Prevent user emitted events to use the new standard topic prefix

Motivation and Context
---
We want the engine to emit default events for vault deposits and withdrawals, to be able to easily query them in the indexer.

This PR updates the engine to emit those events:
* The topics are: `std.vault.withdraw` and `std.vault.deposit`.
* We explicitly forbid users to use topics that start with `std.`
* The event payload includes `amount` (only if non confidential fungible), `resource_address` and `vault_id`
* Events already include `template_address`, `component_address` and `tx_hash` by default

How Has This Been Tested?
---
New unit tests to assert the new events

---
Transfers funds and check the events generated by the transaction

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


dan_layer/engine/src/runtime/impl.rs
```diff
@@ -227,6 +234,41 @@ impl<TTemplateProvider: TemplateProvider<Template = LoadedTemplate>> RuntimeInte
...
+
+        let mut payload = Metadata::new();
+        payload.insert("vault_id", vault_id.to_string());
+        payload.insert("resource_address", resource_address.to_string());
+        if let Some(amount) = amount {
+            payload.insert("amount", amount.to_string());
+        }
+
+        let event = Event::new(component_address, *template_address, tx_hash, topic, payload);
+        log::log!(target: "tari::dan::engine::runtime", log::Level::Debug, "{}", event.to_string());
```
sdbondi:
```suggestion
        debug!(target: LOG_TARGET, "Emitted vault event {}", event);
```

dan_layer/engine/src/runtime/impl.rs
```diff
@@ -227,6 +234,41 @@ impl<TTemplateProvider: TemplateProvider<Template = LoadedTemplate>> RuntimeInte
...
+        state: &mut WorkingState,
+    ) -> Result<(), RuntimeError> {
+        self.invoke_modules_on_runtime_call("emit_event")?;
+
+        let component_address = Ok::<_, RuntimeError>(
+            state
+                .current_call_scope()?
+                .get_current_component_lock()
+                .and_then(|l| l.address().as_component_address()),
+        )?;
```
sdbondi:
```suggestion
        let component_address = state
                .current_call_scope()?
                .get_current_component_lock()
                .and_then(|l| l.address().as_component_address());
```

dan_layer/engine/src/runtime/impl.rs
```diff
@@ -832,6 +879,14 @@ impl<TTemplateProvider: TemplateProvider<Template = LoadedTemplate>> RuntimeInte
 
                 self.tracker.write_with(|state| {
                     let vault_lock = state.lock_substate(&SubstateId::Vault(vault_id), LockFlag::Write)?;
+                    let bucket = state.take_bucket(bucket_id)?;
+
+                    // Emit a builtin event for the deposit
+                    let amount = match bucket.resource_type() {
+                        ResourceType::Fungible => Some(bucket.amount()),
+                        _ => None,
+                    };
```
sdbondi: All resources support amount 

```suggestion
                    let amount = bucket.amount();
```

dan_layer/engine/src/runtime/impl.rs
```diff
@@ -877,12 +932,15 @@ impl<TTemplateProvider: TemplateProvider<Template = LoadedTemplate>> RuntimeInte
...
 
                     let vault_mut = state.get_vault_mut(&vault_lock)?;
-                    let resource_container = match arg {
-                        VaultWithdrawArg::Fungible { amount } => vault_mut.withdraw(amount)?,
-                        VaultWithdrawArg::NonFungible { ids } => vault_mut.withdraw_non_fungibles(&ids)?,
-                        VaultWithdrawArg::Confidential { proof } => vault_mut.withdraw_confidential(*proof)?,
+                    let (resource_container, amount) = match arg {
+                        VaultWithdrawArg::Fungible { amount } => (vault_mut.withdraw(amount)?, Some(amount)),
+                        VaultWithdrawArg::NonFungible { ids } => (vault_mut.withdraw_non_fungibles(&ids)?, None),
+                        VaultWithdrawArg::Confidential { proof } => (vault_mut.withdraw_confidential(*proof)?, None),
```
sdbondi: Each withdraw can have an amount. For non-fungible, it's the number of tokens. For Confidential, it can be the number of revealed tokens.

dan_layer/engine/src/runtime/impl.rs
```diff
@@ -227,6 +234,41 @@ impl<TTemplateProvider: TemplateProvider<Template = LoadedTemplate>> RuntimeInte
...
+                .and_then(|l| l.address().as_component_address()),
+        )?;
+
+        let tx_hash = self.entity_id_provider.transaction_hash();
+        let (template_address, _) = state.current_template()?;
+        let resource_address = state.get_vault(&vault_lock)?.resource_address();
+
+        let mut payload = Metadata::new();
+        payload.insert("vault_id", vault_id.to_string());
+        payload.insert("resource_address", resource_address.to_string());
```
sdbondi, line 257: Maybe include the resource type too

dan_layer/engine/tests/events.rs
```diff
@@ -1,31 +1,151 @@
...
+    assert!(result.finalize.is_accept());
+
+    // a standard event for the withdraw must have been emmitted
+    assert!(result.finalize.events.iter().any(|e| {
+        e.topic() == "std.vault.withdraw".to_owned() &&
+            e.template_address() == ACCOUNT_TEMPLATE_ADDRESS &&
+            e.component_address().unwrap() == sender_address &&
+            *e.payload().get("resource_address").unwrap() == faucet_resource.to_string() &&
+            *e.payload().get("amount").unwrap() == amount.to_string()
+    }));
```
sdbondi: Prefer if each of these are separate asserts, will make it easier to tell what exactly is wrong if the test fails

```rust
assert_eq!(e.topic(), "std.vault.withdraw");
// etc
```

sdbondi (commented):
Looks good, only small nits

sdbondi (approved):

PR #976, opened 2024-03-14
feat(engine)!: add Elgamal verifiable encryption (optional view key) support

sdbondi:
Description
---
feat(engine)!: add Elgamal verifiable encryption (optional view key) support
feat: implement brute force value decryption for confidential outputs using secret view key
refactor: move confidential wallet crypto into new crate
refactor(wallet/sdk): use wallet crypto crate
refactor(test tooling): use wallet crypto crate
tests(engine): new test checking confidential transfers with a view key enabled
tests(wallet/crypto): checks that proof generation, verification and decryption are valid
feat(engine): add simple interface for value lookup table

Motivation and Context
---
Allow a template author to specify a view key on confidential resources optionally.   This allows anyone with the secret key to uncover the balance of commitments generated for the resource.

```rust
        let coins = ResourceBuilder::confidential()
                .initial_supply(confidential_proof)
                .with_view_key(view_key)
                .build_bucket();
```

Wallets MUST generate a ViewableBalanceProof for all confidential outputs for the resource, allowing validators to verify that the encrypted balance was generated correctly without revealing the balance.

All confidential crypto was duplicated in the test tooling and wallet SDK. Since duplicating the Elgamal verifiable encryption scheme could lead to issues down the road, or is just plain ugly, this PR puts all confidential crypto into a crate that is used by the test tooling and wallet SDK.

A value lookup table is passed into the brute force function. A production implementation of may make use of the binary file provided by the new generate_ristretto_value_lookup bin crate to return canonical (compressed) bytes for a value. The implementation can optimise for sequential reads and a low memory footprint. A binary file containing 1 billion entries (assuming 6 decimals, whole values from 0 - 1000) will be 32 x 1B in size (32Gb).

A future PR will add wallet support for attempting to reveal the balance of vaults. 

How Has This Been Tested?
---
New unit tests

---
Create a template using the view key in a resource.
Check transfers work as before.
Fetch the commitments from a vault and reveal the balance (this is a manual process currently without wallet support)

Breaking Changes
---
- [ ] None
- [x] Other - Please specify

BREAKING CHANGES: added field to resource create args, meaning any template using this would need to be recompiled.

### DIFF
--- a/dan_layer/template_lib/src/models/confidential_proof.rs
+++ b/dan_layer/template_lib/src/models/confidential_proof.rs
@@ -7,7 +7,7 @@ use serde_with::{serde_as, Bytes};
 use ts_rs::TS;
 
 use crate::{
-    crypto::{BalanceProofSignature, PedersonCommitmentBytes, RistrettoPublicKeyBytes},
+    crypto::{BalanceProofSignature, PedersonCommitmentBytes, RistrettoPublicKeyBytes, SchnorrSignatureBytes},
     models::Amount,
 };
 
@@ -40,12 +40,11 @@ impl ConfidentialOutputProof {
 }
 
 /// A zero-knowledge proof that a confidential resource amount is valid
-#[serde_as]
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
 #[cfg_attr(feature = "ts", derive(TS), ts(export, export_to = "../../bindings/src/types/"))]
 pub struct ConfidentialStatement {
-    #[serde_as(as = "Bytes")]
-    pub commitment: [u8; 32],
+    #[cfg_attr(feature = "ts", ts(type = "Array<number>"))]
+    pub commitment: PedersonCommitmentBytes,
     /// Public nonce (R) that was used to generate the commitment mask
     #[cfg_attr(feature = "ts", ts(type = "Array<number>"))]
     pub sender_public_nonce: RistrettoPublicKeyBytes,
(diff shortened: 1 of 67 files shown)


dan_layer/template_lib/src/models/confidential_proof.rs
```diff
@@ -54,14 +53,58 @@ pub struct ConfidentialStatement {
...
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+#[cfg_attr(feature = "ts", derive(TS), ts(export, export_to = "../../bindings/src/types/"))]
+pub struct ViewableBalanceProof {
+    /// E = v.G + r.P
+    #[cfg_attr(feature = "ts", ts(type = "Uint8Array"))]
+    pub elgamal_encrypted: RistrettoPublicKeyBytes,
+    #[cfg_attr(feature = "ts", ts(type = "Uint8Array"))]
+    pub elgamal_public_nonce: RistrettoPublicKeyBytes,
+    #[cfg_attr(feature = "ts", ts(type = "Uint8Array"))]
+    pub c_prime: PedersonCommitmentBytes,
```
stringhandler, line 69: Maybe we need a doc or something for these letters

stringhandler (approved):

dan_layer/template_lib/src/models/confidential_proof.rs
sdbondi: Cool will add on the next PR - I had copilot do them at first but they were all wrong (even though they looked right) so I removed them lol

PR #982, opened 2024-03-19
feat!: implements epoch change

Cifko:
Description
---
Add epoch events
- `End` this is added at the last base layer block that is still in the epoch. We have to bury the block to be locked, after that we officially end the epoch
- `Start` this will be in the first block when the epoch switch, its parent will have `End` or it will be genesis block. 

Motivation and Context
---

How Has This Been Tested?
---
I've tested multiple committees and I switched epoch many times. I've checked the blocks and the commands in them to confirm the epoch end/start.
There is also a cucumber test. That checks if the epoch ended with a EpochEnd command and new one started with EpochStart command.

---
Do as above, or go a step further and try to make a transaction during epoch switching when you also add a VNs that changes space division.

Breaking Changes
---

- [ ] None
- [x] Other - The 2nd layer should be restarted, otherwise the blocks validation will fail.


dan_layer/consensus/src/hotstuff/on_propose.rs
```diff
@@ -141,8 +142,28 @@ where TConsensusSpec: ConsensusSpec
...
         };
 
-        let base_layer_block_height =
-            std::cmp::max(qc_block.base_layer_block_height(), current_base_layer_block_height);
+        let locked_block = self.store.with_read_tx(|tx| {
+            let locked_block = LockedBlock::get(tx)?;
+            Block::get(tx, locked_block.block_id())
+        })?;
+        // If epoch has changed, we should first end the epoch with an EpochEvent::End
+        let epoch_end = !locked_block.is_epoch_end() &&
```
stringhandler: Please add some comments to explain what each part of this `if` represents

dan_layer/consensus/src/hotstuff/on_ready_to_vote_on_local_block.rs
```diff
@@ -121,9 +123,19 @@ where TConsensusSpec: ConsensusSpec
...
         let mut locked_blocks = Vec::new();
         let mut finalized_transactions = Vec::new();
+        let parent_block = self.store.with_read_tx(|tx| valid_block.block().get_parent(tx))?;
+        let locked_block = self.store.with_read_tx(|tx| {
+            let locked_block = LockedBlock::get(tx)?;
+            Block::get(tx, locked_block.block_id())
+        })?;
+        let epoch_start = parent_block.epoch() < valid_block.epoch();
 
+        let epoch_end = parent_block.epoch() == valid_block.epoch() &&
```
stringhandler: also here please

applications/tari_validator_node/src/p2p/rpc/sync_task.rs
```diff
@@ -136,11 +140,36 @@ impl<TStateStore: StateStore> BlockSyncTask<TStateStore> {
...
+            loop {
+                let block = self.store.with_read_tx(|tx| {
+                    let locked_block = LockedBlock::get(tx)?;
+                    Block::get(tx, locked_block.block_id())
+                })?;
+                if block.is_epoch_end() && block.epoch().0 + 1 == up_to_epoch.0 {
+                    // If found the epoch end block, break.
+                    break;
+                }
+                tokio::time::sleep(Duration::from_secs(10)).await;
```
sdbondi: Added a comment to explain this

dan_layer/rpc_state_sync/src/manager.rs
```diff
@@ -554,6 +555,122 @@ where TConsensusSpec: ConsensusSpec<Addr = PeerAddress> + Send + Sync + 'static
...
+            Some(leaf_block) => {
+                let block = self
+                    .state_store
+                    .with_read_tx(|tx| Block::get(tx, leaf_block.block_id()))?;
+                block.epoch()
+            },
+            None => Epoch(0),
+        };
+        let current_epoch = self.epoch_manager.current_epoch().await?;
+        if current_epoch > leaf_epoch {
```
sdbondi, line 584: Sync is also required if we fall far behind, say if we shut down for a few minutes. Looks as if this will only sync if we're a whole epoch behind.

sdbondi (commented):
Nice work on what is a tricky problem. Tested 1 VN then adding another, and works in this case.

My updates to this PR:
- update bindings
- update vn web UI for epoch events
- added some comments
- couple of minor code improvements 

NOTES/TODO: 
- Sync should work if fallen behind in the same epoch event after initial sync and register
- Formalise the process for being ready to perform consensus as a new VN. 
  - The strategy in this PR is to "pause" the RPC call until VNs lock the `epochend` block and then sync to the leaf block. This LGTM for now but should be improved.
  - We may want another call that streams blocks as they come in, this could also be useful for the indexer.
  - Perhaps require each new VN to sync state and monitor new blocks, then provide proof that they have synced the state before epoch change. VNs would then include the VNs going in and out for the next epoch inside the end-of-epoch block. This mechanism could also come into play when VNs are causing leader failures and should be skipped for the epoch (as voted by 2/3 majority of VNs)
- Commands should be verbs, not nouns. Maybe `Command::{Start|End}Epoch`
- The last 3 blocks of the epoch all have EpochEvent: End. Maybe not strictly necessary, but no problem there however I suspect that things will break if a VN deviates from this e.g. leader failure after epoch end may break things (unconfirmed)
- Some validations are missing e..g check that no transactions are present in end-of-epoch blocks

applications/tari_validator_node/src/p2p/rpc/sync_task.rs
sdbondi: I've added this

sdbondi:
Further testing revealed a unique constraint issue when validator nodes re-register. Investigating.

dan_layer/storage_sqlite/global_db_migrations/2022-09-30-212244_create_vns_and_committees/up.sql
```diff
@@ -28,3 +28,4 @@ create table validator_nodes (
 );
 
 create index validator_nodes_epoch_index on validator_nodes (epoch);
+create unique index validator_nodes_public_key_unique_index on validator_nodes (public_key);
```
sdbondi: This is a problem if the validator node reregisters (confirmed in testing). I'll update the queries to account for the 1 VN : many registrations relation.

dan_layer/rpc_state_sync/src/manager.rs
sdbondi: Ah ok this tries to resolve which shard range to query for.

sdbondi (dismissed):

sdbondi (approved):

PR #986, opened 2024-03-21
feat: store substate_id in events

mrnaveira:
Description
---
* Updated the `events` database table for the new field
* Updated database API and managers to use substate ids instead of component addresses
* Builtin vault events (e.g. `std.vault.deposit`) now emit two events: one has the vault as `substate_id` and the other has the resource. This is done to allow queries by both criteria

Motivation and Context
---
Up until now events stored the component address related to it. This is not general enough now because we want also to emit events that can be related to other type of substates (e.g. resources or vaults). This PR changes the `component_address` field for `substate_id` for events in the whole project (from database to APIs)

How Has This Been Tested?
---
* Existing unit and integration tests pass.
* Manually inspected the content of the new `substate_id` field in different types of events

---
Emit any type of event and check the new `substate_id` field value

Breaking Changes
---

- [ ] None
- [x] Event fields change (`substate_id` instead of `component_address`) so it requires network reset


dan_layer/engine/src/runtime/impl.rs
```diff
@@ -282,18 +294,22 @@ impl<TTemplateProvider: TemplateProvider<Template = LoadedTemplate>> RuntimeInte
...
                 state
                     .current_call_scope()?
                     .get_current_component_lock()
                     .and_then(|l| l.address().as_component_address()),
             )
         })?;
+        let substate_id = match component_address_option {
+            Some(component_address) => Some(SubstateId::Component(component_address)),
+            None => None,
+        };
```
sdbondi: https://rust-lang.github.io/rust-clippy/master/index.html#/manual_map

```suggestion
        let substate_id = component_address_option.map(SubstateId::Component);
```

sdbondi:
LGTM - will merge once conflicts are fixed and tests pass

sdbondi (dismissed):

sdbondi:
Some standard events are no longer associated with the account component in the cucumber tests. I pushed a change to provide more helpful error output on the test and to remove the unnecessary num_events parameter (num_events is equal to the number of events provided in the check). 

```
Unexpected topic at index 2. Events emitted were component-created, deposit, deposit, deposit, deposit, deposit, deposit. Expected component-created,deposit,std.vault.deposit,deposit,std.vault.deposit,deposit,std.vault.deposit,deposit,std.vault.deposit,deposit,std.vault.deposit,deposit,std.vault.deposit
```

I suggest adding more checks to check the events emitted for the resource, vault etc.

e.g.
`When indexer IDX1 scans the network events for substate NFT/resources/0 with topics std.resource...`
`When indexer IDX1 scans the network events for substate NFT/vault/0 with topics std.vault...`
etc

sdbondi (approved):

PR #994, opened 2024-03-28
feat: add ability to register multiple sidechains on the base layer

stringhandler:
BREAKING CHANGE: Adds Shard to QC's and proposals. Removes merkle proof of VNS


dan_layer/consensus/src/block_validations.rs
```diff
@@ -127,32 +127,40 @@ pub async fn check_quorum_certificate<TConsensusSpec: ConsensusSpec>(
     vote_signing_service: &TConsensusSpec::SignatureService,
     epoch_manager: &TConsensusSpec::EpochManager,
 ) -> Result<(), HotStuffError> {
-    if candidate_block.justify().epoch().as_u64() == 0 {
-        // Ignore genesis block.
+    let qc = candidate_block.justify();
```
stringhandler, line 130: @sdbondi Please have a look at this logic, I expanded it a bit and it's potentially out of scope of this PR

integration_tests/tests/features/claim_fees.feature
```diff
@@ -84,52 +84,3 @@ Feature: Claim Fees
 
     # Claim fees into ACC2
     When I claim fees for validator VN and epoch 1 into account ACC2 using the wallet daemon WALLET_D, it fails
-
-  @serial
-  Scenario: Prevent validator fees claim for unauthorized wallet
```
stringhandler, line 89: I deleted this test as it was basically the same as the one above

sdbondi: This test above checks that the fees can't be double-claimed. This test checks if a wallet does not use/possess the private claim key, it cannot claim fees.

dan_layer/consensus/src/block_validations.rs
sdbondi: Yeah the previous check isn't right. Also agree with the comment that a blanket "LGTM" for all genesis blocks in a number of places in the code is wrong. I think we need to refactor/audit the validations in general, and also relook at if the VN merkle proof in the QC is necessary. Not in scope for this PR so happy with these changes.

dan_layer/consensus/src/block_validations.rs
```diff
@@ -127,32 +127,40 @@ pub async fn check_quorum_certificate<TConsensusSpec: ConsensusSpec>(
...
+    if qc.is_genesis() {
+        // This is potentially dangerous. There should be a check
+        // to make sure this is the start of the chain.
+
         return Ok(());
     }
-    if candidate_block.height() < candidate_block.justify().block_height() {
+    if candidate_block.height() < qc.block_height() {
```
sdbondi, line 137: Should be <=

integration_tests/tests/features/claim_fees.feature
stringhandler: Ah ok, thanks. I've added a line to the test to also check that a non-account holder can't claim the fees

Cargo.toml
```diff
@@ -249,26 +250,26 @@ panic = 'abort'
 # seen in debug mode. Panicking at this time is better than silently using the wrong value.
 overflow-checks = true
 
-# Make a copy of this code, uncomment and replace account and my-branch with the name of your fork and the branch you want to temporarily use
+## Make a copy of this code, uncomment and replace account and my-branch with the name of your fork and the branch you want to temporarily use
```
sdbondi, line 253: Please revert these changes

sdbondi (approved):
utACK

Looks good, some nice improvements to cucumbers.

PR #998, opened 2024-04-03
feat(indexer): scan and store all events from the network

mrnaveira:
Description
---
* The indexer scans the network for events using the VN P2P API:
    1. Uses the `sync_blocks` method to get the blocks across all the network
    2. Inspects the blocks to extract all _finalized_ and _committed_ transactions
    3. Uses the `get_transaction_result` to get the results of all the transactions in the previous blocks
    4. inspects each transaction result to extract all events
* The indexer stores all scanned events into the database, allowing to query them by `substate_id` and `topic`. Supports pagination of results using the `offset` and `limit` parameters.
* The indexer's `getEvents` GraphQL method fetches the requested events from the database.

Motivation and Context
---
We want a way to search all events emitted in the network (e.g. to list all transfers of a particular resource).

This PR implements this feature with the following limitations (to be tackled in future PRs):
* Smarter scanning by treating each shard individually and storing the latest block for each shard, so subsequent scans only start from the latest scanned block onwards.
* Chronological ordering of events, using epoch + block id to order
* There is a lot of code duplicity surrounding event methods (in database, graphql, manager, etc.), unify all of them in each layer with generic methods

How Has This Been Tested?
---
The existing cucumber scenario `Indexer GraphQL filtering and pagination of events` does return the requested events.

---
Invoke the `getEvents` GraphQL method and inspect the result

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


stringhandler (commented):
Looks good, but I think you should track the starting block for each vn individually, because they may be on different tips, but also, the different shards might run at different speeds

applications/tari_indexer/src/event_manager.rs
```diff
@@ -0,0 +1,366 @@
...
+    async fn get_all_vns(&self) -> Result<Committee<PeerAddress>, anyhow::Error> {
+        // get all the committees
+        // TODO: optimize by getting all individual CommiteeShards instead of all the VNs
+        let epoch = self.epoch_manager.current_epoch().await?;
+        let full_range = RangeInclusive::new(SubstateAddress::zero(), SubstateAddress::max());
+        let mut committee = self
+            .epoch_manager
+            .get_committee_within_shard_range(epoch, full_range)
+            .await?;
+        committee.members.shuffle(&mut OsRng);
```
sdbondi, line 296: Committee has a shuffle method which means you can probably remove the `rand` dependency.
```suggestion
        committee.shuffle();
```

sdbondi (approved):

PR #1003, opened 2024-04-05
feat(indexer): smarter event scanning

mrnaveira:
Description
---
* Added a convenience `get_network_committees` method in the epoch manager, so we can get the network committee segmentation
* In the indexer:
    * Added database tables and all needed methods to read/store the latest scanned block for each committee
    * Updated the `EventManager` to treat each committee separately and use/update the last scanned block

Motivation and Context
---
Continuing the work of https://github.com/tari-project/tari-dan/pull/998, this PR implements a smarter approach for event scanning in the indexer:
* Querying each committee individually
* Storing in the database the latest scanned block for each committee, so subsequent scans only start from the latest scanned block onwards.

How Has This Been Tested?
---
Running the existing cucumber scenario `Indexer GraphQL filtering and pagination of events` and inspecting the logs (and SQLite database) of the indexer for the `event_manager` messages

---
See previous section

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi:
Think you'll need to update bindings

sdbondi (approved):
utACK

Nice

PR #1004, opened 2024-04-05
feat: tari swarm

sdbondi:
Description
---
feat: adds tari swarm binary to spawn and manage all processes required for L2

Motivation and Context
---
Port of https://github.com/tari-project/dan-testing to rust. 

Currently implemented: 
- Create an initial config for the swarm
- Compile minotari node binary
- Run minotari node binary
- Additional node types can easily be added to the node
- Web UI brought over from dan-testing
- named port allocation
- minotari node and wallet
- L2 node, wallet and indexer
- JSON-RPC methods, get logs, template uploads
- VN registration
- basic process management
- mining

How Has This Been Tested?
---
`tari_swarm_daemon -c data/swarm.toml init`
`tari_swarm_daemon -c data/swarm.toml start`

---
Above commands

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


applications/tari_swarm_daemon/src/main.rs
```diff
@@ -0,0 +1,309 @@
...
+        result = webserver => {
+            result??;
+            log::info!("Webserver exited");
+        },
+        result = task_handle => {
+            result??;
+            log::info!("Process manager exited");
+        }
+    }
+
```
stringhandler, line 223: do you need to call `pm_handle.terminate_all()` if the second or third branches happened?

stringhandler (dismissed):
nice, will try test it locally

applications/tari_swarm_daemon/src/main.rs
sdbondi: Yeah you are right for the web server (though I enabled kill on drop which will hard terminate the processes). If the process manager exits we cannot make that call, so just added it for the webserver.

sdbondi:
> nice, will try test it locally

Thanks, let me know if there are any features/behaviours that do not match from the python code. I think some calls were only used with the VS code extension which has not been tested.

stringhandler (approved):

PR #1007, opened 2024-04-09
chore: use different hash crate

Cifko:
Description
---
There are 2 hash crates on `feature-dan2` in the `tari` repo. I'm going to remove the old one.

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (dismissed):
Thanks - this replaced tari_hash_domains, which was removed, and includes the borsh hasher from the base layer. We can remove the copy of that code in the dan layer now but can leave that for another PR. This PR would also need to include the other updates for feature-dan2 (https://github.com/tari-project/tari/pull/6225)

sdbondi (commented):
This work is underway in this PR https://github.com/tari-project/tari-dan/pull/994

sdbondi (approved):

PR #1011, opened 2024-04-17
feat: add sidechain config

stringhandler:
Description
---
Adds config for using a specific sidechain.

Motivation and Context
---
This may not make it all the way to mainnet, but it allows us to use a single base layer (igor) to host multiple second layer sidechains

How Has This Been Tested?
---
Tested locally

---
Register a VN on the console wallet using a sidechain_deployment_key, then change the config for the VN and start it.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (dismissed):
utACK

PR #1040, opened 2024-05-22
fix(template-lib)!: improve initial supply api resource builder

sdbondi:
Description
---
Improve API for resource builder with initial supply.
Rename XTR2 constant to XTR
Add `Bucket::join`

Motivation and Context
---

Previously if you provided an initial supply for the resource builder and called `build()` a panic would occur.
This PR changes `initial_supply` to immediately build and return the bucket preventing the user from making the above error at compile time.

Previous API: 
```rust
let bucket = ResourceBuilder::fungible().initial_supply(1000).build_bucket()
```
```rust
// will panic at runtime
let bucket = ResourceBuilder::fungible().build_bucket()
```
New API:
```rust
let bucket = ResourceBuilder::fungible().initial_supply(1000);

let bucket = ResourceBuilder::non_fungible()
       .initial_supply((1..=100).map(NonFungibleId::from_u32));

let confidential = ResourceBuilder::confidential()
       .initial_supply(confidential_supply);
```

How Has This Been Tested?
---
Existing engine tests updated

---
Update existing templates to use the new API

Breaking Changes
---

- [ ] None
- [x] Other - Please specify

BREAKING CHANGE: templates that previously built resources with initial supply will no longer compile. The ABI for buckets adds `Join`. Since the integer representation of the ops shifts by 1 after `Join` some bucket operations will be invalid on previous WASM. i.e. WASM needs to be recompiled.

### DIFF
--- a/applications/tari_validator_node/src/bootstrap.rs
+++ b/applications/tari_validator_node/src/bootstrap.rs
@@ -485,7 +485,7 @@ where
     let substate_id = SubstateId::Resource(CONFIDENTIAL_TARI_RESOURCE_ADDRESS);
     let substate_address = SubstateAddress::from_substate_id(&substate_id, 0);
     let mut metadata = Metadata::new();
-    metadata.insert(TOKEN_SYMBOL, "tXTR2".to_string());
+    metadata.insert(TOKEN_SYMBOL, "tXTR".to_string());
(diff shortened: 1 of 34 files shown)


mrnaveira (approved):
LGTM

PR #1043, opened 2024-06-03
refactor(indexer): event filtering and related substate storing

mrnaveira:
Description
---
* New configuration in the indexer for event filtering
* Event filtering logic in the `event_scanner` module
* Removal of all the code related to substate address monitoring
* If an event is filtered to be stored in the database, the related substate (if any) will also be stored in the database automatically. The related substate is extracted from the corresponding transaction result.
* Refactor of the `get_non_fungibles` functionality to fetch all non fungibles of a collection from the network on the fly
* Refactored the `event_manager` and `substate_manager` to simplify and properly split them, as a lot of functionality was duplicated and overlapped between the two modules.
* Updated TypeScript bindings

Motivation and Context
---
Up until now, the behaviour of the indexer regarding events and substates was:
* Storing in the database _all_ events in the network, with no possibility of filtering.
* Allowing to store substates in the database via the _user manually adding addresses to a watchlist_ using JRPC methods.

This PR refactors the indexer behaviour to be more useful:
* Allowing the user to **filter which events** in the network we want to store in the database.
* Automatically **storing the related substate of each event** in the database. Thus removing all the address watchlist functionality from the indexer, as we will be using the new event filters to decide what to store.

Regarding the event filtering:
* The list of event filters is static, it can be edited in the **indexer config file**. Any change in event filtering requires a restart of the indexer.
* It allows to specify a list of **event filters.** If a new event matches _at least_ one filter, it will be stored in the database.
* Each filter allows to specify an AND condition of values for `topic`, `substate_id`, `template_address` and `entity_id`. All filter fields are optional. An event will match a filter if all of its fields match the values specified in the filter.
* The `entity_id` one can be used when we want to store everything that happens in a particular EntityId no matter the type of substate (component, vault or resource).
* The default behaviour is an empty filter which matches all events, making it store all events that happen in the network.

Regarding the substate storing:
* We can fetch the substate data from the transaction result (`up_substates` field) corresponding to the event we are storing. This is a handy optimization because it avoids a query to the network to fetch the related substate.
* Because we are constantly fetching new events and extracting the related substate from the transaction results, it means we don't need to keep monitoring for the latest versions of each substate.
* In the database we only store the latest version of each substate.

This PR **removes functionality regarding substate monitoring and address watchlist**, as now the substate storing is determined by the events.

This PR keeps existing functionality regarding:
* Arbitrary substate fetching via the JRPC `get_substate` method. Even if the substate is not stored in the database, the indexer will try to fetch it from the network.
* Listing all NFTs from a collection via the `get_non_fungibles` method. This functionality requires to fetch the non fungibles from the network on the fly, as currently we don't emit the proper events regarding NFT minting. Even if we do in a future, we would need to account for NFT burning (thus deleting the substate from the database in that case)

How Has This Been Tested?
---
Manually by:
* Spawning a local network using `tari_swarm`.
* Using different `event_filters` configurations in the indexer.
* Doing transactions (e.g. transfers) and inspecting which events and substates where stored in the indexer database. 

---
See previous section.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


applications/tari_indexer/src/event_scanner.rs
```diff
@@ -0,0 +1,561 @@
...
+            "scan_events",
+        );
+
+        let mut event_count = 0;
+
+        let current_epoch = self.epoch_manager.current_epoch().await?;
+        // let network_committee_info = self.epoch_manager.get_network_committees().await?;
+        // let epoch = network_committee_info.epoch;
+        let current_committees = self.epoch_manager.get_committees(current_epoch).await?;
+        for (shard, committee) in current_committees {
```
sdbondi:
```suggestion
        for (shard, mut committee) in current_committees {
```

applications/tari_indexer/src/event_scanner.rs
```diff
@@ -0,0 +1,561 @@
...
+            info!(
+                target: LOG_TARGET,
+                "Scanning committee epoch={}, shard={}",
+                current_epoch,
+                shard
+            );
+            // TODO: use the latest block id that we scanned for each committee
+            let new_blocks = self
+                .get_new_blocks_from_committee(shard, &mut committee.clone(), current_epoch)
```
sdbondi: Unnecessary clone - mark `committee` as mutable above
```suggestion
                .get_new_blocks_from_committee(shard, &mut committee, current_epoch)
```

applications/tari_indexer/src/event_scanner.rs
```diff
@@ -0,0 +1,561 @@
...
+        epoch: Epoch,
+    ) -> Result<Vec<Block>, anyhow::Error> {
+        // We start scanning from the last scanned block for this commitee
+        let start_block_id = {
+            let mut tx = self.substate_store.create_read_tx()?;
+            tx.get_last_scanned_block_id(epoch, shard)?
+        };
+        let start_block_id = start_block_id.unwrap_or(self.build_genesis_block_id());
+
+        committee.members.shuffle(&mut OsRng);
```
sdbondi: `Committee` has a shuffle method. In future, we'll want to remove the `pub` from the `members` field
```suggestion
        committee.shuffle();
```

applications/tari_indexer/src/event_scanner.rs
```diff
@@ -0,0 +1,561 @@
...
+                        transaction_ids.insert(*t.id());
+                    },
+                    _ => {
+                        // we are only interested in events from confirmed transactions
+                    },
+                }
+            }
+        }
+
+        transaction_ids
```
sdbondi:
```suggestion
 blocks.iter().flat_map(|b| b.all_accepted_transactions_ids()).copied().collect()
```

applications/tari_indexer/src/event_scanner.rs
```diff
@@ -0,0 +1,561 @@
...
+
+        info!(
+            target: LOG_TARGET,
+            "Scanning new blocks since {} from (epoch={}, shard={})",
+            last_block_id,
+            epoch,
+            shard
+        );
+
+        for (member, _) in &committee.members {
```
sdbondi:
```suggestion
        for member in committee.members() {
```

sdbondi (dismissed):
ACK

Tested and works - few minor comments but can go in as is

sdbondi (approved):

PR #1044, opened 2024-06-14
fix(rpc)!: remove chunking

sdbondi:
Description
---
- remove chunking from RPC protocol

Motivation and Context
---
Chunking does not add any benefit especially when not using tor.
Porting over changes from L1: https://github.com/tari-project/tari/pull/6345

How Has This Been Tested?
---
Manually

---
Should work as before

Breaking Changes
---

- [ ] None
- [x] Other - Please specify

BREAKING CHANGE: Any responses which previously triggered chunking (>= 256kb) would not be readable by clients using this PR


mrnaveira (approved):
Tested and works as expected

PR #1047, opened 2024-06-16
fix(engine)!: uuid duplication on the same EntityId

mrnaveira:
Description
---
* Updated the `new_uuid` implementation to include the `transaction_hash`
* New unit test to reproduce the issue
* Clened the `account_nfts` test code

Motivation and Context
---
The engine implementation of the `new_uuid` function, used when generating a random id (e.g. non fungibles), does cause a duplication id issue on objects generated inside the scope of the same EntityId.

This happens because the `new_uuid` only uses a the `entity_id` and a increasing integer to generate a new random uuid. On separated transactions that generate random uuids in the same EntityId this can end up in a duplication.

This PR adds the `transaction_hash` to to the `new_uuid` generation, avoiding separated transactions to  generate the same uuids even if the EntityId is the same.

How Has This Been Tested?
---
New `engine` unit test `mint_multiple_times` that reproduces the problem. With the fix the new test pass.

---
See previous section

Breaking Changes
---

- [ ] None
- [x] **BREAKING CHANGE**: consensus change on engine uuid generation, so VNs without this change will not reach consensus with VNs that have this change.

### DIFF
--- a/dan_layer/engine/tests/account_nfts.rs
+++ b/dan_layer/engine/tests/account_nfts.rs
@@ -1,10 +1,10 @@
 //   Copyright 2023 The Tari Project
 //   SPDX-License-Identifier: BSD-3-Clause
 
-use tari_engine_types::instruction::Instruction;
+use tari_engine_types::{commit_result::ExecuteResult, instruction::Instruction};
 use tari_template_lib::{
     args,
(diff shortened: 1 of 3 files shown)


sdbondi (approved):
utACK 

Thanks

PR #1049, opened 2024-06-18
feat(indexer): new JRPC method to list substates

mrnaveira:
Description
---
* Added new columns `template_address`, `module_name` and `timestamp` to the `substates` SQLite table.
* Added new column `timestamp` to the `events` SQLite table. For now it's populated but not used by any method yet.
* The `event_scanner` fetches and stores in database the values for all the new fields in substates and events
* New JPRC method `list_substates`
* Updated TypeScript bindings related with the new indexer method
* Removed old logic for event storing in the `event_manager`, it's all now done only on the `event_scanner` and controlled by the event filter in the indexer config

Motivation and Context
---
On https://github.com/tari-project/tari-dan/pull/1043 we refactored the Indexer to automatically store all substates related to events. This PR follows on that work and adds a new JRPC method `list_substates` that allows to fetch and filter all the substates stored in the Indexer database. It also adds some metadata (`template_address`, `module_name`, `timestamp`) to substates.

The new method JRPC method `list_substates` has the following _optional_ parameters:
* **filter_by_template**: String with the template address of the components that we want to filter
* **filter_by_type**: String with the only type of substate that we want. Possible vaules are `"Component"`, `"Resource"`, `"Vault"`, `"UnclaimedConfidentialOutput"`, `"NonFungible",` `"TransactionReceipt"` and `"FeeClaim"` 
* **limit**: u64 value for the maximum number of results that we want (intended for pagination)
* **offset**: u64 with the position of the first result that we want (intended for pagination)

How Has This Been Tested?
---
Manually by:
* Spawning a local network using `tari_swarm`
* Performing some transactions to generate events and substates
* Inspecting the Indexer database and also querying the new `list_substate` method 

---
See previous section

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi:
LGTM - looks like a graphQL test is missing the timestamp field or it should be optional

sdbondi (approved):
utACK

PR #1057, opened 2024-06-20
fix(indexer): reuse validator RPC sessions

sdbondi:
Description
---
caches and reuses validator RPC sessions

Motivation and Context
---
When submitting many transactions through the indexer (using tariswap test) new RPC sessions would hit the session limit. This PR keeps hold of RPC sessions and reuses them (requests are queued) instead of creating new sessions.

How Has This Been Tested?
---
Tariswap test works.

---
Tariswap test (`utilities/tariswap_test_bench`)

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/networking/rpc_framework/src/lib.rs
+++ b/networking/rpc_framework/src/lib.rs
@@ -30,7 +30,7 @@
 
 /// Maximum frame size of each RPC message. This is enforced in tokio's length delimited codec.
 /// This can be thought of as the hard limit on message size.
-pub const RPC_MAX_FRAME_SIZE: usize = 3 * 1024 * 1024; // 3 MiB
+pub const RPC_MAX_FRAME_SIZE: usize = 6 * 1024 * 1024; // 3 MiB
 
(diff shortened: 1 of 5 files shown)


networking/rpc_framework/src/lib.rs
```diff
@@ -30,7 +30,7 @@
 
 /// Maximum frame size of each RPC message. This is enforced in tokio's length delimited codec.
 /// This can be thought of as the hard limit on message size.
-pub const RPC_MAX_FRAME_SIZE: usize = 3 * 1024 * 1024; // 3 MiB
+pub const RPC_MAX_FRAME_SIZE: usize = 6 * 1024 * 1024; // 3 MiB
```
mrnaveira:
```suggestion
pub const RPC_MAX_FRAME_SIZE: usize = 6 * 1024 * 1024; // 6 MiB
```

mrnaveira (dismissed):

PR #1060, opened 2024-06-23
refactor(wallet_daemon): use the indexer for substate listing

mrnaveira:
Description
---
* Unified the `SubstateType` enum (duplicated in both the wallet daemon and the indexer) into the `tari_dan_common_types` crate
* Updated the indexer JRPC client to include the `list_substates` method
* Updated the `substates.list` JRPC method in the wallet daemon to proxy the call to the corresponding method in the indexer. The request type now includes pagination support (offset + limit) but the resulting type (`WalletSubstateRecord`) was kept the same for now to avoid conflicts with existing applications.
* Updated bindings for the substate type and the new pagination options for the wallet daemon substate listing

Motivation and Context
---
On https://github.com/tari-project/tari-dan/pull/1049 we added a new indexer JRPC method for listing substates. The indexer substate listing is more advanced than the one in the wallet daemon:
* Is capable of indexing all the network while the wallet only stores substates related to its own account's transactions
* Supports pagination (limit + offset fields)

This PR changes the `substates.list` JRPC method in the wallet daemon to proxy the call to the corresponding method in the indexer. It also adds pagination support.

How Has This Been Tested?
---
Manually by:
* Spawning a local network with `tari_swarm`
* Doing transactions to generate substates
* Query the `substates.list` JRPC in the wallet daemon and inspecting the results

---
See previous sections

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (approved):
utACK

PR #1067, opened 2024-07-05
feat!: state checkpoint sync, fixes, epoch change

### LINKED ISSUE #1037: refactor: epoch change sync to checkpoint
## Problem

When a registered validator joins the network for the first time, there is a high bandwidth and processing 
cost which worsens as the network progresses.

Currently, this includes:
A. Syncing the whole state of the shard. Including DOWN substates.
B. Syncing all historical transactions for the shard/s
C. Syncing the entire block history/chain of one or more shards

(A) is unavoidable and should be optimised to reduce time and bandwidth costs. 

(B) Should not be necessary at all. The main reason it is there is to prevent duplicate transactions from being processed. There
also may be some code that expects the transaction to be available (e.g. in web UIs). However, duplicate transactions are 
already prevented by the TransactionReceipt substate. A validator node may want to generate an index of these as it syncs 
to optimise duplicate transaction checking before it fails in the commit phase. Some archival nodes may want to track and store 
historical transactions, but these are separate concerns and consensus/block producer nodes would generally not do this.

(C) This requires more thought. It is important that new validator node knows that they have the complete and agreed shard state 
at or close to the end of the previous epoch. Beyond that, knowing which transactions were historically processed in which 
block by whom is not necessary to proceed with consensus in a new epoch.

## Proposal 1

Start a new chain for each new epoch and verify a checkpoint proof. 

For a validator to join a shard for an epoch, they request a succinct proof that asserts that two thirds of the validator
set for the shard-epoch have committed a given state.

To achieve this the following is done:
1. During the previous epoch, they contact one or more validators in the shard-epoch, request and verify a checkpoint proof signatures.
2. Download the state to the checkpoint.
3. Validate the state hash against the checkpoint proof.
4. The previous epoch chain has likely progressed over this time. Rince and repeat as necessary until the end of epoch commands are observed

### Checkpoint Proof

_The exact construction of the proof needs to be thought about more_

A shard-epoch checkpoint proof for the current commit block can be generated at any point by a 
participating validator node. The proof contains the jellyfish Merkle root of the state as 
well as the last 3 linked QCs of the block that is represented in the proof. 

For example, the current tip block is 300, the proof contains the _linked_ QCs of 299, 298, and 297.
This proves that block 297 was committed by all non-faulty nodes.

### LINKED ISSUE #1070: bug: dangling bucket errors crash the VNs

sdbondi:
Description
---

feat: adds state sync
fix: remove block sync
feat: start new chain at each epoch
fix: fixes bug that drops consensus messages (inbound messaging not cancel safe)
fix: scope JMT to epoch/shard
fix: only emit epoch changed event once scanning to tip has completed
fix: swarm mines more initial blocks when registering many vns
fix: bug that caused block signature to disappear after unparking the block
feat: adds state checkpoint call
test: fix epoch change and state sync cucumbers

Motivation and Context
---

State sync fetches state transitions and applies them to the local state db. Syncing historical blocks is not feasible across changing shards/epochs and this is removed in this PR.

Closes #1070
Closes #1037 

Outstanding items:
- verify checkpoint on sync
- construct a "tiered" state merkle root consisting of presharded JMT merkle roots applicable to the current shard space.
- leader failure due to race condition when proposal for end of epoch is received before all validators have scanned to the base layer tip  

How Has This Been Tested?
---
Manually: 
1. 2 validators, mine to next epoch
2. 2 validators, send transactions, delete data of one vn and resync
3. 4 validators, delete data from one validator, leader failures, epoch change, resync validator
4. [TODO] new validator joining network
5. [TODO] multishard test 

---
Swarm, mine to next epoch

Breaking Changes
---

- [ ] None
- [x] Other - Please specify


mrnaveira (approved):
LGTM

stringhandler (approved):
tested locally

PR #1069, opened 2024-07-08
chore!: renaming and hasher documentation

AaronFeickert:
Description
---
Updates the documentation for `TariHasher`. Does some renaming relating to hasher outputs.

Motivation and Context
---
The `TariHasher` type has minimal documentation. Because it uses CBOR encoding under the hood to help ensure canonical hasher input, it may be useful to describe its design intent and requirements on input data. This PR adds such documentation.

Separately, one use of the hasher functionality is to [produce vote signature messages](https://github.com/tari-project/tari-dan/blob/9db28d44b9bc328d8523e2665dab4f94e8d53813/dan_layer/consensus/src/traits/signing_service.rs#L21-L25). The naming used suggests this value is a signature challenge, which is not the case. This PR does some renaming to clarify this.

How Has This Been Tested?
---
Existing tests pass.

---
Confirm that the updated documentation is accurate, and that the renaming was done correctly.

BREAKING CHANGE: Changes the `VoteSignatureService` public trait API via a rename.

### DIFF
--- a/applications/tari_validator_node/src/consensus/signature_service.rs
+++ b/applications/tari_validator_node/src/consensus/signature_service.rs
@@ -36,7 +36,7 @@ impl VoteSignatureService for TariSignatureService {
         block_id: &BlockId,
         decision: &QuorumDecision,
     ) -> bool {
-        let challenge = self.create_challenge(leaf_hash, block_id, decision);
-        signature.verify(challenge)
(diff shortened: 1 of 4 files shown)


sdbondi (approved):
Thanks +1

sdbondi (approved):
Thanks +1

PR #1082, opened 2024-07-16
refactor!: remove mempool execute, add faucet component, remove create free test coins

sdbondi:
Description
---
- remove duplicate and incorrect state bootstrapping code in engine
- remove CreateFreeTestCoins instruction
- remove mempool transaction validation
- add validation to consensus-level transaction executor
- remove deferred decision (decision is now made until after execution and locking)
- fixes for local-only-rule concurrent execution
- fix(engine): fix incorrectly marking components as changed when no state change occurred
- adds XtrFaucet builtin template 
- create funded XtrFaucet component in testnet bootstrap 
- update walletd to call XtrFaucet for free test coins
- fix(consensus): only accept missing transactions if they were requested
- fix(transaction-submitter): ignore race condition when retrieving tx results

Motivation and Context
---
CreateFreeTestCoins was a quick way to get test funds. However because it is the only valid transaction that required no inputs, a number of edge cases had to be handled. This PR replaces this with a faucet component that is bootstrapped with funds. This component is only created for non-mainnet networks.

The mempool execution has a number of limitations e.g. cannot handle versionless inputs.

Next steps:
- fee claiming does not currently work. This is because we need a bunch of information from the L1/epoch manager to authorize a claim. We may need to rethink this e.g. at the end of epoch, we create/update an UnclaimedVnFee substate which can later be claimed/swept by validators. The "minting" of the validator fee substate may require co-operation from multiple shards.
- multishard transactions only previously worked for versioned inputs. After this PR they will not work at all because they require a substate pledging protocol. This will be implemented in a subsequent PR.
- there is a small chance that a substate may change shards as the version changes. This chance needs to be completely removed. The easiest way would be to simply ignore the last 8 (u32) bytes of a substate address when mapping to a shard, however that would leave many tiny "holes" in the shard space where state cannot exist.

How Has This Been Tested?
---
Existing tests.
Manually: 5000 transaction stress test on single shard network

---
Run a stress test (remember to regenerate stress test transaction bin file with new XtrFaucet transaction)
Observe that local-only transactions that share the XtrFaucet vault are executed in the same block 

Breaking Changes
---

- [ ] None
- [x] Other - CreateFreeTestCoins is no longer valid, so all clients that submit these will have to change to use the XtrFaucet


stringhandler (approved):

stringhandler (approved):
Nice. utACK

PR #1083, opened 2024-07-17
feat(tari_wallet_daemon): walletconnect support

mrnaveira:
Description
---
Adds support for WalletConnect in the `tari_wallet_daemon_ui`:
* Adds new envvar `VITE_WALLET_CONNECT_PROJECT_ID`
* New button to accept WalletConnect URIs
* Implements the WalletConnect communication logic using the official JavaScript Wallet SDK
* Requests are being forwarded to the wallet daemon

Motivation and Context
---
We want to support WalletConnect in Tari, as this is a de facto standard for web-to-wallet interaction. It will allow any wallet (web or mobile) to add support for Tari without us having to implement anything extra.

In this context, the goal of this PR is to add support for WalletConnect in the Tari Wallet Daemon alongside the exising TariConnector.

For now we have separated UI buttons for both TariConnector and WalletConnect, due do some TariConnect logic being entangled into the UI. It should be possible to refactor to only have one generic button that accepts both types of connection URIs but it was left out of the scope of this PR.

See https://github.com/tari-project/tari.js/pull/27 and https://github.com/tari-project/tari-template-web/pull/6 for more information. 

How Has This Been Tested?
---
Manually by running `tari_swarm`, opening the tari wallet daemon web UI and then paste a valid WalletConnect URI

---
See previous section

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (dismissed):
ACK

Didnt test the wallet connect functionality but confirmed the new functionality doesn't break the wallet web ui

sdbondi (approved):

PR #1086, opened 2024-07-22
chore!: update to use latest feature-dan2

sdbondi:
Description
---

Updates dan to use latest feature-dan2 branch
Skips balance proof signature check for revealed only transfer
Refactored confidential output proof to formalize a reveal only transfer

Motivation and Context
---
Requires https://github.com/tari-project/tari/pull/6420

In the updated version of tari crypto, Schnorr signature verification correctly explicitly rejects a (previously valid) signature signed by a zero key. For our use case, we want to allow this so that a template can easily construct a withdraw proof for revealed funds without needing to sign with the secret (which should never go into template land). 

The wallet would create a proof with a zero excess in some cases when paying fees and/or with transfers with only revealed funds. This PR fixes the wallet to never create a signature with a zero key in these cases.

Breaking Changes
---

- [ ] None
- [x] Other - Please specify

BREAKING CHANGE: many breaking changes in the layer one code. Confidential output encrypted data not backward compatible.


therealdannzor (approved):

leet4tari (dismissed):
utACK - ubuntu updated deps look inline with the tari-suite project.

PR #1092, opened 2024-07-25
feat(consensus)!: implement preshards and shard groups

sdbondi:
Description
---
[feat!(consensus): implement preshards and shard groups](https://github.com/tari-project/tari-dan/pull/1092/commits/6ff275cb3622e24cf3f91b2bbd16d976644880ae)
Implement JMT for each shard.

Motivation and Context
---
Divide the shard space into a number of preshards. Each VN committee is responsible for a group (1 or more) preshards. 

How Has This Been Tested?
---
Existing tests, updated Substate address unit tests, manually

---
Look at the shard group in the database, and check the shared `state_tree` table

Breaking Changes
---

- [ ] None
- [ ] Other - Please specify

BREAKING CHANGE: sharding model changed and not backward compatible


applications/tari_indexer/src/substate_storage_sqlite/schema.rs
```diff
@@ -36,23 +23,42 @@ diesel::table! {
...
 
+diesel::table! {
+    substates (id) {
+        id -> Integer,
+        address -> Text,
+        version -> BigInt,
+        data -> Text,
+        tx_hash -> Text,
+        template_address -> Nullable<Text>,
+        module_name -> Nullable<Text>,
```
therealdannzor, line 51: q: what are these two?

dan_layer/common_types/src/substate_address.rs
```diff
@@ -140,65 +133,90 @@ impl SubstateAddress {
...
-            next_pow_2_shard
-        };
-
-        // u256::MAX address needs to be clamped to the last shard index
-        cmp::min(shard, num_shards - 1).into()
-    }
+            end = start + shards_per_committee;
+            if shards_per_committee_rem > 0 {
+                end += 1;
+            }
```
therealdannzor, line 216: i'm reading this regarding the committee: `the target number of committees to manage the shard space as determined by the number of registered validators for an epoch`, and that `num_committees <= num_shards`.

does this algo try to spread out some shards per committee if there is a remainder, i.e. if `shards_per_committee % num_shards != 0`?

dan_layer/consensus/src/hotstuff/on_ready_to_vote_on_local_block.rs
```diff
@@ -191,8 +195,8 @@ where TConsensusSpec: ConsensusSpec
 
         // Store used for transactions that have inputs without specific versions.
         // It lives through the entire block so multiple transactions can be sequenced together in the same block
-        let tree_store = ChainScopedTreeStore::new(block.epoch(), block.shard(), tx);
-        let mut substate_store = PendingSubstateStore::new(*block.parent(), tree_store);
+        // let tree_store = ChainScopedTreeStore::new(block.epoch(), block.shard_group(), tx);
```
therealdannzor: q: why is the (jm?) tree store not needed anymore with shard groups?

dan_layer/consensus_tests/src/support/helpers.rs
```diff
@@ -1,27 +1,50 @@
...
     let rand_bytes = OsRng.gen::<[u8; ComponentKey::LENGTH]>();
     let component_key = ComponentKey::new(copy_fixed(&rand_bytes));
     let substate_id = SubstateId::Component(ComponentAddress::new(ObjectKey::new(entity_id, component_key)));
     VersionedSubstateId::new(substate_id, 0)
 }
 
+fn random_substate_address_range<R: RangeBounds<SubstateAddress>>(range: R) -> SubstateAddress {
+    let start = match range.start_bound() {
+        std::ops::Bound::Included(addr) => addr.to_u256(),
+        std::ops::Bound::Excluded(addr) => addr.to_u256() + 1,
```
therealdannzor, line 34: i'm unfamiliar with this type, but if a bound `(0..6)` is excluding, you continue at `(7..)`? but this is a new shard group no?

dan_layer/rpc_state_sync/src/manager.rs
```diff
@@ -187,14 +188,15 @@ where TConsensusSpec: ConsensusSpec<Addr = PeerAddress>
                     NodeHeight(0),
                     *checkpoint.block().id(),
                     substate.created_by_transaction,
-                    *created_qc.id(),
+                    // TODO: correct QC ID
+                    QcId::zero(),
```
therealdannzor, line 279: what does quorum certificates do here, and why isn't it needed now? i thought this concerns only consensus

dan_layer/storage/src/consensus_models/block.rs
```diff
@@ -766,20 +773,20 @@ impl Block {
...
-                            created_qc: substate.get_created_quorum_certificate(tx)?,
+                            // created_qc: substate.get_created_quorum_certificate(tx)?,
                             substate: substate.into(),
                         }));
                     } else {
                         updates.push(SubstateUpdate::Destroy(SubstateDestroyedProof {
                             substate_id: substate.substate_id.clone(),
                             version: substate.version,
-                            justify: QuorumCertificate::get(tx, &destroyed.justify)?,
+                            // justify: QuorumCertificate::get(tx, &destroyed.justify)?,
```
therealdannzor, line 783: is this ok to do?

therealdannzor (commented):
looks good to me, just a couple of questions to better understand pre-sharding

dan_layer/common_types/src/substate_address.rs
sdbondi: Yeah that's right, it essentially adds one shard to each initial shard group until there is no remainder. I initially had an implementation with only calculations and no loop but I was unable to get the correct result in all cases, so I went for the direct approach.

dan_layer/consensus/src/hotstuff/on_ready_to_vote_on_local_block.rs
sdbondi: I decoupled the tree calculations from the substate store since I needed to implement a per shard JMT and I wanted that logic to be separate from the substate store. Basically, I looked at it and it didn't make sense to me for it to be in there

applications/tari_indexer/src/substate_storage_sqlite/schema.rs
sdbondi: These are only for debugging purposes in SQLite, so would be omitted in an e.g. RocksDB backend. These are only set for a component substate and refer to the template and module name of that template that created the component.

dan_layer/consensus_tests/src/support/helpers.rs
sdbondi: The start range bound cant actually be Excluded AFAIK anyway, but if the bound is `(0..6)` the inclusive bound would be `[1..5]` (note the start bound is `0 + 1`.)

therealdannzor (approved):

dan_layer/rpc_state_sync/src/manager.rs
sdbondi: The QC that justified a substate is probably just for debugging purposes in general. It was set before to the QC which justifies the block before it, so was not correct anyhow. The idea for future is that we download state and build up a merkle tree per shard. Check the final state against the checkpoint which contains a proof that the block containing the state hash was committed.

therealdannzor: ok, I guess this answers the question below in `consensus_models/block.rs` too

dan_layer/storage/src/consensus_models/block.rs
sdbondi: This is only used when calling RPC block sync, which has been replaced by state sync. The indexer still uses block sync however, but doesn't validate the justify anyway. This part needs to be rethought so easier to leave this out.

dan_layer/common_types/src/num_preshards.rs
```diff
@@ -0,0 +1,78 @@
...
+pub enum NumPreshards {
+    One = 1,
+    Two = 2,
+    Four = 4,
+    Eight = 8,
+    Sixteen = 16,
+    ThirtyTwo = 32,
+    SixtyFour = 64,
+    OneTwentyEight = 128,
+    TwoFiftySix = 256,
```
mrnaveira, line 23: A bit of a nitpick, but I think using the actual numbers on the enum names (e.g. `P128`, `P256`) is more readable in this case

mrnaveira (approved):

dan_layer/common_types/src/num_preshards.rs
sdbondi: Agree with that - Ok if I rename them in the next PR?

PR #1094, opened 2024-07-26
fix(docs): tari_template_lib resource builders not showing on rust docs

mrnaveira:
Description
---
Set to `pub` the visibility of resource builder modules in `tari_template_lib`

Motivation and Context
---
Currently `FungibleResourceBuilder`, `NonFungibleResourceBuilder` and `ConfidentialResourceBuilder` don't show up in `tari_template_lib` docs.

This happened because the visibility of the modules containing those structs was not public, so this PR sets the visibility to public.

How Has This Been Tested?
---
Manually by running `cargo doc --open` on the `tari_template_lib` crate

---
See previous section

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (approved):
Thanks

PR #1097, opened 2024-07-29
feat(consensus): optionally configure max block time

### LINKED ISSUE #1088: [consensus] Optionally specify maximum block time in pacemaker
The pacemaker has a hard coded value for "base" block time of 10s. e.g during GST, a leader failure will occur after 10 + 2^1 seconds.

The 10s block time should be configurable in consensus constants for the network

therealdannzor:
Description
---
Update consensus constants to include the (maximum) base block time, a component of the GST before a leader fails.

Resolves #1088 

Motivation and Context
---

Currently it is hard coded to 10 seconds. This will make it configurable.

How Has This Been Tested?
---

Unit and integration tests.

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_dan_app_utilities/src/consensus_constants.rs
+++ b/applications/tari_dan_app_utilities/src/consensus_constants.rs
@@ -26,7 +26,7 @@ pub struct ConsensusConstants {
     pub committee_size: u32,
     pub max_base_layer_blocks_ahead: u64,
     pub max_base_layer_blocks_behind: u64,
-    pub max_block_time_threshold: u64, // GST in seconds for consensus
+    pub pacemaker_max_base_time: u64, // GST in seconds for consensus
 }
 
 impl ConsensusConstants {
@@ -36,7 +36,7 @@ impl ConsensusConstants {
             committee_size: 7,
             max_base_layer_blocks_ahead: 5,
             max_base_layer_blocks_behind: 5,
-            max_block_time_threshold: 10,
+            pacemaker_max_base_time: 10,
         }
     }
 }
--- a/applications/tari_dan_app_utilities/src/consensus_constants.rs
+++ b/applications/tari_dan_app_utilities/src/consensus_constants.rs
@@ -26,7 +26,7 @@ pub struct ConsensusConstants {
     pub committee_size: u32,
     pub max_base_layer_blocks_ahead: u64,
     pub max_base_layer_blocks_behind: u64,
-    pub pacemaker_max_base_time: u64, // GST in seconds for consensus
+    pub pacemaker_max_base_time: std::time::Duration,
 }
 
 impl ConsensusConstants {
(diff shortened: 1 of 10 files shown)


applications/tari_dan_app_utilities/src/consensus_constants.rs
```diff
@@ -35,6 +36,7 @@ impl ConsensusConstants {
             committee_size: 7,
             max_base_layer_blocks_ahead: 5,
             max_base_layer_blocks_behind: 5,
+            max_block_time_threshold: 10,
```
sdbondi: Can make this a `Duration` type
I like `pacemaker_max_base_block_time` as the name, but pacemaker prefix can be left off if you prefer. `_base_` is there to clarify that the actual block time/leader timeout is always more than this. But open to suggestions or leaving it off.
```suggestion
            pacemaker_max_base_block_time: Duration, 
```

dan_layer/consensus/src/hotstuff/pacemaker.rs
```diff
@@ -20,24 +21,31 @@ use crate::hotstuff::{
...
         let on_force_beat = OnForceBeat::new();
         let on_leader_timeout = OnLeaderTimeout::new();
         let current_height = CurrentView::new();
 
+        let block_time = if max_block_time_threshold == 0 {
+            DEFAULT_BLOCK_TIME
+        } else {
+            Duration::from_secs(max_block_time_threshold)
+        };
+
```
sdbondi: If set to Duration::from_secs(0), it can stay like that (yeah it could cause problems but then again it may be unexpected that it changes to 10 by itself, maybe we want to set 0 i.e. 2 sec block time). The default in the consensus constants will already be set to 10 by default if it is omitted.

sdbondi (commented):
Looking good so far, some very minor comments

applications/tari_dan_app_utilities/src/consensus_constants.rs
therealdannzor: Maybe `pacemaker` prefix is a good thing to help clarify where this time goes? `base` also makes sense since this is not the actual threshold, good spot.

sdbondi (approved):

applications/tari_dan_app_utilities/src/consensus_constants.rs
```diff
@@ -26,6 +26,7 @@ pub struct ConsensusConstants {
     pub committee_size: u32,
     pub max_base_layer_blocks_ahead: u64,
     pub max_base_layer_blocks_behind: u64,
+    pub pacemaker_max_base_time: std::time::Duration,
```
sdbondi, line 29: nit: `use std::time::Duration`

sdbondi (approved):

PR #1109, opened 2024-08-05
fix(engine): finalize errors (e.g. dangling bucket) reject the main transaction only

mrnaveira:
Description
---
Extracted the `validate_finalized` check invocation from the `StateTracker` into the `TransactionProcessor`, to be done automatically at the end of all `instructions` and `fee_instructions`.

Motivation and Context
---
Currently, when the main instruction set has a dangling bucket (or another finalizing error), the whole transaction is rejected. What we want instead is only the main transaction being rejected, and the fees are still paid (`AcceptFeeRejectRest`)

How Has This Been Tested?
---
New unit test (`dangling_bucket_pay_fees`) in the engine to reproduce this scenario

---
Reproduce a dangling bucket error and check that the fees are paid

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (approved):
utACK

Nice!

PR #1113, opened 2024-08-06
fix(bindings): building package

MCozhusheck:
Description
---
[Tari.js](https://github.com/tari-project/tari.js) is depending on this package but after inspecting it's dependencies I found issues with bindings package like incorrect package.json deceleration and project structure.

Motivation and Context
---
This PR is necessary to also fix Tari.js which throws errors in others project depending on it with error message that Tari.js is missing types declarations.

How Has This Been Tested?
---
Locally build bindings and imported it to local Tari.js and after that imported Tari.js to the [Tari-Universe](https://github.com/tari-project/tari-universe) and checked if application was working as expected.

---
Build it locally with `npm run build` command and use it as dependency in other projects.

Breaking Changes
---

- [ ] None
- [x] Other - Please specify

Some types have changed their name to distinguish from each other and everything now it's exported directly from main file `index.ts`  so imports like
```ts
import type { ComponentAddressOrName } from "@tari-project/typescript-bindings/wallet-daemon-client";
```
should be now:
```ts
import type { ComponentAddressOrName } from "@tari-project/typescript-bindings";
```


sdbondi (approved):
Did a quick test this doesnt break any of the WebUIs. LGTM

> I have added all generated files (from npm run build command) to .gitignore file - this should prevent from situations where released package comes from older builds.

Was the gitignore change removed? Just checking if we should continue to run `npm run build` when types change.

Maybe we just need to add a step to build the bindings before running CI - but we can leave that for another PR

MCozhusheck:
> Did a quick test this doesnt break any of the WebUIs. LGTM
> 
> > I have added all generated files (from npm run build command) to .gitignore file - this should prevent from situations where released package comes from older builds.
> 
> Was the gitignore change removed? Just checking if we should continue to run `npm run build` when types change.
> 
> Maybe we just need to add a step to build the bindings before running CI - but we can leave that for another PR

Yes I have reverted changes to the `.gitignore` and github action workflow so we have more control when we are building type bindings

mrnaveira (approved):

PR #1115, opened 2024-08-09
fix(bindings): remove dist from gitignore to expose build project

MCozhusheck:
Description
---
Last PR broke last [@tari-project/typescript-bindings](https://www.npmjs.com/package/@tari-project/typescript-bindings?activeTab=readme) version. package.json points entrypoint as `./dist/index.js` but since CI doesn't builds project every run (as it is costly) the package doesn't have `dist` folder and effectively we don't export anything. 

Motivation and Context
---
Fix package for other projects like tari.js which relies on typescript-bindings.

How Has This Been Tested?
---
Imported locally build typescript-bindings with dist folder present.
Also imported the tari.js to the [Tari Universe](https://github.com/tari-project/tari-universe) and to the [faucet tapplet](https://github.com/MCozhusheck/faucet-tapplet) to check if project depending also on tari.js would work.

---
Eg. change typescript-bindings dependency for tari.js to local (in my case it's `"@tari-project/typescript-bindings": "file:../tari-dan/bindings",`) and check if it's working

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (dismissed):
I think you need to bump the version to get the package to re-release

sdbondi (approved):

PR #1116, opened 2024-08-09
feat(watcher): start validator node process

therealdannzor:
Description
---
Spawns the the `tari_validator_node` through a tokio `Command` but has not been tested fully with actual config. (It doesn't persist the child or even outlives its scope.)

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


applications/tari_watcher/src/port.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub fn new() -> Self {
+        Self {
+            validator: ValidatorPorts::new(),
+            wallet: MinotariPorts::new(),
+        }
+    }
+
+    pub async fn open_at(&mut self, instance: InstanceType, name: &'static str) -> io::Result<u16> {
+        let addr = TcpListener::bind(SocketAddr::from(([127, 0, 0, 1], 0))).await?;
+        let port = addr.local_addr()?.port();
```
sdbondi: Ports should be configurable by the user. The use cases for the watcher and the swarm are very different.
The swarm is for development ("As a developer, I want to start up a DAN network so that I can test my changes"), the watcher should be a tool that validator node operators can use ("As a node operator, I want to ensure my VN is registered and running, so that I can earn fees").

The main watcher feature is taking care of validator registration for you.

RE the ports, an operator may want to specify and expose ports as needed, so should be left up to them how to configure it. They can configure it by editing the config file. We could think about adding a feature down the line to allow the watcher to configure things, but for now we can keep this dead simple and have the validator node process use the configured ports in it's config - the watcher simply has to start the process and monitor it.

applications/tari_watcher/src/forker.rs
```diff
@@ -0,0 +1,190 @@
...
+    process::{Child, Command},
+};
+
+use crate::{
+    config::{ExecutableConfig, InstanceType},
+    port::PortAllocator,
+};
+
+const TARI_L2_VN: InstanceType = InstanceType::TariValidatorNode;
+const MINO_L1_WALLET: InstanceType = InstanceType::MinoTariConsoleWallet;
```
sdbondi: nit: rather avoid the extra indirection of a const (i.e. I have to look at what TARI_L2_VN is). An "non-struct" enum variant is already in essence a const

applications/tari_watcher/src/forker.rs
```diff
@@ -0,0 +1,190 @@
...
+
+        let listen = IpAddr::V4(Ipv4Addr::from([127, 0, 0, 1]));
+        self.listen_ip = Some(listen);
+
+        match self.app {
+            TARI_L2_VN => {
+                self.port.open_at(TARI_L2_VN, "jrpc").await?;
+                self.port.open_at(TARI_L2_VN, "web").await?;
+            },
+            MINO_L1_WALLET => {
```
sdbondi: There is no need to monitor the minotari wallet process. 
Either, we should assume the minotari wallet is running with gRPC call to register a VN is available. The watcher simply needs to have the gRPC address to call. We can 100% do this if it's easier. 

Or optionally, we could run a command to send the registration [see here](https://github.com/tari-project/tari/blob/35573840c3b40f7e87b4021f8bed754be8e859cf/applications/minotari_console_wallet/src/cli.rs)

The watcher would not monitor the process but simply execute the command when we need to register.

Happy for either approach.

applications/tari_watcher/src/port.rs
therealdannzor: Thanks for the clarification. Would you have me add the additional fields to the config for the validator node: `vn_json_rpc_port` and `vn_http_port`? Since we don't need to monitor the wallet.

applications/tari_watcher/src/forker.rs
therealdannzor: removed

applications/tari_watcher/src/forker.rs
therealdannzor: > The watcher simply needs to have the gRPC address to call. We can 100% do this if it's easier.

Sounds like this is the easier approach for now and we can iterate later if needed

applications/tari_watcher/src/port.rs
sdbondi: These ports are already configurable in the VN config file, so I think we can leave it to the VN code.

applications/tari_watcher/src/port.rs
```diff
@@ -1,58 +1,76 @@
...
-            InstanceType::MinoTariConsoleWallet => {
-                if name == "p2p" {
-                    self.wallet.p2p = Some(port)
-                } else if name == "grpc" {
-                    self.wallet.grpc = Some(port)
-                } else {
-                    log::error!("Invalid port name for {} instance: {}", instance, port);
-                }
-            },
+    fn vn_json_rpc_port(&self) -> Option<u32> {
```
sdbondi: Is there a reason the watcher needs to know the VN's JSON-RPC address?

I don't want the user to have to set some environment vars to setup/start the watcher. 

The watcher doesn't need to talk to the VN's JSON RPC to register it. We only need to read the registration.json file.

The approach I had in mind for determining whether the VN needs to be re-registered was that the watcher periodically polls the L1 minotari node gRPC for the validator node registration (see Registration Management in https://github.com/tari-project/tari-dan/issues/1112). The advantage here is that these calls already exist.

Another approach may be to poll the VN JSON-RPC to check L1 chain progression since VN's does already know this. Then the watcher wouldn't need to talk to the minotari node at all.  Happy to do that, though we'd need to add a new call or modify get_epoch_manager_stats to give the expiry epoch and current block height to allow this. We'd also need to know what port the VN is using for JSON RPC which will be tricky unless we set it in the watcher. 

There are a few ways to do this. Open to thoughts.

applications/tari_watcher/src/port.rs
sdbondi: For now the watcher can simply start a validator node e.g `/path/to/tari_validator_node -c path/to/config.toml`. It does not need to know the JSON-RPC or web UI ports.

applications/tari_watcher/src/port.rs
therealdannzor: Likely makes sense to do it with the same pattern for now, unless we otherwise know a clearly better approach.

It is not needed [for watcher to know VN json rpc address], I was just not certain exactly where we needed what data at what point. Thanks for clarifying earlier.

applications/tari_watcher/src/port.rs
therealdannzor: I will remove the env stuff, I was unsure where it would get the port data from. We can work under the assumption now that it already knows it then, as you mentioned earlier.

applications/tari_watcher/src/forker.rs
```diff
@@ -44,25 +39,19 @@ impl Forker {
...
-        let web_ui_address = instance.port.validator.web_ui_address(instance.listen_ip.unwrap());
-
         let mut command = self
             .get_command(
                 config.executable_path.unwrap(),
                 "esmeralda".to_string(), // TODO: add network to cfg
                 base_node_grpc_address,
-                json_rpc_public_address.unwrap(),
-                web_ui_address.unwrap(),
+                vn_public_json_rpc_address,
```
sdbondi: The watcher doesn't need the VN JSON rpc address.

You can put this in place of get_command which isn't really needed. This code comes from the swarm which needs to abstract different processes, watcher only needs to execute one process

```rust
        let mut command = Command::new(config.executable_path.unwrap_or_else(|| Path::new("tari_validator_node"))
// Add host envs
            .envs(env::envs())
// This is technically optional, the VN will default to ~/.tari/{network}/config.toml
            .arg(format!("--config={validator_node_config_path}"))
```

applications/tari_watcher/src/config.rs
```diff
@@ -23,6 +23,12 @@ pub struct Config {
     /// The Minotari console wallet gRPC address
     pub base_wallet_grpc_address: String,
 
+    // The L2 validator node JSON RPC address
+    pub vn_public_json_rpc_address: String,
+
+    // The L2 validator node web GUI server address
+    pub vn_gui_http_address: String,
+
```
sdbondi: These are not needed

sdbondi (approved):

PR #1118, opened 2024-08-14
feat(watcher): registration management prototype

therealdannzor:
Description
---
Checks the active set of keys on the network before sending a [RegisterValidatorNode](https://github.com/tari-project/tari/blob/feature-dan2/applications/minotari_app_grpc/proto/wallet.proto#L82) request, and backs off until expected registration expires, to register the node again.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (approved):
utACK - nice

PR #1119, opened 2024-08-15
fix(consensus)!: substate pledges

sdbondi:
Description
---
fix(consensus)!: multi shard substate pledges
Version is not considered when calculating which preshard a substate belongs to
Fix several race conditions in consensus
Adds a test for all local inputs, foreign outputs
Build transactions with inputs in tests
Fixed mempool gossip

Motivation and Context
---

Implements cross-shard exchange and pledges for inputs/outputs.

TODO: 
- output shard group does not sequence an unknown transaction from foreign proposals (PR #1120)
- number of cucumbers scenarios do not include any inputs. This is now completely invalid, so they have been ignored for now until we switch them to use the wallet daemon.

How Has This Been Tested?
---
Manually - multi-shard-group test, existing tests, new tests

---
Submit transactions to a multi-shard-group network

Breaking Changes
---

- [ ] None
- [x] Other - Please specify

BREAKING CHANGE: SubstateAddress has increased in size by 4 bytes, new commands


mrnaveira (approved):
Tested with `tari_swarm`, no issues found.

PR #1121, opened 2024-08-21
fix: indexer event scanning is broken after epoch change

### LINKED ISSUE #1111: bug: indexer event scanning is broken after epoch change PR
```
2024-08-05T13:13:44.593390000Z [Indexer-#0] WARN  Could not get blocks from VN 12D3L7AVAS8oxPHEmrSbCQimAVhRv3TzubU5ieT7g7b2oV8VvVur (epoch=Epoch(52), shard=0): NotFound: start_block_id f4cd8143918f422fa44│ab76da58e54877fb9830f7bf442ba47b67385e85c542b is after locked block LockedBlock(NodeHeight(1), b322ced2553861f289ac5b1b7721b759f61f253aa945ec37b43a656c22b91deb)
2024-08-05T13:13:44.593478000Z [Indexer-#0] WARN  Could not get blocks from any of the VNs of the committee (epoch=Epoch(52), shard=0)
```

Steps:
1. Start swarm
2. submit transactions
3. restart swarm
4. broken

mrnaveira:
Description
---
* In the indexer:
    * Refactored the event scanner to take into account epoch changes
    * New db methods `get_oldest_scanned_epoch` and `delete_scanned_epochs_older_than` to support the event scanning
* In the validator node:
    * Refactored the block sync task to take into account epoch changes and to use the existing `up_to_epoch` parameter (it was ignored previously)
    * New db method `blocks_get_first_in_epoch` to support the new block sync logic

Motivation and Context
---
Recently the epoch change logic was reworked, which broke the event scanning functionality in the indexer. This PR fixes it.

Fixes https://github.com/tari-project/tari-dan/issues/1111

How Has This Been Tested?
---
Manually by running `tari_swarm`, doing transactions, and stopping/resuming the network

---
See previous section

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


dan_layer/state_store_sqlite/src/reader.rs
```diff
@@ -720,6 +720,36 @@ impl<'tx, TAddr: NodeAddressable + Serialize + DeserializeOwned + 'tx> StateStor
         block.try_convert(qc)
     }
 
+    fn blocks_get_first_in_epoch(&self, epoch: Epoch) -> Result<Option<Block>, StorageError> {
```
sdbondi: We typically use the [optional combinator](https://github.com/tari-project/tari-dan/blob/development/dan_layer/common_types/src/optional.rs) instead of returning Option<T>

e.g.
```
// .optional() changes Result<T, StorageError> to Result<Option<T>, StorageError> if a NotFound error is returned
tx.blocks_get_first_in_epoch(epoch).optional()?
```

This both simplifies the implementation a bit and allows the caller to choose if they expect a NotFound error 

```rust
 let row = blocks::table
            .left_join(quorum_certificates::table.on(blocks::qc_id.eq(quorum_certificates::qc_id)))
            .select((blocks::all_columns, quorum_certificates::all_columns.nullable()))
            .filter(blocks::epoch.eq(epoch.as_u64() as i64))
            .filter(blocks::height.eq(0))
            .first::<(sql_models::Block, Option<sql_models::QuorumCertificate>)>(self.connection())
            .map_err(|e| SqliteStorageError::DieselError {
                operation: "blocks_get_first_in_epoch",
                source: e,
            })?;

            let qc = qc.ok_or_else(|| SqliteStorageError::DbInconsistency {
                operation: "blocks_get_first_in_epoch",
                details: format!(
                    "block {} references non-existent quorum certificate {}",
                    block.id, block.qc_id
                ),
            })?;
            let res = block.try_convert(qc)?;
            Ok(res)
```

sdbondi: We could also name this `blocks_get_genesis_for_epoch`

applications/tari_validator_node/src/p2p/rpc/block_sync_task.rs
```diff
@@ -110,11 +110,32 @@ impl<TStateStore: StateStore> BlockSyncTask<TStateStore> {
...
             loop {
-                let children = tx.blocks_get_all_by_parent(&current_block_id)?;
-                let Some(child) = children.into_iter().find(|b| b.is_committed()) else {
+                let current_block = tx.blocks_get(&current_block_id)?;
+
+                // Find the next block in the database
+                let child = if current_block.is_epoch_end() {
+                    // The current block is the last one in the epoch,
+                    // so we need to find the first block in the next expoch
+                    tx.blocks_get_first_in_epoch(current_block.epoch() + Epoch(1))?
```
sdbondi:
```suggestion
                    tx.blocks_get_first_in_epoch(current_block.epoch() + Epoch(1)).optional()?
```

sdbondi (dismissed):
LGTM small nits on the storage API

dan_layer/state_store_sqlite/src/reader.rs
sdbondi: It is possible that the VN is not registered for the next epoch (therefore does not have the next genesis) or doesnt have blocks for the previous epoch(s) (newly synced or we may delete them when the epoch ends). We may have to request historical transaction receipts rather than blocks in future.

sdbondi (approved):
utACK

PR #1122, opened 2024-08-23
feat(watcher): add monitored events to be used by notifier

therealdannzor:
Description
---
Creates the following events, to be used by the notifier subsequently:
- `Exited`: manually cancelling of the child process, returning its status code
- `Crashed`: the child process finished for some reason but was not cancelled manually
- `InternalError`: the watcher received an error when attempting to check the child process
- `Submitted`: the watcher has submitted a registration transaction for the node (tracks `tx_id` and `block`)
- `WarnExpiration`: the watcher emits a warning if the current block is near the expiration block of a registration
- `Running`: the child process is running as expected, do nothing

Picks up the events from the backend and logs them in `monitoring.rs`.

In addition, whenever it spawns a new child process it also saved its process id to a file `validator.pid`. This is updated automatically and can be read from when we want to check whether a node process already exists. Currently, the lifetime of a child is dependent on the parent process.

How Has This Been Tested?
---
Running the `tari_watcher` together with `tari_swarm_daemon`. It then displayed (through logs) the node being registered and then warned when it was near expiration (less than 100 blocks from the expiration block, defined as `registered_at_block + 1000`).


sdbondi (approved):
utACK

Nice, will test but this looks good to merge

PR #1124, opened 2024-08-27
fix!: propose and mint burnt L1 UTXOs in consensus

sdbondi:
Description
---
Proposes burnt UTXOs in local blocks `MintConfidentialOutput` command
swarm: adds the ability to burn funds on L1 wallet to L2 wallet

Motivation and Context
---
Previously UTXOs would be recreated as UP if the VN resynced L1. Validators must agree and commit (JMT) to state changes

How Has This Been Tested?
---
Manually

---
1. Start swarm
2. Create an account
3. Burn funds into the account using swarm
4. Submit a claim instruction from the L2 wallet
5. Profit

Breaking Changes
---

- [ ] None
- [x] Other 

BREAKING CHANGE: new command, changes to JSON serialisation that effect the SQLITE db


therealdannzor (approved):
This looks good to me 👍

PR #1125, opened 2024-08-28
fix(consensus)!: defer proposing txs with conflicting unversioned inputs

sdbondi:
Description
---
fix(consensus)!: defer proposing txs with conflicting unversioned inputs
refactor: locks take into account if the requested input was versioned or not

Motivation and Context
---
A transaction may provide an unversioned input. In the multi-shard case, if this input is locked by another transaction, the first transaction should not be proposed until that lock is cleared.

How Has This Been Tested?
---
Manually, submitting transactions with conflicting unversioned inputs.

---
Submit transactions with transaction_generator and transaction_submitter. The default generated transactions use unversioned inputs.

Breaking Changes
---

- [ ] None
- [ ] Other - Please specify


dan_layer/common_types/src/lock_intent.rs
```diff
@@ -0,0 +1,80 @@
...
+    }
+}
+
+/// Substate lock flags
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[cfg_attr(
+    feature = "ts",
+    derive(ts_rs::TS),
+    ts(export, export_to = "../../bindings/src/types/")
+)]
```
therealdannzor, line 34: i notice [ts-rs](https://docs.rs/ts-rs/latest/ts_rs/trait.TS.html) is used but when do we need to export ts bindings generally?

therealdannzor (approved):
LGTM, just one minor question

dan_layer/common_types/src/lock_intent.rs
sdbondi: Basically when it is part of the public API in any way. For this type it is part of a chain of types that are in the Transaction and Block types. Since you can query for Block and Transactions using JSON-RPC we need ts_rs. Rust will not compile if TS rs is used on a "parent" type and not on a child type so you'll never need to figure that out for yourself. You can also manually define TS types on fields e.g.

```rust
    #[cfg_attr(feature = "ts", ts(type = "string"))]
    pub public_key: PublicKey,
```

PR #1127, opened 2024-08-29
feat(watcher): telegram alerting and automatically persist val node

therealdannzor:
Description
---

* Add Telegram support for alerting on various events
* Add automatic restart of `tari_validator_node` if it crashes and configuration to disable

How Has This Been Tested?
---
Create TG channel and populate `credentials` (bot token) and `channel_id` in the config file. Go through previous steps of starting `tari_swarm_daemon` and have the `tari_watcher` register the node.

Look at the watcher logs for the spawned process pid, or use `ps aux | grep 'tari_validator_node'` and look for the process with the default VN folder, and call `kill -9 $PID`. This induces a crash of the node and if `auto_restart` is enabled, the `tari_watcher` will start a new validator node. It will keep the state of registry since before, as long as the `tari_watcher` does not exit, and send registration transaction as usual.

### DIFF
--- a/applications/tari_watcher/README.md
+++ b/applications/tari_watcher/README.md
@@ -8,7 +8,7 @@
 
 ### Quickstart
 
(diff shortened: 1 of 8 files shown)


sdbondi (approved):
utACK

PR #1128, opened 2024-08-30
fix(watcher): register before expiry and use network consts

therealdannzor:
Description
---
* Send a VN registration transaction one epoch (10 blocks on `esmeralda`) before it is due to expire, rather than reacting to not being in the validator set
* Use consensus constants received from L1 as part of the check rather than using hardcoded ones for the `esmeralda` network to generalize the solution for any network

How Has This Been Tested?
---
1. Run `tari_swarm_daemon` and `tari_watcher`
2. Observe the process sending a registration tx
3. If the registration expires at block $B_E$, observe warning(s) from $B_E - 100$ (default value) and up
4. Mine until any block in  $\left[B_E - 10, B_E - 1\right]$ is reached, and observe the registration tx sent

### DIFF
--- a/applications/tari_watcher/README.md
+++ b/applications/tari_watcher/README.md
@@ -8,9 +8,9 @@
 
 ### Quickstart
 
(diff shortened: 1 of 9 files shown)


sdbondi (approved):
utACK

Will test it out but this can go in as is

PR #1130, opened 2024-09-03
fix(consensus)!: state change on foreign proposals in local blocks

sdbondi:
Description
---
Propose foreign proposals before associated transaction pending state changes occur.
Evidence structure now holds a QC ID for prepare and accept phases for each substate.
Indexer will submit transactions to two validators instead of just one.
Remove unused `should_propagate` option from the mempool.

Motivation and Context
---
Previously foreign proposals would appear after the transaction that they justify.
Foreign proposals should be proposed and the transaction stage transactions included in the change set for the block.

How Has This Been Tested?
---
Existing tests, manually

---
Observe blocks contain foreign proposals that affect LocalPrepare and LocalAccept pledging at or before the actual stage transaction is proposed.

Breaking Changes
---

- [ ] None
- [ ] Other - Please specify


therealdannzor (approved):
LGTM

PR #1135, opened 2024-09-05
feat(consensus)!: add sidechain id in the genesis block

mrnaveira:
Description
---
* Created the definition of a `ExtraData` (following Bitcoin and Ethereum naming) type for storing additional information inside blocks:
    * It contains a key-value map of arbitrary binary data
    * For now only a key exists for sidechain ids, but it can be extended in the future
    * The binary value max size is limited by definition via `MaxSizeBytes` (copied from the minotari codebase)
* Updated the definition for `Block`, `ParkedBlock` and `ForeignProposal` to include an optional `extra_data` field
* Updated the SQLite storage for the new field
* Updated the protobuf definitions with the new field
* Updated the typescript bindings
* Genesis blocks now encode the sidechain id ( if present in the config) in the new `extra_data` field
* New `check_sidechain_id` validation in consensus block validation, to check the sidechain id in the extra data of genesis blocks

Motivation and Context
---
We should have the sidechain id in the genesis block for the shard and epoch. Including it in all blocks would be wasteful, so only genesis blocks should include it.

To make it more general for the future, we want to create an optional field in blocks to store any type of data, and use it to store the sidechain id.

How Has This Been Tested?
---
Manually with `tari_swarm`:
* Setting up a `sidechain_deployment_key` in the `minotari_wallet`
* Settiup up a `validator_node_sidechain_id` in `validator_node`
* Setting up `sidechain_id` in the `indexer`
* Performing transactions to advance the network and inspecting the blocks in the database

Also tested with no sidechain

---
See previous section

Breaking Changes
---

- [ ] None
- [x] Other - Network reset as the block definition and validation has changed


dan_layer/storage/src/consensus_models/block.rs
```diff
@@ -293,12 +317,20 @@ impl Block {
...
     }
 
+    fn extra_data_from_sidechain_id(sidechain_id: Option<RistrettoPublicKey>) -> Result<Option<ExtraData>, BlockError> {
+        let extra_data = sidechain_id
+            .map(|id| ExtraData::new().insert_sidechain_id(id).cloned())
+            .transpose()?;
+        Ok(extra_data)
+    }
+
     pub fn calculate_hash(&self) -> FixedHash {
```
sdbondi, line 334: extra fields needs to go into the hash (BlockId) to prevent malleability and to commit to the genesis sidechain id.

sdbondi (commented):
Looks great - we just need to add the extra_data to the block hash calculation

sdbondi (approved):

PR #1136, opened 2024-09-09
test(cucumber): enable counter template scenario

therealdannzor:


integration_tests/src/wallet_daemon_cli.rs
```diff
@@ -721,3 +767,55 @@ pub async fn get_auth_wallet_daemon_client(world: &TariWorld, wallet_daemon_name
...
+    let wait_req = TransactionWaitResultRequest {
+        transaction_id: submit_resp.transaction_id,
+        timeout_secs: Some(120),
+    };
+    let resp = client
+        .wait_transaction_result(wait_req)
+        .await
+        .unwrap_or_else(|_| panic!("Waiting for the transaction when calling component failed"));
+
+    if let Some(reason) = resp.result.clone().and_then(|finalize| finalize.reject().cloned()) {
```
sdbondi, line 816: nit: can use references

```suggestion
    if let Some(reason) = resp.result.as_ref().and_then(|finalize| finalize.reject()) {
```

sdbondi (approved):
LGTM

integration_tests/src/wallet_daemon_cli.rs
therealdannzor: will use in next PR

PR #1138, opened 2024-09-11
feat!: new account constructor for custom access rules

mrnaveira:
Description
---
* Updated the account template constructor to allow custom owner and access rules
* Updated the `CreateAccount` instruction type
* Updated the transaction builder for the new account creation fields
* Updated the transaction processor to use the new fields
* Updated the protobuf definitions with the new fields in the account creation instruction type

Motivation and Context
---

When creating accounts we want to be able to specify custom access rules. 

This PR refactors the account template constructor (`create` function) to allow specifying the following fields:
* `public_key_token`: mandatory field. Needed always for component creation. If the `owner_rule` is not set, it will be used to set up the default owner rule
* `owner_rule`: optional field. Used to specify custom logic for the account component ownership (e.g. update the access rules after creation)
* `access_rules`: optional field. Used to specify custom access control rules over the account component methods
* `bucket`: optional field. Initial funds of the account

How Has This Been Tested?
---
* New unit test for accounts: `custom_access_rules`
* Manually spawning a network with `tari_spawn` and checking it works
* Existing unit and integration tests pass

---
See previous section

Breaking Changes
---

- [ ] None
- [x] Other - Requires network reset as account creation (including the related instruction type) has changed


sdbondi (approved):
Nice LGTM

PR #1167, opened 2024-10-08
fix: syncing from epoch 0 before consensus has started

sdbondi:
Description
---

Motivation and Context
---
Swarm will sometimes fail to start because syncing from epoch 0 will fail when starting a new network.

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


ksrichard:
LGTM!

PR #1172, opened 2024-10-14
feat: consistent transaction abort reason

ksrichard:
Description
---
There was no exact reason when a transaction was aborted, so added all currently possible abort reasons.

Motivation and Context
---
There was no explicit reason/reason code provided for aborted transactions.

How Has This Been Tested?
---
- Submitted a transaction with invalid versions of inputs:
image

- Submitted 2 concurrent transactions which are causing one an abort error on locking inputs:
image

---

Breaking Changes
---

- [ ] None
- [x] Other - Please specify - Since a new data structure is used it can be inconsistent with older version of code, so a fresh start with deleted data/processes directory would be needed.


mrnaveira (dismissed):

dan_layer/consensus/src/hotstuff/on_propose.rs
```diff
@@ -694,7 +695,7 @@ where TConsensusSpec: ConsensusSpec
                     "⚠️ Transaction is LOCAL-ONLY EARLY ABORT, proposing LocalOnly({}, ABORT)",
                     tx_rec.transaction_id(),
                 );
-                tx_rec.set_local_decision(Decision::Abort);
+                tx_rec.set_local_decision(Decision::Abort(AbortReason::EarlyAbort));
```
sdbondi, line 698: This is due to local inputs not existing or being down
```suggestion
                tx_rec.set_local_decision(Decision::Abort(AbortReason::LockInputsFailed));
```

sdbondi (approved):
ACK

PR #1174, opened 2024-10-16
feat(swarm): auto mining

ksrichard:
Description
---
Added auto mining feature to swarm daemon UI. It needs to be started by hand but afterwards it runs until stopped from UI.

https://github.com/user-attachments/assets/90a00c45-3fef-47c8-a598-7e6ce7415363

Motivation and Context
---
To be able to simulate a real L1 network, we need to have constant mining on that, so we can see epoch changes and see how the network/consensus moves forward.

How Has This Been Tested?
---

https://github.com/user-attachments/assets/90a00c45-3fef-47c8-a598-7e6ce7415363

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (approved):
NIce!

PR #1176, opened 2024-10-17
feat(template_lib)!: new access rules macro

mrnaveira:
Description
---
* Created a new rule! macro in `tari_template_lib` to build `AccessRule` enums
* Updated all usages of access rules construction to use the new macro, including builtin templates (e.g. `account` and `account_nft`)

Motivation and Context
---
We want a better way of constructing `AccessRules`. For example, to restrict a resource to be minted only by a specific component, we need something like:
```
ResourceBuilder::fungible()
                .mintable(AccessRule::Restricted(RestrictedAccessRule::Require(
                    RequireRule::Require(address.into()),
                )))
```

Instead, we want a new macro that allows us to build access rules in simpler and more readable way:
```
ResourceBuilder::fungible()
                .mintable(rule!(component(address)))
```

This PR adds a new macro rule! for this purpose. All variants of rules can be build as follows:
* `rule!(allow_all)`
* `rule!(deny_all)`
* `rule!(component(address))`
* `rule!(resource(address))`
* `rule!(template(address))`
* `rule!(non_fungible(address))`
* `rule!(any_of(component(c_address)), template(t_address)))`

How Has This Been Tested?
---
* New unit test that covers all possible outputs of the new macro
* Existing templates (e.g. account) now use the new macro and all the related tests still pass

---
See previous section

Breaking Changes
---

- [ ] None
- [x] Other - As the `tari_template_lib` crate changes, all the builtin templates (e.g. account) also change.


sdbondi (approved):
Nice utACK

PR #1181, opened 2024-10-21
feat(consensus)!: suspend nodes that fail to participate

sdbondi:
Description
---
- suspend nodes that fail to participate in 3 rounds (configurable)
- resume nodes after voting in 5 rounds 
- refactor and simplify leader failure

Motivation and Context
---
Reduce the impact of a nodes going offline by enabling nodes to agree to suspend the offending nodes from consensus rounds. Once suspended, nodes will immediately send a newview to the next functioning node.

How Has This Been Tested?
---
Unit test, manually 5 nodes taking one offline, observe suspension, go online, wait for resumed node instruction

---
As above

Breaking Changes
---

- [ ] None
- [x] Other - Please specify

BREAKING CHANGE: 2 new block commands added


leet4tari (dismissed):
utACK - Can only say that the changes to the workflow should not break anything.

PR #1182, opened 2024-10-21
feat(validators): trickle in validators

ksrichard:
Description
---
Allow only a configurable amount of validator nodes to be active in an epoch, if we reach the limit, simply adding new validator nodes to a later epoch.

Example:
Let say that the max number of vns to be registered in an epoch is `2` and there are `10` new validator nodes are trying to to join at the same time.

Validator nodes table on other nodes (epochs are relative):
1. epoch: new node 1, start epoch: 1, end epoch: 101
2. epoch: new node 2, start epoch: 1, end epoch: 101
3. epoch: new node 3, start epoch: 2, end epoch: 102
4. epoch: new node 4, start epoch: 2, end epoch: 102
5.  epoch: new node 5, start epoch: 3, end epoch: 103
etc...

Motivation and Context
---
Validators were registered all at once, so when the registration form base layer has been processed, the new validators are available from the next epoch. The problem is that if there are thousands of thousands of validators joining the same time, it would cause the consensus to stuck for a longer period of time until it gets running again healthy.

How Has This Been Tested?
---
Current limit for developer networks is `50`, so setting the number of validator nodes to a higher number than `50` is needed. Then starting up a fresh swarm using this new configuration (delete processes dir) and after the first 50 nodes are up an running, open `data/swarm/processes/validator-node-00/localnet/data/validator_node/global_storage.sqlite` local database and check `validator_nodes` table content.
It will nicely show that the first 50 nodes are going to the next epoch, then the rest in the one after next one.
Then start periodic mining to get to a new epoch and see if everything still runs well with new validator nodes.

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


dan_layer/epoch_manager/src/base_layer/base_layer_epoch_manager.rs
```diff
@@ -20,12 +20,25 @@
...
+    collections::HashMap,
+    mem,
+    num::NonZeroU32,
+    ops::DerefMut,
+    rc::Rc,
+    sync::Arc,
+};
 
-use log::*;
+use log::{__private_api::loc, *};
```
sdbondi:
```suggestion
use log::*;
```

dan_layer/storage_sqlite/src/global/backend_adapter.rs
```diff
@@ -1,24 +1,26 @@
+use std::ops::Add;
```
sdbondi: Below license

dan_layer/epoch_manager/src/base_layer/base_layer_epoch_manager.rs
```diff
@@ -36,16 +49,16 @@ use tari_dan_common_types::{
...
-use tokio::sync::{broadcast, oneshot};
+use tokio::sync::{broadcast, oneshot, Mutex};
 
 use crate::{base_layer::config::EpochManagerConfig, error::EpochManagerError, EpochManagerEvent};
 
 const LOG_TARGET: &str = "tari::dan::epoch_manager::base_layer";
 
 pub struct BaseLayerEpochManager<TGlobalStore, TBaseNodeClient> {
-    global_db: GlobalDb<TGlobalStore>,
+    global_db: Arc<GlobalDb<TGlobalStore>>,
```
sdbondi: Global db already has an inner Arc - what is the reason for using an outer Arc?

dan_layer/epoch_manager/src/base_layer/base_layer_epoch_manager.rs
```diff
@@ -194,11 +230,20 @@ impl<TAddr: NodeAddressable + DerivableFromPublicKey>
...
-        let next_epoch = constants.height_to_epoch(block_height) + Epoch(1);
-        let next_epoch_height = constants.epoch_to_height(next_epoch);
+
+        let constants = self.base_layer_consensus_constants().await?;
+        let mut next_epoch = constants.height_to_epoch(block_height) + Epoch(1);
         let validator_node_expiry = constants.validator_node_registration_expiry;
 
+        // find the next available epoch
+        let mut next_epoch_vn_count = self.validator_nodes_count(next_epoch, registration.sidechain_id())?;
+        while next_epoch_vn_count == self.config.max_vns_per_epoch_activated {
```
sdbondi: What if the count > `self.config.max_vns_per_epoch_activated`?

dan_layer/epoch_manager/src/base_layer/base_layer_epoch_manager.rs
ksrichard: ah sorry it was a leftover, let me update

dan_layer/storage_sqlite/src/global/backend_adapter.rs
ksrichard: its a leftover, just did a cleanup

dan_layer/epoch_manager/src/base_layer/base_layer_epoch_manager.rs
ksrichard: its a leftover, just did a cleanup

dan_layer/epoch_manager/src/base_layer/base_layer_epoch_manager.rs
ksrichard: in theory it should not be greater than maximum, but let me change it to `>=`

sdbondi: Oh I see, this is counting the active set for each epoch.

sdbondi (approved):
ACK didn't test > 50 nodes, but tested with 20 VNs and working as before

PR #1185, opened 2024-10-24
feat(swarm): register templates on startup automatically if not already registered

ksrichard:
Description
---
At startup we check for every registered and active templates, then compare with what we have in templates folder and then we will register automatically all the missing templates. If any new template in the meantime added to the folder and not registered, at the next startup they will be registered automatically as well.
We have a new parameter to be able to pass to `start` command which is `--disable-template-auto-register` which disables the automatic registration of templates from the templates directory (by default this feature is on).

Motivation and Context
---
When we start from a fresh swarm (by deleting `<base_dir>/processes` folder) all the previously registered templates are gone and not registered again. We need to have the previously uploaded and registered templates already in place, so any test or anything can use them.

How Has This Been Tested?
---
### Prepare
  1. Start a swarm locally
  2. Upload a compiled template through swarm daemon UI (http://127.0.0.1:8080/) - I used simply counter example
  3. Stop swarm
  4. delete `<base_dir>/processes` folder (base_dir is mostly `./data`)
  5. Make sure that in the `<base_dir>/templates` folder there is the previously uploaded `.wasm` file (example: `counter-wasm-e8bd952191282de51ecec1bfdcea1260db3c08bf3ff3b3b072484db4da7fe0c2.wasm`)

### Test
  1. Start a swarm locally again
  2. Wait for it to register all nodes and start consensus
  3. Check any of the Validator Node's UI templates list if we have the new templates. (example URL: `http://localhost:12072/templates`)

Also tested the scenarios when I skip registration by passing the param above and checked also when I start multiple times without touching data directory whether the template is registered multiple times (it's not).

---
See test.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (approved):
LGTM - not tested but think it can go in

PR #1208, opened 2024-12-02
feat: new PublishTemplate instruction

ksrichard:
Description
---
As part of https://github.com/tari-project/tari-dan/issues/1207 , this PR contains the basic instruction and skeleton for instruction processing. There is a validation implemented for the new PublishTemplate instruction (where also we validate size of wasm binary).

Please note that in this PR there is no new substates or anything generated yet.

Motivation and Context
---
https://github.com/tari-project/tari-dan/issues/1207

How Has This Been Tested?
---
Send a new transaction to VN:

```rust
    let AccountGetResponse { account, public_key, } = client
        .accounts_get(ComponentAddressOrName::Name("acc"))
        .await?;
    let account_component_address = account.address
        .as_component_address()
        .expect("Failed to get component address");

    // publish wasm template
    let wasm_binary = fs::read("./templates/counter/target/wasm32-unknown-unknown/release/counter.wasm")?;
    let tx = Transaction::builder()
        .fee_transaction_pay_from_component(account_component_address, Amount(1000))
        .publish_template(wasm_binary)
        .build_unsigned_transaction();
    let transaction_submit_req = TransactionSubmitRequest {
        transaction: tx,
        signing_key_index: Some(account.key_index),
        detect_inputs: true,
        detect_inputs_use_unversioned: true,
        proof_ids: vec![],
        autofill_inputs: vec![],
    };
    let resp = client.submit_transaction(transaction_submit_req).await?;
    println!("Submit RESP: {resp:?}");
    let wait_req = TransactionWaitResultRequest {
        transaction_id: resp.transaction_id,
        timeout_secs: Some(120),
    };
    let wait_resp = client.wait_transaction_result(wait_req).await?;
    println!("TX RESP: {wait_resp:?}");
```

Check printed out result and transaction result on UIs.

Examples:

- Invalid wasm template binary
image

- Size exceeding wasm template (manually set a lower limit to check error)
image

- Success
image

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


dan_layer/engine/Cargo.toml
```diff
@@ -19,6 +19,7 @@ tari_template_builtin = { workspace = true }
 tari_template_lib = { workspace = true }
 tari_utilities = { workspace = true }
 tari_transaction = { workspace = true }
+tari_consensus = { workspace = true }
```
sdbondi: Consensus and execution must remain decoupled. 

I see you're using consensus constants to dictate the maximum permitted size of a template. 

I suggest creating a `TransactionProcessorConfig` struct that you pass into the processor that we can configure  using consensus constants in the validator/indexer's `bootstrap.rs`. The network can also go into that config as well as any future config.

```rust
pub struct TransactionProcessorConfig {
   pub network: Network,
  pub template_binary_max_size_bytes: usize
}
```

sdbondi (commented):
LGTM - you need to regenerate typescript bindings

dan_layer/engine/Cargo.toml
ksrichard: okay, totally makes sense, let me update

ksrichard: just pushed

ksrichard:
@sdbondi I just pushed changes, now it should be okay

dan_layer/template_test_tooling/src/template_test.rs
```diff
@@ -474,12 +474,14 @@ impl TemplateTest {
             initial_ownership_proofs: proofs,
         };
         let processor = TransactionProcessor::new(
+            TransactionProcessorConfigBuilder::new()
```
sdbondi: nit: A nice pattern is to return the builder from the type being built

```rust
TransactionProcessorConfig::builder().with_network(..).build()
```

sdbondi (dismissed):
utACK

dan_layer/template_test_tooling/src/template_test.rs
ksrichard: yepp true, updated

sdbondi (approved):

PR #1222, opened 2024-12-19
feat(engine)!: syncing published templates

ksrichard:
Description
---

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


applications/tari_validator_node/src/bootstrap.rs
```diff
@@ -251,11 +252,20 @@ pub async fn spawn_services(
     }
     handles.push(epoch_manager_join_handle);
 
+    let validator_node_client_factory = TariValidatorNodeRpcClientFactory::new(networking.clone());
+
     info!(target: LOG_TARGET, "Template manager initializing");
     // Template manager
+    let (tx_template_sync, rx_template_sync) = broadcast::channel::<TemplateSyncRequest>(100);
```
sdbondi, line 259: use an unbounded mpsc channel. Broadcast channel is not appropriate as it is typically used for events and if there are more than 100 requests some of them may start to be dropped (broadcast channel lagging)

dan_layer/state_store_sqlite/src/substate.rs
```diff
@@ -0,0 +1,10 @@
+// Copyright 2024 The Tari Project
+// SPDX-License-Identifier: BSD-3-Clause
+
+use tari_dan_storage::StorageError;
+use tari_engine_types::substate::SubstateId;
+
+/// General trait for different state stores to do operations on substates.
+pub trait SubstateStore {
```
sdbondi, line 8: This doesnt seem right. Firstly, this is in the SQLite implementation of the state store which will be a problem when we switch to rocksDB. Templates can always be version 0, so not sure this is necessary, I'll need to look more closely at the rest of the code. There is currently no special process for upgrading the template code, you simply publish a new template at v0.

applications/tari_dan_app_utilities/src/template_manager/implementation/manager.rs
```diff
@@ -301,10 +350,34 @@ impl<TAddr: NodeAddressable + Send + Sync + 'static> TemplateProvider for Templa
...
+                let start = SystemTime::now();
+                let mut template = None;
+                while template.is_none() {
+                    let elapsed = start.duration_since(UNIX_EPOCH).expect("Time went backwards");
+                    if elapsed.gt(&self.config.pending_templates_wait_timeout()) {
+                        break;
+                    }
+                    if let Some(TemplateResult::Template(fetched_template)) = self.fetch_template(address).optional()? {
+                        template = Some(fetched_template);
+                    }
```
sdbondi, line 372: break here?

applications/tari_dan_app_utilities/src/template_manager/implementation/manager.rs
```diff
@@ -301,10 +350,34 @@ impl<TAddr: NodeAddressable + Send + Sync + 'static> TemplateProvider for Templa
...
+                    // sleeping here to not overload the local database while waiting for the template to be ready
+                    thread::sleep(std::time::Duration::from_millis(100));
```
sdbondi, line 374: Not ideal. Will think about how we should handle pending templates in a transaction.

PR #1223, opened 2024-12-23
fix(transactions)!: adds versioned transaction schema

sdbondi:
Description
---

fix(transactions)!: adds versioned transaction schema
Updated bindings (1.2.0), walletd client (1.2.0) and web UIs
Fix clippies in latest stable rust

Motivation and Context
---
Allows us to make versioned changes to the transaction schema/model while still supporting older schemas.

V1 may include breaking changes until mainnet.

How Has This Been Tested?
---
Manually, existing tests

---
Should work as before

Breaking Changes
---

- [ ] None
- [ ] Other - Please specify


sdbondi:
The Web UI is failing to compile because the bindings are not published

PR #1228, opened 2024-12-26
feat!: global publish template, malleability fix, pub templates in wallet

sdbondi:
Description
---
fix(transactions)!: fix transaction signatures malleability 
feat(wallet): add support for publishing templates
feat(consensus): publish templates to all shards
feat(wallet): add UI for publishing templates
fix(consensus):  correct input involvement info in evidence
Wallet daemon now has a transactions.publish_template JSON-RPC call

Motivation and Context
---
Currently all signatures but one could be surreptitiously removed from the transaction - this PR fixes this by allowing a pre-agreed signer to "seal" the transaction. 

Wallet daemon now has a `transactions.publish_template` JSON-RPC call.

Transactions containing publish template commands are considered "global" i.e. involve all shard groups. 
This provides the following benefits:
1. All validators immediately have the templates without having to sync them (#1222) and handle the associated edge cases
2. The proof that a template was previously committed is provided by existing mechanisms and consensus 
3. Simple to implement

Downside is the extra communication overhead of a global transaction

image

Next steps:
- a syncing node does not currently sync templates - syncing the entire template set should be done before the node registers and joins the network and reconciled with synced substates. 
- remove the actual binary BLOB from the on-chain substate (binary is currently stored twice) - this is currently used when adding the published template to the template store. Instead, one could obtain the BLOB from the transaction instead of from the substate.
- swarm should publish templates using the wallet daemon
- remove template web URL downloading from base layer UTXO, ignore base layer template registration UTXO

How Has This Been Tested?
---
Manually, publishing templates through the wallet UI on a multi-shardgroup network.

---
As above

Breaking Changes
---

- [ ] None
- [x] Other - Please specify

BREAKING CHANGE: wire-protocol for transactions, database is not compatible, submit transaction JSON-RPC call is not backward-compatible.


sdbondi:
New bindings release required, so web components fail to build

ksrichard:
LGTM!

utACK

PR #1242, opened 2025-01-15
feat(template): sync templates when syncing substates

ksrichard:
Description
---
Synchronizing when a VN is just added to the network or syncing new substates. We try to fetch template binary from its owning committee to make our chance higher to get it.
There is also a logic put to periodically check for pending templates, so if we know that we must have a template ready, just try to fetch again from other VNs.

Motivation and Context
---
When new nodes are joining the network, they did not have the template binaries as so we needed synchronization.

How Has This Been Tested?
---
1. Start a local swarm with X VNs
2. Publishing a new template (via CLI is simpler)
3. Adding a new VN (or more)
4. Mine to next epoch to make VN(s) available
5. See if new VN(s) have the new template already registered

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


applications/tari_dan_app_utilities/src/template_manager/implementation/manager.rs
```diff
@@ -190,6 +241,28 @@ impl<TAddr: NodeAddressable> TemplateManager<TAddr> {
...
+    pub fn add_pending_template(
+        &self,
+        template_address: tari_engine_types::TemplateAddress,
+        epoch: Epoch,
+    ) -> Result<(), TemplateManagerError> {
+        let template = DbTemplate::empty_pending(template_address, epoch);
+
+        let mut tx = self.global_db.create_transaction()?;
+        let mut templates_db = self.global_db.templates(&mut tx);
+        match templates_db.get_template(&template.template_address)? {
```
sdbondi, line 253: Couldn't you use `template_exists` here?
```suggestion
        match templates_db.template_exists(&template.template_address)? {
```

applications/tari_dan_app_utilities/src/template_manager/implementation/service.rs
```diff
@@ -41,31 +59,46 @@ use super::{
...
-use crate::template_manager::interface::{TemplateExecutable, TemplateManagerError, TemplateManagerRequest};
+use crate::template_manager::interface::{
+    SyncTemplatesResult,
+    Template,
+    TemplateExecutable,
+    TemplateManagerError,
+    TemplateManagerRequest,
+};
 
 const LOG_TARGET: &str = "tari::template_manager";
```
sdbondi: This was incorrect previously, please update this to the correct LOG_TARGET (see log4rs_sample it logs `tari::dan`) - just noticed I'm not seeing logs from this service.
```suggestion
const LOG_TARGET: &str = "tari::dan::template_manager";
```

sdbondi (commented):
Testing by adding a new VN worked. 

I tested as follows but the template was not listed:

1. start a swarm (3 SGs, 21 VNs)
2. publish a template
3. mine to next epoch
4. delete data on VN0
5. start VN0
6. After several minutes, did not observe the template in the web UI

sdbondi (approved):
After discussing offline, discovered that templates are (obviously) not synced on every shard - something that I didn't really think about when adding the global template publishing. This PR can be merged and we'll address this in a new PR.

PR #1243, opened 2025-01-15
feat: add TryFrom impls for SubstateId and Addresses

humb1t:
## Description

Associated error type for `impl TryFrom<SubstateId> for A` was changed from `SubstateId` to an error type which wraps id. Implementations for main variants were provided.

## Motivation and Context

For users of Tari SDK it's important to convert SubstateId into one of subvariants (for example into ComponentAddress). Unfortunately `as_T` methods return not an error type, but `Option`, which makes it verbose to manually wrap into own error types for handling. Existing implementation for ComponentAddress has associated type for error which doesn't implement error trait from Rust standard library. To improve developers experience new error type with some out of the box implementations proposed in this PR.

## How Has This Been Tested?

- Cargo builds
- CI

This change should not affect anything, so additional testing may not be needed.

## Breaking Changes

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/dan_layer/engine_types/src/substate.rs
+++ b/dan_layer/engine_types/src/substate.rs
@@ -21,6 +21,7 @@
 //   USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 use std::{
+    any,
     fmt::{Display, Formatter},
     str::FromStr,
 };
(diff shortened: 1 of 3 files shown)


dan_layer/engine_types/src/substate.rs
```diff
@@ -346,13 +341,105 @@ impl From<PublishedTemplateAddress> for SubstateId {
     }
 }
 
+#[derive(Debug, thiserror::Error)]
+#[error("Invalid substate id variant'{0}'")]
```
sdbondi:
```suggestion
#[error("Invalid substate ID variant '{0}'")]
```

I would be nice if we could know what variant is expected in the error message.

```rust
#[derive(Debug, thiserror::Error)]
#[error("Could not convert substate ID variant '{substate_id}'  to {expected}")]
pub struct InvalidSubstateIdVariant{ substate_id: SubstateId, expected: &'static str };

//(...)
InvalidSubstateIdVariant::new(value, "ResourceAddress")
```

sdbondi (dismissed):
utACK

Thanks

dan_layer/engine_types/src/substate.rs
humb1t: @sdbondi Thank you for the comment, I made a change, please check if it satisfies project need.

sdbondi: Perfect thanks

sdbondi (approved):

PR #1247, opened 2025-01-17
feat(template-sync-test): updated tests to use new template syncing method

ksrichard:
Description
---
Updated tests to use the new template publishing method and possibly fixed the random cucumber test hanging without even exiting.

**Important:** https://github.com/tari-project/tari-dan/pull/1242 must be merged to development before this PR could be

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


integration_tests/src/template.rs
```diff
@@ -11,17 +12,93 @@ use minotari_app_grpc::tari_rpc::{
...
+            TransactionResult::AcceptFeeRejectRest(_, reason) | TransactionResult::Reject(reason) => {
+                format!("Status: {}, Reason: {}", tx_resp.status, reason)
+            },
+            TransactionResult::Accept(_) => String::new(), // does not happen here
+        };
+        return Err(anyhow!(format!(
+            "Invalid transaction {}: {error_status:?}",
+            response.transaction_id
+        )));
+    }
```
sdbondi:
```suggestion
    if let Some(reason) = finalize_result.result.full_reject() {
        return Err(anyhow!(format!(
            "Invalid transaction {}: Status: {}, Reason: {}", 
            response.transaction_id, tx_resp.status, reason
        )));
    }
```

integration_tests/src/template.rs
```diff
@@ -11,17 +12,93 @@ use minotari_app_grpc::tari_rpc::{
...
+    // look for the new UP template substate
+    let mut result = None;
+    if let TransactionResult::Accept(diff) = finalize_result.result {
+        for (substate_id, _) in diff.up_iter() {
+            if let SubstateId::Template(addr) = substate_id {
+                result = Some(addr.as_hash());
+                break;
+            }
+        }
+    }
```
sdbondi:
```suggestion
    let template_id = finalize_result
      .result
      .accept()
      .and_then(|result| result.up_iter().find_map(|(substate_id, _)| substate_id.as_published_template()))
      .map(|id| id.as_hash())
      .ok_or_else(||  anyhow!("Transaction result did not contain a  published template"))?;
```

integration_tests/tests/features/committee.feature
```diff
@@ -5,6 +5,7 @@
 @committee
 Feature: Committee scenarios
 
+  @dev
```
sdbondi:
```suggestion
```

sdbondi (commented):
LGTM Reviewed this one first by mistake - will review the base PR

integration_tests/src/template.rs
ksrichard: Ah makes sense thanks! I did not check `full_reject()` call impl before as I thought it only checks against full reject and not partial one too.

ksrichard: Just pushed changes applied.

integration_tests/src/template.rs
ksrichard: Thanks, just pushed changes applied.

integration_tests/src/template.rs
sdbondi: Yeah the names are really bad - maybe we can make them fee_transaction_accept (prev. full_accept), accept, fee_transaction_reject (prev. full_reject), reject. These always confuse me and have caused bugs (even found one related to that today)

sdbondi (approved):
utACK

PR #1249, opened 2025-01-17
feat(wallet): using correct network

### LINKED ISSUE #1241: Add network to wallet
## Problem

Currently all transactions are for network '0' and the network is not validated. This means that transactions for e.g. igor are valid on e.g. mainnet

## Proposal 1

**Validator**
Validate that the transaction network byte is correct for the currently configured network. Otherwise reject the transaction.

**walletd**

The network byte field already exists on transactions. Add configuration and a cli arg `--network` to the wallet daemon and use the correct network byte.

ksrichard:
Description
---
For now we use the correct network configured and also validating on the validator node side in mempool.

Motivation and Context
---
Closes #1241 

How Has This Been Tested?
---
Start a fresh swarm locally and send any transaction through wallet daemon. It is expected to still work.

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (dismissed):
LGTM

sdbondi:
Think the integration tests need to be updated - looks like the validators are on Esmeralda and wallet is on localnet in cucumber. Localnet should be used for cucumbers.

Also need to update the wallet swarm command to pass in the `--network {network}` arg

ksrichard:
> Think the integration tests need to be updated - looks like the validators are on Esmeralda and wallet is on localnet in cucumber. Localnet should be used for cucumbers.
> 
> Also need to update the wallet swarm command to pass in the `--network {network}` arg

Okay will check the integration tests and the command passing, just forgot about that thanks!

sdbondi (dismissed):

applications/tari_dan_wallet_daemon/src/handlers/accounts.rs
```diff
@@ -122,7 +122,7 @@ pub async fn handle_create(
     );
 
     let max_fee = req.max_fee.unwrap_or(DEFAULT_FEE);
-    let transaction = Transaction::builder()
+    let transaction = transaction_builder(context)
```
humb1t, line 126: Sorry, just in case, idiomatic Rust use of functions assumes the following syntax: `module_name::function_name`, so I think it will be more idiomatic to use `types::transaction_builder(context)`

applications/tari_dan_wallet_daemon/src/handlers/helpers.rs
```diff
@@ -150,3 +151,8 @@ pub(super) fn invalid_params<T: Display>(field: &str, details: Option<T>) -> any
     )
     .into()
 }
+
+/// Returns a TransactionBuilder with the current network configured.
+pub fn transaction_builder(context: &HandlerContext) -> TransactionBuilder {
```
humb1t, line 167: In my humble opinion it would be nice to make the name more verbose to what it's really do and maybe associate the method with `TransactionBuilder` type. For example `TransactionBuilder::with_context(context)`.

applications/tari_dan_wallet_daemon/src/handlers/accounts.rs
```diff
@@ -323,6 +323,7 @@ pub async fn handle_reveal_funds(
 
     // If the caller aborts the request early, this async block would be aborted at any await point. To avoid this, we
     // spawn a task that will continue running.
+    let ctx = context.clone();
```
humb1t, line 323: Maybe instead of cloning the whole context we can instantiate builder here and move it into task's scope? Like `let builder_for_task = TransactionBuilder::with_contex(context);`?

applications/tari_validator_node/src/transaction_validators/error.rs
```diff
@@ -44,4 +45,8 @@ pub enum TransactionValidationError {
     TransactionNotSigned { transaction_id: TransactionId },
     #[error("Network error: {0}")]
     NetworkingError(#[from] NetworkingError),
+    #[error("Unknown network byte \"{0:?}\": {1}")]
+    UnknownNetwork(u8, ConfigurationError),
+    #[error("Network mismatch! Current network: {0}, Transaction network: {1}")]
+    NetworkMismatch(Network, Network),
```
humb1t: I know it may be overengineering, but it's very easy (at least for myself) to mess up expected and actual sides. Maybe use struct variant here?

```rust
NetworkMismatch {
   expected: Network,
   actual: Network,
}
```

sdbondi: I prefer this too

applications/tari_dan_wallet_daemon/src/handlers/helpers.rs
sdbondi: This is just a convinence method for the JSON-RPC handlers so not related to the transaction builder.

Please make the visibility of the function `pub(super)` to prevent it being used outside of the handlers module.

applications/tari_dan_wallet_daemon/src/handlers/accounts.rs
sdbondi: I suggest `let builder = transaction_builder(context);` outside the spawn to avoid the clone (not really fussed about it though :wink:) .

sdbondi (approved):

PR #1261, opened 2025-01-23
feat(validator): emit component-updated event

ksrichard:
Description
---
Now we emit a new event when a component has been changed, so indexer will save it to its local DB.

Motivation and Context
---
Indexer was caching an old version of a component (only saved when it was created) and indexer clients were not able to get back the latest version of the component.

How Has This Been Tested?
---
- Start a swarm locally
- Publish a new template
- Create a new component out of the template
- Update the component by calling any method on that
- Use indexer client to get latest version of the component
Example code:
```rust
let counter_comp_address = ComponentAddress::from_hex("ccc9d5676a692c4b43ee63e049704692841279a0867a6c4fdfd6d1f34852ee88").unwrap();
let mut indexer = tari_indexer_client::json_rpc_client::IndexerJsonRpcClient::connect("http://127.0.0.1:12014/")?;
    let result = indexer.get_substate(GetSubstateRequest{
        address: SubstateId::Component(counter_comp_address),
        version: None,
        local_search_only: false,
    }).await?;
    println!("substate: {:?}", result);
```

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,7 +137,7 @@ jobs:
 
       - name: install typescript
         run: |
-          npm install typescript -g
+          npm install typescript
 
       - name: toolchain
         uses: dtolnay/rust-toolchain@nightly
@@ -169,7 +169,7 @@ jobs:
 
       - name: install typescript
         run: |
-          npm install typescript -g
+          npm install typescript
 
       - name: toolchain
         uses: dtolnay/rust-toolchain@stable
@@ -226,7 +226,7 @@ jobs:
 
       - name: install typescript
         run: |
-          npm install typescript -g
+          npm install typescript
 
       - name: toolchain
         uses: dtolnay/rust-toolchain@stable
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,10 +135,6 @@ jobs:
       - name: checkout
         uses: actions/checkout@v4
 
-      - name: install typescript
-        run: |
-          npm install typescript
-
       - name: toolchain
         uses: dtolnay/rust-toolchain@nightly
         with:
@@ -166,10 +162,6 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@v4
-
-      - name: install typescript
-        run: |
-          npm install typescript
 
       - name: toolchain
         uses: dtolnay/rust-toolchain@stable
@@ -223,10 +215,6 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@v4
-
-      - name: install typescript
-        run: |
-          npm install typescript
 
       - name: toolchain
         uses: dtolnay/rust-toolchain@stable
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -135,6 +135,10 @@ jobs:
       - name: checkout
         uses: actions/checkout@v4
 
+      - name: install typescript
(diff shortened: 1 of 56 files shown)


humb1t (commented):
Really nice!

dan_layer/engine/src/runtime/impl.rs
```diff
@@ -660,6 +660,17 @@ impl<TTemplateProvider: TemplateProvider<Template = LoadedTemplate>> RuntimeInte
...
                     })?;
 
+                    // add event to indicate that there is a change in component
+                    let (template_address, module_name) =
+                        state.current_template().map(|(addr, name)| (*addr, name.to_string()))?;
+                    state.push_event(Event::new(
+                        Some(component_lock.address().clone()),
+                        template_address,
+                        state.transaction_hash(),
+                        "component-updated".to_string(),
```
mrnaveira: I would use something like "std.component.updated" as topic. The "std." event namespace is reserved for system events, and we also prevent user events from using that prefix. See https://github.com/tari-project/tari-dan/pull/975/files for more info

mrnaveira (changes requested):

dan_layer/engine/src/runtime/impl.rs
ksrichard: Okay makes sense thanks!
Just pushed.

mrnaveira (dismissed):

dan_layer/engine/src/runtime/impl.rs
```diff
@@ -660,6 +660,17 @@ impl<TTemplateProvider: TemplateProvider<Template = LoadedTemplate>> RuntimeInte
                         true
                     })?;
 
+                    // add event to indicate that there is a change in component
+                    let (template_address, module_name) =
+                        state.current_template().map(|(addr, name)| (*addr, name.to_string()))?;
+                    state.push_event(Event::new(
```
sdbondi: Event is also needed  in SetAccessRules

You could emit the event in `modify_component_with` which will cover both usages and will not emit the event unless it actually changes.

```rust
pub fn modify_component_with<F: FnOnce(&mut ComponentHeader) -> bool>(
 //(...snip)
        self.validate_component_state(Some(&before), &after)?;
// Emit an event for the component update
        let (address, _) = self.current_template()?;
        self.push_event(Event::new(
            Some(locked.address().clone()),
            *address,
            self.transaction_hash(),
            std_event("component.updated"),
            Metadata::new(),
        ));

        Ok(())
    }
```

ksrichard: ah true, okay let me change

ksrichard: Just pushed, thanks!

.github/workflows/ci.yml
```diff
@@ -135,6 +135,10 @@ jobs:
       - name: checkout
         uses: actions/checkout@v4
 
+      - name: install typescript
```
sdbondi: I'm not sure this is necessary due to this fix (merged) https://github.com/tari-project/tari-dan/pull/1260/files#diff-04646d5069f9923612c7ce0f122758738e2f809a5ff3628e25fc691e683dca93

ksrichard: ah okay, yeah yesterday the branch was broken, let me rollback

sdbondi (approved):

PR #1270, opened 2025-01-29
feat!: accumulate leader fees in substate, impl claim fees in wallet

sdbondi:
Description
---
Accumulate leader fees in ValidatorFeePool substate as blocks are proposed
Implement fee claim transactions in wallet daemon
Update bindings
Implement claim fee in wallet web UI
Implement fee scanning in wallet for a given key
Swarm uses key 0 of the wallet daemon as claim key for all validators

Motivation and Context
---
For non-zero block fees, a round leader must include a (Down->)Up transition in the state Merkle tree for their ValidatorFeePool substate (as defined by their claim public key).

How Has This Been Tested?
---
New (unignored) cucumber tests, manually

---
Claim fee feature in wallet web UI

Breaking Changes
---

- [ ] None
- [ ] Other - Please specify


sdbondi:
Due to bindings release 1.4.0 cargo check and cucumber build fails

PR #1288, opened 2025-02-13
chore: wallet daemon and npm pain points

mrnaveira:
Description
---
* In the wallet daemon
    * Removed the old Tari connector button (deprecated), leaving only the WalletConnect option. This is to avoid user confusion.
    * Hardcoded the WalletConnect ProjectId as default, to make wallet daemon binaries automatically use it without needing to specify it in the envvars (it can still be overridden in the envvars if needed).
    * Make `Igor` the default network if no network is specified from CLI 
* Adding metadata (repository, homepage and issues) to the npm packages (`typescript-bindings`, `wallet_jrpc_client`) so they are linked back from the npm registry. Bumped the version number to trigger a npm publish.

Motivation and Context
---
We identified some pain points for users/developers related to wallet daemon and npm packages.

How Has This Been Tested?
---
* Manually by starting a new localnet with tari spawn and connect with walletconnect
* For the default network value, spawned a wallet daemon directly from CLI and inspected the logs

---
See previous section

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_dan_wallet_daemon/src/config.rs
+++ b/applications/tari_dan_wallet_daemon/src/config.rs
@@ -83,7 +83,7 @@ impl Default for WalletDaemonConfig {
     fn default() -> Self {
         Self {
             override_from: None,
-            network: Default::default(),
+            network: Network::Igor,
             json_rpc_address: Some(SocketAddr::from(([127u8, 0, 0, 1], 9000))),
             ui_connect_address: None,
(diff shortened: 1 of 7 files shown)


sdbondi (approved):
utACK

Thanks

sdbondi:
Checks failing due to bindings package which is not yet released - merging and will make a follow-up PR to update package lockfiles

PR #1343, opened 2025-03-20
fix(wallet-daemon): authored templates are not saved always to local DB

### LINKED ISSUE #1323: feat: track templates published from the wallet
## Problem

There is currently no way to see what templates were published by the wallet. This can lead to poor UX when publishing templates and attempting to find the published template.

## Proposal 1

Add a new table to track published templates and provide the relevant JSON-RPC call (`templates.get`, `templates.list`) to query authored templates. Templates should be associated with a particular key_index used to sign the publish transaction.

ksrichard:
Description
---
I implemented a simple retry mechanism to try to fetch template definition (with increasing wait time to not spam indexer) for a new template published in a transaction.

Motivation and Context
---
It is not guaranteed that when we receive an event for a transaction that contains a published template that we have the template available on indexer. Because of this, new authored templates were not always showing on UI (and also saved to local DB).

Closes https://github.com/tari-project/tari-dan/issues/1323

How Has This Been Tested?
---
1. Started a fresh swarm with 10 nodes for having a longer time while templates are synchronized
2. Create an account
3. Open in another window the templates in wallet daemon UI
4. Publish a template
5. Wait for wallet daemon UI to be updated

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

## Summary by CodeRabbit

- **New Features**
  - Enhanced wallet event processing by adding asynchronous template fetching with robust retry and error handling for improved reliability.

- **Refactor**
  - Streamlined template management by simplifying dependencies and reducing configuration complexity for easier integration.

### DIFF
--- a/applications/tari_dan_wallet_daemon/src/services/template_monitor.rs
+++ b/applications/tari_dan_wallet_daemon/src/services/template_monitor.rs
@@ -63,6 +63,9 @@ where
                 Ok(template_def) => Some(template_def),
                 Err(error) => {
                     error!(target: LOG_TARGET, "Failed to fetch template definition: {}, retry...", error);
+                    if self.shutdown_signal.is_triggered() {
+                        return Err(anyhow!("shutdown during fetch template definition"));
+                    }
                     tokio::time::sleep(current_wait_time).await;
                     if current_wait_time < max_wait_time {
                         current_wait_time = current_wait_time.add(wait_step);
--- a/applications/tari_dan_wallet_daemon/src/services/template_monitor.rs
+++ b/applications/tari_dan_wallet_daemon/src/services/template_monitor.rs
@@ -1,22 +1,26 @@
 // Copyright 2025 The Tari Project
 // SPDX-License-Identifier: BSD-3-Clause
 
+use std::{ops::Add, time::Duration};
+
+use anyhow::anyhow;
 use log::error;
 use tari_common_types::types::PublicKey;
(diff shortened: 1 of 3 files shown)


applications/tari_dan_wallet_daemon/src/services/template_monitor.rs
```diff
@@ -42,6 +46,35 @@ where
...
+        while template_definition.is_none() {
+            template_definition = match network_interface
+                .fetch_template_definition(template_address)
+                .await
+                .map_err(|error| TransactionApiError::NetworkInterfaceError(format!("{}", error)))
+            {
+                Ok(template_def) => Some(template_def),
+                Err(error) => {
+                    error!(target: LOG_TARGET, "Failed to fetch template definition: {}, retry...", error);
+                    tokio::time::sleep(current_wait_time).await;
```
sdbondi, line 69: May want to add a check for shutdown so that this loop will exit <= 1s if shutdown
```suggestion
if self.shutdown_signal.is_triggered() {
return Err(anyhow!("shutdown during fetch template definition"));
}
```

ksrichard: ah true makes sense

ksrichard: pushed

sdbondi (approved):

PR #1404, opened 2025-04-17
feat(testing): add option to compile templates in cargo workspace

ksrichard:
Description
---
Added an optional custom target directory, so compilation will work even with a cargo workspace project.

Motivation and Context
---
Compiling templates were only possible in a non-workspace cargo project, so if a template project is compiled in a different `target` directory (in workspace case in the root) then the compilation fails because it tried to load the wasm binary from the wrong place.

How Has This Been Tested?
---
Manually.

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

## Summary by CodeRabbit

- **New Features**
  - Added the ability to specify a custom target directory when compiling templates.
  - Introduced a new way to create template tests with a shared target directory.

- **Refactor**
  - Updated method signatures to support optional target directory parameters for template addition and testing.
  - Streamlined internal logic for template compilation and test initialization.

### DIFF
--- a/dan_layer/engine/src/wasm/compile.rs
+++ b/dan_layer/engine/src/wasm/compile.rs
@@ -70,7 +70,15 @@ crate-type = ["cdylib", "lib"]
     compile_template(temp_dir.path(), features)
 }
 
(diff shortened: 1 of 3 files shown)


sdbondi (approved):
LGTM

PR #1423, opened 2025-05-07
feat(wallet): automatic wasm template size optimization

### LINKED ISSUE #1385: Use wasm-opt to optimise WASM size
https://docs.rs/wasm-opt/latest/wasm_opt/

Cli example optimising for size:
```
wasm-opt -Os -g --strip-debug --strip-dwarf --strip-producers -o faucet.wasm target/wasm32-unknown-unknown/release/faucet.wasm
```

faucet reduced from 140k to 102k

ksrichard:
Description
---
I added the mentioned [wasm_opt](https://docs.rs/wasm-opt/latest/wasm_opt/) crate and used to optimize wasm templates when sent via wallet-daemon.

Motivation and Context
---
Trying to make the published templates smaller to save space.

Closes #1385 

How Has This Been Tested?
---
### Manually
1. Start a fresh swarm
2. Create the default account
3. Publish a template through UI
4. Check the logs of wallet-daemon, there should be a message like (that shows how much the template was optimized):
` WASM template optimized, original size: 461614 bytes, new size: 312781 bytes`

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

## Summary by CodeRabbit

- **New Features**
  - Added automatic optimization of WebAssembly (WASM) template binaries before publishing to reduce size and improve performance.
  - Introduced enhanced error handling and logging for WASM binary optimization.
  - Integrated a new module for WASM optimization within the wallet daemon.

- **Bug Fixes**
  - Ensured original WASM binaries are published if optimization fails, maintaining workflow continuity.

### DIFF
--- a/applications/tari_dan_wallet_daemon/src/handlers/transaction.rs
+++ b/applications/tari_dan_wallet_daemon/src/handlers/transaction.rs
@@ -38,11 +38,13 @@ use tari_wallet_daemon_client::{
     },
 };
 use tokio::time;
+use wasm_opt::OptimizationOptions;
 
 use super::context::HandlerContext;
 use crate::{
     handlers::{
         helpers::{get_account, get_account_or_default, invalid_params, transaction_builder},
+        wasm_optimizer::{Error, WasmTemplateOptimizer},
         HandlerError,
     },
     services::WalletEvent,
@@ -479,12 +481,24 @@ pub async fn handle_publish_template(
 
     let fee_account = get_account_or_default(req.fee_account, &sdk.accounts_api())?;
 
+    // trying to optimize WASM binary
+    let wasm_binary = match WasmTemplateOptimizer::new().optimize(req.binary.as_slice()).await {
+        Ok(optimized) => {
+            info!(target: LOG_TARGET, "WASM template optimized, original size: {} bytes, new size: {} bytes", req.binary.len(), optimized.len());
+            optimized
+        },
+        Err(error) => {
+            warn!(target: LOG_TARGET, "Error while optimizing WASM template (using original version now): {}", error);
+            req.binary
+        },
+    };
+
     let transaction = transaction_builder(context)
         .fee_transaction_pay_from_component(
             fee_account.address.as_component_address().unwrap(),
             req.max_fee.try_into()?,
         )
-        .publish_template(req.binary)
+        .publish_template(wasm_binary)
         .build_unsigned_transaction();
 
     if req.dry_run {
--- a/applications/tari_dan_wallet_daemon/src/handlers/transaction.rs
+++ b/applications/tari_dan_wallet_daemon/src/handlers/transaction.rs
@@ -38,13 +38,12 @@ use tari_wallet_daemon_client::{
     },
 };
 use tokio::time;
-use wasm_opt::OptimizationOptions;
 
 use super::context::HandlerContext;
 use crate::{
     handlers::{
         helpers::{get_account, get_account_or_default, invalid_params, transaction_builder},
-        wasm_optimizer::{Error, WasmTemplateOptimizer},
+        wasm_optimizer::WasmTemplateOptimizer,
         HandlerError,
     },
     services::WalletEvent,
--- a/applications/tari_dan_wallet_daemon/src/handlers/transaction.rs
+++ b/applications/tari_dan_wallet_daemon/src/handlers/transaction.rs
@@ -43,7 +43,7 @@ use super::context::HandlerContext;
 use crate::{
     handlers::{
         helpers::{get_account, get_account_or_default, invalid_params, transaction_builder},
-        wasm_optimizer::WasmTemplateOptimizer,
+        wasm_optimizer::optimize_wasm_template,
         HandlerError,
     },
     services::WalletEvent,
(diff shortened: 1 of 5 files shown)


applications/tari_dan_wallet_daemon/src/handlers/wasm_optimizer.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    pub async fn optimize(&self, template_binary: &[u8]) -> Result<Vec<u8>, Error> {
+        let temp_dir = tempdir()?;
+
+        // create temporary input file
+        let input_file_path = temp_dir.path().join("input.wasm");
+        let output_file_path = temp_dir.path().join("output.wasm");
+        let mut input_file = File::create(input_file_path.as_path()).await?;
+        input_file.write_all(template_binary).await?;
+        input_file.flush().await?;
```
sdbondi: input_file will only close at the end of the function
```suggestion
        tokio::fs::write(&input_file_path, template_binary).await?;
```

applications/tari_dan_wallet_daemon/src/handlers/wasm_optimizer.rs
```diff
@@ -0,0 +1,65 @@
...
+
+        OptimizationOptions::new_optimize_for_size().run(input_file_path, output_file_path.as_path())?;
+
+        let result = tokio::fs::read(output_file_path).await?;
+
+        if result.is_empty() {
+            return Err(Error::InvalidFile("Empty file".to_string()));
+        }
+
+        temp_dir.close()?;
```
sdbondi: nit: AFAIK You dont need to close a tempfile - it will delete the path when dropped

ksrichard: yeah it is true, but it's doc says that in that case we do not know if that operation was successful or not, so this way we will have an error when failed to delete folder

applications/tari_dan_wallet_daemon/src/handlers/transaction.rs
```diff
@@ -479,12 +480,24 @@ pub async fn handle_publish_template(
 
     let fee_account = get_account_or_default(req.fee_account, &sdk.accounts_api())?;
 
+    // trying to optimize WASM binary
+    let wasm_binary = match WasmTemplateOptimizer::new().optimize(req.binary.as_slice()).await {
```
sdbondi: nit: The WasmTemplateOptimizer struct is a glorified function. Up to you but I cant see a need now or in the future for some stateful struct for optimisation. 
```suggestion
    let wasm_binary = match optimize_wasm_template(req.binary.as_slice()).await {
```

applications/tari_dan_wallet_daemon/src/handlers/wasm_optimizer.rs
```diff
@@ -0,0 +1,65 @@
...
+        let temp_dir = tempdir()?;
+
+        // create temporary input file
+        let input_file_path = temp_dir.path().join("input.wasm");
+        let output_file_path = temp_dir.path().join("output.wasm");
+        let mut input_file = File::create(input_file_path.as_path()).await?;
+        input_file.write_all(template_binary).await?;
+        input_file.flush().await?;
+
+        OptimizationOptions::new_optimize_for_size().run(input_file_path, output_file_path.as_path())?;
```
sdbondi: What effect does [new_optimize_for_size_aggressively](https://docs.rs/wasm-opt/latest/wasm_opt/struct.OptimizationOptions.html#method.new_optimize_for_size_aggressively) have?

ksrichard: Using that for some reason the resulting binary was invalid then and actually there was no difference in size in the end.

applications/tari_dan_wallet_daemon/src/handlers/transaction.rs
ksrichard: Yeah I was thinking the same, let me convert that as it could be still a struct if we would optimize for different purposes.

applications/tari_dan_wallet_daemon/src/handlers/wasm_optimizer.rs
ksrichard: true, its better thanks!

ksrichard:
@sdbondi just pushed changes, thanks for the suggestions!

applications/tari_dan_wallet_daemon/src/handlers/wasm_optimizer.rs
sdbondi: Ah interesting 👍 ok let's not use that then lol

applications/tari_dan_wallet_daemon/src/handlers/wasm_optimizer.rs
```diff
@@ -0,0 +1,54 @@
...
+    let input_file_path = temp_dir.path().join("input.wasm");
+    let output_file_path = temp_dir.path().join("output.wasm");
+    tokio::fs::write(&input_file_path, template_binary).await?;
+
+    OptimizationOptions::new_optimize_for_size().run(input_file_path, output_file_path.as_path())?;
+
+    let result = tokio::fs::read(output_file_path).await?;
+
+    if result.is_empty() {
+        return Err(Error::InvalidFile("Empty file".to_string()));
```
sdbondi, line 48:
```suggestion
        return Err(Error::InvalidFile("The WASM optimiser output an empty file".to_string()));
```

sdbondi (approved):

PR #1427, opened 2025-05-09
feat(wallet)!: prevent dry run transactions submitted as real

### LINKED ISSUE #1376: Prevent dry run transaction from being submitted as a real transaction
## Problem

A dry run submitted to the indexer could be submitted as a transaction to the network.
Ultimately it must be up to the user to do so, not any other party.

## Solution

Either a different transaction type that does not require signatures (e.g. just lists public keys that would sign for the executor) or a dry run flag on all transactions and a VN mempool validation to reject all dry runs (NB: dry run flag is part of the signature message)

ksrichard:
Description
---
Added a new `dry_run` flag to transaction (that goes to the hash of that too, because it's body is already used to generate tx hash) and this new flag is check by both mempool validator and dry run transaction executor.

Closes https://github.com/tari-project/tari-dan/issues/1376

Motivation and Context
---
Previously transactions created could be sent both to dry run transaction processor and be sent to the network too.
We should prevent this in order to avoid mixing up transactions wanted to be dry runs only (to have fees or something) and real executable transactions.

How Has This Been Tested?
---
Manually

---

Breaking Changes
---

- [ ] None
- [ ] Other - Please specify

## Summary by CodeRabbit

- **New Features**
  - Added support for "dry run" transactions, enabling users to simulate transfers and submissions without actual execution.
  - Transactions can now be marked as dry runs across wallet, API, and validator interfaces.
- **Bug Fixes**
  - Improved validation to reject dry run transactions where not allowed, preventing unintended processing.
- **Developer Improvements**
  - Enhanced error handling and validation for dry run transactions.
  - Updated transaction builder, storage, and related components to fully support the dry run flag.


dan_layer/wallet/storage_sqlite/src/models/transaction.rs
```diff
@@ -61,6 +61,7 @@ impl Transaction {
                         min_epoch: self.min_epoch.map(|epoch| Epoch(epoch as u64)),
                         max_epoch: self.max_epoch.map(|epoch| Epoch(epoch as u64)),
                         is_seal_signer_authorized: true,
+                        dry_run: false,
```
sdbondi:
```suggestion
                        dry_run: self.is_dry_run,
```

ksrichard: ah good catch, just pushed

sdbondi (approved):

PR #1468, opened 2025-06-20
fix(ci): rename packages, only build necessary

sdbondi:
Description
---
tari_network_suite -> tari_ootle
Only include wallet, indexer, vn and swarm in builds 

Motivation and Context
---
Bins like tari_generate and wallet cli are not used or not critical at this stage.


leet4tari (approved):
utACK - LGTM

PR #1472, opened 2025-06-23
fix(consensus)!: use `jmt` crate 

sdbondi:
Description
---

Motivation and Context
---

How Has This Been Tested?
---

---

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi:
On hold for now due to the amount of work involved, the JMT restoration procedure is different and will require a rethink for state sync.

PR #1669, opened 2025-12-11
fix: add an error case for dropped subscriber

stringhandler:
The previous implementation, if all senders had dropped (should never happen), the function would create a subscription that would never emit new items. Although this is completely fine in this case, this may be surprising in future and potentially lead to bugs. 

This PR instead chooses to "crash" the node on initialisation. The error message indicates to developers that there is a bug.

### DIFF
--- a/applications/tari_validator_node/src/event_subscription.rs
+++ b/applications/tari_validator_node/src/event_subscription.rs
@@ -22,6 +22,12 @@
 
 use tokio::sync::broadcast;
 
+#[derive(Debug, thiserror::Error)]
+pub enum EventSubscriptionError {
+    #[error("The sender has been dropped")]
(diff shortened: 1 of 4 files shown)


applications/tari_validator_node/src/event_subscription.rs
```diff
@@ -22,6 +22,12 @@
 
 use tokio::sync::broadcast;
 
+#[derive(Debug, thiserror::Error)]
+pub enum EventSubscriptionError {
+    #[error("The sender has been dropped")]
```
sdbondi:
```suggestion
    #[error("All event senders have dropped on initialzation. This indicates a bug in the node.")]
```

sdbondi (approved):

PR #1680, opened 2025-12-15
fix: remove some expects

stringhandler:
Minor PR to remove some more expects


sdbondi (commented):
These assertions are intentional, and are bugs if triggered. Did you encounter a panic?

PR #1681, opened 2025-12-15
fix: remove another expect

stringhandler:
Another expect removed


sdbondi (approved):
I think I should remove the mpsc channel entirely from the implementation, probably there more due to rushing to get something done more than being the "right way"

PR #1805, opened 2026-02-25
fix(indexer): remove add connection from api

sdbondi:
Description
---
fix(indexer): remove add connection from api
fix(ci): redeploy docs on each new non-draft release
tests(cucumber): add peer in process instead of through indexer API
docs(indexer): updated open api and generated docs
tests(integration): fixed incorrect wallet JSON-RPC URL 
tests(integration): fixed compile errors in example templates

Motivation and Context
---
`add_connection` should not be publicly callable. `add_connection` was used in the web ui (removed from web ui) and cucumber tests (uses in-process communication i.e. mpsc channel) instead.

How Has This Been Tested?
---
Integration tests

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


leet4tari (approved):
utACK - Only the workflow review - LGTM

PR #1955, opened 2026-03-31
fix(walletd): only mark burn proof as claimed on successful transaction

sdbondi:
## Description

Fix a bug in `handle_claim_burn` where burn proofs were moved to the "claimed" directory immediately after submitting the transaction, without waiting for it to succeed. If the transaction was rejected, the proof was permanently consumed and could not be retried.

## Motivation

A failed claim burn transaction should be retryable. The burn proof should only be marked as claimed once the transaction has been confirmed on-chain.

## Changes

- Subscribe to events before submitting the transaction
- Await `wait_for_result` to get the finalized transaction outcome
- Return an error if the transaction is fee-rejected or otherwise rejected
- Only mark the burn proof as claimed after confirming success

### DIFF
--- a/applications/tari_walletd/src/handlers/accounts.rs
+++ b/applications/tari_walletd/src/handlers/accounts.rs
@@ -587,7 +587,20 @@ pub async fn handle_claim_burn(
         });
     }
 
+    let mut events = context.notifier().subscribe();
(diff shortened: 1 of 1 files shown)


sdbondi:
This has some shortcomings (the result is thrown away) - will make a better PR for 0.28

PR #2017, opened 2026-04-14
feat(engine)!: implement epoch hash mixing for randomness

lucenera75:
Implements [OIP-0002] end-to-end: exposes the per-epoch L1-derived block hash as a deterministic, manipulation-resistant randomness beacon accessible from template code via Consensus::current_epoch_hash()

The epoch hash is already present in every BlockHeader (epoch_hash field, derived from the L1 base-layer block that triggers the epoch transition). This PR makes it accessible inside WASM template execution by:

Adding a new VirtualSubstateId::CurrentEpochHash variant that is pre-injected before each transaction executes, alongside the existing CurrentEpoch.
Threading the value from its two source sites — EpochState::epoch_hash in on_propose.rs and block.epoch_hash() in on_ready_to_vote_on_local_block.rs — through ConsensusTransactionManager down to the engine runtime.
Surfacing it as Consensus::current_epoch_hash() in the template lib.
A two-component reference betting template (EpochBettingHouse + EpochBet) is included to demonstrate the pattern in practice, along with an OIP RFC and a developer guide.

Motivation and Context
Smart contract templates frequently need a source of randomness that is deterministic (all validators agree), unpredictable in advance, and manipulation-resistant. Prior to this PR the only consensus-layer value available to templates was Consensus::current_epoch(). There was no randomness primitive satisfying all three properties.

The epoch hash fills this gap with zero additional consensus overhead: the value already exists in every signed block header and is fixed for the entire epoch. The key insight is the commit-reveal pattern: a player commits to a bet in epoch N, and the outcome is resolved using the epoch N+1 hash — a value that was entirely unknowable at commit time. No single validator can choose the epoch hash; it is fixed by L1 mining at the epoch boundary.

Use cases unlocked: betting and prediction markets, lotteries, fair NFT trait assignment, commit-reveal games.

The keeper pattern is also introduced as the idiomatic way to enable automatic settlement without a native scheduling primitive: anyone can call settle once the next epoch arrives and earn a small protocol fee, incentivising keeper bots to settle bets on behalf of players.

How Has This Been Tested?
Automated tests — all passing:

cargo test -p tari_engine --test test "consensus::"
5 tests in mod consensus of crates/engine/tests/test.rs:

current_epoch — existing test, unaffected
test_epoch_hash_is_injected — returned hash matches the injected virtual substate
test_epoch_hash_differs_across_epochs — two different epoch hashes produce different values
test_epoch_hash_constant_within_epoch — same epoch hash returns the same value across calls
test_epoch_hash_not_available_without_injection — missing virtual substate produces VirtualSubstateNotFound

cargo test -p tari_engine --test epoch_betting
5 tests in crates/engine/tests/epoch_betting.rs:

test_settle_bet_requires_later_epoch — settling in the same epoch as placement is rejected
test_keeper_can_settle — third-party keeper receives the keeper fee; player receives 2 × (stake − fee)
test_player_settles_own_bet_win — player settling their own winning bet receives 2× stake with no fee deducted
test_player_settles_own_bet_loss — losing stake goes to house reserve; player balance decreases by stake
test_cannot_settle_twice — a settled bet cannot be settled again
Run the new tests:

cargo test -p tari_engine --test test "consensus::"
cargo test -p tari_engine --test epoch_betting
All 10 tests should pass.

Review the RFC at docs/proposals/consensus/OIP-0002-epoch-hash-randomness.md for the full rationale, security analysis, and backwards-compatibility discussion.

Trace the data flow — the epoch hash flows from two entry points:

on_propose.rs → epoch_state.epoch_hash() → prepare_transaction(…, epoch_hash, …) → manager.prepare(…, epoch_hash, …)
on_ready_to_vote_on_local_block.rs → block.epoch_hash() → execute_transaction(…, epoch_hash, …) → manager.execute(epoch, epoch_hash, pledged) Both paths converge in ConsensusTransactionManager::execute_or_fetch, which injects VirtualSubstate::CurrentEpochHash before calling the engine.
Verify the template API in crates/template_lib/src/consensus.rs — current_epoch_hash() mirrors the existing current_epoch() pattern exactly.

Check the reference template in crates/engine/tests/templates/epoch_betting_house/ and epoch_betting/ to validate the two-component keeper-pattern design (no token burning, cross-component settlement via ComponentManager::invoke).

Breaking Changes
BlockTransactionExecutor::execute trait gains a current_epoch_hash: FixedHash parameter. The only production implementor (TariBlockTransactionExecutor) is updated in this PR; any external forks implementing this trait will need to add the parameter.
VirtualSubstateId and VirtualSubstate gain new variants. Existing match arms with #[non_exhaustive] handling are unaffected; exhaustive matches elsewhere in this repo are updated.
Templates compiled against the new ABI calling current_epoch_hash() require validator nodes that implement GetCurrentEpochHash. Nodes that do not will return VirtualSubstateNotFound, rejecting the transaction. Validator nodes must be upgraded before such templates are deployed.


sdbondi:
Thanks for the PR!

I've been wanting to look at adding this mainly to "mix" the additional entropy with the GetRandomBytes engine call (which calls new_uuid in https://github.com/tari-project/tari-ootle/blob/c6340779fccba2e8d5e406d4aa7b40c00a0125ba/crates/engine_types/src/id_provider.rs)

Does doing this simplify your use cases (using `rand::get_random_bytes` in templates)?
Could you add that to this PR (new_uuid can take in `entropy: &[u8]` and pass in the epoch hash)?

docs/developer-docs/src/content/docs/guides/epoch-hash-randomness.mdx
```diff
@@ -0,0 +1,239 @@
...
+
+Smart contract templates frequently need randomness that is:
+
+1. **Deterministic** — every validator node must agree on the same value for a given transaction.
+2. **Unpredictable in advance** — the value must not be knowable at the time a user commits to a
+   bet or game outcome.
+3. **Manipulation-resistant** — no single validator or small coalition should be able to bias the
+   value.
+
+Tari Ootle provides `Consensus::current_epoch_hash()` as a built-in randomness primitive that
```
sdbondi, line 18: `tari_template_lib::rand::get_random_bytes(len) -> Vec<u8>` is used for this. It uses the transaction hash (tx entropy), component entity id (context entropy), and a counter (changes for each call), but you are correct that it should mix the epoch hash into the hash derivation (cannot be easily/cheaply predicted)

sdbondi:
Busy working through this one, and since it touches consensus, I'll be extra careful.

One snafu, there is a chance that the epoch and epoch hash will mismatch in this implementation. We need to lock the epoch hash along with the locked epoch number to avoid a possibly inconsistent execution context across shards (edge case, race condition). This requires changes to the sidechain crate on the L1 repo, so I'll need to coordinate that. No action required, I'll work though that part.

Moving to draft for now.

lucenera75:
> Busy working through this one, and since it touches consensus, I'll be extra careful.
> 
> One snafu, there is a chance that the epoch and epoch hash will mismatch in this implementation. We need to lock the epoch hash along with the locked epoch number to avoid a possibly inconsistent execution context across shards (edge case, race condition). This requires changes to the sidechain crate on the L1 repo, so I'll need to coordinate that. No action required, I'll work though that part.
> 
> Moving to draft for now.

Like very much what you did here!!!, and I see what you meant. thanks

sdbondi (dismissed):
Ack

I've unfortunately removed oip 0002 and the developer docs because they were misleading  (epoch hash comes from the base layer and is known for some time ahead)

sdbondi (approved):

PR #2028, opened 2026-04-15
fix(cucumber): fix VN registration detection timeout in state sync test

### LINKED ISSUE #1978: Fix VN registration detection timeout in state sync test
## Bounty: Fix VN Registration Detection Timeout in State Sync Test

**Tier:** M — 60,000 XTM

### Description

The "New validator node registers and syncs" cucumber test times out waiting for VN2 to report itself as registered after sending a registration transaction. The miner mines to the next epoch, but VN2 gets stuck at block height 40 and never picks up its registration. This could be a race condition where VN2 hasn't processed the block containing its registration, or the epoch boundary logic doesn't trigger registration pickup in time.

### Acceptance Criteria

- [ ] The `the validator node VN2 is listed as registered` step passes
- [ ] "New validator node registers and syncs" scenario passes end-to-end
- [ ] Investigation confirms whether the `mines to the next epoch` step mines enough blocks for VN2 to see the registration

### Context

- Issue opened by @sdbondi
- Error: `Timed out waiting for validator node to pick up registration (current block height: 40)`
- Affected step: `integration_tests/tests/steps/validator_node.rs:255`
- Affected scenario in `tests/features/state_sync.feature`

### How to Claim

1. Comment on this issue to signal intent (courtesy, not a lock)
2. Fork the repo and do the work
3. Submit a PR that meets the acceptance criteria
4. First PR that passes review and gets merged wins the bounty
5. On acceptance, XTM payment is processed

### Notes

- AI-assisted development is expected and encouraged
- If you get stuck, ask in Discord
- If you can't complete it within 14 days of claiming, let us know so someone else can pick it up

0xskr33p:
Description
---
This PR fixes the `@state_sync` cucumber scenario `New validator node registers and syncs` in two steps

The first change fixes a false timeout in the `the validator node VN2 is listed as registered` assertion by accounting for the validator scanner's intentional base-layer confirmation lag

The follow-up changes address the issues that became visible once that timeout was removed: registration timing around the epoch boundary, validator-set update ordering in the base layer oracle, and checkpoint/genesis-root consistency for nodes joining through state sync. Together these changes make the scenario pass end-to-end instead of only progressing past the initial registration check

Key Changes:
- **`oracle.rs`** (base layer scanner): Reordered block scanner logic to check for VN Merkle Root changes *before* epoch boundaries, ensuring registrations mined in the last block of an epoch are processed before committee assignment.
- **`on_receive_local_proposal.rs`** (consensus): Changed genesis block state root source from `valid_tip_block` to `eoe_block` so state-synced and in-consensus nodes produce identical genesis hashes.
- **`worker.rs`** (hotstuff): Allowed genesis block creation with a calculated state root when previous overlapping committees are empty or the local node is the sole validator.
- **`validator_node.rs`** (integration tests): Fixed registration timeout by computing `lagged_height`; added mempool sync wait; increased retry/timeout values; corrected epoch filter in `list_blocks_paginated`.
- **`state_sync.feature`**: Uses a new account (`ACC4`) for free coin claims to avoid duplicate transaction errors.
- **`rpc_impl.rs`**, **`state_sync.rs`**: Removed temporary diagnostic logs left over from troubleshooting.

Motivation and Context
---
Fixes #1978 

The original failure was a timeout while waiting for VN2 to detect its own registration in the state sync integration test. Investigation showed that the test compared the validator's scanned base-layer height to the base node tip height even though the scanner intentionally trails the tip by `base_layer_confirmations` blocks

After fixing that, the scenario still failed later because VN2 could miss activation for epoch 4 if the registration transaction had not yet reached the mempool before mining, or if validator-set changes were processed after the epoch-change event at the boundary block. Once those ordering issues were fixed, a final follow-up was needed to ensure checkpoint-based state sync and epoch genesis creation use the same canonical state roots across in-consensus and state-synced nodes

How Has This Been Tested?
---
`cargo test --release --test cucumber -p integration_tests -- --tags "@state_sync"`

- 1 feature
- 1 scenario (1 passed)
- 20 steps (20 passed)

This scenario was used throughout the investigation to verify both the initial registration-timeout fix and the follow-up state sync / epoch transition fixes

---
1. Check out this branch on top of `development`
2. Run `cargo test --release --test cucumber -p integration_tests -- --tags "@state_sync"`
3. Verify that the scenario `New validator node registers and syncs` passes end-to-end
4. If useful, inspect the progression through these milestones in the test output:
   - `Then the validator node VN2 is listed as registered`
   - `Then the validator node VN2 has started epoch 4`
   - the post-sync account creation steps
   - the final leaf-height checks for both validators
5. Optionally review the two commits separately:
   - the first commit only adjusts the registration detection wait logic in the cucumber step
   - the second commit contains the follow-up fixes for mempool timing, oracle event ordering, and checkpoint/genesis consistency

Breaking Changes
---

- [ ] None
- [x] Other - Please specify


0xskr33p:
Chronology of the Fix:

1. Problem: The "New validator node registers and syncs" test times out at the step VN2 is listed as registered. VN2 gets stuck at height 40 while tip = 43;

2. Root cause: The VN scanner intentionally lags behind the tip by base_layer_confirmations (3 blocks). The condition current_block_height >= tip (40 >= 43) can never be satisfied;

3. Fix: Compare against lagged_height = tip - base_layer_confirmations, increase timeout from 20 to 40 iterations. (validator_node.rs);

---

4. Problem: The test progresses further, but VN2 fails to start Epoch 4 — remains Idle, is_valid=false, start_epoch=None;

5. Root cause: Two race conditions:
- The registration tx was sent without waiting for mempool inclusion → could be mined in a later epoch than intended;
- In the scanner, on an epoch boundary block, EpochChanged was processed before ActiveValidatorNodeSetChanged → committee assignment ran without the new VN;

6. Fix: Added mempool wait after sending the registration tx (validator_node.rs). Reordered scanner logic — validator set update now happens before epoch change (oracle.rs);

---

7. Problem: VN and VN2 produce different genesis blocks for Epoch 4. VN2 rejects proposals from VN: Bad proposal, unable to find dummy blocks;

8. Root cause: VN takes the state root from valid_tip_block, while VN2 (after state sync) computes it from its local state tree. The hashes don't match;

9. Fix: Changed the state root source for genesis to eoe_block (End-of-Epoch block) — deterministic and identical across all nodes. (on_receive_local_proposal.rs);

---

10. Problem: The wallet daemon returns 500: Transaction invalid when creating an account. The VN keeps cycling Running → Syncing → Running;

11. Root cause: HotstuffWorker refuses to create a genesis block without an EpochCheckpoint from the previous epoch, even though the node is the sole validator and no checkpoint is needed;

12. Fix: Allowed genesis creation with a calculated state root without a checkpoint when the previous committee is empty or the node is the only validator (worker.rs). Also fixed the test – duplicate free coins claim replaced with a new account ACC4 (state_sync.feature);

0xskr33p:
I'll be back today to look at the merge conflict and retest the full scenario

sdbondi:
- [x] The `the validator node VN2 is listed as registered` step passes
- [x] "New validator node registers and syncs" scenario passes end-to-end
- [x] Investigation confirms whether the `mines to the next epoch` step mines enough blocks for VN2 to see the registration

0xskr33p:
Hi @sdbondi!
Currently, the committee scenario is not working in this branch (unlike development) – I'm looking into it right now
(oh, it's shown in a pipeline also, nvm – thought that there was only clippy error (I'll fix))

sdbondi:
I'm inclined to revert the consensus changes since they carry risk, the test passes without them, and the logic behind them is not clear to me without being able to reproduce/more details - wdyt?

0xskr33p:
@sdbondi 
Sure, as you wish. Besides, I was thinking that I’ve gone beyond the scope of the task, and I can continue digging into the essence in another branch and document everything well along the way.
Actually at the time I was working on this issue, the entire scenario didn't work without changes in consensus (but since then there have been commits in development that affected some parts, and perhaps they actually solved a number of problems)

0xskr33p:
@sdbondi 
I'll check if changes in oracle/validator_node affects other tests

0xskr33p:
Hi @sdbondi !
Previously (on last CI check) introduced problems were `VN1 did not start consensus in time: status: Sleeping, epoch: Epoch(0)` due to events reordering in oracle (and it's correct if I'm right) and here is a summary from an AI:

> The rpc_state_sync change is intentionally narrow. It does not change validator registration, activation epoch, committee assignment, epoch transition, or consensus rules. It only fixes how state sync interprets one specific edge case introduced by the earlier oracle event ordering change.
> 
> After registrations became visible before EpochChanged, a node catching up could observe epoch N+1 and try to fetch a checkpoint for epoch N from the previous committee, even when that committee consisted entirely of validators whose start_epoch was N. In that case, missing checkpoint availability is not necessarily a broken sync condition; it can be the first active committee bootstrap case.
> 
> Previously rpc_state_sync treated this as a fatal sync failure and consensus could end up in Sleeping. The fix teaches rpc_state_sync to treat missing previous-epoch checkpoints as bootstrap only when all of the following hold: our node starts in prev_epoch, local state has not reached prev_epoch yet, the previous committee consists entirely of validators activated in prev_epoch, and the peer response indicates temporary checkpoint unavailability rather than an arbitrary hard failure. Outside of that narrow case, checkpoint requirements remain unchanged and hard failures are still propagated.

Please, LMK if this seems valid from your POV. If so, I would like to suggest to express this policy somewhere in the consensus/epoch layer so that state sync doesn't have to infer it indirectly and can be told explicitly whether previous-epoch checkpoint sync is required, temporarily unavailable, or not applicable in the first-committee bootstrap case

0xskr33p:
Also, after state_sync fix, I pushed fixes for the two remaining failing tests, so the tests is now fully green

sdbondi:
@0xskr33p The problem makes sense, it was noted in the code before and never solved properly. Ideally sync should not start if it's obvious that there is nothing to sync, so I think some of this logic belongs in the epoch manager, which can track the "birthday" epoch. Then we simple check if we're past the birthday epoch in `check_sync`. I've added a commit for this. I think the explicit check for syncing from nodes which themselves are syncing is another good change - Thanks! Think this is ready for merge.

sdbondi (dismissed):

0xskr33p:
@sdbondi Thanks for the clarification and for the refactor - it looks way, way better! 🔥

0xskr33p:
@sdbondi jfyi – I'm still working on the problem, so I will move MR to draft temporarily. State sync fallback was a suitable temporary solution but I still need some more time before I can avoid going back to it

0xskr33p:
@sdbondi, gm!
I'll update this soon along with fmt fix and will ask for a rerun

0xskr33p:
Although maybe I should wait for the current test run - so far it’s not hanging like last time

0xskr33p:
@sdbondi The job got stuck at some point, so I pushed without waiting for it to finish. Could you please restart pipeline?

0xskr33p:
Looks like it’s ok now, and if I’m not mistaken, the remaining failure is something from yesterday on the development branch. I can take a look at it and submit a separate MR, since I’d like to get this one over the line 🙃

sdbondi (approved):
ACK 🚀 Thank you!

Issue #1957, opened 2026-04-01
wallet: authentication that is friendly for agents


sdbondi:
# Problem

Currently the only authentication mecahnism is webauthn. Agents are not able to authenticate on the user's behalf using this method, and it requires a user to copy a short lived token from the wallet. 

# Proposal 1

Allow the user to create a permission-scoped api key that agents can use to authenticate, the user provides the api key to their agent.
![Image](https://github.com/user-attachments/assets/61afad03-3126-4b34-83fe-6eb2d1561f14)

abisheik687:
I’d like to work on this bounty.

Plan:
- Extend the existing wallet daemon auth/JWT permission model with Admin-managed API keys rather than adding a parallel auth system.
- Store only hashed API keys, show the raw key once on creation, and support create/list/revoke with scoped permissions and timestamps.
- Allow agents to exchange an API key for a scoped JWT via `auth.request`, then rely on the existing permission checks for JSON-RPC calls.
- Add Rust client, JS client, wallet CLI/UI surfaces, integration coverage, and docs for the agent setup flow.

I’ll keep the PR scoped to the acceptance criteria and include explicit tests for non-admin rejection, scope rejection, and revocation.

abisheik687:
PR opened for this bounty: https://github.com/tari-project/tari-ootle/pull/2108

InFirePro:
Hi @sdbondi 👋

Picking this up. Sharing a high-level design before I start coding so we can align early — happy to iterate based on your feedback.

**Proposed design:**
- **Format:** `taro_<base32(32 bytes CSPRNG)>` prefix-tagged keys; only a blake2b hash with domain separator `tari.walletd.api_key.v1` is stored.
- **Storage:** new diesel migration `api_keys(id, name, hash, permissions, created_at, last_used_at, revoked_at)` in `tari_ootle_wallet_storage_sqlite`.
- **JSON-RPC:** `auth.create_api_key | list_api_keys | revoke_api_key` (admin-only), `auth.login_with_api_key` issues a scoped JWT via the existing `JwtApi`. Scopes are an arbitrary subset of `JrpcPermission`; granting `Admin` requires an explicit `allow_admin: true` flag in the create request.
- **Clients:** mirrored APIs in `tari_ootle_walletd_client` (Rust) and `@tari-project/wallet_jrpc_client` (JS), plus `with_api_key` / `loginWithApiKey` constructors that authenticate and cache the JWT.
[...]

PR #2108, opened 2026-05-07
feat(walletd): add scoped API keys for agent auth (#1957)

abisheik687:
Description
---
Adds Admin-managed, permission-scoped API keys for wallet daemon agent authentication. Raw API keys are returned only at creation, only hashes are persisted, and API-key-backed JWTs include the key id so revocation is checked immediately during auth checks.

Bounty issue: https://github.com/tari-project/tari-ootle/issues/1957

Motivation and Context
---
The wallet daemon currently requires WebAuthn or a manually copied short-lived token, which does not work well for AI agents. This extends the existing JSON-RPC permission/JWT model instead of adding a separate auth path.

Implemented:
- persistent wallet API key storage with create/list/revoke and timestamps
- auth.request API-key credential exchange for scoped JWTs
- Admin-only management endpoints and explicit Admin grant confirmation
- Rust wallet daemon client methods and authenticate-with-key helper
- JavaScript wallet daemon client methods and authenticateWithApiKey helper
- wallet CLI auth api-key create/list/revoke/authenticate commands
- wallet UI API key creation/list/revoke surface under access tokens
- docs for agent setup flow and API endpoints

How Has This Been Tested?
---
- git diff --cached --check
- Added storage coverage for create/list/touch/revoke API-key persistence.
- Added permission subset and AddressBook permission round-trip tests.

Local limitations: cargo/rustup are not installed in this Windows shell, so I could not run cargo +nightly-2025-06-25 fmt --all or cargo test locally. pnpm is present, but repo dependencies are not installed, so local prettier/tsc also could not run.

---
- cargo +nightly-2025-06-25 fmt --all
- cargo test -p tari_ootle_wallet_storage_sqlite api_keys_can_be_created_listed_touched_and_revoked
- cargo test -p tari_ootle_walletd_client permissions_subset_requires_every_requested_permission address_book_permission_round_trips_from_display
- cargo check -p tari_ootle_walletd -p tari_ootle_walletd_client -p tari_wallet_cli -p tari_ootle_wallet_storage_sqlite
- Generate TS bindings, then run the wallet web UI/client TypeScript build.
- Manual JSON-RPC smoke flow: authenticate as Admin, call auth.api_key_create, call auth.request with {ApiKey:{api_key:...}}, verify in-scope request succeeds, out-of-scope request fails, auth.api_key_revoke invalidates the key-backed JWT.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_walletd/src/handlers/auth/handlers.rs
+++ b/applications/tari_walletd/src/handlers/auth/handlers.rs
@@ -4,20 +4,34 @@
 //   Copyright 2023 The Tari Project
 //   SPDX-License-Identifier: BSD-3-Clause
 
+use std::time::{SystemTime, UNIX_EPOCH};
+
 use axum_extra::{extract::CookieJar, headers::authorization::Bearer};
 use axum_jrpc::JsonRpcResponse;
-use tari_ootle_wallet_sdk::models::AuthLoginRequestEvent;
+use rand::RngCore;
+use sha2::{Digest, Sha256};
+use tari_ootle_wallet_sdk::{
+    models::{AuthLoginRequestEvent, NewApiKeyModel},
+    storage::{WalletStore, WalletStoreReader, WalletStoreWriter},
+};
 use tari_ootle_walletd_client::{
-    permissions::JrpcPermission,
+    permissions::{JrpcPermission, JrpcPermissions},
     types::{
+        AuthApiKeyInfo,
+        AuthCreateApiKeyRequest,
+        AuthCreateApiKeyResponse,
         AuthGetMethodRequest,
         AuthGetMethodResponse,
+        AuthListApiKeysRequest,
+        AuthListApiKeysResponse,
         AuthListSessionsRequest,
         AuthListSessionsResponse,
         AuthLoginRequest,
         AuthLoginResponse,
         AuthMethod,
         AuthRefreshRequest,
+        AuthRevokeApiKeyRequest,
+        AuthRevokeApiKeyResponse,
         AuthRevokeTokenRequest,
         AuthRevokeTokenResponse,
         AuthSessionInfo,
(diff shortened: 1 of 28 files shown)


applications/tari_walletd/src/handlers/auth/handlers.rs
```diff
@@ -126,3 +224,75 @@ pub async fn handle_get_auth_method(
...
+        .credentials
+        .as_api_key()
+        .ok_or_else(|| unauthorized("API key credentials are required"))?;
+    let api_key_id = parse_api_key_id(&credentials.api_key)?;
+    let api_key = context
+        .wallet_sdk()
+        .store()
+        .with_read_tx(|tx| tx.api_keys_get_active(api_key_id))?;
+    let supplied_hash = hash_api_key(&credentials.api_key);
+    if !hashes_equal(&supplied_hash, &api_key.key_hash) {
```
sdbondi, line 242: `subtle` is already a transitive dep, I think we should use it even though your implementation looks correct.
```suggestion
    if !supplied_hash.ct_eq(&api_key.key_hash) {
```

applications/tari_walletd/src/handlers/auth/handlers.rs
```diff
@@ -93,6 +113,84 @@ pub async fn handle_revoke(
...
+    if request.permissions.is_empty() {
+        anyhow::bail!("At least one permission is required");
+    }
+    if request.permissions.contains(&JrpcPermission::Admin) && !request.allow_admin {
+        anyhow::bail!("Granting Admin to an API key requires explicit confirmation");
+    }
+
+    let id = uuid::Uuid::new_v4().to_string();
+    let secret = generate_api_key_secret();
+    let api_key = format!("twda_{}_{}", id, secret);
```
sdbondi, line 134: I think we can simplify this:

`twda_{base64_no_padding(32 random bytes)}` (zeroize this on drop)

We can insert the api key hash and get a new numeric ID (normal PRIMARY KEY AUTO INCREMENT, not a credential just an identifier)
and use that as the ID in the claim (revocation etc). For auth we do a unique lookup by the API key hash.

```rust
  base64::engine::general_purpose::URL_SAFE_NO_PAD.encode(&secret);
```

Does that make sense?

sdbondi (commented):
Thanks for the PR, looking good. I think we can simplify a little. I've not completely reviewed the PR yet, but submitting initial findings

sdbondi:
Thanks for the contribution and the effort that went into this — we had several strong submissions for the bounty and ultimately merged #2109. Hope to see you on a future one.

PR #2109, opened 2026-05-11
feat(walletd): permission-scoped API keys for agent authentication (closes #1957)

### LINKED ISSUE #1957: wallet: authentication that is friendly for agents

0xPepeSilvia:
## Summary

Closes #1957 — adds a long-lived, permission-scoped API key authentication path so AI agents (and other automated clients) can authenticate to the wallet daemon's JSON-RPC API without needing the WebAuthn interactive ceremony.

Agents call `auth.request` with `AuthCredentials::ApiKey(raw_key)` and receive a normal short-lived JWT scoped to the key's stored permissions. No parallel auth path, no permanent credential in the JWT itself — just a much simpler way to obtain the same kind of session a human gets via WebAuthn.

## Trust model

- Raw key bytes leave the daemon exactly **once** (the response of `auth.create_api_key`). Only a SHA-256 hex digest is persisted, so a database leak does not grant access.
- Scopes on a key are an arbitrary subset of `JrpcPermissions`. `Admin` is allowed as a grant but the request must set `confirm_admin: true` — a deliberate speed-bump so a UI checkbox can be the gate rather than the JSON-RPC payload alone.
- Revocation is **immediate**: storage filters `revoked_at IS NULL` inside `api_key_find_active_by_hash` so a revoked key can't authenticate even when a request is mid-flight.
- The auth path returns the same error and same timing for ''no such key'' and ''key revoked'' so an attacker can't enumerate valid hashes.
- The `permissions` field of the underlying login request is **ignored** for API-key auth — scopes always come from the persisted row, so an agent cannot self-promote.

## New JSON-RPC endpoints (all Admin-gated)

| Method | Purpose |
|---|---|
| `auth.create_api_key` | Mint a key. Returns the raw key exactly once. Rejects empty perms, empty name, Admin-without-confirm. |
| `auth.list_api_keys` | Enumerate active + revoked keys with `last_used_at` so admins can spot stale credentials. |
| `auth.revoke_api_key` | Soft-delete by id. Audit row preserved. |

## Test coverage (12 new tests)

**Storage layer** (`crates/wallet/storage_sqlite/tests/api_keys.rs`, 8 tests):
- insert + find round-trips
- unknown hash returns None (no enumeration leak)
- revoke is immediate for subsequent lookups
- double-revoke is a NotFound no-op (preserves original timestamp)
- touch_last_used updates timestamp
- touch on revoked id is no-error (race-safe with auth path)
- list_all includes active and revoked with sensible ordering

**Handler module** (`api_keys.rs`, 4 unit tests):
- Minted keys have correct shape and are unique
- Hashing is deterministic and distinguishes inputs
- Permission round-trip through `format_permissions` ↔ `JrpcPermissions::from_str`
- Permission formatting is deterministic regardless of input order

All 12/12 pass. Clippy clean with `-D warnings -A clippy::too_many_arguments` (the suppression covers pre-existing lints, nothing introduced here).

## Acceptance criteria from the issue

- [x] An Admin user can create a named, permission-scoped API key
- [x] Available scopes cover existing `JrpcPermissions`
- [x] Admin cannot grant scopes broader than `Admin`; granting `Admin` requires explicit `confirm_admin`
- [x] API key shown exactly once at creation; only a hash is persisted
- [x] Agents authenticate via `auth.request` with `AuthCredentials::ApiKey` and receive a scoped session JWT
- [x] Admin can list active keys (name, scopes, created/last-used) and revoke; revocation is immediate
- [x] Non-admin sessions rejected at key management endpoints with permission error
- [x] Requests outside the key's granted scopes rejected with clear permission error
- [x] Rust client supports create/list/revoke + authenticate-with-key
- [x] JavaScript client supports the same surface
- [x] Integration tests cover correctness invariants
- [x] Documentation update covering the agent setup flow (`docs/developer-docs/src/content/docs/guides/agent-api-keys.mdx`)

## Bounty

Bounty payment address: `123JbAxCZ4Vrh4jCjFLXTu4z8sLeggUR87fejwNunsB55NMBSH9RpkjNiw1WL2GkKr8JhqRZhbrsSREchGanchw2Nhb`

## Test plan

- [ ] CI passes (cargo build / test / clippy / fmt)
- [ ] Manual: run the daemon, mint a key via WebAuthn-admin client, authenticate a second client with the raw key, confirm scope-bounded access and immediate revocation
- [ ] Manual: confirm JS bindings export and `authenticateWithApiKey` works end-to-end


0xPepeSilvia:
Audited the PR against the issue's 12 acceptance criteria and found two gaps. Closed in 16bf0f145:

1. **Wallet CLI commands.** The criterion says ''wallet UI/CLI'' — I had the JSON-RPC endpoints and the Rust/JS clients done, but no `tari_ootle_wallet_cli` subcommands. Added `auth api-key {create,list,revoke}` with `--confirm-admin` gating on the create path.

2. **Handler-level e2e tests.** The criterion lists five specific scenarios (`admin key creation`, `non-admin attempt to create a key`, `authenticated calls within scope`, `rejection out of scope`, `revocation`). The initial commit had storage-layer invariants tests but didn't directly exercise these. Added an `e2e_tests` module in `api_keys.rs` that ties together the in-memory `SqliteWalletStore` + `authenticate_api_key` + `JwtApi` to cover each scenario by name. 5/5 pass.

Updated total: 17 new tests (9 handler + 8 storage). Clippy clean.

0xPepeSilvia:
Second-pass audit with fresh eyes turned up two real issues. Both fixed in f8b7278f2.

**Critical: refresh cookie bypassed revocation.**

The original `handle_login_request` was issuing the in-memory refresh cookie for API-key-authenticated sessions, same as for WebAuthn. But `handle_token_refresh` only consults `RefreshTokenStore`, not the `api_keys` table. So a revoked API key could keep minting fresh JWTs for the full refresh-token-store lifetime (1h default) by riding the cookie — directly breaking the issue's ''revocation is immediate'' acceptance criterion.

Fix: don't issue a refresh cookie for API-key auth. Agent re-presents its key on JWT renewal (one DB lookup per `jwt_expiry` window — 5min default). The active-row filter on `api_key_find_active_by_hash` means revocation takes effect on the very next `auth.request` call. Added `api_key_credentials_must_not_be_issued_a_refresh_cookie` regression test.

**Ship-blocker: TS barrel re-exports missing.**

`bindings/src/wallet-types.ts` is the barrel the JS client imports from. `cargo test export_bindings` writes the individual type files but does NOT update the barrel — that's only done by `bindings/build.sh`. Without the barrel update the JS client's `import { AuthCreateApiKeyRequest } from "@tari-project/ootle-ts-bindings"` would fail to resolve at `tsc`. Added explicit re-exports for the 7 new types.

10 tests still pass, clippy + fmt clean.

applications/tari_walletd/src/handlers/auth/api_keys.rs
```diff
@@ -0,0 +1,345 @@
...
+
+    Ok(AuthRevokeApiKeyResponse {})
+}
+
+/// Round-trip a `JrpcPermissions` back to the comma-separated textual form
+/// that `JrpcPermissions::from_str` accepts. We keep the order
+/// deterministic (sorted by display form) so a re-list returns the same
+/// string the user supplied at creation time.
+fn format_permissions(p: &JrpcPermissions) -> String {
+    let mut parts: Vec<String> = p.clone().into_vec().iter().map(|p| p.to_string()).collect();
```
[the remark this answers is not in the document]
0xPepeSilvia: Good catch. The suggested form needed a small upstream change — `JrpcPermissions` didn't expose an `iter()` and its inner `HashSet` is private. Added `JrpcPermissions::iter()` in the client crate and used the clean form here. Done in 624bc051b.

applications/tari_walletd/Cargo.toml
```diff
@@ -33,6 +33,7 @@ tari_utilities = { workspace = true }
 
 async-trait = { workspace = true }
 anyhow = { workspace = true }
+base64 = "0.22.1"
```
sdbondi:
```suggestion
base64 = { workspace = true }
```

crates/wallet/storage_sqlite/src/writer.rs
```diff
@@ -1781,6 +1782,88 @@ impl WalletStoreWriter for WriteTransaction<'_> {
...
+        let now = time::OffsetDateTime::now_utc();
+        let now = time::PrimitiveDateTime::new(now.date(), now.time());
+
+        // `affected = 0` is not an error here: a stale `last_used` bump on
+        // an already-revoked key is a harmless no-op. Auth has already
+        // succeeded so we must not surface a write failure to the caller.
+        let _ = diesel::update(api_keys::table.filter(api_keys::id.eq(id)))
+            .set(models::ApiKeyLastUsedChangeset {
+                last_used_at: Some(now),
+            })
```
sdbondi: If the key has been revoked, it should not update last_used_at


```suggestion
.filter(api_keys::revoked_at.is_null())
            })
```

docs/developer-docs/src/content/docs/guides/agent-api-keys.mdx
```diff
@@ -0,0 +1,184 @@
...
+}
+
+// Admin: revoke a key (effective immediately).
+admin_client
+    .auth_revoke_api_key(AuthRevokeApiKeyRequest { id: 7 })
+    .await?;
+```
+
+Revoke is immediate: the storage layer filters revoked rows out of the authentication lookup, so any in-flight
+authentication using a just-revoked key will fail at the next request.
```
sdbondi, line 171: Is this strictly true?

If the api key was used to obtain a JWT, that JWT would still be valid until it expires.

applications/tari_walletd/src/handlers/auth/api_keys.rs
```diff
@@ -0,0 +1,345 @@
...
+    // transaction to bump `last_used_at`. SQLite supports concurrent
+    // readers but a single writer; holding a read tx alongside a write tx
+    // on the same connection-pool slot would deadlock under load.
+    let api_key = {
+        let mut tx = store
+            .create_read_tx()
+            .context("Failed to open read transaction for API key lookup")?;
+        let row = tx.api_key_find_active_by_hash(&hash)?;
+        row.ok_or_else(|| anyhow!("Authentication failed: API key is invalid or revoked"))?
+    };
```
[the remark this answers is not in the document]
sdbondi: Fair assesment - not worth fixing because as mentioned a timing attack is not practical with secure key entropy.

docs/developer-docs/src/content/docs/guides/agent-api-keys.mdx
sdbondi: Maybe binding the api key id to the JWT claims would make sense for immediate revocation (this is what https://github.com/tari-project/tari-ootle/pull/2108 does)

0xPepeSilvia: Good catch, you're right. The current "revocation is immediate" framing is only true at the `auth.request` boundary: the active-row filter on `api_key_find_active_by_hash` means the next credential exchange rejects a revoked key. But a JWT minted off a key before revocation keeps working until `jwt_expiry` (5 min default), and the docs read like that window doesn't exist.

Two options for closing this. Happy with either, would rather not duplicate effort against #2108:

1. **Doc-only fix.** Tighten line 171 to describe the actual window honestly: revocation takes effect on the very next `auth.request` call (one DB lookup per `jwt_expiry`), and outstanding JWTs survive until their existing expiry. Note that immediate JWT-level revocation is being explored on #2108 via claim binding.

2. **Implement claim binding here.** Add an optional `api_key_id` to the JWT claims (only set on API-key-authenticated sessions), and have the JWT verification path check `api_key_find_active_by_id` on each authenticated request. Brings #2109 in line with #2108's revocation model and lets the "immediate" wording stand truthfully.

I'd lean toward (1) on this PR since #2108 is already doing the claim-binding work, but if you'd rather have both PRs converge on the (2) behaviour I'll do that here instead. Which do you prefer?

applications/tari_walletd/Cargo.toml
0xPepeSilvia: Done in c092fc504. Already declared at the workspace root, so the local crate just tracks it now.

crates/wallet/storage_sqlite/src/writer.rs
0xPepeSilvia: Good catch, done in c092fc504. Added the `revoked_at IS NULL` filter to the update query. Also strengthened the existing `touch_last_used_on_revoked_key_is_noop_not_error` test to assert the revoked row's `last_used_at` actually stays None (the old test only checked the call didn't error).

applications/tari_walletd/web_ui/src/App.tsx
```diff
@@ -232,6 +233,7 @@ function App() {
             path="access-tokens"
             element={<GuardedRoute redirect="/access-tokens" component={AccessTokensLayout} />}
           />
+          <Route path="api-keys" element={<GuardedRoute redirect="/api-keys" component={ApiKeysLayout} />} />
```
sdbondi, line 236: Is there a link to this page anywhere?

0xPepeSilvia: Good catch, it wasn't linked anywhere. Added a sidebar entry in 4582421e0 (`MainListItems`, with a `LuKeyRound` icon, slotted between Address Book and Settings). Picked the same pattern Address Book uses so the placement is consistent.

sdbondi:
Apologies that expiry was not part of the aceptance criteria, you dont have to implement it, but I'd like the initial database to include an optional expiry even if for now it's always set to None/null. 

I'd like it if the permissions were selectable rather than typed in by the user. It doesnt have to fuly support the fine grain permissions, e.g. AccountList(some component), we can just support AccountList - I think I want to rework permissions at some point since I've never liked how they were implemented, but that is for another day.

Current UI:

image

Mock up:

image

0xPepeSilvia:
Both addressed in fad930e8c.

**1. Optional expiry column.** Added `expires_at DATETIME NULL` to the existing migration (rather than a follow-up migration since the PR hasn't merged), threaded through the diesel schema, storage `ApiKey` Queryable, SDK `ApiKey` model, both `map_api_key_row`/`api_key_from_row` converters, the `IssuedApiKey` wire type as `Option<i64>` unix-seconds, and all three `IssuedApiKey` construction sites. Always written as NULL by current code; the active-row filter on `api_key_find_active_by_hash` does NOT yet check it, with a comment in the migration documenting the deferred enforcement semantics so adding the check later is a one-line code change.

**2. Permissions picker.** Replaced the free-text input with a labelled checkbox list. The 10 unparameterised forms `JrpcPermissions::from_str` accepts are surfaced as individual checkboxes with a short description each (`AccountInfo`, `AccountList`, `KeyList`, `TransactionGet`, `TransactionSend`, `SubstatesRead`, `TemplatesRead`, `NftGetOwnershipProof`, `GetNft`, `StartWebrtc`). Variants that require a substate/resource/component address are intentionally out of scope here — matches your "we can just support AccountList" note.

`Admin` is in its own section below a divider with warning-coloured styling, and grants only after the user ticks both the Admin checkbox *and* a separate "I understand" confirmation checkbox that only appears once Admin is selected. The Create button stays disabled until something is selected.

Submit order is stable (`PERMISSION_OPTIONS` order, regardless of click order) so the wire payload is deterministic.

applications/tari_walletd/src/handlers/auth/api_keys.rs
```diff
@@ -0,0 +1,736 @@
...
+    let permissions_str = format_permissions(&permissions);
+
+    let stored = {
+        let mut tx = context
+            .wallet_sdk()
+            .store()
+            .create_write_tx()
+            .context("Failed to open write transaction")?;
+        let row = tx
+            .api_key_insert(&request.name, &hash, &permissions_str)
```
sdbondi:
```suggestion
            .api_key_insert(request.name.trim(), &hash, &permissions_str)
```

sdbondi (approved):
Thanks for the great work on this — the implementation is thorough, well-tested, and the responsiveness to review feedback has been excellent. Approved.

PR #2146, opened 2026-05-21
feat(ootle-wasm): decrypt viewable balances from persisted UTXOs

martinserts:
Description
---

Adds two WASM exports — `decryptViewableBalance` and `decryptViewableBalanceBatch` —
  that read the value of a viewable-resource UTXO directly from what the chain stores,
  using only the holder's view key.

Motivation and Context
---

The viewable-balance read/audit path was unreachable through the WASM bridge: the
  only existing decrypt export needed proof data that the chain discards after a
  transaction is accepted, so it could never be called against a real on-chain UTXO.
  This blocked every consumer (Python `ootle`, JS `@tari-project/ootle-wasm`). The new
  exports are additive and don't change the existing one.

How Has This Been Tested?
---

Unit tests covering: reading a real (persisted) UTXO returns the correct value and
  matches the old path's result, out-of-range values return nothing, and the batch
  variant handles a mix of readable and out-of-range entries. Verified it builds for
  the WASM target.

---

  Run `cargo test -p ootle-wasm-core stealth::viewable_balance`.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/crates/ootle_wasm/core/src/stealth/viewable_balance.rs
+++ b/crates/ootle_wasm/core/src/stealth/viewable_balance.rs
@@ -5,7 +5,12 @@
 
 use ootle_byte_type::ConvertFromByteType;
 use tari_crypto::ristretto::pedersen::PedersenCommitment;
-use tari_engine_types::crypto::{ValueLookupTable, validate_elgamal_verifiable_balance_proof};
+use tari_engine_types::crypto::{
+    ElgamalVerifiableBalance,
+    ElgamalVerifiableBalanceBytes,
+    ValueLookupTable,
+    validate_elgamal_verifiable_balance_proof,
+};
 use tari_ootle_wallet_crypto::{
     GenerateValueLookup,
     viewable_balance_proof::generate_elgamal_viewable_balance_proof as crypto_generate,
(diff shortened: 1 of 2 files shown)


crates/ootle_wasm/core/src/stealth/viewable_balance.rs
```diff
@@ -86,6 +91,86 @@ pub fn decrypt_elgamal_viewable_balance(
...
+
+    let verifiable = ElgamalVerifiableBalance {
+        encrypted: public_key_from_bytes(encrypted)?,
+        public_nonce: public_key_from_bytes(public_nonce)?,
+    };
+    let view_sk = secret_key_from_bytes(view_secret_key)?;
+
+    // GenerateValueLookup computes `v.G` on-the-fly (no precomputed file), so this works in WASM.
+    let mut lookup = GenerateValueLookup;
+    verifiable
```
sdbondi, line 130: This could take years :)

sdbondi (commented):
Elgamal value decryption is probably not needed for this environment and not needed in general (only for issuers of stealth resources who have set a view key). It's mostly implemented in the wallet daemon as a proof of concept and the wallet daemon uses a pre generated lookup file and mmap to greatly improve performance (and even that is slow if we have to scan a large numeric space). Generating value commitments on the fly makes this incredibly slow so I worry that even adding this API would result in freezes and crashes from people trying to use it expecting it to just work. 

I assume we're only adding this for completeness - in which case, is it work the extra binary size of the wasm?

For reference, just generating the `v.H` values on the fly would take:

| value `v`                       | native (single core) | WASM (browser) |
|---------------------------------|----------------------|----------------|
| 10^6  (~1 TARI in microtari)    | ~10 sec              | ~2 min         |
| 10^9  (~1000 TARI)              | ~3 hours             | ~30 hours      |
| 10^12 (~1 M TARI)               | ~115 days            | ~3 years       |
| 10^15                           | ~300 years           | millennia      |
| u64::MAX (~1.84 × 10^19)        | ~5.8 M years         | ~58 M years    |

sdbondi:
Closed after discussion with @martinserts

PR #2261, opened 2026-06-18
feat(walletd): add account balance change history

### LINKED ISSUE #1949: walletd: Log of balance changes

tzmWW:
## Summary

- persist one structured balance-change row per changed vault, with atomic vault/log writes, non-zero enforcement, signed deltas, deterministic pagination, and transaction replay idempotency
- attribute finalized transactions, ordinary scans, and the real account-recovery path with a typed, internally tagged `BalanceChangeSource`; include NFT counts while excluding stealth-only balances without vaults
- promote the newest exact-balance unlinked scan when the corresponding finalized transaction arrives, preserving the transaction link even when an older scan reached the same balance
- expose an account-authorized `accounts.get_balance_changes` RPC with resource/transaction filters, filtered totals, generated TypeScript bindings, and Rust/JavaScript client methods
- add a shared React history table for account-wide history and resource-filtered vault dialogs, including safe transaction links and graceful scan/recovery/unknown labels

## Validation

- `cargo build --workspace`
- targeted SQLite, SDK, source-plumbing, client serialization, and real walletd handler tests
- `cargo lints clippy --all-targets --all-features`
- `cargo +nightly-2025-12-05 fmt --all -- --check`
- full TypeScript binding generation/export tests and bindings/client builds
- `pnpm --recursive run build`
- `./scripts/file_license_check.sh`
- `git diff --check`
- browser QA for account-wide and resource-filtered views, revealed/confidential and NFT rows, transaction links, empty/unknown states, and narrow layouts

`applications/tari_walletd/web_ui/dist` was rebuilt successfully for verification and intentionally remains uncommitted per maintainer guidance.

Fixes #1949


tzmWW:
Just to confirm @sdbondi, should `web_ui/dist` be committed here? And is that the source of the CR-too_long label?

sdbondi:
@tzmWW Thanks for the PR - no dist folders should be committed (weird that it's there, should be excluded by applications/tari_walletd/webui/.gitignore). The too long label applies (IIRC) for any PR over 1000 lines, but no need to worry about it, PR is still mergable with this tag

tzmWW:
> @tzmWW Thanks for the PR - no dist folders should be committed (weird that it's there, should be excluded by applications/tari_walletd/webui/.gitignore). The too long label applies (IIRC) for any PR over 1000 lines, but no need to worry about it, PR is still mergable with this tag

Corrected it and pushed some minor fixes

tzmWW:
Fixed:
- Added a transaction-history existence check and used it in `scanner.rs` before NFT/confidential-output refresh side effects, so replayed transactions no longer mutate NFT state after the vault transaction row already exists.
- Tightened scan-to-transaction promotion in `storage_sqlite/src/writer.rs`: promotion now only happens when exactly one matching final-balance history row exists and it is an unlinked Scan.
- Updated `BalanceChangeHistory.tsx` to mask deltas and new balances when `showBalance` is off.

tzmWW:
Final hardening is pushed on `f7d79fdf4`:

- transaction attribution now promotes the newest exact-balance unlinked scan, with regression coverage for repeated historical balances and replay
- the walletd handler test no longer fails the full clippy gate
- the bindings package is bumped to `1.42.0` after the updated base consumed `1.41.0`

Local validation passes: workspace build, full clippy, targeted balance-history tests, bindings generation/export, recursive pnpm build, nightly fmt, licenses, and diff checks. The wallet UI `dist` was rebuilt for verification and remains uncommitted as requested.

@sdbondi Could you approve the fork CI/PR workflows for the latest head when convenient?

crates/wallet/storage_sqlite/src/models/balance_change.rs
```diff
@@ -0,0 +1,98 @@
...
+            "scan" => BalanceChangeSource::Scan,
+            "recovery" => BalanceChangeSource::Recovery,
+            value => {
+                return Err(WalletStorageError::DecodingError {
+                    operation: OPERATION,
+                    item: "source_type",
+                    details: format!("unknown balance change source '{value}'"),
+                });
+            },
+        };
```
sdbondi: nit: implement FromStr

sdbondi:
## Review — account balance change history

Overall a careful, well-tested PR. The schema, FK setup (`PRAGMA foreign_keys=ON` is set; `transactions.transaction_id` is UNIQUE so the FK is valid; the tx row always exists before a `Transaction`-sourced insert because `TransactionFinalized` only fires for locally-stored pending txs), diesel column/field ordering, and timestamp parsing all check out. The notable issues are in the **attribution/promotion design** rather than crash bugs.

Things I checked and consider non-issues: negative-`Amount` in `signed_delta` (`Amount` is `u128`), the FK firing on a missing tx row, and replay skipping confidential/NFT work (the balance-change row is written *last*, so if it exists those steps already succeeded).

### Correctness / design (highest value)

**1. Scan→Transaction promotion is a brittle balance-string heuristic that can mis-attribute the delta.**
`balance_changes_promote_scan_to_transaction` (`crates/wallet/storage_sqlite/src/writer.rs`) correlates a finalizing transaction to a prior scan purely by matching the scan row's `revealed_after`/`confidential_after` against the *current* vault balance, then promotes the **latest matching** scan — with no causal link to the transaction. Under a round-trip balance (10→20→10) or interleaved scans this promotes the wrong row. The test `transaction_source_promotes_latest_matching_scan_for_same_balance` enshrines it: after scans 10/20/10, the arriving transaction is attributed `revealed_delta == "-10"` even though there's no evidence the transaction caused a −10 change. The principled fix is to carry the transaction id at the source (the scan path doesn't know it) rather than reverse-engineering it from balance equality.

**2. A finalized transaction with no net balance change (and no matching scan to promote) is silently absent from history.** In `AccountsApi::update_vault_balance_and_record_change` (`crates/wallet/sdk/src/apis/accounts.rs`), when the new balance equals the current balance the code takes the promote branch; if no unlinked scan row matches, it returns `Ok(false)` and writes **no** transaction row. A real transaction that touches a vault without changing its net balance never appears in the transaction-filtered history. Defensible as "no balance change → no row", but worth a deliberate decision for a feature whose `Source` column links to `/transactions/<id>`.

**3. Promotion excludes `recovery` rows.** `balance_changes_promote_scan_to_transaction` filters `source_type = 'scan'` only. A recovery-then-transaction sequence at the same balance promotes nothing and inserts nothing, leaving the transaction unlinked. New gap, because recovery is now its own source (it used to share the Scan path).

**4. Source attribution is account-refresh-scoped, not per-event.** `refresh_account_with_source` (`crates/wallet/sdk_services/src/account_monitor/scanner.rs`) passes one `BalanceChangeSource` to `refresh_vault` for *every* vault it touches. A periodic `Scan`/`Recovery` refresh that runs while a transaction's effects land will stamp the change `Scan`/`Recovery` unless promotion later corrects it — so the attribution column is advisory rather than authoritative for non-Transaction rows.

### UI / behavior

**5. History table flickers to a spinner on every page / rows-per-page change.** `useAccountsGetBalanceChanges` (`web_ui/src/services/api/hooks/useAccounts.ts`) has no `placeholderData: keepPreviousData`, and the `queryKey` includes `offset`/`limit`, so each pagination action is a brand-new query (`isLoading=true, isRefetching=false`) and `FetchStatusCheck`'s `isLoading && !isRefetching` hides the whole table. Add `placeholderData: keepPreviousData`.

**6. `get` and `count` run as two separate read transactions.** `handle_get_balance_changes` (`applications/tari_walletd/src/handlers/accounts.rs`) calls `get_balance_changes` then `count_balance_changes`; in the SQLite layer each opens its own read tx and re-resolves `account_id`. Under concurrent monitor writes (amplified by the 5 s poll) `total` can disagree with the returned page (off-by-one pager / empty last page). One tx returning both SELECT + COUNT would fix consistency and halve the work.

**7. The `source_type` / `transaction_id` request filters are dead UI plumbing.** The hook accepts `sourceType` and hard-codes `transaction_id: null`, but no component ever passes `sourceType` — so the backend filters (fully implemented in Rust + SQL + bindings) are unreachable from the web UI. Either wire a control or drop the param.

**8. Every mounted history view polls every 5 s indefinitely.** `refetchInterval: 5000` is hard-coded, and `BalanceChangeHistory` renders both inline on `AccountDetails` *and* inside each open dialog (from `AccountDetails.tsx` and `Tokens.tsx`). Several `get_balance_changes` (each = two read txs, see #6) fire every 5 s and aren't paused when hidden. History rows are immutable, so invalidating on the existing balance-refresh path would avoid the steady-state load.

### Cleanup / altitude

**9. Stored signed-delta is redundant state.** `revealed_delta`/`confidential_delta` (migration `up.sql` + `BalanceChange`) are fully derivable from the before/after columns and are *also* re-derived in the TS layer via `BigInt`. They force an extra `CHECK` to stay consistent and four columns to move together in any future amend path. Consider deriving on read.

**10. Duplicated account-id-by-address resolution.** The two new reader methods (`crates/wallet/storage_sqlite/src/reader.rs`) each copy the same ~12-line `accounts::table.select(id)…optional().ok_or(NotFound)` block (already repeated ~13× in the file). Extract a private `account_id_by_address` helper.

**11. `BalanceChangeSource` and `BalanceChangeSourceType` are near-duplicate enums** (`crates/wallet/sdk/src/models/balance_change.rs`) with two identical `as_key_str` mappings, themselves duplicating the SQL `CHECK` list. `SourceType` could derive from `Source`.

**12. Dead zero-change guard in `balance_changes_insert`** (`crates/wallet/storage_sqlite/src/writer.rs`): the equality → `bad_query` branch is unreachable because the only caller already returns early on equal balances and the SQL `CHECK` enforces it — a third encoding of the same invariant.

**13. TS duplication.** `formatSignedDelta` re-implements magnitude+symbol formatting that `formatCurrency`/`bigintToDecimalString` already provide (it only needs the sign prefix), and the "View changes" button + dialog wiring is copy-pasted verbatim into both `AccountDetails.tsx` and `Tokens.tsx`'s `BalanceRow`.

### Latent / minor

**14. Audit history is less durable than the data it audits.** Rows `ON DELETE CASCADE` with their vault/account, so re-importing an account or deleting+re-adding a vault silently purges history. Meanwhile `transaction_id`'s FK has **no** `ON DELETE` action (RESTRICT) — an undeclared asymmetry that would block any future `transactions` deletion. Pick a deliberate policy (e.g. `SET NULL` for the tx link).

**15. `token_symbol`/`divisibility` are read from the live `vaults` join, not stored on the row** (`reader.rs`), so historical entries retroactively re-render with a vault's current metadata if symbol/divisibility ever changes.

Also worth a quick look (low confidence): the new `balance_change_handler_tests` boots full `spawn_services` against a non-existent indexer (`127.0.0.1:18300`) with a 5 s shutdown-timeout `.unwrap()` — the test pre-seeds data and doesn't need the services, so this adds connection-refused noise and CI flakiness risk; and the inner `NotFound` from the SDK account-id lookup may surface as a 500 rather than the structured 404 the handler intends (dead today since `get_account` already resolved it).

---

**Verdict:** No crash/data-loss bug that blocks merge. The promotion/attribution items (#1–#4) are the substantive ones — a design trade-off (heuristic correlation, account-scoped source) that's safe today but will produce confusing/incorrect attribution under round-trips and interleaving, worth a decision before this becomes a user-facing "audit" feature. The UI fixes (#5–#7) are quick wins.

tzmWW:
## Implementation Changes

1. Redesigned the unmerged balance-history migration to store account/vault addresses, vault version, resource metadata, and balances directly. Removed cascading FKs and stored delta columns; retained transaction IDs durably and derived signed deltas on read.
2. Implemented a unique vault-address/version key. When a finalized transaction arrives, atomically find or promote only the row for that exact vault version, whether Scan or Recovery, and skip replay side effects. Removed balance-string promotion and its obsolete tests.
3. Passed vault version through `refresh_vault`, `AccountsApi::update_vault_balance_and_record_change`, and storage insertion. Document Scan/Recovery as observation contexts; zero-net transactions remain absent per acceptance criteria.
4. Made `BalanceChangeSource::as_key_str` delegate through `BalanceChangeSourceType`; removed the redundant insertion guard and duplicated account-ID lookup made obsolete by snapshot addressing.
5. Introduced an SDK `BalanceChangePage { changes, total }`. Replaced separate get/count APIs with one SQLite read transaction returning both, while preserving the existing JSON-RPC request and response wire shapes.
6. Hardened the handler test’s structured 404 assertion and replaced full network-facing service startup with dormant actor handles plus the minimal idle worker required by `HandlerContext`.
7. Added `keepPreviousData` pagination behavior, source and validated transaction-ID filter controls, and reset pagination when filters change.
8. Removed history’s five-second polling. Invalidate all active history queries for an account only when the existing balance poll observes changed balance data.
9. Extracted the duplicated “View changes” button/dialog state into one shared action component and reuse `formatCurrency` for signed-delta formatting.

## Tests Passed Successfully

- Storage: exact version attribution across round trips, Recovery promotion, mismatched versions, replay idempotence, derived deltas, metadata snapshots, deletion durability, filters, and atomic page totals.
- SDK/services: multi-account/vault transactions, non-zero-only behavior, rollback, source plumbing, and replay-side-effect protection.
- Walletd: authorization, filtering, pagination totals, and structured not-found behavior without indexer connections.
- Workspace build, full clippy, nightly formatting, license and machete gates, recursive pnpm build, explicit wallet UI Prettier check, and `git diff --check`.
- Browser-check source/transaction filtering, pagination without spinner flicker, account/resource views, privacy masking, and absence of steady history polling.

_P.S. GitGuardian now fails because of placeholder secrets in one of the crates. Can be ignored imo._

sdbondi:
### Balance-change log doesn't capture confidential balances received via the async UTXO scan flow

Flagging a design-level omission (applies identically to the parallel #1949 implementation in #2257), as it's more fundamental than the line-level findings.

Balance-change recording is hooked **only** into `refresh_vault` — i.e. transaction finalization (`process_result`) and the periodic scan when a vault's on-chain substate **version advances**. The confidential balance is read via `get_unspent_balance(vault)` *after* `verify_and_update_confidential_outputs(...)` has recovered our outputs, so confidential deltas **are** correctly logged for:

- transactions we submit/process (outputs recovered synchronously at finalization, before the balance read), and
- changes that bump the vault substate version.

What's **not** covered: confidential/stealth UTXOs picked up via the **async stealth UTXO scan flow** (one-sided payments from others, where we didn't submit the transaction):

- the scanner/recovery worker writes the outputs into `confidential_outputs` (changing `get_unspent_balance`) **without bumping the vault substate version** and **without calling `refresh_vault` / `update_vault_balance_and_record_change` / the balance-change writer**, and
- the resulting `UtxoRecovered` / `UtxoSpent` / `UtxoRecoveryCompleted` events are ignored by the account monitor (`on_event → … => {}`).

So an inbound stealth confidential receipt changes the wallet's confidential balance but produces **no history row** — and won't be picked up later unless an unrelated vault-version bump happens to run `refresh_vault` and capture the now-higher balance (at which point it's mis-attributed to whatever triggered that refresh). The version-keyed attribution rework doesn't help here, since the vault version doesn't move when these outputs are scanned.

Net: the log is effectively a **revealed**-balance history; confidential changes are only captured when they coincide with a transaction we process or a vault-version change. Closing it needs a recording hook on the confidential-output path itself — e.g. handle `UtxoRecovered`/`UtxoSpent` in the monitor, or record the delta inside the UTXO recovery worker after it updates `confidential_outputs`.

tzmWW:
## Stealth UTXO balance changes + Hardening

Thanks for flagging the async stealth UTXO gap — that's now closed, along with a few related issues turned up while auditing the surface.

### Stealth UTXO recording (the flagged gap)
- `UtxoRecovered / UtxoSpent` are now handled in the account monitor. Each event resolves the vault for (`account, resource`) and records a Scan-sourced balance change; vault-less stealth resources and non-stealth vaults are skipped.
- Stealth vault balances are now read from the account/resource-scoped `stealth_outputs` table (what the UI displays), not `confidential_outputs` (which stays empty for stealth resources). So the recorded `confidential_after` no longer diverges from the displayed balance.
- The last observed `vault_version` is persisted on the vaults row (new migration), so event-driven updates key on the correct version without a network fetch — a one-sided payment no longer needs a vault substate version bump to be recorded.

### Same-version history semantics
- `Scan/Recovery` writes at an already-recorded version now upsert the row's after-balances (keeping the original `before`), so multiple recoveries at one version collapse into one cumulative row.
- A cumulative net-zero round-trip deletes the row (honouring the "zero-net → no row" rule).
- Transaction-attributed rows are never clobbered; a duplicate finalized transaction remains a replay no-op.
- The vault balance snapshot is updated even when the history insert is a no-op, fixing a latent stale-snapshot regression.

### Other hardening
- `accounts.get_balance_changes` now caps page size at 200 (was an unbounded u32).

## Tests added

- Same-version upsert + net-zero deletion, transaction-row immutability, SDK same-version snapshot update, account/resource-scoped stealth balance helper, and handler limit cap. 
- Targeted storage/SDK/walletd tests, clippy, nightly fmt, and git diff --check pass. 
- No wire/binding shape changed (`Scan` is reused; `vault_version` is storage-only).

One known limitation worth stating explicitly: multiple stealth transitions at the same vault version collapse into a single cumulative history row, since the model remains one-row-per-(vault, version). Per-UTXO granularity would break the snapshot model and converges to the same net under serialized event processing.

crates/wallet/storage_sqlite/migrations/2026-06-18-160000_account_balance_changes/up.sql
```diff
@@ -0,0 +1,35 @@
+CREATE TABLE account_balance_changes
+(
+    id                  INTEGER  NOT NULL PRIMARY KEY AUTOINCREMENT,
+    account_address     TEXT     NOT NULL,
+    vault_address       TEXT     NOT NULL,
```
sdbondi: Vaults do not necessarily exist for a resource - you'll need to key this table by account and resource - probably (account_id, resource_id)

sdbondi (commented):
I still dont see balance changes when doing stealth transfers.

How I test on a fresh wallet:
1. create an account
2. claim faucet funds
3. transfer from revealed to stealth utxo
4. create another account + faucet funds
5. stealth transfer from second account to first account

I should see that 2 confidential balance increases on the first account and 1 balance decrease each on the first and second accounts.

tzmWW:
- Made balance history account/resource keyed with nullable vault metadata 
- Added FromStr for BalanceChangeSourceType and recorded stealth resource changes after UTXO state updates
- Preserved transaction replay protection per account/resource/transaction, regenerated bindings, and made the UI expose history actions for vaultless stealth rows

sdbondi:
## Follow-up review — stealth balance-change commits (`a5ff76a` + `d7a11ac`)

Reviewed the two new balance-change commits since the last round (the rest of the range is development merges): recording stealth UTXO/resource balance changes via `UtxoRecovered`/`UtxoSpent` + the tx `verify_and_update_outputs` path, reading stealth balances from `stealth_outputs`, storing `vault_version` on `vaults`, making `vault_address`/`vault_version` nullable (stealth resources with no vault), and the new upsert / net-zero-delete / never-clobber-transaction writer path.

**Good:** decoupling the vault-snapshot update from the history insert fixes the latent coupling flagged last round; the page limit is capped at 200; `vaults_update` callers all pass the correct version; `verify_and_update_outputs`'s return-type change has a single caller; UtxoSpent/Recovered ordering is fine (outputs commit before the driving event).

### 1. [Med] Balance-change read queries no longer hit an index

`balance_changes_get_page_by_account` (and its count query) filter by `account_address`/`resource_address` and `ORDER BY created_at DESC, id DESC`, but this commit re-keyed both composite ordering indexes to start with `account_id`/`resource_id`:

```
account_created_idx           (account_id, created_at DESC, id DESC)
account_resource_created_idx  (account_id, resource_id, created_at DESC, id DESC)
```

There is no index on `account_address` (only `vault_address` and `transaction_id` have single-column indexes), so every history page load and every `total` count is now a full-table scan + filesort — reading all accounts' rows to serve one account. `balance_changes_insert` already resolves `account_id`/`resource_id`, so the fix is to filter the read path on those ids (matching the indexes), or keep the ordering indexes on the address columns.

### 2. [Med] A stealth balance change for an uncached, vault-less resource aborts the whole event/transaction loop

`record_resource_balance_change`'s no-vault branch does `tx.resources_get(&resource_address)?`, and `balance_changes_insert` resolves `account_id`/`resource_id` with `.first::<i32>()` (not `.optional()`). If the wallet receives a stealth output for a resource that has no vault yet *and* whose `Resource` substate was never cached (stealth outputs store the resource address as raw text and don't require a `resources` row), the lookup returns `NotFound`. Because both `process_result`'s touched-balances loop and the monitor's `UtxoRecovered`/`UtxoSpent` handler propagate it with `?`, a single unrecordable balance change aborts *all* of transaction/event processing rather than skipping that one record. Prefer skip-and-log here. The common stealth-TARI case is likely cached, which is why this is med rather than high.

### 3. [Low-Med] The `2026-06-18-160000` migration was edited in place again

It now has nullable vault columns, new NOT-NULL `account_id`/`resource_id`, a new CHECK, and re-keyed indexes — plus a separate new `2026-06-25-150000_vault_version` migration. Diesel keys applied migrations by name, not checksum, so any dev DB that already ran an earlier version of this migration keeps the old schema and then errors on `account_id`/`resource_id` at insert/select time. Fresh-DB-only and pre-launch, but each in-place edit widens the drift window — a new migration would be safer.

### 4. [Low] No-vault `before`-chaining ignores source and vault-backing

For a vault-less resource, `record_resource_balance_change` derives `before` from `balance_changes_get_latest_by_account_resource`, which returns the latest row for the (account, resource) regardless of whether it was vault-backed or a Transaction row. Across a vault-added/removed transition or interleaved Transaction rows, the derived `before` (and thus the recorded delta) can be off.

### 5. [Low] Touched-stealth-loop vs `refresh_vault` dedup is ordering-dependent

For a vault-backed stealth resource in the diff, `refresh_vault` records the Transaction row first and the touched-balances loop then no-ops via the `(account_id, resource_id, transaction_id)` unique index + never-clobber-transaction upsert. Correct today, but the single-row guarantee rests on refresh_vault running first rather than anything explicit.

---

**Net:** solid extension that also fixes last round's coupling note. #1 and #2 are the ones worth addressing before merge — #1 is a straightforward index/read-path alignment, #2 is a graceful-degradation fix so an uncached resource can't stall transaction processing.

_(Note on a dropped item: burnt outputs being counted in `get_unspent_balance_for_account_resource` was considered, but that method uses the same `status = Unspent` query as the existing stealth-balance display path, so it mirrors established semantics rather than introducing a new divergence — not a regression here.)_

tzmWW:
1. Added a new SQLite migration that replaces the balance-history read indexes with address-keyed indexes, preserving durable address-based reads.
2. Made uncached vaultless resource metadata a logged skip (`Ok(false)`) rather than an event-loop abort.
3. Tightened no-vault “before” chaining so it only chains from prior vaultless rows.
4. Added focused tests for the skip path, the no-vault chaining guard, and final index definitions.
5. Conflict resolved by keeping `@tari-project/ootle-ts-bindings `at `1.43.0` as required by this branch.

sdbondi (approved):

metalaureate:
👋 @tzmWW — your PR was merged and this bounty is queued for an XTM payout. To receive it, **reply to this PR with your Tari (XTM) payout address** (base58 or emoji ID) on its own line, for example:

`XTM: <your address>`

For your protection we only accept an address posted by **@tzmWW** (the merged-PR author). Comments from anyone else are ignored.

tzmWW:
Thank you @metalaureate 

XTM: `128zDKJpimGBJPjgSzETLev2RsuXa5eACBkoo5KwPyCY7bWqpN4nfAiQ3AVNCUQza5KFiuY6xgRvi1mxMAMntrnm6cW`

PR #2265, opened 2026-06-19
refactor(consensus): avoid double validation of tx signatures

sdbondi:
## Motivation

On a validator node, transaction signatures (and every other structural check) were verified **twice** for the normal path. `validate` was a method on `BlockTransactionExecutor`, and the consensus validation chain (`create_consensus_transaction_validator`) wrapped the *entire* mempool chain via `map_context`. So a mempool-originated transaction was:

1. fully validated at mempool ingress (`TransactionSignatureValidator` et al.), then
2. fully validated **again** when sequenced into consensus (`OnReceiveNewTransaction::try_sequence_transaction` → `executor.validate`).

`TariBlockTransactionExecutor::validate` didn't even use the executor's state (its `_tx` arg was ignored) — validation is a pure function of `(epoch, transaction)`, so it never belonged on the executor.

## Change

- **Executor is execute-only.** `BlockTransactionExecutor` loses `validate`; `TariBlockTransactionExecutor` drops its validator field and generic.
- **Validation is its own concern:** a new `BlockTransactionValidator` trait, injected via `ConsensusSpec::TransactionValidator` (mirroring how the executor is injected).
- Validation is split into its two orthogonal parts:
  - **structural / content-only** (network, dry-run, basic, weight, **signature**, template-exists): immutable over a transaction's lifetime → run once.
  - **epoch / time-dependent** (epoch range): the current epoch advances while a transaction waits in the pool → must be re-checked at sequencing time.
- The handler selects scope by **provenance** (no skip-flags leaking into the validator chain):
  - peer-requested **missing** transactions never passed local validation → validated **in full**.
  - mempool-originated transactions were already fully validated on ingress → only the **epoch** rules are re-checked when sequenced.

Net result: structural checks, including signature verification, now run **exactly once** per transaction per node on every path. `create_consensus_transaction_validator`, `ConsensusTransactionValidator`, and the `ValidationContext`/`WithContext` plumbing are removed.

## Safety

Skipping structural re-validation on the mempool→consensus path is sound only if everything delivered via `ConsensusHandle::notify_new_transaction` has already passed mempool validation. Audited: `notify_new_transaction` has a single caller (`MempoolService::handle_new_transaction`), reached only after `before_execute_validator.validate()` succeeds, and the channel sender is private to `ConsensusHandle`. The invariant holds.

## Testing

- `cargo check` clean for `tari_consensus`, `tari_validator_node`, `consensus_tests`.
- `cargo nextest run -p consensus_tests`: **62/62 pass**, including `node_requests_missing_transaction_from_local_leader` (full-validation peer path), `single_transaction*` (mempool→sequence epoch-only path), and `single_transaction_epoch_expired` (epoch validator on the sequence path).

### DIFF
--- a/applications/tari_validator_node/src/bootstrap.rs
+++ b/applications/tari_validator_node/src/bootstrap.rs
@@ -78,7 +78,6 @@ use tari_ootle_transaction_validation::{
     TransactionValidationError,
     TransactionWeightValidator,
     Validator,
-    WithContext,
 };
 use tari_rpc_framework::RpcServer;
 use tari_shutdown::ShutdownSignal;
@@ -102,7 +101,7 @@ use crate::{
         self,
         ConsensusHandle,
         TariBlockTransactionExecutor,
-        ValidationContext,
+        TariBlockTransactionValidator,
         spec::ValidatorTemplateProvider,
     },
     file_l1_submitter::FileLayerOneSubmitter,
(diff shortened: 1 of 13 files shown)


sdbondi:
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #2331, opened 2026-07-09
feat(wallet): external ElGamal view keys for traffic-sim init and decrypt

sdbondi:
## Motivation

The traffic sim always instantiated its stablecoin with a wallet-derived ElGamal view key and could only decrypt UTXOs with that same derived key. This adds the ability to drive the sim with an **externally-generated** view key — useful when the view key is managed out-of-band and must match a key that already exists elsewhere.

## Changes

**traffic-sim**
- `init` gains `--view-key <hex>`: when supplied, the secret's public key overrides the wallet-generated view key used to instantiate the stablecoin.
- `decrypt-utxos` gains `--view-key <hex>`: the secret is imported into the exchange wallet and used to decrypt UTXOs created with an external view key. Without it, behaviour is unchanged (wallet-derived `ElgamalEncryptionViewKey/0`).
- After `init` it prints the suggested `setup` command; after `setup` it prints the suggested `run` commands.

**walletd `keys.import` JSON-RPC**
- New RPC that imports an externally-generated secret into the wallet keystore and returns its `KeyId` + public key. Gated on `Permission::Keys(Crud::Create)`.
- Added the matching `import_key` method to `wallet_daemon_client`. The secret travels on the wire as hex via `tari_crypto`'s serde.

**Idempotency (keyed on the public key)**
- The public key is the stable identity of an imported key (the secret is encrypted at rest, so it can't be queried by derived pubkey). A migration adds a unique `public_key` column to `key_manager_imported_keys` and drops the old unique-on-`label` index.
- The insert is now an upsert on `public_key`: re-importing the same secret updates the (mutable) label and returns the existing id, so repeated `decrypt-utxos --view-key …` runs no longer collide on the unique index.

## Testing

- `cargo check` / `clippy` clean across the affected crates.
- New storage test `import_key_is_idempotent_on_public_key` asserts same-pubkey → same id (with a changed label) and distinct-pubkey → distinct id; it also exercises the migration end-to-end.

## Notes

The `key_manager_imported_keys` table has never been insertable before this change (no RPC exposed it), so it is empty in practice and the migration is safe.


sdbondi:
Pushed follow-ups:

**CI fix (`ethnum` 1.5.2 → 1.5.3).** The red checks were not from this PR — `development` fails identically. Stable rolled to Rust 1.97, and `ethnum 1.5.2`'s error helpers do `mem::transmute(())` (zero-sized → non-zero-sized), which 1.97 rejects with E0512. 1.5.3 builds the std error variants from real failing parses instead. The existing `1.5.0` requirement already permits it, so only the lockfile changes. Folded in here to unblock this PR's CI; happy to split it into its own PR if preferred.

**Review feedback addressed:**
- `key_manager_insert_imported_key` now overwrites `label`, `encrypted_secret` and `key_type` on an idempotent re-import (not just the label).
- Row id converted with `u64::try_from` instead of an `as` cast.
- `decrypt_utxos` connects the exchange wallet once and reuses it for the key import and both decrypt paths.

The idempotency test was extended to assert the key type and encrypted secret are updated on re-import.

PR #2352, opened 2026-07-17
feat(transaction)!: exclude seal signature witness data from the transaction id

sdbondi:
## Motivation

The seal signature carries a fresh random Schnorr nonce per seal, so hashing it into the
transaction id gives every re-seal of the identical body + co-signature set a distinct id.
Since the network dedups by transaction id (mempool `transaction_exists` + finalized-record
checks), a sealer holding collected co-signatures could seal the same authorized body N
times and produce N distinct, individually-valid transactions - with unversioned inputs,
an approve-once-execute-many replay. This is particularly relevant to multi-party flows
(walletd transaction requests #2348, SDK co-signing) where co-signers authorize a body
once: a co-signature should authorize one id.

## Changes

- `TransactionV1::calculate_id` and `PrunedTransactionV1::calculate_id` now hash only the
  seal signer's **public key**, not the full `TransactionSealSignature` - the signature
  scalar/nonce is excluded as witness data. Same reasoning as Bitcoin's SegWit txid/wtxid
  split, and Radix Babylon's intent hash, which excludes all signatures including the
  notary's.
- The public key stays in the projection: with zero co-signatures nothing else binds the
  sealer, and excluding the key would let a different sealer produce a colliding id for a
  semantically different transaction (different authorized-signer set) - a griefing vector
  where a re-sealed variant consumes the id of a legitimate transaction.
- Both id computations now delegate to a single shared `pub(crate) fn calculate_id_v1`
  helper in `crates/transaction/src/v1/transaction.rs` documenting the projection
  invariants, keeping the full and pruned forms provably identical.
- Three new tests in `transaction_id_tests` pin the semantics:
  - `id_is_stable_across_reseals` - same body + co-signatures re-sealed by the same key
    (fresh nonce each time) -> identical id
  - `id_binds_seal_signer_public_key` - zero co-signatures, same body sealed by two
    different keys -> different ids
  - `id_binds_co_signatures` - dropping/adding a co-signature changes the id

## Effects

- One authorization set = one id: re-sealing identical bytes is recognised as the same
  transaction and dedups instead of double-executing.
- Identical logical re-runs now require a body distinguisher (e.g. a different epoch
  window) - resubmission after a timeout becomes idempotent-safe rather than a
  double-spend risk.
- Committed transactions already collide on the `TransactionReceipt` substate address
  (derived from the tx id), so the receipt becomes the on-chain enforcer of commit-once
  for the stable id.

## Breaking change

**This is a consensus-breaking change** - transaction ids change for all existing
transactions. Intended for the same testnet-reset batch as #2350. No `TransactionV2`
gating - this is landing directly on `TransactionV1` by explicit decision. External
id-asserting fixtures (notably ootle-go golden vectors) will need re-vendoring in a
coordinated follow-up.

## Verification

- `cargo nextest r -p tari_ootle_transaction --release` - 49/49 pass, including the three
  new id tests and the existing pruned-form tests asserting full <-> pruned id equality
- `cargo check --all-targets` and `cargo lints clippy -p tari_ootle_transaction
  --all-targets` clean
- Formatted with `cargo +nightly-2025-12-05 fmt`
- A full workspace `nextest` sweep (excluding `integration_tests`) is running; results
  will be reported on this PR

### DIFF
--- a/applications/tari_walletd/src/handlers/context.rs
+++ b/applications/tari_walletd/src/handlers/context.rs
@@ -287,7 +287,13 @@ impl HandlerContext {
     }
 
     /// Returns a TransactionBuilder with the current network configured.
+    ///
+    /// The builder is stamped with a random nonce so that repeated identical intents (same
+    /// instructions, inputs and signer) produce distinct transaction ids. Paths that submit
+    /// caller-provided bytes replace the whole unsigned transaction via
+    /// `with_unsigned_transaction`, which discards the stamp — the caller's bytes are preserved
+    /// verbatim there.
     pub fn transaction_builder(&self) -> TransactionBuilder {
-        Transaction::builder(self.config().network.as_byte())
+        Transaction::builder(self.config().network.as_byte()).with_nonce(rand::random())
     }
 }
--- a/bindings/package.json
+++ b/bindings/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tari-project/ootle-ts-bindings",
-  "version": "1.47.1",
+  "version": "1.48.0",
   "private": false,
   "publishConfig": {
     "access": "public"
--- a/bindings/src/types/PrunedUnsignedTransactionV1.ts
+++ b/bindings/src/types/PrunedUnsignedTransactionV1.ts
@@ -30,4 +30,8 @@ export type PrunedUnsignedTransactionV1 = {
    * archives that didn't record sizes.
    */
   blob_sizes: Array<number>;
+  /**
+   * Mirrors `UnsignedTransactionV1::nonce` of the full form.
+   */
+  nonce: bigint;
 };
--- a/bindings/src/types/UnsignedTransactionV1.ts
+++ b/bindings/src/types/UnsignedTransactionV1.ts
@@ -23,4 +23,11 @@ export type UnsignedTransactionV1 = {
    * signature verifiability or transaction id.
    */
   blobs: Blobs;
+  /**
+   * Distinguishes otherwise-identical transactions. The transaction id excludes the seal
+   * signature, so two identical bodies sealed by the same key are the *same* transaction —
+   * an intent builder (e.g. a wallet) that wants each submission to execute independently
+   * must stamp a distinct nonce per intent. Part of the signing and id domains.
+   */
+  nonce: bigint;
 };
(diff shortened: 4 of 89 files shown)


sdbondi:
Full workspace sweep (`cargo nextest r -E "not package(integration_tests)" --release --no-fail-fast`): 1656/1662 passed. The 6 failures were all tests relying on the old semantics (re-seals producing distinct ids):

- `consensus_tests::{single_shard_input_conflict, multishard_unversioned_input_conflict}` built their conflicting pair as byte-identical bodies sealed by the same key — now a single transaction by definition. Fixed by sealing with distinct keys (be2591689).
- `tari_engine::{test_state, cross_template::*}` — the `TemplateTest` harness built repeated identical transactions, which now collide on id-derived component addresses. Fixed by stamping each harness-built transaction with a unique `max_epoch` (the engine does not evaluate epoch bounds, so this is behaviourally inert) (be2591689).

After the fixes, the affected suites (`consensus_tests`, `tari_engine`, `tari_template_test_tooling`) pass 361/361.

sdbondi:
Fixed the integration-test failure (\`UNIQUE constraint failed: transactions.transaction_id\` on the @concurrency scenario).

**Cause:** with the seal signature out of the id, 30 concurrent identical \`counter.increase\` calls built identical unsigned bodies sealed by the same wallet key → one transaction id → walletd rejected the duplicates. Any intent builder now needs a per-intent distinguisher (the same role Radix Babylon gives its intent-header nonce).

**Fix (7322f465a):** added a \`nonce: u64\` to \`UnsignedTransactionV1\`, part of the signing and id domains, stamped randomly at the two intent-builder layers (walletd \`HandlerContext::transaction_builder\`, integration-test \`transaction_builder\`). Paths that submit caller-provided bytes verbatim (\`with_unsigned_transaction\`) discard the stamp, so frozen transaction-request bytes are untouched.

**Scope note:** this makes the PR a **signing-domain** change, not id-only — the ledger signing preimage gains a \`Nonce\` field (renumbering \`BlobHashes\`/\`Signatures\`), so the coordinated follow-up now also covers the Ledger app, not just id-asserting fixtures. TS bindings bumped 1.47.0 → 1.48.0 + dist rebuilt; SDK golden vectors regenerated.

**Verification:** \`@concurrency\` cucumber scenario passes (7/7 steps, the 30 concurrent calls included); \`tari_ootle_transaction\` + \`ootle_ledger_client\` + \`ootle_ledger_common\` 54/54; \`ootle_sdk_core\` 283/283 post-regen; workspace \`cargo check --all-targets\` clean.

PR #2354, opened 2026-07-19
fix(engine): fix race condition in stealth claim_validator_fees (closes #2149)

### LINKED ISSUE #2149: Race in stealth claim_validator_fees between fee-pool snapshot and on-chain claim
## Background

PR #2147 added a stealth-output path to `validators.claim_fees` (`applications/tari_walletd/src/handlers/validator.rs`). For each shard's fee pool, the wallet daemon:

1. Fetches the pool's current `amount` from the network (`build_self_stealth_plan` → `get_substates_from_network`).
2. Builds a `StealthTransferStatement` that commits to that exact `amount` as revealed input and produces a stealth output (plus a `max_fee` revealed-output bucket on one carrier pool).
3. Submits the transaction.

The `claim_validator_fees` instruction in the engine sweeps the **current** pool balance into a revealed bucket. If additional fees were deposited to the pool between step 1 and the on-chain execution, the bucket contains MORE than what the stealth balance proof commits to → the proof fails and the transaction is rejected (fees still charged).

## Impact

- Most painful on validators that are actively producing blocks while a user claims, because fees accumulate every committed block.
- A user retry hits the same race unless they happen to land between deposits.
- The revealed-output path (`output_to_revealed: true`) is unaffected — it just deposits whatever the bucket holds into the account vault, no commitment to a specific amount.

## Possible fixes (rough sketch, not committed)

1. **New engine instruction `claim_validator_fees_up_to(address, max_amount)`** — engine claims at most `max_amount`, leaves the rest in the pool. Wallet then commits to `max_amount` exactly. Simplest semantic fix.
2. **Workspace-amount-on-claim** — engine writes the actual claimed amount into the workspace so subsequent instructions can read it. Stealth_transfer would need to be reworked to consume the amount dynamically (and the balance proof too — likely a much bigger change).
3. **Pre-flight + retry** — wallet calls a network 'lock pool at version N' / does versioned read of the pool and includes the version in the input. If the actual version on-chain differs, the input lock fails before the claim runs. Requires versioned-input semantics.

(1) is the smallest blast radius and probably the right next step.

## Acknowledged in

PR #2147 review comment: https://github.com/tari-project/tari-ootle/pull/2147#discussion_r3281710240

## Acceptance

- A repeated claim against a hot validator pool succeeds deterministically (no balance-proof rejection).
- Old `claim_validator_fees(address)` keeps working for the revealed-output path (no change needed there).

webhop123:
### Description
Fixes a race condition in stealth \claim_validator_fees\ when additional blocks land between fetching the fee pool balance from the network and transaction execution on-chain (#2149).

### Changes
- Added \withdraw_up_to(&mut self, max_amount: Amount)\ to \ValidatorFeePool\ in \	ari_engine_types\.
- Added optional \max_amount: Option<Amount>\ to \Instruction::ClaimValidatorFees\ and added \claim_validator_fees_up_to\ on \TransactionBuilder\.
- Updated \StateTracker\ and \RuntimeInterface\ to support partial pool withdrawals up to \max_amount\.
- Updated \	ari_walletd\ stealth fee claim handler to pass the snapshot fee balance as \max_amount\, guaranteeing total claimed funds match the stealth commitment proof.

Closes #2149.

### DIFF
--- a/applications/tari_indexer/README.md
+++ b/applications/tari_indexer/README.md
@@ -80,6 +80,7 @@ The following options are available (use `tari_indexer --help` for the most up-t
 
 #### API Server
 - `-r, --api-listen-address <API_LISTEN_ADDRESS>`: Bind address for REST API server
+- `--metrics-listen-address <METRICS_LISTEN_ADDRESS>`: Bind address for Prometheus metrics server
 - `--web-ui-public-api-url <WEB_UI_PUBLIC_API_URL>`: Public API URL for web UI [env: `TARI_INDEXER_WEB_UI_PUBLIC_API_URL`]
 - `--web-ui-public-graphql-url <WEB_UI_PUBLIC_GRAPHQL_URL>`: Public GraphQL URL for web UI [env: `TARI_INDEXER_WEB_UI_PUBLIC_GRAPHQL_URL`]
 
--- a/applications/tari_indexer/docs/configuration.md
+++ b/applications/tari_indexer/docs/configuration.md
@@ -47,6 +47,9 @@ Main indexer application settings.
 # API server listening address (default: "127.0.0.1:18300")
 #api_listen_address = "127.0.0.1:18300"
 
+# Prometheus metrics server listening address (default: "127.0.0.1:18302")
+#metrics_listen_address = "127.0.0.1:18302"
+
 # GraphQL server address (default: "127.0.0.1:18301")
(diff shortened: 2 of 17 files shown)


webhop123:
Closing - PR was based on wrong branch and included unrelated indexer changes. Will reopen clean from development.

PR #2356, opened 2026-07-19
fix(engine)!: fix race condition in stealth claim_validator_fees (closes #2149)

### LINKED ISSUE #2149: Race in stealth claim_validator_fees between fee-pool snapshot and on-chain claim

webhop123:
### Description

Fixes a race condition in stealth \claim_validator_fees\ when additional blocks land between fetching the fee pool balance from the network and transaction execution on-chain (#2149).

When using stealth claims, the wallet daemon fetches each fee pool's current balance to build a stealth commitment proof. If new blocks arrive before execution, the on-chain pool balance grows larger than the committed amount, causing the stealth statement validation to fail.

### Solution

Add an optional \max_amount\ cap to \ClaimValidatorFees\ so the engine withdraws **at most** the snapshot balance used in the stealth proof, leaving any surplus in the pool.

### Changes

**Engine Types** (\crates/engine_types/src/validator_fee.rs\)
- Added \withdraw_up_to(&mut self, max_amount: Amount)\ to \ValidatorFeePool\
- \withdraw_all()\ now delegates to \withdraw_up_to(Amount::MAX)\

**Transaction** (\crates/transaction/\)
- Added \max_amount: Option<Amount>\ field to \Instruction::ClaimValidatorFees\
- Added \claim_validator_fees_up_to()\ builder method on \TransactionBuilder\

**Engine Runtime** (\crates/engine/src/runtime/\)
- Updated \RuntimeInterface::claim_validator_fees\ signature to accept \max_amount: Option<Amount>\
- Added \withdraw_fees_from_pool_up_to()\ to \StateTracker\

**Wallet Daemon** (\pplications/tari_walletd/src/handlers/validator.rs\)
- Stealth claim path now passes fetched pool balance as \max_amount\ per pool

Closes #2149.


sdbondi (commented):
Thanks for the PR — and for picking up #2149. The diagnosis is right and the approach is the right minimal fix: capping the withdrawal at the snapshot the stealth proof commits to, rather than trying to make the snapshot fresher, is exactly the correct shape for this race.

I verified the core logic: snapshot amounts are collected per pool and zero/unfetchable pools are rejected before `pool_amounts` is built, so every cap is >= 1; the statement total is `sum(snapshots) - max_fee`, and the engine withdraws exactly the snapshot whenever the pool has grown, which closes the race. If a pool *shrinks* (competing claim) the claim under-fills and fails as it did before, which is correct. `withdraw_all -> withdraw_up_to(Amount::MAX)` is lossless given `Amount` is a `u128` newtype and pool balances are `u64`. The untouched match sites in `unsigned.rs` and `transaction.rs` use `..` patterns and `RuntimeInterfaceImpl` is the only trait implementor, so nothing was missed. Leaving the `output_to_revealed` path uncapped is right, since revealed claims don't commit to a snapshot.

A few things to fix before this can go in.

**1. The new field needs `serde(default)` and `cbor(default)` — `crates/transaction/src/v1/instruction.rs:93`**

Without `serde(default)`, JSON deserialization of a `ClaimValidatorFees` instruction that omits `max_amount` — i.e. every existing client — will fail, because serde requires struct-variant fields unless defaulted. The precedent is a few lines down in the same enum, `PublishTemplate.metadata_hash`:

```rust
#[n(1)]
#[cfg_attr(feature = "serde", serde(default))]
#[cbor(default)]
max_amount: Option<Amount>,
```

**2. The committed TS bindings need updating**

`bindings/src/types/Instruction.ts:34` still reads `{ ClaimValidatorFees: { address: ValidatorFeePoolAddress } }`. This repo hand-maintains the bindings rather than regenerating them, so please add `max_amount: Amount | null`, bump the bindings package version, and rebuild `dist`.

**3. Derive the withdrawal record without the lossy round-trip — `crates/engine/src/runtime/working_state.rs:1068`**

```rust
let amount = u64::try_from(resource_container.unlocked_amount().to_u128()).unwrap_or(0);
```

Copilot flagged this too and it's the one comment of the three worth acting on. It can't fail today, since the container's revealed amount originated as a `u64`, but `ValidatorFeeWithdrawal` is what consensus uses to settle pool balances — if the fallback ever fired we'd record a zero-amount withdrawal while funds actually left the pool, which is silent inflation. Rather than a checked conversion that returns a `RuntimeError`, I'd prefer removing the fallible conversion altogether: have `withdraw_up_to` return the withdrawn `u64` alongside the container (or compute the pool-balance delta before/after).

**4. `Some(0)` semantics are inconsistent — `crates/engine_types/src/validator_fee.rs`**

An empty pool errors with `InsufficientBalance`, but `max_amount = 0` against a non-empty pool succeeds and mints an empty stealth bucket. It's owner-authorized so not exploitable, but a zero cap is always a caller bug. Either reject it or document the clamp and empty-pool semantics on the method.

**5. Tests**

The new engine semantics are untested and cheap to cover in `validator_fee.rs`: cap below balance leaves the remainder, cap at or above balance withdraws everything, empty pool errors. Ideally also an engine-level test asserting the surplus remains in the `ValidatorFeePool` substate and that `ValidatorFeeWithdrawal.amount` equals the capped amount — that's the invariant point 3 is protecting, and it's consensus-relevant accounting, so it should land with the change.

**6. `fmt` is red**

rustfmt wants the long `write!` in the `Display` impl broken across lines (`instruction.rs:420`). `cargo +nightly-2025-12-05 fmt --all` will sort it.

Minor, no action needed: adding the field changes the borsh encoding of the variant even when `None`, because of `Option`'s discriminant byte, which shifts any borsh-derived hashes. That's fine pre-launch, just noting it's a wire change. On Copilot's other two comments — the `StateTracker::` prefix on those `InvariantError`s is indeed stale, since the code lives in `WorkingState`, and "an ValidatorFeePool" should be "a". Both are pre-existing and the file's convention is mixed anyway, so fix them while you're in there if you like, but I don't consider them blocking.

Nice work — the hard part (identifying that the cap belongs on the instruction rather than in the wallet's retry logic) is done and done well.

sdbondi:
@webhop123 FYI I've proposed this issue be given a bounty size M for a successful merge

webhop123:
The requested changes and unit/integration tests have been added and verified to pass. The panic has been fully resolved and we verified that draining the pool successfully triggers its destruction without leaving dangling empty substates. 

cc: @sdbondi /claim

.agents/AGENTS.md
```diff
@@ -0,0 +1,17 @@
+# Tari Ootle Workspace Guidelines & Memory
```
sdbondi: Please remove

webhop123: Done !

sdbondi:
Thank you 

Just a note for general reference: Once this is merged and released (to all validators and indexers), wallets will need to upgrade too to be able to claim vn fees. Any rederivation of the tx id will yield a different result (extra None byte in preimage), but I dont expect this to be a big issue for testnet (would rather not reset if we dont have to)

sdbondi:
Use `cargo +nightly-2025-12-05 fmt --all` for formatting. ~~The clippy lints look unrelated (due to our use of stable rust, which detected new lints) sdo not a blocker~~ Were related, pushed a fix

sdbondi (approved):
Tested - thanks!

webhop123:
Hi @sdbondi, thank u for merging the PR! Since this fixes #2149 (Bounty Size M), could you please lemme  know the process for claiming the bounty or where I should share my payment/wallet details? Thanks again!

sdbondi:
@webhop123 Thanks again for the fix — and sorry for the delay in answering.

#2149 is now recorded as a bounty at size **M — 60,000 XTM**, with you credited as the contributor.

To claim: reach out to **@metalaureate** — `possum` on Discord — with your Tari address. Processing can take up to a week.

The bounty board is at https://github.com/tari-project/bounties if you want to pick up anything else — open bounties across the org are listed there.

metalaureate:
👋 @webhop123 — your PR was merged and this bounty is queued for an XTM payout. To receive it, **reply to this PR with your Tari (XTM) payout address** (base58 or emoji ID) on its own line, for example:

`XTM: <your address>`

For your protection we only accept an address posted by **@webhop123** (the merged-PR author). Comments from anyone else are ignored.

webhop123:
@metalaureate ok thanks 

XTM : 122VtMBiuKMk7tXoYbri1BLM2PcN4fhtpF326fscuEFMCN55zhA7uzXF53k9cuU4C5rUYw9HDf3bZj2K8tbD6sreAyT

PR #2375, opened 2026-07-21
Fix WalletDaemonClient not sending the token on auth.refresh

chironbuilds:
## Summary

`WalletDaemonClient.sendRequest()`'s 401 auto-refresh path calls `this.transport.sendRequest({...})` with only the request payload, omitting the `{ token: this.token, timeout_millis: null }` options argument every other call site in this method passes. `FetchRpcTransport.sendRequest()` only attaches an `Authorization: Bearer <token>` header `if (options?.token)`, so the `auth.refresh` request goes out with no token at all — the server has nothing to look up a session from, and refresh unconditionally fails with `Unauthorized: No refresh token`, even when the refresh grant itself is still valid.

Net effect: once an access token expires, every subsequent call from that client permanently fails with a 401 instead of transparently refreshing and retrying, defeating the entire point of `setReauthenticationEnabled(true)`.

## Fix

Pass the same `{ token: this.token, timeout_millis: null }` options object to the `auth.refresh` call that every other `transport.sendRequest` call in this method already uses.

## Verification

Reproduced and confirmed the fix against a live `tari_ootle_walletd` (esmeralda) from a small Node script using this client directly:

- **Before**: connecting with an expired/invalid token → `auth.refresh` request → daemon logs `Refresh resp { error: { code: 401, message: "...Unauthorized: No refresh token" } }` → client throws the original 401.
- **After**: same scenario → refresh succeeds, returns a new token, and the original request is retried and succeeds.

This surfaced while building a Chrome extension wallet that relays to `tari_ootle_walletd` via this client — a daemon connection that had been running for a while (tokens expire after 5 minutes per this repo's own `CLAUDE.md`) became permanently unusable on the very next call, with no way to recover short of a brand new `auth.request`.

## Test plan

- [x] `npx prettier --check` passes on the changed file (matches existing formatting)
- [x] Manually verified end-to-end against a running `tari_ootle_walletd` as described above
- [ ] No existing automated test coverage for this client to extend — happy to add one if there's a preferred harness/mocking approach for this package

### DIFF
--- a/clients/javascript/wallet_daemon_client/src/index.ts
+++ b/clients/javascript/wallet_daemon_client/src/index.ts
@@ -569,12 +569,15 @@ export class WalletDaemonClient<T extends RpcTransport = FetchRpcTransport> {
         throw origError;
       }
       const id = this.id++;
-      const refreshResp = (await this.transport.sendRequest<any>({
-        method: "auth.refresh",
-        jsonrpc: "2.0",
-        id,
-        params: {},
-      })) as RpcResponse<AuthLoginResponse>;
+      const refreshResp = (await this.transport.sendRequest<any>(
+        {
+          method: "auth.refresh",
+          jsonrpc: "2.0",
+          id,
+          params: {},
+        },
+        { token: this.token, timeout_millis: null },
+      )) as RpcResponse<AuthLoginResponse>;
 
       if (refreshResp.error) {
         console.debug("Refresh resp", refreshResp);


sdbondi:
Thanks for the PR, and for the detailed writeup — but I don't think this is the bug.

By design, auth.refresh doesn't refresh from the previous (valid or expired) JWT. It reads an HttpOnly, SameSite=Strict cookie set on successful login. The "No refresh token" error means the cookie wasn't on the request, which is exactly what we want: JavaScript and other origins can't get at that session by design.

Your underlying problem is real though — a Chrome extension genuinely can't hold a session this way, and that's intentional rather than something we want to loosen. For an external client, use an API key with limited permissions instead: pass it as the bearer and it works on every route with no refresh required. One caveat, minting a key requires an Admin session from a real user login, so you'll need to create it from the local wallet UI once and configure the extension with it. Wallets handle money, we have to be careful with auth.

Also worth knowing: WebAuthn is the default auth method, and the RP origin defaults to http://localhost:{json_rpc_port}, so browser login only works from that origin regardless.

Closing this one, but happy to help if the API key path has rough edges. I'll also think about if we can safely support chrome extensions, apart from assigning it an API key, or at least improve UX (e.g. your extension asks to "pair" with the wallet, and you have to click approve in the web UI, and it issues the API key)

Issue #2343, opened 2026-07-13
Architecture discussion: selected-account stealth ScriptPath HTLCs through walletd


GSXRspartan:
I built and tested a proof of concept for selected-account stealth UTXO ScriptPath HTLC operations through walletd using an Ootle v0.35.2 fork.

The proof demonstrates real walletd-driven:

- selected-account stealth source lookup
- ScriptPath HTLC funding
- hashlock claim
- redeem
- epoch-based refund
- exact fee convergence
- terminal and retry-safe result handling

I am opening this issue to document the working proof and discuss the preferred upstream division of responsibility between transaction construction, approval, signing, and submission. Initial maintainer feedback suggested a generic unsigned-transaction flow with separate `request to sign` and privileged `approve + submit` steps, using scoped walletd API-key permissions.

This is a proof of concept. It is not presented as production-ready or as an upstream-ready pull request.

## Validated snapshot

- Ootle base version: `v0.35.2`
- Validated local commit: `10ea70205b292de71c27e2a14c8868c3c50a9954`
- Validated walletd SHA256: `3B7EACC16EAD3EC106E2BFC6B6C0409C7FE4ED3A4E3DA44247F3DA21EDA750C5`

The source commit is currently local. A neutral patch or branch can be provided for review once the preferred upstream architecture is agreed.

## Proof-of-concept RPCs

- `accounts.stealth_htlc_source_status`
- `accounts.stealth_htlc_fund`
- `accounts.stealth_htlc_claim`
- `accounts.stealth_htlc_redeem`
- `accounts.stealth_htlc_refund`

The current proof performs the following inside walletd:

1. validates the explicitly selected account;
2. locates an owned key-path stealth source;
3. constructs the ScriptPath HTLC transaction;
4. signs using wallet-controlled keys;
5. submits the transaction;
6. returns transaction, output, fee, and terminal-state evidence.

Private account and stealth signing keys do not leave walletd.

## Validated resource

Native testnet TARI:

`resource_0101010101010101010101010101010101010101010101010101010101010101`

### Funding and hashlock claim

- Funding transaction: `db11892a652a5dedd2a9f2a79d725f6598911ba6b52773512c5ac2c25fb121c3`
- Funding required fee: `585`
- Claim transaction: `0e03730e1e90b252d70117cfd4dff865bcbc85bb535905d2e9e25eb66907153c`
- Claim required fee: `409`
- Amount recovered from a 10,000-unit HTLC: `9591`

[...]

PR #2386, opened 2026-07-23
test(walletd): prove PayTo conditions statement construction

GSXRspartan:
## Summary

Adds an isolated walletd handler test proving that
`accounts.create_stealth_transfer_statement` can construct a blinded
`PayTo::Conditions` output from exact `InputSelection::Specific` key-path
inputs.

The test verifies:

- exact caller-selected inputs and promised input ordering;
- positional UTXO signer ordering;
- a fully blinded output with zero revealed amount;
- an HTLC-style claim and refund condition tree;
- `SpendAuthorization::Script` using the canonical condition root;
- no key-path authorization on the conditions output;
- successful input locking under the returned wallet lock ID.

## HTLC conditions

Claim path:

- SHA-256 hash-lock digest;
- `BeforeEpoch(refund_epoch)`;
- claimant public-key access rule.

Refund path:

- `AfterEpoch(refund_epoch)`;
- refunder public-key access rule.

## Scope

This is test-only. It does not change production handlers, RPC methods,
permissions, lock duration, wallet storage, network clients, signing,
submission, or broadcast behavior.

The test uses the existing temporary SQLite wallet harness. The exercised path
resolves its resource from the local cache and does not contact an indexer or
other network service.

## Validation

- targeted test: `1 passed; 0 failed`
- walletd package tests: `23 passed; 0 failed`
- walletd all-targets check: passed
- nightly formatting check: passed
- walletd clippy with repository lint policy: passed
- validation performed with Cargo offline

Related to #2343.

### DIFF
--- a/applications/tari_walletd/src/handlers/accounts.rs
+++ b/applications/tari_walletd/src/handlers/accounts.rs
@@ -1791,7 +1791,11 @@ mod create_stealth_transfer_statement_handler_tests {
 
     use axum_extra::headers::{Authorization, authorization::Bearer};
     use axum_jrpc::error::{JsonRpcError, JsonRpcErrorReason};
-    use tari_engine_types::resource::Resource;
+    use tari_crypto::{keys::SecretKey as _, ristretto::RistrettoSecretKey};
+    use tari_engine_types::{
+        resource::Resource,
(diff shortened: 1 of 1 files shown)


sdbondi (approved):
Thanks

PR #2390, opened 2026-07-31
fix(wallet): bind stealth auth signatures to one-time key (P_seal), not account key (K)

m4r1m0:
## Problem

The engine verifies every authorization signature against the seal signature's public key. For **account-sealed** transactions that is the account key (K); for **stealth-sealed** transactions it is the seal signer's *one-time* key (P), **not** the account key.

The previous implementation in `WalletStealthAuthorizer::create_authorizations` (`stealth.rs`) always bound authorizations to K. This is invisible with a single stealth input (no authorizations are produced, so the K-vs-P mismatch doesn't matter) but produces an **invalid signature** as soon as a second stealth input requires an authorization signature.

## Impact

Any transaction with **two or more stealth inputs from different transfers** fails with "Invalid transaction signature" on real submission (the dry-run skips signature validation, so it falsely passes). No existing example or test covers this case, which is why it went unnoticed.

This blocks a class of privacy-preserving applications that need multiple stealth inputs in one transaction — for example, a confidential voting template where a voter spends a stealth vote-token UTXO and a stealth TARI fee UTXO in the same transaction (so the whole tx is ephemeral-sealed with no identity leak).

## Fix

When the transaction is stealth-sealed (`must_sign_with_account_key == false`), derive the one-time stealth owner public key P and bind authorization signatures to P instead of K.

### Changes

1. **`transaction/signer.rs`** — Add `derive_stealth_owner_public_key` to the `TransactionSigner` trait. Default implementation returns an error (for signers like hardware wallets that can't derive the public key without signing). Local software providers override it.

2. **`key_provider/local/generic_impls.rs`** — Override `derive_stealth_owner_public_key` in `LocalKeyProvider`. The one-time public key is derived solely from the account secret and the public nonce, so we call `sign_prehash_with_stealth_key` with a dummy prehash and discard the signature, keeping only the public key.

3. **`wallet/ootle.rs`** — Expose `derive_stealth_owner_public_key` on `OotleWallet` so `WalletStealthAuthorizer` can call it.

4. **`wallet/stealth.rs`** — In `create_authorizations`: when `must_sign_with_account_key` is true, use the account key K (unchanged); when false (stealth-sealed), derive P_seal via `derive_stealth_owner_public_key` and pass it to `authorize_transaction_with_stealth_key` instead of the account key.

## Testing

Validated end-to-end on the Esmeralda testnet with a confidential voting integration test: a voter spends two stealth inputs (vote-token UTXO + TARI fee UTXO) in one transaction, sealed with the vote one-time key. All transactions committed successfully with the fix; without the fix, the two-input spend fails with "Invalid transaction signature."


crates/wallet/ootle-rs/src/wallet/stealth.rs
```diff
@@ -56,16 +56,31 @@ impl WalletStealthAuthorizer<'_, OotleWallet> {
         &self,
         unsigned: &UnsignedTransaction,
     ) -> signer::Result<Vec<TransactionAuthorization>> {
-        let Some(seal_signer) = self.seal_signer() else {
-            // There are no inputs to sign for
-            return Ok(vec![]);
+        // The engine verifies every authorization signature against the seal signature's public
+        // key. For account-sealed transactions that is the account key (K); for stealth-sealed
+        // transactions it is the seal signer's *one-time* key (P), NOT the account key. The
+        // previous implementation always bound authorizations to K, which is invisible with a
```
Kelketek, line 62: This comment about the previous implementation isn't relevant for the code as it stands-- it should be removed, and the comment should only explain what's happening here. The comment about the previous implementation can be on the PR to explain why this version is more suitable.

Kelketek (changes requested):
@m4r1m0 I don't have enough context on the security implications of the code you're working on to evaluate whether it's a good idea. @sdbondi will have to cover that.

However, I found one area of improvement in a comment, and there are two things you should do:

1. Add tests for this new functionality to verify it works as you expect. At a baseline, these functions should all be called and verified to work or error handle in the way you expect.
2. Sign your commits. The CI here requires signatures on all commits to pass. There are some guides to doing so online-- you can typically use your SSH keys to sign.

sdbondi:
@m4r1m0 Thank you for finding this issue and for the PR - if it's ok with you, I'm going to take a crack at fixing this in a new PR. Your diagnosis is right (authorizations were binding to the account key K while stealth-sealed transactions seal under the one-time key P), but digging into it turned up two more problems in the same area: 

`TransactionRequest::build()` never calls `create_authorizations` at all, so the ergonomic path silently drops them, and `StealthTransfer`'s `inputs_to_spend` is keyed by owner address, so a second input from the same address overwrote the first, meaning multi-input was unbuildable in the first place. This all came from stealth support being added as a layer on top of an existing API, so I'd like to rework the design with the multi-input case in mind rather than patch it.

m4r1m0:
@sdbondi that works and is even preferable. I understand the problem here, but in terms of the fix I had no idea what I was doing. Purely "vibe coded" in an attempt to get my Ootle application working. Thanks for taking a look! :)

Kelketek:
@sdbondi Should this be closed now that https://github.com/tari-project/tari-ootle/pull/2403 is merged?

sdbondi:
@m4r1m0 thanks all the same! Let me know if the latest release of ootle-rs works for your use case. Closing this

PR #2403, opened 2026-08-10
fix(ootle-rs): bind stealth authorizations to the key that seals

sdbondi:
Supersedes #2390, which fixed the message binding but left the same class of bug reachable through `authorization_message()` and did the derivation eagerly. This takes the same premise further and reworks the surrounding design.

## The bug

A stealth-sealed transaction seals with the one-time key `P` that `seal_transaction_with_stealth` derives from the spent input's sender nonce. `WalletStealthAuthorizer::seal_signer()` returned the owner's **account key `K`** instead. Since `UnsealedTransactionV1::verify_all_signatures` checks every authorization against `seal_signature().public_key()`, any transaction needing both a stealth seal and an authorization was rejected.

Two more defects kept that path from ever working — and hid each other:

2. **`TransactionRequest::build` only sealed.** It never asked the authorizer for the authorizations its inputs require, so they were silently dropped. This is the path both stealth examples use.
3. **`StealthTransferSpec::inputs_to_spend` was a `HashMap<Address, StealthInput>`.** A second input owned by the same address overwrote the first, so a multi-input transfer from one wallet was unbuildable.

Net effect: only the single-stealth-input shape — which needs zero authorizations — could be constructed at all, which is why this never showed up in testing.

## The fix

**Signature requirements are resolved once.** `must_sign_with_account_key` + `Option<seal_signer>` + `other_signers()`-with-a-skip becomes a `SealSource` (account key / stealth input / ephemeral) plus the authorizers that sign against it. Promoting the first input to seal signer happened in three places (the builder, `seal_signer()`, and `other_signers()`); it now happens once, at construction.

**The seal key is knowable before signing.** `TransactionStealthKeySigner::stealth_public_key` derives the one-time key without producing a signature. `LocalKeyProvider` derives it straight from `owner_stealth_dh_secret`. `LedgerSigner` reports it unsupported — `GetPublicKey` takes only `(account, index, key_type)`, so the device has no way to return `c + k` except alongside a signature it makes with it. Multi-input stealth on Ledger needs a new APDU; single-input (zero authorizations) is unaffected.

**Sealing and authorizing live behind one trait.** `TransactionSealSigner::create_authorizations` defaults to none, and `TransactionRequest::build` calls it. Only whoever chooses the seal key can produce signatures that commit to it, so the two steps cannot be separated by a caller. `NetworkWallet::sign_transaction` for the authorizer collapses to that same path, dropping a duplicated seal branch that also errored on the ephemeral case.

**The ephemeral key is drawn when the authorizer is built**, not at seal time, so an authorization taken from `authorization_message` commits to the key that actually seals. That case previously returned `None` and was unusable.

## Cases

Three, not two — the third is degenerate but real:

| | seals | authorizes |
|---|---|---|
| account component accessed (revealed input bucket) | account key `K` | every stealth input |
| stealth inputs only | first input's one-time key `P` | the remaining inputs |
| nothing spent, no account access | fresh ephemeral key | nothing |

## Tests

`stealth_authorizations_verify_against_the_sealing_one_time_key` is the regression test: three stealth inputs owned by one address, sealed via `TransactionRequest::build`, asserting `verify_all_signatures()` and that the seal key is `P` and not `K`. Reintroducing the old binding fails it (verified). Plus coverage for the single-input, account-key and ephemeral cases, external authorizations against all three seal sources, and `stealth_public_key` agreeing with the key the signing methods actually use.

`examples/stealth_transfer.rs` now splits the faucet funds across two outputs and spends both, so the example itself exercises the multi-input path.

## Notes

- `ootle-rs` is a published crate and this is a breaking API change: `SignatureRequirements`' constructors and accessors, `authorization_message` becoming `async` + `Result`, and the two new trait methods. Nothing else in the workspace depends on it.
- The pre-existing `src/lib.rs` doctest failure (missing trait imports in the module-level example) is unrelated and left alone.

### DIFF
--- a/crates/wallet/ootle-rs/examples/stealth_transfer.rs
+++ b/crates/wallet/ootle-rs/examples/stealth_transfer.rs
@@ -29,7 +29,7 @@ async fn main() {
     //     .init();
 
     // This is the address that we will transfer to (Feel free to change this another address!)
-    let recipient = address!( "otl_loc_1c370ayp9849gzmj9gwelyyd086ntrt84w5nkclu32tzyr2pvcfpre43z8z2xvupm6wltw9k5e8tzay3qqf9nfj9v5xuxwcpcxmg22vqlvz86l" );
+    let recipient = address!( "otl_loc_1em0npr8f3uzs3fznygglr4eujwl0qgr6e30hu3whr9fpfh2w99r743uxd7qg6f7jchgm6pdht7lpcwcggcjgfejfpd4jgvfmzve9vec5xep5t" );
 
     let indexer_api_url = default_indexer_url(recipient.network());
 
@@ -67,14 +67,18 @@ async fn main() {
     // The faucet funds are split across two outputs so that the transfer below spends two stealth inputs. One of them
     // seals that transaction and the other must attach an authorization signature committing to the seal signer's
     // one-time public key.
-    const FEE_BUDGET: u64 = 2000;
+
+    // The revealed amount each transaction reserves to pay its fee. It is deliberately generous rather than fitted to
+    // a dry-run estimate: the budget is part of the transfer statement, so spending an estimate would change the
+    // transaction it was estimated from. Whatever is not charged is refunded (see the receipt's overcharge line).
+    const FEE_BUDGET: u64 = 20_000;
     const FIRST_INPUT_AMOUNT: u64 = 6 * TARI;
-    const SECOND_INPUT_AMOUNT: u64 = 10 * TARI + 2 * FEE_BUDGET - FEE_BUDGET - FIRST_INPUT_AMOUNT;
+    const SECOND_INPUT_AMOUNT: u64 = 10 * TARI + FEE_BUDGET - FIRST_INPUT_AMOUNT;
     // // This builder creates a stealth transfer statement (spend proof). This is added to the transaction later.
     let (faucet_transfer, required_signers) = StealthTransfer::new(tari_token, &provider)
         // Tell the transfer to expect 10 TARI (plus a fee budget for this transaction and the transfer below) as revealed funds from a bucket (the faucet looks at this value and automatically provides the bucket).
         .spend_revealed_input(10 * TARI + 2 * FEE_BUDGET)
-        // The transfer will output the fee budget as revealed funds to pay for the fee. Creating each stealth output costs a substate-creation premium plus its stored bytes, so this has to cover two of them.
+        // The transfer will output the fee budget as revealed funds to pay for the fee.
         .to_revealed_output(FEE_BUDGET)
         // Spend the remaining value (10 TARI - fee) into outputs for the sender address. NOTE: the sender address is not actually included in the output (privacy!),
         // but a supporting wallet that holds the secret key would be able to spend the output.
@@ -120,7 +124,7 @@ async fn main() {
         // Together these are worth 10 TARI plus the second transaction's fee budget.
         .spend_stealth_input(sender_address.clone(), inputs_to_spend[0].commitment())
         .spend_stealth_input(sender_address.clone(), inputs_to_spend[1].commitment())
-        // The transfer will output the fee budget as revealed funds to pay for the fee. Two inputs means an extra authorization signature to verify on top of the two new outputs.
+        // The transfer will output the fee budget as revealed funds to pay for the fee.
         .to_revealed_output(FEE_BUDGET)
         // Spend to a new output (8 TARI) that we'll generate for the recipient address.
         .to_stealth_output(
@@ -160,6 +164,8 @@ async fn main() {
         .unwrap();
     let _diff = result.expect_success();
     println!("Dry run successful!");
+    // Informational: the fee budget above is not derived from this number. Spending the estimate would mean changing
+    // the transfer statement, which would change the transaction the estimate was taken from.
     println!(
         "Estimated fees for transfer: {}",
         result.finalize.fee_receipt.total_fees_charged()
--- a/crates/wallet/ootle-rs/examples/stealth_transfer.rs
+++ b/crates/wallet/ootle-rs/examples/stealth_transfer.rs
@@ -67,14 +67,15 @@ async fn main() {
     // The faucet funds are split across two outputs so that the transfer below spends two stealth inputs. One of them
     // seals that transaction and the other must attach an authorization signature committing to the seal signer's
     // one-time public key.
+    const FEE_BUDGET: u64 = 2000;
     const FIRST_INPUT_AMOUNT: u64 = 6 * TARI;
-    const SECOND_INPUT_AMOUNT: u64 = 10 * TARI + 1000 - 500 - FIRST_INPUT_AMOUNT;
+    const SECOND_INPUT_AMOUNT: u64 = 10 * TARI + 2 * FEE_BUDGET - FEE_BUDGET - FIRST_INPUT_AMOUNT;
     // // This builder creates a stealth transfer statement (spend proof). This is added to the transaction later.
     let (faucet_transfer, required_signers) = StealthTransfer::new(tari_token, &provider)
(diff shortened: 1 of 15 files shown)


crates/wallet/ootle-rs/src/types/transaction/request.rs
```diff
@@ -66,8 +66,11 @@ impl TransactionRequest<WithTx> {
     }
 
     pub async fn build(self, seal_signer: &dyn TransactionSealSigner) -> WalletResult<Transaction> {
-        let builder = self.state.0;
-        let unsealed = builder.with_signatures(self.authorizations.into_iter().map(|a| a.into_signature()).collect());
+        let unsigned = self.state.0;
+        let mut authorizations = self.authorizations;
+        authorizations.extend(seal_signer.create_authorizations(&unsigned).await?);
```
sdbondi: **medium** — `create_authorizations` kept its exact name and signature when it moved from an inherent method on `WalletStealthAuthorizer` to a `TransactionSealSigner` method, and `build` now calls it itself. Downstream 0.16 code following the previous example:

```rust
let a = authorizer.create_authorizations(&tx).await?;
// ...
.with_authorizations(a).build(&authorizer)
```

still compiles whenever `TransactionSealSigner` is in scope, and now attaches every stealth authorization twice. Nothing dedupes by public key — `verify_all_signatures` and the mempool validators accept duplicates — so the transaction silently carries N extra signatures (extra weight plus `per_signature_verification_cost`, 10 µT each) instead of erroring.

Renaming the trait method, or erroring when `self.authorizations` already contains a signature for a key the signer produces, would turn this into a compile/runtime error rather than a silent overpay.

crates/wallet/ootle-rs/src/stealth/builder.rs
```diff
@@ -89,34 +89,36 @@ impl<'a, P: WalletProvider<Wallet = OotleWallet>> StealthTransfer<'a, P> {
...
         let found_substates = self
             .provider
-            .fetch_substates(substate_id_to_addr_map.keys().cloned())
+            .fetch_substates(inputs_by_utxo.keys().cloned())
             .await
             .map_err(|e| StealthProviderError::UnexpectedError {
                 details: format!("Failed to fetch stealth input substates: {}", e),
             })?;
-        if found_substates.len() != self.spec.inputs_to_spend.len() {
+        if found_substates.len() != inputs_by_utxo.len() {
```
sdbondi: **low** — this guard can't catch duplicate inputs. `inputs_by_utxo` is keyed by `SubstateId`, so two `spend_stealth_input` calls naming the same commitment (the same UTXO with two different witnesses, or mistakenly with two different owner addresses) collapse to one entry, last-write-wins. Comparing `found_substates.len()` against `inputs_by_utxo.len()` compares against the already-deduplicated count, so the caller silently gets one input resolved with the *second* entry's witness/owner.

The same class of collapse existed before keyed by address; the key changed here but the guard wasn't updated to compare against the caller's original `Vec` length.

crates/wallet/ootle-rs/src/stealth/builder.rs
```diff
@@ -89,34 +89,36 @@ impl<'a, P: WalletProvider<Wallet = OotleWallet>> StealthTransfer<'a, P> {
...
             return Err(StealthProviderError::UnexpectedError {
                 details: "Some stealth inputs could not be found in the provider substates".to_string(),
             }
             .into());
         }
 
         let mut required_signers = IndexSet::with_capacity(found_substates.len());
```
sdbondi: **low** — nondeterministic seal selection. `required_signers` is populated by iterating `found_substates`, which is a `std::collections::HashMap` (`Provider::fetch_substates` returns `HashMap<SubstateId, Substate>`). Since `SignatureRequirements::stealth_seal` promotes the *first* element, which stealth input seals is nondeterministic across runs, as is the order of `resolved_inputs` in the statement.

Concretely: a wallet holding a mix of key providers (one local, one `LedgerSigner`) spending two stealth inputs will succeed or fail with `"the Ledger device cannot derive a stealth public key without signing"` depending purely on hash iteration order, because `create_authorizations` resolves `seal_public_key()` from whichever input happened to come first. Iterating `inputs_by_utxo` (or sorting by `SubstateId`) would make it deterministic.

crates/wallet/ootle-rs/src/wallet/stealth.rs
```diff
@@ -1,105 +1,120 @@
...
+    StealthInput(StealthSignerRequirement),
+    Ephemeral(EphemeralKeySigner),
+}
+
+impl From<SealSource> for Seal {
+    fn from(source: SealSource) -> Self {
+        match source {
+            SealSource::AccountKey => Self::AccountKey,
+            SealSource::StealthInput(signer) => Self::StealthInput(signer),
+            SealSource::Ephemeral => Self::Ephemeral(EphemeralKeySigner::random()),
```
sdbondi, line 35: **low** — the ephemeral key is now drawn once here and retained (`EphemeralKeySigner::seal_transaction` changed from `self` to `&self`), so a single authorizer seals every transaction it builds with the same key. The documented flow — `provider.sign_and_send_dry_run_with(&authorizer, tx.clone())` followed by `TransactionRequest::…build(&authorizer)` — therefore submits two transactions carrying an identical seal public key, linking them and defeating the stated purpose of the ephemeral seal.

(Nonce reuse is not a risk here: `RistrettoSchnorr::sign` samples randomly.)

Related: the struct doc at `crates/wallet/ootle-rs/src/transaction/ephemeral_signer.rs:12` still says the signer "signs a transaction and throws the secret away", which is no longer true.

crates/wallet/ootle-rs/examples/stealth_transfer.rs
```diff
@@ -64,25 +64,32 @@ async fn main() {
...
+    // seals that transaction and the other must attach an authorization signature committing to the seal signer's
+    // one-time public key.
+    const FIRST_INPUT_AMOUNT: u64 = 6 * TARI;
+    const SECOND_INPUT_AMOUNT: u64 = 10 * TARI + 1000 - 500 - FIRST_INPUT_AMOUNT;
     // // This builder creates a stealth transfer statement (spend proof). This is added to the transaction later.
     let (faucet_transfer, required_signers) = StealthTransfer::new(tari_token, &provider)
         // Tell the transfer to expect 10 TARI (+1000 to cover fees) as revealed funds from a bucket (the faucet looks at this value and automatically provides the bucket).
         .spend_revealed_input(10 * TARI + 1000)
         // The transfer will output 500 micro TARI as revealed funds to pay for the fee
         .to_revealed_output(500u64)
```
sdbondi: **low** — both transactions kept a 500 µT fee budget while getting materially bigger. TX1 now creates two stealth UTXO substates instead of one (`per_substate_create_cost` 25 µT plus `per_byte_storage_cost` 1 µT/byte over the second output's commitment, nonce, tag and padded encrypted data), and TX2 gains a second input plus an extra authorization signature (`per_signature_verification_cost` 10 µT plus ~96 bytes of transaction weight). If the added cost exceeds the previous headroom under 500 µT, the example aborts on insufficient fee at the `pay_fee_from_bucket("fees")` step.

sdbondi (commented):
Review of the multi-input stealth signing fix.

The core fix checks out: `stealth_public_key` derives via the same `kdfs::owner_stealth_dh_secret` path that `sign_prehash_with_stealth_key` signs with, `authorization_message`'s `create_message_v1` is byte-identical to the `to_signing_message` used by `sign_authorization_with_stealth`, and the new regression test passes along with the rest of the lib tests.

Six findings inline, plus one that has no diff line to anchor to:

**`build_unsealed()` silently drops authorizations** (`crates/wallet/ootle-rs/src/types/transaction/request.rs:63`) — it returns `self.state.0.finish()`, discarding `self.authorizations` and never invoking `create_authorizations`. Now that authorization production is the seal signer's responsibility inside `build`, this is the only path that yields an unsealed transaction with zero stealth authorizations: `TransactionRequest::default().with_transaction(tx).add_authorization(a).build_unsealed()` produces a transaction whose inputs are unauthorized, with no error.

PR #2410, opened 2026-08-11
feat(engine)!: split the stealth balance proof challenge for later re-verification

sdbondi:
## Motivation

The stealth balance proof's challenge hashed both statements wholesale:

```rust
engine_hasher64(StealthBalanceProof)
    .chain(public_excess).chain(public_nonce)
    .chain(stealth_inputs_statement)
    .chain(stealth_outputs_statement)
```

Re-deriving that challenge therefore requires reproducing both statements byte-for-byte. Two of the fields make that infeasible after execution:

- each input's `SpendWitness`, bounded only by `STEALTH_LIMITS.max_witness_data_len` (4096 B) **per input**, and inputs are capped at 1000;
- each output's `viewable_balance_proof`, of which only the ElGamal balance — not the proof — survives into the `Utxo` substate.

Nothing smaller than the whole statement can reconstruct the challenge, and statements are discarded once executed (transaction payloads are pruned after `epoch_history_length`). So a stealth balance proof is verified exactly once, by the committees that execute it, and is unverifiable by anyone thereafter.

This is a prerequisite for retaining stealth proofs on-chain so that indexers can re-verify the confidential set rather than trusting that every quorum in every shard group in every epoch behaved.

## What changed

The preimage is now the value-carrying fields plus a digest:

```
domain(StealthBalanceProof)
  || public_excess || public_nonce
  || [input commitments] || revealed_in
  || [(output commitment, minimum_value_promise)] || revealed_out
  || aux_digest
```

where `aux_digest = H_StealthBalanceProofAux(everything the explicit half does not bind)` — the per-input spend witnesses, the per-output authorisation, tag, sender nonce, encrypted data and viewable-balance proof, and the aggregate range proof.

`minimum_value_promise` is in the explicit half because re-verifying the aggregate range proof needs it verbatim: a verifier working from a retained record holds it regardless, so folding it into an opaque digest would leave that verifier using a value it can only check for internal consistency (fail-closed, but never signer-attested). Binding it explicitly makes it the value the signer signed, at no cost to the record.

**The same data is bound as before, each field exactly once.** The two halves partition the statements; nothing is hashed twice. The split only changes how it is presented to the hasher — an intercepted statement still cannot be re-targeted by rewriting an output's `auth`, which is the malleability that matters here (`auth` does not enter the excess, so the commitments and hence the balance proof are untouched by such a rewrite). The sealed transaction covers the statement too; this binding is what protects the pre-seal window in which partial statements circulate during multi-party assembly.

What changes is that a verifier holding only `{commitments, revealed amounts, 32-byte digest}` can now rebuild the challenge via `stealth_balance_proof64_from_parts`.

### Scope of what the from-parts path attests

**Value conservation and range-proof validity.** A verifier working from a retained record can confirm that the signer conserved value across the commitments and revealed amounts in that record, and can re-verify the aggregate range proof against the signed minimum value promises.

It cannot confirm that the `Utxo` substates the committee wrote carry the `auth`, `tag` or `sender_public_nonce` the signer signed — those remain folded into the opaque `aux_digest`, which cannot be recomputed from surviving state since it also covers input witnesses and viewable-balance proofs. See "Open design decision" below.

### Three design points worth review

1. **The aux half destructures exhaustively rather than listing fields.** A hand-maintained field list inside a signature preimage rots — add a field to `UnspentOutput`, forget a line, and it silently becomes malleable. Writing the split as exhaustive `let StealthUnspentOutput { output, auth, tag } = …` / `let UnspentOutput { commitment: _, … } = …` patterns makes the compiler enforce it: a new field is `error[E0027]: pattern does not mention field`, verified by temporarily adding one. So the partition stays disjoint *and* fail-safe, with no field bound twice.

2. **Commitments are bound explicitly**, even though `public_excess` is a function of them. The excess is a lossy sum, so binding the individual commitments prevents a signature being carried over to a different statement that happens to share an excess.

3. **`StealthBalanceProofParts` takes iterators**, so the consensus path projects commitments out of the live statements without materialising a list. `ExactSizeIterator` supplies the length prefix without a walk; `Clone` lets `serialize(&self)` re-iterate. Every realistic caller — slice/vec iterators and `Map` over them, whether projecting from a live statement or from a deserialised record — satisfies both. Deriving the prefix by counting a cloned walk instead was considered and rejected: it substitutes the `Clone` contract for the `ExactSizeIterator` contract rather than removing the assumption, and costs an O(n) pass on every consensus verification of up to 1000 inputs. A mismatched prefix is fail-closed in any case — it makes a valid proof fail to verify, never the reverse.

`stealth_balance_proof64` keeps its signature, so no call sites moved.

## Open design decision (before the retention record lands)

With `minimum_value_promise` moved, the aux digest still covers four fields that survive into the `Utxo` substate: `auth`, `tag`, `sender_public_nonce` and `encrypted_data`. Only the input witnesses and the output viewable-balance proofs are genuinely unreconstructible.

Moving `auth`, `tag` and `sender_public_nonce` into the explicit half too would let the from-parts path attest output *targeting* as well as value conservation — extending the audit from supply integrity to "the committee wrote a UTXO whose `auth` the sender never signed", which is provable, attributable theft detection. It costs ~100 B per output in the retained record.

Note this is not achievable by having the record carry `auth` and comparing it against the `Utxo` substate: both are written by the same execution, so a committee re-targeting an output writes a consistent lie in both. Only a binding under the signer's signature makes it checkable, which means moving the fields into the explicit half.

This is a preimage change either way, so it wants deciding before the record lands rather than after.

## Breaking change

The challenge preimage has changed. Signatures produced by earlier versions no longer verify; signers, verifiers and any external implementation must be upgraded together. Pre-mainnet, so no `ProtocolVersion` activation gate is provided — if this needs to change again after mainnet, that is the mechanism, along with a new `EngineHashDomainLabel` variant so old and new challenges are uninterpretable as each other.

Not included: binding the `resource_address`. It closes no attack I could construct — inputs resolve as `(resource, commitment)`, so a statement replayed under another resource either references UTXOs that do not exist or spends the attacker's own — and it would mean plumbing the resource address through `validate_transfer` and its ~15 callers plus the WASM binding signature. Happy to add it if reviewers disagree.

## Tests

Six new tests in `crates/wallet/crypto/tests/stealth_transfer_statement.rs`:

- `balance_proof_binds_output_auth` — the load-bearing one. Without it, an unsealed statement can be re-targeted at an attacker's spend key with the commitments untouched.
- `balance_proof_binds_output_tag`, `balance_proof_binds_sender_public_nonce`, `balance_proof_binds_input_witness` — the other fields that do not enter the excess and are protected only by the digest.
- `challenge_binds_minimum_value_promise` — asserted against the challenge directly rather than through `validate_transfer`, since tampering with the promise in a statement also invalidates the aggregate range proof and would mask the binding.
- `challenge_reconstructs_from_retained_parts` — asserts the statement path and the from-parts path produce identical challenges. That equality is the invariant the retention work will rest on, and nothing else would catch it drifting.

Verified: `tari_ootle_wallet_crypto` (36 + 10 + 5), `tari_engine --test stealth` (19), `ootle_sdk_core` golden vectors (27, unchanged — no fixture pins challenge bytes), `ootle-wasm-core` (52), `ootle-rs` (27). `cargo check -p tari_engine_types` with default and no-default features. Clippy clean on the touched files. The `ootle-rs` doctest failures are pre-existing on `development`.

## Follow-ups

This lands the preimage only. The retention work it unblocks — a `StealthTransferProof` record on `TransactionReceipt`, part-hashed so it can be pruned independently, plus an indexer verification worker — is separate.


sdbondi (commented):
Reviewed through `ddaaa604e`. No correctness findings outstanding. Posting as a comment rather than an approval since it's my own PR — recording the review for the trail.

## What was verified

- The split is preimage-equivalent in what it binds. The two halves partition the statements: the aux serializer destructures every type exhaustively and discards exactly `{StealthInput.commitment, StealthInputsStatement.revealed_amount, UnspentOutput.commitment, UnspentOutput.minimum_value_promise, StealthOutputsStatement.revealed_output_amount}`, which is precisely what `StealthBalanceProofParts` binds. Every field is covered once.
- Encoding is unambiguous. Both sequences are `u32` count-prefixed; commitments are fixed 32 bytes, `minimum_value_promise` is a `u64` (fixed 8), `Amount` is a fixed 16-byte `u128`, and every remaining aux field is self-delimiting under borsh (enum variant tags, `Option` tags, length-prefixed byte fields).
- Exhaustive destructuring is compile-time enforcement of the new-field safety that hashing-whole previously gave incidentally — a field added to any covered type fails to build rather than silently escaping the binding.
- No second implementation of the challenge exists in the repo: `ootle-rs`, `ootle_sdk_core` and `ootle_wasm` all route through `messages::stealth_balance_proof64`, and `crates/wallet/ledger` has no balance-proof code. No fixture pins a balance proof (`balance_proof` is `null` in every `ootle_sdk_core` stealth fixture), and no doc or spec restates the preimage. The `output_commitments` → `outputs` rename breaks no callers — the test is the only site outside `messages.rs` constructing `StealthBalanceProofParts`.
- `EngineHashDomainLabel` is only ever consumed via `as_label()`, so inserting `StealthBalanceProofAux` mid-enum is safe.
- The tampering tests are non-vacuous: `validate_stealth_outputs_statement` does not check `auth` or `tag`, and `validate_transfer` does not evaluate input witnesses, so those tests can only fail via the balance proof.
- The `391b701ce` development merge is clean — no evil-merge edits, and the net diff against the merge base is still the three files.

One claim from the description I checked and can confirm: folding the aux fields behind a 256-bit digest does not weaken the binding. The intuition that it drops collision resistance from 2^256 to 2^128 doesn't hold, because the 64-byte challenge is reduced to a Ristretto scalar by `sign_raw_uniform` — the effective challenge-collision bound was already ~2^126, so the 2^128 aux collision is not the new minimum.

## Findings and resolutions

**1. `ExactSizeIterator::len()` as the length prefix — raised, withdrawn.**

I flagged that `StealthBalanceProofParts::serialize` derives the count from a safe trait's contract, which an implementor can get wrong, and suggested counting the walk instead. That suggestion is worse than the code. `serialize(&self)` must clone to iterate at all, so deriving the prefix from `clone().count()` and writing the payload from a second traversal counts a *different* walk — it substitutes `Clone`'s contract for `ExactSizeIterator`'s rather than removing the assumption, and a hand-written `len()` and a hand-written `Clone` are the same class of implementor bug. The failure mode is also fail-closed: a prefix disagreeing with its payload makes a valid proof fail to verify, never the reverse. Paying an unconditional O(n) pass over up to 1000 commitments per transfer per validator to defend a hypothetical caller against a violated safe-trait contract is the wrong trade. `ExactSizeIterator` states exactly the invariant needed. Withdrawn.

**2. Aux digest scope — partially taken, remainder decided.**

The original point was that fields surviving into the `Utxo` substate are folded into the opaque digest, so a verifier working from a retained record can attest value conservation but cannot check what the signer actually signed for those fields.

`minimum_value_promise` moved into the explicit half in `a6c25123d`, and the argument that settled it is stronger than the one I made: re-verifying the aggregate range proof needs the promise verbatim, so any verifier working from a record holds it regardless. Leaving it in the digest meant that verifier used a value it could only check for internal consistency — fail-closed, since a tampered promise fails the range proof, but never signer-attested. Moving it costs nothing in the record. The from-parts path now attests value conservation *and* range-proof validity.

`auth`, `tag`, `sender_public_nonce` and `encrypted_data` stay in the digest, deliberately. They would add ~100+ B per output to permanent unpruned consensus state in perpetuity, and they buy a property outside what the retention is being paid for: retention exists for supply integrity, the one failure undetectable no matter how much data is held. Output re-targeting is theft — it has a victim who notices, and sits in the same detectability class as every other way a byzantine committee can fabricate state. Worth recording *why* the property can't be had more cheaply: comparing a record-carried `auth` against the `Utxo` substate proves nothing, since the same execution writes both — only a binding under the signer's signature is checkable. That is exactly why the option is expensive, and why it's declined rather than approximated.

**3. `challenge_binds_minimum_value_promise` didn't pin its stated property — fixed in `ddaaa604e`.**

As originally written the test tampered the *statement* and compared two `stealth_balance_proof64` results, which passes identically under either design: were the promise inside the aux digest, tampering would change that digest and the challenges would differ just the same. It asserted "bound somewhere", not "bound by the explicit half".

The replacement asserts it in the shape that distinguishes the two, which is also the actual threat model — a record carries the promise separately from the digest, so the attack is a record that alters the promise while replaying the digest verbatim. `challenge_from_record` takes the output values from the (tampered) record while computing `aux_digest` from the untampered statements, so a promise bound only inside the digest would contribute nothing and the challenge would reproduce. Confirmed non-vacuous by moving the field back into the digest, which fails this test while `challenge_reconstructs_from_retained_parts` still passes.

## Minor, no action

`stealth_balance_proof_aux32` is `pub` and, standing alone, no longer binds the commitments or revealed amounts the way it did when it hashed the statements whole. Intentional — the retention record computes and stores it at write time — and the doc comment states that it is only meaningful as one half of the pair.

PR #2411, opened 2026-08-11
feat(receipt)!: add an intent commitment to TransactionReceipt

sdbondi:
## Motivation

A transaction receipt is already bound to its transaction transitively — the receipt is addressed by the transaction id, and `TransactionV1::calculate_id` chains `TransactionSignatureFields` (network, fee instructions, instructions, inputs, epochs, flags). But *proving* that link requires reproducing the whole id preimage, which also includes `signatures` and `seal_signature` — so the proof leaks the signers' and the seal signer's public keys.

This adds a dedicated 32-byte commitment that breaks that entanglement: anyone holding a transaction can prove it produced a given receipt, without revealing who authorized it.

## What is committed

The existing `TransactionSignatureFields` projection — everything except the signatures — plus the schema version and the per-blob commitments, under a new `EngineHashDomainLabel::TransactionIntent`.

Reusing the projection rather than hashing instructions alone:
- it is already canonical and already streamed field-by-field to hardware signers via `PreimageField`/`PreimageSegment`, so there is no second encoding of the same data to keep in sync;
- it inherits the right blob behaviour (blob hashes, never blob bytes);
- instructions thread through workspace slots (`WorkspaceOffsetId`), so a single instruction in isolation is semantically incomplete — a Merkle root over instructions would buy little.

## Changes

- `EngineHashDomainLabel::TransactionIntent`.
- `TransactionReceipt::intent_commitment`, additive CBOR index `#[n(7)]`. Serialized in full by `TransactionReceiptHashMessage` — part-hashing 32 bytes is meaningless.
- Populated in the engine where the receipt is built (both the `Commit` and `FeeIntentCommit` paths), threaded from the `Executable` through `StateTracker`/`WorkingState`.
- New `TransactionIntent` trait with `calculate_intent_commitment` and `verify_receipt_intent`, implemented for `Transaction`/`TransactionV1` and `PrunedTransaction`/`PrunedTransactionV1`. The pruned form produces the same commitment as the full form, matching the existing `TransactionId` invariant.

## Tests

- `crates/transaction/src/v1/intent.rs`: commitment is stable across pruning (with and without blobs), binds every projected field (including blob contents), and is unaffected by the number/identity of signers while the transaction id still differs.
- `crates/engine/tests/transaction_receipt.rs`: end-to-end — a committed receipt and a fee-intent receipt both verify against the executing transaction (full and pruned), and a different transaction's intent does not.

## Scope notes

- This is a commitment, not retention. The network still prunes transaction payloads after `epoch_history_length` (`keep_transaction_history` defaults false), so a receipt is provable by whoever kept the transaction — not by anyone.
- Aborted transactions produce no receipt (`FinalizeOutcome` is only `Commit`/`FeeIntentCommit`), so this does not cover proving intent when execution failed outright. Out of scope here.

## Breaking

`TransactionReceipt` gains a field, changing its CBOR encoding and its substate hash. Pre-mainnet, so no `ProtocolVersion` activation gate.


crates/transaction/src/v1/intent.rs
```diff
@@ -0,0 +1,297 @@
...
+    /// The 32-byte commitment to everything the signers authorized.
+    fn calculate_intent_commitment(&self) -> Hash32;
+
+    /// Verifies that `receipt` was produced by this transaction.
+    ///
+    /// This proves the link between a held transaction and a receipt without reproducing the
+    /// transaction id, and so without revealing the signers. It says nothing about whether the
+    /// receipt itself is authentic — the caller is responsible for obtaining the receipt from the
+    /// chain.
+    fn verify_receipt_intent(&self, receipt: &TransactionReceipt) -> Result<(), IntentCommitmentMismatch> {
```
sdbondi, line 69: **medium — the doc claims more than the commitment provides.**

`verify_receipt_intent` is documented as proving "that `receipt` was produced by this transaction", but the commitment binds no signer material and has no blinding factor, so it is neither a proof of authorship nor hiding.

Concrete false positive: two users each submit a transaction with byte-identical `TransactionSignatureFields` — same network, same instructions, same *unversioned* inputs (`VersionedSubstateId::unversioned`, e.g. a faucet claim), same epoch bounds and flags — but different signers and seal signers. The two transactions have different `TransactionId`s and different receipts, yet identical intent commitments, so Alice's transaction returns `Ok(())` against Bob's receipt.

Separately, because the commitment is unblinded and published on-chain over a guessable preimage, any mempool observer who saw the transaction can recompute it and construct the same "proof".

Suggest the module and trait docs state that this is a linkage/lookup aid over a guessable preimage — it hides the *signers*, which is the stated goal — rather than an authorship proof.

crates/engine_types/src/transaction_receipt.rs
```diff
@@ -37,6 +37,17 @@ pub struct TransactionReceipt {
...
+    /// Commitment to the transaction's intent: every field the signers authorized (network, fee
+    /// instructions, instructions, inputs, epoch bounds, flags and blob commitments), excluding the
+    /// signatures themselves.
+    ///
+    /// The receipt is already bound to the transaction transitively — it is addressed by the
+    /// transaction id — but reproducing that id requires the signatures and the seal signature, so
+    /// proving the link reveals the signers. This commitment is over the same projection minus the
+    /// signatures, so whoever holds the transaction can prove it produced this receipt without
+    /// revealing who authorized it.
+    #[n(7)]
```
sdbondi, line 50: **low — no `#[cbor(default)]` on the new field.**

`TransactionReceipt` uses array (not map) CBOR encoding, so every already-persisted receipt — indexer sqlite, wallet stores, VN state store — decodes as a 7-element array and now fails with a missing-value error rather than defaulting.

The repo's own precedent for exactly this is `PrunedUnsignedTransactionV1::blob_sizes`, which carries `#[cbor(default)]`. If a network reset is assumed this is fine, but it's worth being a deliberate call rather than an omission.

crates/wallet/ootle-rs/src/provider/tx_watcher.rs
```diff
@@ -460,6 +461,10 @@ impl PendingTransaction {
                             .map(|res| res.finalize.fee_receipt.clone())
                             .unwrap_or_default(),
                         epoch: execute_epoch,
+                        // Locally reconstructed from the execution result, which does not carry the
+                        // commitment. Zero marks it as absent — a caller wanting to prove the
+                        // transaction produced this receipt must fetch the indexed receipt.
+                        intent_commitment: Hash32::default(),
```
sdbondi: **low — the zero sentinel is invisible at the API boundary.**

A caller who does the obvious `tx.verify_receipt_intent(&receipt)?` on the result of `PendingTransaction::get_receipt()` gets `IntentCommitmentMismatch { actual: 0x00…00 }` and cannot distinguish "this is the wrong transaction / a forged receipt" from "the commitment was never populated".

Consider returning `PendingTransactionError::ReceiptNotFound` here (the commented-out line just below), or making the field `Option`-shaped for locally reconstructed receipts.

crates/transaction/src/v1/transaction.rs
```diff
@@ -239,6 +243,17 @@ impl TransactionV1 {
     }
 }
 
+impl TransactionIntent for TransactionV1 {
+    fn calculate_intent_commitment(&self) -> Hash32 {
+        let unsigned = self.body.unsigned_transaction();
+        calculate_intent_commitment_v1(
+            self.schema_version(),
+            &TransactionSignatureFields::from(unsigned),
+            &unsigned.blobs.hashes(),
```
sdbondi: **low — re-hashes blob payloads on the execution hot path.**

`calculate_intent_commitment` calls `unsigned.blobs.hashes()`, and `Blobs::hashes()` is uncached (`self.0.iter().map(hash_blob).collect()`). `processor.rs:180` now calls this once per execution *in addition to* `calculate_id()`, which already hashes the same payloads — so a `PublishTemplate` transaction carrying a multi-megabyte WASM blob gets its blobs fully re-hashed on every validator, during execution.

The pruned form avoids this by using the stored `BlobHashes`; the full form could take the hashes it already has to hand, or cache them.

sdbondi (commented):
Code review of `577008db7`. Builds pass (`tari_ootle_transaction`, `tari_engine`, with tests); all `TransactionReceipt` literal sites and `Executable` impls are updated.

The core mechanic looks sound: the commitment reuses `TransactionSignatureFields` + `schema_version` + `BlobHashes` under a distinct engine hash domain (`TransactionIntent`, `com.tari.ootle.engine`) that cannot collide with the transaction-crate signing domain; the full and pruned forms genuinely produce the same bytes; `EngineHashDomainLabel` is string-keyed so inserting a variant mid-enum is safe; and the fee-checkpoint path carries `intent_commitment` because it clones `WorkingState`.

Four findings inline — one medium (doc claim vs. what the commitment actually proves) and three low.

crates/transaction/src/v1/intent.rs
sdbondi: Agreed on the doc, not on it being a defect — excluding signer material is the point, and the collision is inherent to that. Reworded rather than redesigned (b4702130e8).

Module docs now carry a "What it is not" section: not a proof of authorship (transactions differing only in who signed them share a commitment and each verifies against the other's receipt), and not hiding (the preimage is the gossiped transaction body). `verify_receipt_intent` now says `Ok(())` means the receipt came from *a* transaction with exactly this intent, not that this transaction produced it. The receipt field doc and the test names/comments that said "prove" were reworded to match.

Blinding it would need a nonce the receipt cannot carry publicly, so the linkage-aid framing is the honest one.

crates/engine_types/src/transaction_receipt.rs
sdbondi: Deliberate — leaving it without a default.

The substate hash changed too (the field is serialized by `TransactionReceiptHashMessage`), so a receipt written before this change is already inconsistent with the new schema. `#[cbor(default)]` would let it decode silently as a receipt carrying a zero commitment whose recomputed substate hash no longer matches what consensus committed — the same invisible-sentinel problem raised on the tx_watcher thread, but in consensus-relevant data. A hard decode failure is the correct loud break for a schema change that is breaking on purpose, pre-mainnet, with a reset assumed.

`blob_sizes` defaults because it is explicitly outside the signing/id domain and cosmetic; this field is inside the substate hash, so the precedent does not carry over.

crates/wallet/ootle-rs/src/provider/tx_watcher.rs
sdbondi: Fixed in b4702130e8, and better than the placeholder: the execution result's accept diff carries the receipt substate itself, so `get_receipt` now returns that — real commitment, events, logs and fee receipt — instead of reconstructing the fields by hand. `ReceiptNotFound` is returned only when there is no execution result to take it from, which is the case where the reconstruction was near-empty anyway (outcome `Commit`, epoch 0, no fees).

Note this is a small behaviour change: that path used to always return `Ok` with a fabricated receipt.

crates/transaction/src/v1/transaction.rs
sdbondi: Fixed in b4702130e8. `TransactionV1::calculate_id_and_intent_commitment` derives `BlobHashes` once and feeds both digests, with `calculate_id` / `calculate_intent_commitment` now thin wrappers over private `_with_blob_hashes` helpers — same idiom as `create_message_v1_with_blob_hashes`. `Executable::to_id_and_intent_commitment` carries it to the processor (default impl calls both, overridden for `Transaction`), so execution is back to one pass over the blobs.

`paired_derivation_matches_the_separate_entry_points` pins the two paths together — drift there would mean a receipt committing to a different intent than a verifier computes.

sdbondi (commented):
Re-reviewed at `21ff854d4`, after the rebase onto #2352 and the follow-ups. All four findings from the previous review are resolved, and the rebase did not disturb them. LGTM from my side — recorded as a comment rather than a formal approval, since GitHub does not accept a self-approval on one's own PR; this still needs an approving review from another maintainer.

**Findings from the previous round**

- **Doc claim (medium)** — resolved as a wording change, which is the right call: the collision is inherent to excluding signer material, so there was nothing to redesign. The "What it is not" section and the reworded `verify_receipt_intent` doc (`Ok(())` means the receipt came from *some* transaction with exactly this intent) now match what the digest actually establishes.
- **`#[cbor(default)]` (low)** — rejection accepted, and the reasoning is stronger than the finding. `intent_commitment` is serialized by `TransactionReceiptHashMessage` (`substate_hasher.rs:145`), so it is inside the substate hash; defaulting would let a pre-change receipt decode as a zero commitment whose recomputed hash disagrees with what consensus committed. A hard decode failure is the correct break for a deliberately breaking, pre-reset schema change, and `blob_sizes` is not a precedent because it sits outside the signing/id domain.
- **tx_watcher sentinel (low)** — resolved better than proposed: `get_receipt` returns the receipt substate from the execution result's accept diff, with `ReceiptNotFound` as the fallback. Worth carrying into the PR description — that path previously always returned `Ok` with a fabricated receipt, so any caller relying on the infallible path now sees an error.
- **Blob re-hashing (low)** — resolved via `calculate_id_and_intent_commitment` and `Executable::to_id_and_intent_commitment` (overridden for `Transaction`), with the processor using the paired entry point, so execution is back to one pass over the blobs. `paired_derivation_matches_the_separate_entry_points` guards drift between the paired and separate paths, which is the failure that would actually matter — a receipt committing to a different intent than a verifier computes.

**Delta since**

The rebase onto #2352 lands the intent commitment on top of the reworked id derivation, and the two compose correctly: `calculate_id_with_blob_hashes` now delegates to the shared `calculate_id_v1` (signatures plus the seal signer's public key, no seal witness data), while the commitment path is untouched, and both still share the single `blobs.hashes()` pass. #2352 narrows what the id binds but keeps signatures and the seal signer in it, so the premise the commitment exists to work around — reproducing the id reveals who authorized the transaction — still holds. The `code nit` commit rewrites the `get_receipt` fallback as `.cloned().ok_or_else(...)`, same behaviour.

No new findings. Merge gate is CI plus a second reviewer.

PR #2416, opened 2026-08-11
fix(engine)!: charge storage for the transaction receipt and drop its logs

sdbondi:
> **Stack:** this PR ← #2417 (charge finalization fees against the state actually persisted). Merge this one first.

## Problem

The transaction receipt is persisted like any other substate — it is never downed, and it is streamed to every syncing node ([`block_sync_task.rs:324-340`](https://github.com/tari-project/tari-ootle/blob/development/applications/tari_validator_node/src/p2p/rpc/block_sync_task.rs)) — but nothing charged for its bytes.

`FeeModule::on_before_finalize` prices the substates in `mutated_substates`, and the receipt only joins the diff afterwards, in `tracker.finalize`. It therefore escaped both the per-byte storage charge and the per-substate create charge. Since `emit_log` costs a flat `per_module_call_cost` (1 µT on testnet) regardless of message size, the engine limits (`max_logs: 256`, `max_log_size_bytes: 32 KiB`) put **8 MiB of permanent, replicated, unpriced state within reach of any transaction for 256 µT**.

## Charge for the receipt

`on_before_finalize` now adds the receipt's encoded size to the storage tally and counts its slot in the create charge.

The size cannot be measured exactly at that point: the receipt embeds the `FeeReceipt` that this very charge feeds into, so an exact figure would need a fixed point. `TransactionReceipt::encoded_size_upper_bound` instead measures the parts already fixed at charge time (events, fee withdrawals, epoch, intent commitment) and substitutes worst-case stand-ins for the two that are not — `FeeReceipt::widest()` and a max-width version on each diff-summary entry. It never comes in under the receipt that is actually persisted; unit tests hold the overshoot below 256 bytes.

## Remove `logs` from the receipt

Logs are `println!` for template authors, not contract semantics — events already carry the indexable, subscribable record, and forcing debug output through events would pollute what indexers consume without shrinking anything.

Logs remain on `FinalizeResult`, so **the wallet CLI, dry runs, the SDK's `LogSummary` and the indexer's transaction-result endpoint are unaffected**. Two consumers did read them off the persisted receipt and now show events only:

- `ootle-rs`'s `PendingTransaction::get_receipt()` — the `fungible_transfer` and `stealth_transfer` examples
- the indexer web UI's receipt page (the Transaction page still renders `finalize.logs`)

## Breaking

`TransactionReceipt` drops `logs` and renumbers the CBOR indices after it, and the log digest leaves the receipt hash. Existing receipts do not decode and their substate hashes differ — this needs a network reset.

## Known follow-up: the reject path over-charges

Every finalize-time charge is computed against the *live* state, but on the reject path the diff that gets persisted comes from the fee checkpoint. `Storage`, `SubstateCreate` and `TemplatePublish` are all priced over substates that are then discarded, and `ExhaustBurn` inherits the inflation since it is derived from `total_fee_charges()`.

This predates the PR — the per-substate storage charge has always behaved this way — but pricing the receipt amplifies it, because the bound scales with the discarded state's substate count and events. In `fail_partial_paid_fees` that is ~1000 µT of storage against a ~160 µT receipt.

Fixing it properly means running the finalize-time charges twice — provisionally against the live state to gate the commit/reject decision, then authoritatively against the state actually persisted — which needs `set` rather than `add` semantics on the finalize-time fee sources and a hook fired after state selection (splitting `tracker.finalize`, since the tracker does not own the modules). Left out of this PR to keep it reviewable.

## Testing

`tari_engine` (all suites), `engine_types`, `transaction`, `ootle_sdk_core` including golden vectors, `storage`, `tari_bor` and `consensus_tests` pass. Three fee tests needed real updates: two paid fixed `max_fee`s now short of the requirement, and `fail_partial_paid_fees` asserted total fees stay near the amount paid, which now excludes storage.

### DIFF
--- a/applications/tari_indexer/web_ui/src/routes/TransactionReceipts/components/TransactionReceiptDetails.tsx
+++ b/applications/tari_indexer/web_ui/src/routes/TransactionReceipts/components/TransactionReceiptDetails.tsx
@@ -26,7 +26,6 @@ import { CURRENCY } from "../../../utils/constants";
 import { formatCurrency } from "../../../utils/helpers";
 import EventsContent from "../../Transaction/components/EventsContent";
 import FeeReceipt from "../../Transaction/components/FeeReceipt";
-import LogsContent from "../../Transaction/components/LogsContent";
 import SubstateChanges from "./SubstateChanges";
 
 function TransactionReceiptDetails({ address }: { address: string }) {
@@ -41,7 +40,7 @@ function TransactionReceiptDetails({ address }: { address: string }) {
     setExpandedPanels((prev) => (isExpanded ? [...prev, panel] : prev.filter((p) => p !== panel)));
   };
 
-  const expandAll = () => setExpandedPanels(["p1", "p2", "p3", "p4", "p5"]);
+  const expandAll = () => setExpandedPanels(["p1", "p3", "p4", "p5"]);
   const collapseAll = () => setExpandedPanels([]);
 
   const receipt = data?.receipt;
@@ -128,17 +127,6 @@ function TransactionReceiptDetails({ address }: { address: string }) {
                   </AccordionSummary>
                   <AccordionDetails>
                     <EventsContent data={receipt.events} />
-                  </AccordionDetails>
-                </Accordion>
-              )}
-
-              {receipt.logs.length > 0 && (
-                <Accordion expanded={expandedPanels.includes("p2")} onChange={handleChange("p2")}>
-                  <AccordionSummary>
-                    <Typography variant="h5">Logs ({receipt.logs.length})</Typography>
-                  </AccordionSummary>
-                  <AccordionDetails>
-                    <LogsContent data={receipt.logs} />
                   </AccordionDetails>
                 </Accordion>
               )}
--- a/applications/tari_walletd/web_ui/src/routes/AssetVault/Tokens/components/ClaimCoinsButton.tsx
+++ b/applications/tari_walletd/web_ui/src/routes/AssetVault/Tokens/components/ClaimCoinsButton.tsx
@@ -91,7 +91,10 @@ function ClaimCoinsButton() {
     claimTestnetFaucetFunds(
       {
         account: { ComponentAddress: accountAddress },
-        fee: 1000,
+        // The claim creates the account and funds it, so it pays for a component, a vault, a badge
+        // and the receipt. Allocated well above that — the excess is refunded, and too low a value
+        // rejects the claim after the funds appear to have arrived.
+        fee: 50_000,
       },
       {
         onSuccess: async (resp: AccountsCreateFreeTestCoinsResponse) => {
--- a/crates/engine_types/src/transaction_receipt.rs
+++ b/crates/engine_types/src/transaction_receipt.rs
@@ -4,6 +4,7 @@
 use std::{fmt, fmt::Display, str::FromStr};
 
 use serde::{Deserialize, Serialize};
+use tari_bor::adapters::boxed_slice;
 use tari_template_lib::types::Hash32;
 
 use crate::{
@@ -11,7 +12,6 @@ use crate::{
     ValidatorFeeWithdrawal,
     events::Event,
     fees::FeeReceipt,
-    logs::LogEntry,
     substate::{SubstateDiff, SubstateId, hash_substate},
 };
 
@@ -31,11 +31,8 @@ pub struct TransactionReceipt {
     #[cbor(with = "tari_bor::adapters::boxed_slice")]
     pub events: Box<[Event]>,
     #[n(4)]
-    #[cbor(with = "tari_bor::adapters::boxed_slice")]
-    pub logs: Box<[LogEntry]>,
-    #[n(5)]
     pub fee_receipt: FeeReceipt,
-    #[n(6)]
+    #[n(5)]
     pub epoch: Epoch,
     /// Commitment to the transaction's intent: every field the signers authorized (network, fee
     /// instructions, instructions, inputs, epoch bounds, flags and blob commitments), excluding the
@@ -47,7 +44,7 @@ pub struct TransactionReceipt {
     /// projection minus the signatures, so whoever holds the transaction can link it to this
     /// receipt without revealing who authorized it. It identifies the intent, not the signers:
     /// transactions differing only in who signed them share a commitment.
-    #[n(7)]
+    #[n(6)]
     pub intent_commitment: Hash32,
 }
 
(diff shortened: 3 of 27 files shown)


crates/engine_types/src/transaction_receipt.rs
```diff
@@ -83,6 +76,49 @@ impl TransactionReceipt {
...
+    /// parts that are not yet known are replaced by worst-case stand-ins — a fee receipt whose
+    /// amounts all encode at full varint width, and a max-width version on each diff-summary entry.
+    /// Everything else (`events`, `fee_withdrawals`, `epoch`, `intent_commitment`) is measured as it
+    /// will actually be encoded.
+    ///
+    /// `upped` is the substates that will appear in the [`DiffSummary`] — one entry each. Nothing
+    /// joins that set after the charge is computed: fee settlement only mutates substates already in
+    /// it, and building the diff can only drop entries, never add them. The receipt is up'd after
+    /// its own summary is built, so it is absent from both.
+    pub fn encoded_size_upper_bound<'a>(
```
SWvheerden, line 95: **The "nothing joins the upped set after the charge" invariant does not hold on the fee-intent-commit path.**

The bound is computed from the *live* working state (`WorkingState::transaction_receipt_size` passes `self.store.mutated_substates().keys()`), but on the reject path `StateTracker::finalize` builds the receipt from `checkpoint_state`, whose `new_substates` is not necessarily a subset of the live one: `WorkingStateStore::down_utxo` (state_store.rs:205) and `down_confidential_output` (state_store.rs:237) both `shift_remove` from `new_substates`.

Scenario: a UTXO / confidential output is mutated during the fee instructions (so it lands in the checkpoint's `new_substates`), the main instructions then spend it (removing it from the live set), and the transaction is rejected. The receipt actually persisted carries an `UpSubstate` entry (~70-80 bytes with a `SubstateId`, `u32` version and 32-byte hash) that the bound never counted. With enough such entries the fixed slack from `FeeReceipt::widest()` is exhausted and the bound comes in *under* the persisted receipt — the case the doc comment says can never happen.

utilities/traffic-sim/src/main.rs
```diff
@@ -1,6 +1,13 @@
 //   Copyright 2025 The Tari Project
 //   SPDX-License-Identifier: BSD-3-Clause
 
+/// Fee allowance for every transaction this simulator submits.
+///
+/// An upper bound, not a payment — the unspent remainder is refunded — so it is set well clear of
+/// what these cost rather than tuned to them. A tight value stops the simulator dead the next time
+/// the engine's pricing moves.
+pub const MAX_FEE: u64 = 50_000;
```
SWvheerden, line 9: **Three `max_fee` values in this same binary were left behind.**

`MAX_FEE` is introduced here and applied at `sim.rs:322` and `sim.rs:504`, but the simulator still hardcodes:

- `sim.rs:302` — `create_free_test_coins(... max_fee: 1500)` in the exchange-wallet bootstrap
- `sim.rs:460` — `create_free_test_coins(... max_fee: 1500)` in the per-wallet funding loop
- `sim.rs:255` — `accounts_stealth_transfer(... max_fee: 10000)`

The two `1500`s are exactly the operation this PR bumped from `1000` to `50_000` in `ClaimCoinsButton.tsx`, with the comment "too low a value rejects the claim after the funds appear to have arrived". The claim creates a component, a vault, a badge and now pays for the receipt, so `1500` is below the required fee: the simulator's bootstrap/funding step will be rejected with `Insufficient fees paid` and the run dies before it generates any traffic — the exact failure mode the new doc comment says `MAX_FEE` exists to prevent.

applications/tari_walletd/web_ui/src/routes/AssetVault/Tokens/components/ClaimCoinsButton.tsx
```diff
@@ -91,7 +91,10 @@ function ClaimCoinsButton() {
     claimTestnetFaucetFunds(
       {
         account: { ComponentAddress: accountAddress },
-        fee: 1000,
+        // The claim creates the account and funds it, so it pays for a component, a vault, a badge
+        // and the receipt. Allocated well above that — the excess is refunded, and too low a value
+        // rejects the claim after the funds appear to have arrived.
+        fee: 50_000,
```
SWvheerden, line 97: **The CLI path for this same claim was not bumped.**

`applications/tari_wallet_cli/src/command/account.rs:190` still does:

```rust
max_fee: args.fee.unwrap_or(1500),
```

for the identical `AccountsCreateFreeTestCoinsRequest`. By the reasoning in the comment added here, `wallet-cli account free-coins` without an explicit `--fee` will now be rejected for insufficient fees — and per the same comment, it fails after the funds appear to have arrived.

integration_tests/src/wallet_daemon_client.rs
```diff
@@ -86,7 +94,7 @@ pub async fn claim_fees(
     let request = ClaimValidatorFeesRequest {
         account: Some(ComponentAddressOrName::Name(account_name)),
         claim_key_index: None,
-        max_fee: 1500,
+        max_fee: MAX_FEE,
```
SWvheerden, line 97: **The CLI default for the same `claim_validator_fees` call is still `1500`.**

`applications/tari_wallet_cli/src/command/validator.rs:102`:

```rust
max_fee: max_fee.map(Into::into).unwrap_or(1500),
```

If `1500` is now too low for this operation in the integration harness (which is why it is being raised to `50_000` here), the CLI's default is too low as well and `wallet-cli validator claim-fees` without `--max-fee` will be rejected.

applications/tari_indexer/web_ui/src/routes/TransactionReceipts/components/TransactionReceiptDetails.tsx
```diff
@@ -41,7 +40,7 @@ function TransactionReceiptDetails({ address }: { address: string }) {
     setExpandedPanels((prev) => (isExpanded ? [...prev, panel] : prev.filter((p) => p !== panel)));
   };
 
-  const expandAll = () => setExpandedPanels(["p1", "p2", "p3", "p4", "p5"]);
+  const expandAll = () => setExpandedPanels(["p1", "p3", "p4", "p5"]);
```
SWvheerden, line 43: **A third receipt-logs consumer was missed.**

`applications/tari_indexer/web_ui/src/routes/Substates/Substates.tsx:245` renders logs off a `TransactionReceipt` substate:

```tsx
{data.logs?.length > 0 && (
  <>
    <Typography variant="subtitle2">Logs ({data.logs.length})</Typography>
    <CodeBlock>{renderJson(data.logs)}</CodeBlock>
  </>
)}
```

`data.logs` is now always `undefined`, so it is dead markup (the optional chain keeps it from throwing). Worth removing alongside the `LogsContent` panel here so the substate viewer and the receipt page stay in sync.

PR #2417, opened 2026-08-11
fix(engine)!: charge finalization fees against the state actually persisted

sdbondi:
> **Stack:** #2416 ← this PR. Based on #2416; review the top commit only. Merge #2416 first.

## Problem

Every finalization charge is computed against the **live** working state. But a transaction that cannot pay for its main intent commits only its fee intent, and that diff comes from the **fee checkpoint** — a snapshot taken when the fee instructions ended.

So `Storage`, `SubstateCreate` and `TemplatePublish` were priced over substates that are then thrown away, and `ExhaustBurn` inherited the inflation since it is taken over the running total. This predates #2416; pricing the receipt just made it visible, because the receipt bound scales with the discarded state's substate count and events.

Measured on `fail_partial_paid_fees`: **1019 µT of storage charged against a receipt and substate set that occupy 444 µT**.

## Why the obvious fix doesn't work

Charging only at the end cannot fix this on its own. The charge has to be in place before the paid-in-full check in `select_finalized_state`, or nothing gates a transaction that commits state it never paid for. But the charge decides the outcome, the outcome picks the state, and the state determines the charge — a cycle.

## Charge twice

- **`on_before_finalize`** runs against the live state, as before. Its only job is to gate the commit/reject decision. Pricing what *would* be persisted is the right question to gate on: to commit that state you must be able to pay for it.
- **`on_before_persist`** (new hook) runs once the state has been chosen and before its fees are settled, replacing those charges with the cost of what is really written.

On a commit the second pass sees the same state and changes nothing. On a fee-intent commit it sees only what the fee intent touched.

For the replacement to be sound the finalization charges are now *assigned* rather than accumulated — `FeeBreakdown::set` — so a second pass over a different state replaces the first instead of doubling it. `ExhaustBurn` is zeroed before its total is taken, so it never compounds across passes.

## Structure

`StateTracker::finalize` is split into `select_finalized_state` (chooses the working state or the fee checkpoint, and carries the live fee state onto it) and `finalize` (settles fees, builds the diff and receipt), so the runtime can drive its modules in between. The two near-identical commit and fee-intent-commit tails collapse into one. `WorkingState` becomes `pub(crate)` and gains `count_newly_created_substates`, so the whole computation can run against a detached state.

## Consequences worth reviewing

Two, in opposite directions.

**Refunds.** A transaction rejected for insufficient fees now gets refunded the storage it was provisionally charged for state that was never written. Execution charges (`WasmExecution`, `NativeExecution`, `RuntimeCall`, …) are untouched — the work was done and is still paid for.

**Free compute.** This is the direction that changes what an attacker gets, so it needs saying plainly. Rejecting the fee intent turns "cannot cover its own commit" from `AcceptFeeRejectRest`, where the fee was collected, into `Reject`, where **nothing is collected**. Execution funded by the payment-capped compute allowance therefore gets done and never paid for, repeatably.

The bound is the whole payment. `StateTracker::wasm_point_allowance` is `points_funded_by(total_payments)` — the full payment, with nothing reserved for the fee intent's own commit cost — so a main intent that runs to its allowance leaves `WasmExecution` alone at roughly the payment, and the fee intent's charges on top put the total above it. A transaction that spends its allowance is therefore *guaranteed* to finish unaffordable: this is the ordinary outcome for compute-heavy transactions, not an edge case, and the same funds work again next time.

Two things still bound it: the per-block metering budget caps throughput, and the same was already true whenever the fee instructions themselves failed, which has always produced a `Reject`.

It is still a widening, and the trade is deliberate: unpaid *state* is a permanent liability replicated across the shard and re-streamed to every syncing node, while unpaid *compute* is a one-off already bounded by block limits. Trading more of the second for none of the first is the right side of that. Reserving the commit cost from the compute allowance closes it properly and is the first follow-up.

## Breaking

`Amount` drops its hand-rolled two-element digit-array codec for minicbor's native `u128` — a compact
integer up to `u64::MAX`, a bignum above. Decimal-string decoding is kept; the array form is not, so
every `Amount` already on disk or in flight is undecodable. `Amount` is embedded in vaults, resources,
confidential outputs, fee receipts and every persisted transaction receipt, so this needs the same
network reset #2416 does.

`FinalizeResult` gains `total_fees_required` (additive on the JSON wire, defaulted when absent), and
`FeeBreakdown` no longer carries rows for sources charged zero.

## Testing

New regression test `a_fee_intent_commit_is_not_charged_for_the_state_it_abandons` runs the same transaction at two fee levels and asserts the rejected run's storage charge is less than half the committed run's. With `on_before_persist` stubbed out it fails at **1769 vs 1772** — the two are within three bytes, which is exactly the bug.

`tari_engine`, `engine_types`, `transaction`, `ootle_sdk_core`, `storage` and `consensus_tests` all pass with no other test changes.

## Follow-up

Storage is still charged in one tally at finalization, so a transaction that cannot afford its state only finds out after doing all the work — and `checkpoint_fee_intent`'s paid-in-full check still sees no storage charge at all, since none has been computed by then. Charging incrementally during execution would close both. It composes cleanly with this PR: with the second pass authoritative, the running charge only has to be an over-estimate.


crates/engine/src/runtime/impl.rs
```diff
@@ -279,6 +279,34 @@ impl<TStore: StateReader + Clone + 'static, TTemplateProvider: TemplateProvider<
...
+    /// which on a fee-intent commit holds only what the fee intent touched. A transaction is
+    /// therefore gated on the cost of the state it asked to commit, but pays for the state that is
+    /// really persisted.
+    fn finalize_with(&mut self, failure: Option<RejectReason>) -> Result<FinalizeResult, RuntimeError> {
+        self.invoke_modules_on_before_finalize()?;
+        let mut finalized = self.tracker.select_finalized_state(failure)?;
+        // A commit persists the very state the first pass charged against, so charging it again
+        // would recompute the same numbers from the same inputs. Only a fee-intent commit swaps the
+        // state out from under those charges, and only it needs them redone.
+        if !finalized.outcome().is_commit() {
```
SWvheerden, line 297: **A fee-intent commit now reports a `required_fees()` that can never succeed.**

On the `FeeIntentCommit` path `on_before_persist` *replaces* `Storage`/`SubstateCreate`/`TemplatePublish`/`ExhaustBurn` with the checkpoint's numbers, and that reduced breakdown is what `finalize_fees_and_refunds` puts into the `FeeReceipt` returned with `AcceptFeeRejectRest`. The new `fail_partial_paid_fees` assertions pin the consequence exactly: `total_fees < FEE_PAID` and `payment.is_paid_in_full()`.

So for a transaction rejected with `InsufficientFeesPaid`, the receipt now says the fee intent's cost — which is by construction *below what was already paid* — while the amount actually needed (the working-state charge that `select_finalized_state` gated on) survives only inside the free-text reason string.

Concrete failure: `handle_submit_dry_run`, `handle_publish_template`, `handle_claim_burn` and `TransactionApi::…dry_run` all surface `fee_receipt.required_fees()` as *the* `max_fee` to submit with. A dry run whose main intent fails (or is under-funded) takes this same path, so the caller is handed, say, 807 µT for a transaction that needs 2000 µT, resubmits at 807, is rejected again with the identical number, and the loop never converges. Before this PR the receipt carried the working-state charges — an over-estimate, but a usable one.

Secondary contradiction from the same cause: the result now reads `AcceptFeeRejectRest(diff, InsufficientFeesPaid("Required fees 2000 but 1000 paid"))` while its own receipt reports `is_paid_in_full() == true` and issues a refund. `ExecuteResult::expect_fees_paid_in_full` and any wallet UI pairing the two see mutually exclusive facts.

Consider keeping the gating (working-state) total reachable on the receipt — e.g. record it before the second pass overwrites it, the way commit c844287 already does for the `Reject` path.

crates/engine/src/fees/fee_table.rs
```diff
@@ -1,6 +1,23 @@
 //   Copyright 2023 The Tari Project
 //   SPDX-License-Identifier: BSD-3-Clause
 
+use tari_ootle_transaction::LITERAL_BYTE_DIVISOR;
+
+/// The narrowest `max_fee` a run can meter at is `1`, which a dry run clamps to. `Amount` encodes as
+/// a native CBOR integer, so that is a single byte.
+const NARROWEST_FEE_LITERAL_BYTES: u64 = 1;
```
SWvheerden, line 8: **"which a dry run clamps to" — nothing clamps it, and the allowance depends on that premise.**

The derivation deliberately leaves the residual-vault-balance term out of `fee_estimate_allowance`, on the grounds that it is one-directional: "a dry run holds `max_fee` at its lowest and so already sees the widest residual". I can't find anything that establishes that. `submit_transaction_dry_run` / `DryRunTransactionProcessor` execute the transaction exactly as submitted — the `max_fee` is whatever literal the caller put in the `pay_fee` instruction — and there is no clamp anywhere in the engine, indexer, walletd or SDK (grep for `is_dry_run` / `dry_run` turns up no fee adjustment).

The shipped flow is the opposite of the assumed one: `handle_submit_dry_run` and `handle_publish_template` dry-run at the caller's *generous* `max_fee` and hand back `required_fees()` as the fee for the real submission. Dry-run-high → submit-low means the residual **grows**, and storage with it.

Worked case, testnet table (`per_byte_storage_cost = 1`, `storage_cost_divisor = 1`): vault holds 1_000_010; dry run at `max_fee = 1_000_000` leaves a residual of 10, which `Amount`'s new native-CBOR encoding writes in 1 byte. The submission at `required_fees() ≈ 400_000` leaves 600_010 — 5 bytes. That is +4 µT of `Storage`, and the exhaust burn re-multiplies it (+8 at the 10_000 bps ceiling the allowance is derived against), against a `FEE_ESTIMATE_ALLOWANCE` of 7. The submission is rejected as underpaid — the exact failure this constant exists to prevent.

`required_fees_covers_a_real_run_at_any_max_fee` doesn't catch it because it takes `receipts[0]` (the *smallest* `max_fee`) as the estimate, i.e. it asserts the property only under the assumption in question.

Either make the comment true (execute dry runs with the fee literal clamped to its narrowest) or fold the residual's width span into `fee_estimate_allowance` the way the literal's span already is.

crates/template_lib_types/src/amount/amount.rs
```diff
@@ -73,28 +37,20 @@ impl<'b, C> minicbor::Decode<'b, C> for Amount {
                 s.parse::<Amount>()
                     .map_err(|e| minicbor::decode::Error::message(format!("Amount: invalid string '{}': {}", s, e)))
             },
-            other => Err(minicbor::decode::Error::message(format!(
-                "Amount: unexpected CBOR datatype {:?}",
-                other
-            ))),
+            _ => d.u128().map(Amount::new),
```
SWvheerden, line 40: **Every `Amount` already on disk becomes undecodable, with no fallback arm.**

The decoder previously accepted `Type::Array`/`ArrayIndef` and rebuilt the value via `from_le_digits`. That arm is gone: an array now falls through to `_ => d.u128()`, which errors with a datatype mismatch. `Amount` is embedded in vaults, resources, confidential outputs, fee receipts and every persisted transaction receipt, so any state store written before this commit fails to decode after it — as does any in-flight message from a peer that hasn't upgraded.

The commit message calls this out ("This changes the encoded form of every `Amount` on the wire"), but the PR body's "Consequences worth reviewing" section — which does discuss the refund and free-compute trades in detail — doesn't mention it, and the change rides in on a commit whose stated subject is deriving the fee allowance. Worth confirming a network reset is planned in the same release; if not, keeping the array arm as a read-only decode fallback costs nothing and makes the encoding change one-way rather than a hard cut.

crates/engine/src/fees/fee_module.rs
```diff
@@ -47,6 +47,113 @@ impl FeeModule {
...
+        // Native verification is priced in the same points and charged at the same rate, under its
+        // own source so the breakdown distinguishes crypto verification from template execution.
+        let native_units = state.fee_state().accumulated_native_points() / self.fee_table.wasm_points_cost_divisor();
+        let native_cost = native_units
+            .checked_mul(self.fee_table.per_wasm_point_cost())
+            .ok_or_else(|| RuntimeModuleError::Overflow("Overflow calculating native execution cost".to_string()))?;
+
+        let fee_state = state.fee_state_mut();
+        fee_state.set_charge(FeeSource::Storage, storage_cost);
+        fee_state.set_charge(FeeSource::TemplatePublish, template_publish_cost);
```
SWvheerden, line 141: The old code guarded this with `if template_publish_cost > 0`. Assigning unconditionally is required for the second pass to be able to *clear* a charge the first pass made, so the guard can't simply come back — but the side effect is that every transaction that publishes nothing now carries a `TemplatePublish: 0` entry in its `FeeBreakdown`.

That breakdown is persisted inside the transaction receipt substate and re-streamed to every syncing node, and it shows up in every JSON/TS `FeeBreakdown` a wallet renders. It's a couple of bytes and a spurious zero row on every transaction on the network, forever, for a charge that didn't happen.

A zero-suppressing assignment (`if amount == 0 { self.breakdown.shift_remove(&source) } else { … }` in `FeeBreakdown::set`) keeps the replace-on-second-pass semantics while leaving absent sources absent. Same applies to the `set_charge(FeeSource::ExhaustBurn, 0)` seed just below when the burn rate is zero.

crates/engine/src/runtime/impl.rs
sdbondi: Fixed in `fc77320a`.

`FinalizeResult::total_fees_required` now carries the figure the commit-or-reject decision was gated on, captured in `select_finalized_state` before the second pass re-derives the charges. `FinalizeResult::required_fees()` is built from it, so a resubmission gets a number that clears.

It lives on the result rather than on the receipt. I tried the receipt first, and it cost ~10 µT of storage on *every* transaction on the network — `TransactionReceipt::encoded_size_upper_bound` prices `FeeReceipt::widest()`, so any field added there is charged at its widest form on every transaction forever. `template_load_fee_charged_once_per_template_per_transaction` caught it by tipping from 1000 to 1001 µT. This figure is execution metadata, not settled state, so it stays out of the persisted receipt.

One correction to the report: the dry-run loop doesn't arise. `select_finalized_state` short-circuits on `is_dry_run()`, so an under-funded dry run never reaches the fee-intent path — only a genuinely failing main intent does, and that transaction is broken regardless of the fee. The damage was to real submissions that got `AcceptFeeRejectRest`, which is what the fix addresses.

The receipt itself still reports the fee intent's own charges alongside `is_paid_in_full()`, and that stays: it accurately describes what was collected and refunded. The contradiction was that nothing carried the other number, not that this one was wrong.

crates/engine/src/fees/fee_table.rs
sdbondi: You're right, and this was my error — fixed in `fc77320a`.

I wrote "which a dry run clamps to" after misreading `req.max_fee.max(1)` as a clamp. It's a floor guarding against zero, as the comment directly above it says. Verified the rest of your point too: `dry_run` only sets a `FeeState` flag that skips the paid-in-full gate and disables the payment-funded compute allowance — nothing in the engine, `dry_run/processor.rs`, walletd or the SDK rewrites the fee instructions.

So the residual term is live in both directions, and the shipped flow moves it the way the allowance didn't cover. `fee_estimate_allowance` now folds it in at `per_byte_storage_cost / storage_cost_divisor`, plus its own rounding boundary, with the burn re-multiplying the pair. For the shipped tables that takes the allowance from 4 to 25.

`required_fees_covers_a_real_run_at_any_max_fee` was assuming the premise, exactly as you say. It now asserts that *any* run's estimate covers *every* run's cost, rather than taking `receipts[0]`.

I've also corrected the note that led me here, so I don't repeat it.

crates/engine/src/fees/fee_module.rs
sdbondi: Fixed in `fc77320a`, taking your suggestion. `FeeBreakdown::set` drops the source when the amount is zero rather than recording a row against it, which keeps the replace-on-second-pass semantics that the removed `> 0` guard can't come back for. `get` already reads an absent source as zero, so it's invisible to readers.

While in there: `get_total` used `.sum()`, which panics in debug and *wraps* in release — a total below the charges it's made of. It saturates now.

PR #2421, opened 2026-08-13
fix(engine)!: account for confidential commitments in total_supply

sdbondi:
## Description

`Resource::total_supply` was mis-accounted for `Confidential` resources. Supply tracking is enabled by default for them (`resource/builder/confidential.rs`), so this was the default configuration rather than an opt-in edge case.

Both sides of the accounting went through `ResourceContainer::unlocked_amount()`, which for `Confidential` returns the **revealed amount only** — commitments are invisible to it:

- mint — `working_state.rs` → `increase_total_supply(resource_container.unlocked_amount())`
- burn — `impl.rs`, `BucketAction::Burn` → `decrease_total_supply(bucket.unlocked_amount())`

That would be merely incomplete-but-self-consistent, except value crosses the revealed ⇄ confidential boundary freely inside `ResourceContainer::withdraw_confidential` with no supply call at all. Two consequences:

**Silent over-reporting.** Mint a commitment worth 1000, burn the bucket holding it → the decrement is 0. The value is destroyed and `total_supply` keeps reporting it, permanently.

**Underflow panic.** Mint a commitment worth 500 with 0 revealed (`total_supply += 0`), convert it to revealed via `output_revealed_amount`, burn the resulting bucket → `decrease_total_supply(500)` against a tracked supply of 0. `Amount` is an unsigned `u128` and `decrease_total_supply` called `.expect()` on the `checked_sub`, so this panicked rather than wrapping.

The blast radius of that panic is the whole process: the validator node installs a panic hook that exits on a panic in *any* thread (`applications/tari_validator_node/src/main.rs`, "This makes a panic in any thread 'crash' the system instead of silently continuing"). Since execution is deterministic, one such transaction takes down every node that executes it. Burning does require passing the resource's `Burn` access rule, so it is not open to arbitrary callers on every resource — but `rule!(allow_all)` is a common choice.

## Approach

`total_supply` now means **revealed + confidential**, which is what `Stealth` already does: `UtxoAction::Burn` requires a `StealthValueProof` whenever supply tracking is enabled, verifies it, and decrements by the recovered value. The primitive existed; it just wasn't wired to the confidential paths.

- **Mint** — `MintArg::Confidential` carries a proof per minted commitment, keyed by commitment, required when the resource tracks total supply since the engine cannot otherwise learn the minted value. Every created output is iterated rather than assuming a single one, so the accounting does not depend on `mint_confidential`'s change-rejection invariant holding. Verification is charged against the payment-funded allowance (`PER_VALUE_PROOF`) before the crypto runs, matching the stealth burn.

- **Burn** — `BucketAction::Burn` takes a `BurnBucketArg` carrying a value proof per commitment, and the supply is decreased by the revealed funds plus the proven commitment value. The proofs are verified inside `spend_confidential_outputs`, where the commitments are already enumerated, so a commitment cannot be destroyed without its value being accounted for whichever path reaches it. Revealing a commitment first is still supply-neutral and remains a valid alternative for a holder who knows the masks.

  One case stays impossible by construction: a commitment recalled from a resource with no view key. The recaller has no mask, so nobody but the original owner can prove its value. `total_supply` remains correct — the value still exists, held by whoever holds the commitment — but it cannot be destroyed.

- **Underflow** — `decrease_total_supply` returns `false` instead of `.expect()`ing, mirroring `increase_total_supply`. Both call sites raise the new `RuntimeError::ResourceSupplyWouldUnderflow`, turning a node crash into a transaction rejection. Worth having independently of the accounting change.

`StealthValueProof` is renamed to `CommitmentValueProof` and `validate_value_proof` moved from `engine_types::stealth` to `engine_types::crypto` — it is no longer stealth-specific, and leaving a stealth-named type on the confidential mint path would be misleading. Its error is now a generic `ResourceError::InvalidValueProof` rather than the UTXO-specific `UtxoBurnFailed`. Mechanical, ~20 call sites; happy to split it out if you'd rather review it separately.

## Breaking changes

- `MintArg::Confidential` gains a `value_proofs` field: a `BTreeMap<PedersenCommitmentBytes, CommitmentValueProof>`.
- `BucketAction::Burn` gains a `BurnBucketArg` argument carrying the same map.
- `ResourceManager::mint_confidential` and `ConfidentialResourceBuilder::initial_supply_with_value_proofs` take the map. `initial_supply` remains, and without tracking disabled only accepts a revealed initial supply.
- `Bucket::burn()` keeps its signature but now sends a `BurnBucketArg`; `Bucket::burn_with_value_proofs` burns commitments on a supply-tracking resource.
- `StealthValueProof` → `CommitmentValueProof` (`bindings/src/types/` renamed to match).

These are wire-format changes, not just Rust API changes: an old-encoding `MintArg::Confidential` decodes with an empty `value_proofs` and is rejected at runtime, as is an old `BucketAction::Burn` with no argument. Accepted deliberately — a testnet reset is planned, so no migration path is provided.

## Test plan

- `mint_initial_commitment` asserted `total_supply == 0` after minting a commitment worth 100, carrying a `// TODO: confidential total_supply tracking only tracks revealed funds`. It now asserts 100 and the TODO is gone.
- `minting_and_burning_a_commitment_tracks_total_supply` — mint 500 into a commitment (supply 500), reveal it (supply unchanged — the conversion is neutral), burn the revealed bucket (supply 0). This is the case that used to panic.
- `burning_a_bucket_holding_commitments_without_a_value_proof_is_rejected` and `burning_a_bucket_holding_commitments_with_value_proofs_decreases_total_supply` — cover both sides of the new burn rule.
- `minting_a_commitment_without_a_value_proof_is_rejected` — the mint-side equivalent.
- `generate_reveal_proof` added to the test tooling: the existing `generate_withdraw_proof` helpers always emit an output commitment, even for a zero output amount, so they cannot express "spend this commitment entirely into revealed funds".
- Value-proof helpers moved from `support::stealth` to a new `support::value_proof`.

`cargo test -p tari_engine` passes in full. `cargo check --workspace --all-targets` is clean.

Two notes on the test changes:

- `it_does_not_overflow_when_minting_more_then_amount_max_confidential_tokens` deliberately mints past `Amount::MAX`, which correct accounting now rejects as a supply overflow before the vault can be exercised. Supply tracking is disabled on the `Fungible` test template's confidential resource so the test keeps covering the vault-level overflow it was written for.
- `it_burns_all_resource_types` minted a confidential supply that was both revealed 1000 *and* a commitment worth 1000, then burned the lot — Symptom A in miniature. Its confidential initial supply is now revealed-only, and the commitment-burn path is covered by the new dedicated test.

## Versioning

Not done here. `scripts/crate_versioning.py impact tari_template_lib_types --breaking` asks for a workspace rollup to 0.40.0 plus minor bumps on 14 independent crates. The two most recent breaking commits on `development` (#2414, #2410) don't carry those bumps, so this looks batched at release time — say if you'd like it in this PR instead.

## Review follow-ups

- **Value proofs required structurally, not positionally.** Burn verification lives in `spend_confidential_outputs`, the function that enumerates and downs the commitments, so the only code path able to destroy a commitment is the one that demands a proof for it. The mint iterates every created output. Neither side depends on a caller remembering a check, nor on an invariant enforced elsewhere.
- **`claim_burn` asymmetry** — documented at the mint site. It rests on two guards, not one: the genesis TARI resource disables supply tracking *and* denies `Burn`.
- **Migration** — no path provided; see breaking changes above.
- **ElGamal value proof forgeability** — fixed separately in #2425. This branch needs a rebase once that lands, since it changes `validate_value_proof`'s signature and reorders the stealth burn arm.

### DIFF
--- a/crates/engine/src/runtime/impl.rs
+++ b/crates/engine/src/runtime/impl.rs
@@ -489,7 +489,11 @@ impl<TStore: StateReader + Clone + 'static, TTemplateProvider: TemplateProvider<
         has_view_key: bool,
         is_total_supply_tracking_enabled: bool,
     ) -> Result<(), RuntimeError> {
-        let MintArg::Confidential { statement, .. } = mint_arg else {
+        let MintArg::Confidential {
+            statement,
+            value_proofs,
+        } = mint_arg
+        else {
             return Ok(());
         };
 
@@ -499,9 +503,17 @@ impl<TStore: StateReader + Clone + 'static, TTemplateProvider: TemplateProvider<
                 has_view_key,
             ))?;
 
-        if is_total_supply_tracking_enabled && statement.output.is_some() {
-            self.tracker
-                .charge_native_execution(tari_engine_types::limits::NativeExecutionPoints::PER_VALUE_PROOF)?;
+        if is_total_supply_tracking_enabled {
+            // One proof is verified per minted commitment; the price depends on each proof's variant.
+            let points = statement
+                .output
+                .iter()
+                .filter_map(|output| value_proofs.get(&output.commitment))
+                .map(crypto::value_proof_native_points)
+                .fold(0u64, u64::saturating_add);
+            if points > 0 {
+                self.tracker.charge_native_execution(points)?;
+            }
         }
 
         Ok(())
@@ -2936,25 +2948,27 @@ where
                     self.invoke_resource_access_hook(auth_hook, auth_caller, ResourceAuthAction::Burn)?;
                 }
 
-                // The hook may have altered the bucket, so the commitments are counted after it runs: the charge
-                // must cover the proofs actually verified below, not those held when the hook was scheduled.
-                let num_value_proofs = self.tracker.write_with(|state_mut| {
+                // The hook may have altered the bucket, so the commitments are read after it runs: the charge must
+                // cover the proofs actually verified below, not those held when the hook was scheduled.
+                let value_proof_points = self.tracker.write_with(|state_mut| {
                     let bucket = state_mut.get_bucket(bucket_id)?;
-                    // A value proof is verified per held commitment, and only when supply tracking is enabled.
-                    let num_commitments = if tracks_supply {
-                        bucket.number_of_confidential_commitments()
-                    } else {
-                        0
-                    };
-                    Ok::<_, RuntimeError>(num_commitments)
+                    if !tracks_supply {
+                        return Ok::<_, RuntimeError>(0);
+                    }
+                    // One proof is verified per unlocked commitment; the price depends on each proof's variant.
+                    let points = bucket
+                        .get_confidential_commitments()
+                        .into_iter()
+                        .flatten()
+                        .filter_map(|commitment| arg.value_proofs.get(commitment))
+                        .map(crypto::value_proof_native_points)
+                        .fold(0u64, u64::saturating_add);
+                    Ok(points)
                 })?;
 
                 // Charge the value proofs' native verification against the payment-funded allowance before they run.
-                if num_value_proofs > 0 {
-                    self.tracker.charge_native_execution(
-                        tari_engine_types::limits::NativeExecutionPoints::PER_VALUE_PROOF
-                            .saturating_mul(num_value_proofs as u64),
-                    )?;
+                if value_proof_points > 0 {
+                    self.tracker.charge_native_execution(value_proof_points)?;
                 }
 
                 self.tracker.write_with(|state| {
--- a/crates/engine/src/runtime/impl.rs
+++ b/crates/engine/src/runtime/impl.rs
@@ -2948,10 +2948,12 @@ where
                     self.invoke_resource_access_hook(auth_hook, auth_caller, ResourceAuthAction::Burn)?;
                 }
 
-                // The hook may have altered the bucket, so the commitments are read after it runs: the charge must
-                // cover the proofs actually verified below, not those held when the hook was scheduled.
+                // The hook may have altered the bucket, so it is re-inspected after the hook runs and before
+                // anything is charged: a hook that locked funds makes the burn fail, and the charge must cover the
+                // proofs actually verified below rather than those held when the hook was scheduled.
                 let value_proof_points = self.tracker.write_with(|state_mut| {
                     let bucket = state_mut.get_bucket(bucket_id)?;
+                    Self::check_bucket_is_burnable(bucket_id, bucket)?;
                     if !tracks_supply {
                         return Ok::<_, RuntimeError>(0);
                     }
@@ -2973,9 +2975,8 @@ where
 
                 self.tracker.write_with(|state| {
                     let bucket = state.take_bucket(bucket_id)?;
-                    Self::check_bucket_is_burnable(bucket_id, &bucket)?;
 
-                    state.burn_bucket(bucket, &resource_lock, &arg.value_proofs)?;
+                    state.burn_bucket(bucket_id, bucket, &resource_lock, &arg.value_proofs)?;
 
                     state.unlock_substate(resource_lock)?;
 
--- a/crates/engine/src/runtime/impl.rs
+++ b/crates/engine/src/runtime/impl.rs
@@ -31,6 +31,7 @@ use tari_engine_types::{
     commit_result::{FinalizeResult, RejectReason},
     component::Component,
     confidential::{ClaimBurnOutputData, ClaimedOutputTombstone, MinotariBurnClaimProof},
+    crypto,
     crypto::OutputBody,
     entity_id_provider::EntityIdProvider,
     events::Event,
@@ -68,6 +69,7 @@ use tari_template_lib::{
         BucketGetAmountArg,
         BucketRef,
         BuiltinTemplateAction,
+        BurnBucketArg,
         BurnStealthUtxoArg,
         CallAction,
         CallFunctionArg,
@@ -81,6 +83,7 @@ use tari_template_lib::{
         FreezeResourceArg,
         GenerateRandomAction,
         InvokeResult,
+        MintArg,
         MintResourceArg,
         NonFungibleAction,
         PayFeeArg,
@@ -462,6 +465,33 @@ impl<TStore: StateReader + Clone + 'static, TTemplateProvider: TemplateProvider<
             function: function.to_string(),
             details: e.to_string(),
         })
+    }
+
+    /// Charges the native verification cost of a confidential mint against the payment-funded allowance before any
+    /// of its proof crypto runs. The charged work must match what `WorkingState::mint_resource` goes on to verify:
+    /// the value proof is only checked when the resource tracks supply and the statement mints a commitment.
+    fn charge_confidential_mint(
+        &mut self,
+        mint_arg: &MintArg,
+        has_view_key: bool,
+        is_total_supply_tracking_enabled: bool,
+    ) -> Result<(), RuntimeError> {
+        let MintArg::Confidential { statement, .. } = mint_arg else {
+            return Ok(());
+        };
+
+        self.tracker
+            .charge_native_execution(tari_engine_types::confidential::statement_native_points(
+                statement,
+                has_view_key,
+            ))?;
+
+        if is_total_supply_tracking_enabled && statement.output.is_some() {
+            self.tracker
+                .charge_native_execution(tari_engine_types::limits::NativeExecutionPoints::PER_VALUE_PROOF)?;
+        }
+
+        Ok(())
     }
 
     fn check_resource_auth_hook(&mut self, hook: &AuthHook) -> Result<(), RuntimeError> {
@@ -1393,12 +1423,12 @@ where
 
                 // Charge the initial mint's native verification cost against the payment-funded
                 // allowance before its proof crypto runs.
-                if let Some(tari_template_lib::args::MintArg::Confidential { statement }) = arg.mint_arg.as_ref() {
-                    self.tracker
-                        .charge_native_execution(tari_engine_types::confidential::statement_native_points(
-                            statement,
-                            arg.view_key.is_some(),
-                        ))?;
+                if let Some(mint_arg) = arg.mint_arg.as_ref() {
+                    self.charge_confidential_mint(
+                        mint_arg,
+                        arg.view_key.is_some(),
+                        arg.is_total_supply_tracking_enabled,
+                    )?;
                 }
 
                 self.tracker.write_with(|state_mut| {
@@ -1503,7 +1533,7 @@ where
                         })?;
                 let mint_resource: MintResourceArg = args.assert_one_arg()?;
 
-                let (resource_lock, maybe_auth_hook, auth_caller, has_view_key) =
+                let (resource_lock, maybe_auth_hook, auth_caller, has_view_key, tracks_supply) =
                     self.tracker.write_with(|state_mut| {
                         let resource_lock = state_mut.write_lock_substate(SubstateId::Resource(resource_address))?;
 
@@ -1517,7 +1547,14 @@ where
 
                         let auth_caller = state_mut.get_auth_caller()?;
                         let has_view_key = resource.view_key().is_some();
-                        Ok::<_, RuntimeError>((resource_lock, resource.auth_hook().cloned(), auth_caller, has_view_key))
+                        let tracks_supply = resource.is_supply_tracking_enabled();
+                        Ok::<_, RuntimeError>((
+                            resource_lock,
+                            resource.auth_hook().cloned(),
+                            auth_caller,
+                            has_view_key,
+                            tracks_supply,
+                        ))
                     })?;
 
                 if let Some(auth_hook) = maybe_auth_hook {
@@ -1526,13 +1563,7 @@ where
 
                 // Charge the mint's native verification cost against the payment-funded allowance
                 // before its proof crypto runs.
-                if let tari_template_lib::args::MintArg::Confidential { statement } = &mint_resource.mint_arg {
-                    self.tracker
-                        .charge_native_execution(tari_engine_types::confidential::statement_native_points(
-                            statement,
-                            has_view_key,
-                        ))?;
-                }
+                self.charge_confidential_mint(&mint_resource.mint_arg, has_view_key, tracks_supply)?;
 
                 self.tracker.write_with(|state_mut| {
                     let mint_arg = mint_resource.mint_arg;
@@ -2084,7 +2115,7 @@ where
                             .output
                             .as_ref()
                             .and_then(|o| o.output.viewable_balance.as_ref());
-                        let value = stealth::validate_value_proof(
+                        let value = crypto::validate_value_proof(
                             &commitment,
                             maybe_view_key.as_ref(),
                             elgamal_proof,
@@ -2094,8 +2125,7 @@ where
                         if value.is_positive() {
                             let resource_lock =
                                 state_mut.write_lock_substate(SubstateId::Resource(resource_address))?;
-                            let resource_mut = state_mut.get_resource_mut(&resource_lock)?;
-                            resource_mut.decrease_total_supply(value);
+                            state_mut.decrease_total_supply(&resource_lock, value)?;
                             state_mut.unlock_substate(resource_lock)?;
                         }
                     } else {
@@ -2860,26 +2890,50 @@ where
                     reason: "Burn bucket action requires a bucket id".to_string(),
                 })?;
 
-                let (resource_lock, maybe_auth_hook, auth_caller) = self.tracker.write_with(|state_mut| {
-                    let bucket = state_mut.get_bucket(bucket_id)?;
+                let arg: BurnBucketArg = args.assert_one_arg()?;
 
-                    let resource_lock =
-                        state_mut.write_lock_substate(SubstateId::Resource(*bucket.resource_address()))?;
+                let (resource_lock, maybe_auth_hook, auth_caller, num_value_proofs) =
+                    self.tracker.write_with(|state_mut| {
+                        let bucket = state_mut.get_bucket(bucket_id)?;
+                        let num_commitments = bucket.number_of_confidential_commitments();
 
-                    let resource = state_mut.get_resource(&resource_lock)?;
+                        let resource_lock =
+                            state_mut.write_lock_substate(SubstateId::Resource(*bucket.resource_address()))?;
 
-                    state_mut.authorization().check_resource_access_rules(
-                        ResourceAuthAction::Burn,
-                        resource.as_ownership(),
-                        resource.access_rules(),
-                    )?;
+                        let resource = state_mut.get_resource(&resource_lock)?;
 
-                    let auth_caller = state_mut.get_auth_caller()?;
-                    Ok::<_, RuntimeError>((resource_lock, resource.auth_hook().cloned(), auth_caller))
-                })?;
+                        state_mut.authorization().check_resource_access_rules(
+                            ResourceAuthAction::Burn,
+                            resource.as_ownership(),
+                            resource.access_rules(),
+                        )?;
+
+                        // A value proof is verified per held commitment, and only when supply tracking is enabled.
+                        let num_value_proofs = if resource.is_supply_tracking_enabled() {
+                            num_commitments
+                        } else {
+                            0
+                        };
+
+                        let auth_caller = state_mut.get_auth_caller()?;
+                        Ok::<_, RuntimeError>((
+                            resource_lock,
+                            resource.auth_hook().cloned(),
+                            auth_caller,
+                            num_value_proofs,
+                        ))
+                    })?;
 
                 if let Some(auth_hook) = maybe_auth_hook {
                     self.invoke_resource_access_hook(auth_hook, auth_caller, ResourceAuthAction::Burn)?;
+                }
+
+                // Charge the value proofs' native verification against the payment-funded allowance before they run.
+                if num_value_proofs > 0 {
+                    self.tracker.charge_native_execution(
+                        tari_engine_types::limits::NativeExecutionPoints::PER_VALUE_PROOF
+                            .saturating_mul(num_value_proofs as u64),
+                    )?;
                 }
 
                 self.tracker.write_with(|state| {
@@ -2892,11 +2946,8 @@ where
                             locked_amount: bucket.locked_amount(),
                         });
                     }
-                    let burnt_amount = bucket.unlocked_amount();
-                    state.burn_bucket(bucket)?;
 
-                    let resource_mut = state.get_resource_mut(&resource_lock)?;
-                    resource_mut.decrease_total_supply(burnt_amount);
+                    state.burn_bucket(bucket, &resource_lock, &arg.value_proofs)?;
 
                     state.unlock_substate(resource_lock)?;
 
@@ -3319,7 +3370,11 @@ where
             let address = ClaimedOutputTombstoneAddress::from_commitment(claim.commitment);
             state_mut.new_substate(address, ClaimedOutputTombstone { value: claim.value })?;
 
-            // 3. Create the stealth UTXO
+            // 3. Create the stealth UTXO.
+            // This mints value with no `increase_total_supply` counterpart, which is sound only because the genesis
+            // TARI resource both disables supply tracking and denies `Burn` (see `get_stealth_tari_resource`).
+            // Enabling either would need an increase here to balance the decrease that
+            // `ResourceAction::StealthUtxoBurn` applies when the UTXO is burnt.
             let address = UtxoAddress::new(TARI_TOKEN, claim.commitment.into());
             let utxo = Utxo::new(UtxoOutput {
                 output: OutputBody {
--- a/crates/engine/src/runtime/impl.rs
+++ b/crates/engine/src/runtime/impl.rs
@@ -28,6 +28,7 @@ use tari_crypto::{ristretto::RistrettoPublicKey, tari_utilities::ByteArray};
 use tari_engine_types::{
     Utxo,
     UtxoOutput,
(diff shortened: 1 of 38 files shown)


crates/engine_types/src/crypto/value_proof.rs
```diff
@@ -8,95 +8,69 @@ use tari_crypto::{
...
 
             if !sig.verify(&public_mask, message) {
-                return Err(ResourceError::UtxoBurnFailed {
-                    id: UtxoId::from(*commitment_bytes),
-                    details: "Invalid mask knowledge proof".to_string(),
-                });
+                return Err(invalid("Invalid mask knowledge proof".to_string()));
             }
         },
         ValueKnowledgeProof::ElgamalEncrypted { reveal_key } => {
```
sdbondi: **ElGamal value proof is forgeable, now load-bearing for total_supply**

The `ElgamalEncrypted` variant only checks `encrypted - reveal_key == value·G`; `public_nonce` is never used and nothing binds `reveal_key` to the ElGamal randomness or view key. For any commitment carrying a viewable balance `E`, a prover can set `reveal_key = E - v'·G` for an arbitrary claimed `v'`. On the new mint path (`working_state.rs` passes `body.viewable_balance` for view-key resources) a minter can mint a commitment worth X but prove v' ≠ X, so `total_supply` moves by v' while X circulates — hidden inflation, or spurious `ResourceSupplyWouldUnderflow` on later honest burns. This was pre-existing on the stealth-burn path, but this PR extends the trust to mint and makes it the basis of the invariant it introduces.

crates/engine/src/runtime/impl.rs
```diff
@@ -2873,11 +2896,35 @@ where
...
                     }
+
+                    let resource_address = *bucket.resource_address();
+                    let is_total_supply_tracking_enabled =
+                        state.get_resource(&resource_lock)?.is_supply_tracking_enabled();
+
+                    // `unlocked_amount` only sees revealed funds, so a commitment's value would be destroyed
+                    // without being subtracted from the supply. Revealing a commitment is balance-proven and
+                    // therefore supply-neutral, so the holder can always reveal first and then burn.
+                    let num_commitments = bucket.number_of_confidential_commitments();
```
sdbondi: **Recalled confidential commitments become permanently unburnable**

Bucket burn blanket-rejects any bucket holding confidential commitments on a supply-tracking resource. For a resource built `.recallable(...).burnable(...)` with default supply tracking: an admin recalls a user's confidential outputs into a bucket, then `bucket.burn()` always fails with `CannotBurnConfidentialCommitments`. The comment's 'reveal first' escape requires the commitment masks, which the recalling admin does not hold — so confiscated value can never be destroyed and stays in `total_supply` forever. The sibling stealth-UTXO burn path solved the same problem by accepting an optional `CommitmentValueProof`; consider the same here.

crates/engine/src/runtime/impl.rs
```diff
@@ -2873,11 +2896,35 @@ where
...
+                        return Err(RuntimeError::CannotBurnConfidentialCommitments {
+                            resource_address,
+                            num_commitments,
+                        });
+                    }
+
                     let burnt_amount = bucket.unlocked_amount();
                     state.burn_bucket(bucket)?;
 
                     let resource_mut = state.get_resource_mut(&resource_lock)?;
```
sdbondi: **Unconditional get_resource_mut persists an unchanged resource substate**

When supply tracking is disabled, `decrease_total_supply` is a no-op but `get_resource_mut` still moves the resource into the mutated set, so an identical resource is written into the diff with a version bump — extra bytes hashed, gossiped and persisted per burn. `is_total_supply_tracking_enabled` is already computed two lines up; the mint path gates its `get_resource_mut` on that flag explicitly 'to prevent needless mutation', and the stealth-burn arm gates on `value.is_positive()`. This arm should gate the same way.

crates/engine/src/runtime/impl.rs
```diff
@@ -1393,12 +1394,19 @@ where
...
-                if let Some(tari_template_lib::args::MintArg::Confidential { statement }) = arg.mint_arg.as_ref() {
+                if let Some(tari_template_lib::args::MintArg::Confidential { statement, value_proof }) =
+                    arg.mint_arg.as_ref()
+                {
                     self.tracker
                         .charge_native_execution(tari_engine_types::confidential::statement_native_points(
                             statement,
                             arg.view_key.is_some(),
                         ))?;
+                    if value_proof.is_some() {
```
sdbondi: **PER_VALUE_PROOF charged even when the proof is never validated**

Both mint arms charge `PER_VALUE_PROOF` as soon as a proof is attached, but `working_state::mint_resource` only runs `validate_value_proof` when supply tracking is enabled AND a commitment was created. A mint on a tracking-disabled resource, or a revealed-only mint, with a proof attached pays a full Schnorr verification's worth of points for crypto that never runs. Gate the charge on the same predicate the validation uses, or move it next to `validate_value_proof` so charge and work cannot diverge.

crates/engine_types/src/crypto/value_proof.rs
```diff
@@ -8,95 +8,69 @@ use tari_crypto::{
...
+                .encrypted
+                .try_from_byte_type()
+                .map_err(|e| invalid(format!("Invalid encrypted balance bytes: {e}")))?;
+
+            let reveal_key: RistrettoPublicKey = reveal_key
+                .try_from_byte_type()
+                .map_err(|e| invalid(format!("Invalid reveal key bytes: {e}")))?;
 
             // E - R.p = v.G
             let check_value = encrypted - reveal_key;
```
sdbondi, line 115: **ElGamal path accepts full u128 while Commitment path bounds at u64::MAX**

`ValueKnowledgeProof::Commitment` errors for value > `u64::MAX` (via `commit_amount` returning `None`), but the `ElgamalEncrypted` arm runs `convert_amount_to_secret` over the whole `u128` with no bound. Combined with the unauthenticated ElGamal check, a mint/burn can assert a supply delta of up to 2^128−1 — e.g. pushing `total_supply` near `u128::MAX` so all subsequent honest mints fail with `ResourceSupplyWouldOverflow`. Bound the ElGamal value to `u64::MAX` like the commitment path.

crates/engine/src/runtime/working_state.rs
```diff
@@ -801,15 +816,19 @@ impl<TStore: StateReader> WorkingState<TStore> {
...
-            // Increase the total supply of the resource
-            if !resource_mut.increase_total_supply(resource_container.unlocked_amount()) {
-                return Err(RuntimeError::ResourceSupplyWouldOverflow {
-                    resource_address,
-                    current_supply: resource_mut
-                        .total_supply()
-                        .expect("Resource supply tracking is enabled"),
-                    amount: resource_container.unlocked_amount(),
-                });
+            let current_supply = resource_mut
```
sdbondi, line 953: **Two-step supply increase is non-atomic and reports stale current_supply**

`current_supply` is captured before `for amount in [unlocked_amount(), minted_commitment_value]`. If the first increase succeeds and the second overflows, the `ResourceSupplyWouldOverflow` error reports the pre-loop supply (understated by the revealed amount) and the first increase is left applied in memory. State is discarded on error so the effect is diagnostic-only, but summing the two with `checked_add` and doing a single `increase_total_supply` is simpler, atomic, and reports the true total.

crates/engine/src/runtime/impl.rs
```diff
@@ -2873,11 +2896,35 @@ where
...
+                            num_commitments,
+                        });
+                    }
+
                     let burnt_amount = bucket.unlocked_amount();
                     state.burn_bucket(bucket)?;
 
                     let resource_mut = state.get_resource_mut(&resource_lock)?;
-                    resource_mut.decrease_total_supply(burnt_amount);
+                    if !resource_mut.decrease_total_supply(burnt_amount) {
```
sdbondi: **Duplicated decrease-or-underflow block; check belongs on Resource**

The `decrease_total_supply`-or-`ResourceSupplyWouldUnderflow` block (including the subtle `.expect("Resource supply tracking is enabled")` re-read) is copy-pasted at the UTXO-burn site (~impl.rs:2096) and here, and mirrors the overflow twin in `working_state.rs:825`. A `#[must_use]` bool API forces every call site to rebuild the payload, and a future burn path can discharge the must_use with a log instead of a rejection. A checked method on `Resource` returning `Result<(), ResourceError>` carrying the supply values would centralize the error shape and the expect's safety argument.

crates/engine/src/runtime/impl.rs
```diff
@@ -1393,12 +1394,19 @@ where
 
                 // Charge the initial mint's native verification cost against the payment-funded
                 // allowance before its proof crypto runs.
-                if let Some(tari_template_lib::args::MintArg::Confidential { statement }) = arg.mint_arg.as_ref() {
+                if let Some(tari_template_lib::args::MintArg::Confidential { statement, value_proof }) =
```
sdbondi: **Confidential-mint charge block duplicated at both mint sites**

This ~12-line `if let MintArg::Confidential { statement, value_proof } … charge statement_native_points … if value_proof.is_some() charge PER_VALUE_PROOF` block appears both here (CreateResource) and in the ResourceInvoke mint arm (~impl.rs:1537), differing only in how `has_view_key` is obtained. A future change to mint charging (e.g. pricing by proof variant) must be applied in both places or the two paths silently charge differently. Consider extracting `fn charge_confidential_mint_points(mint_arg, has_view_key)`.

sdbondi (commented):
Automated review (Claude Code, verified findings).

**Migration hazard (no single line to anchor):** `MintArg::Confidential`'s new `#[n(1)]` value-proof field decodes old encodings as `None`, and that `None` then triggers `ConfidentialMintRequiresValueProof` — so any already-published template minting a confidential commitment on a supply-tracking resource breaks at runtime with no recompile path.

**Findings on lines outside the diff:**
- `crates/engine/src/runtime/impl.rs:3348` — **claim_burn mint/burn supply asymmetry rests silently on tracking being disabled**: `StealthUtxoBurn` now decreases `total_supply`, but the mint side (`claim_burn` creating a TARI UTXO) adds no matching increase. This is safe only because the TARI genesis resource has supply tracking disabled. If TARI supply tracking were ever enabled, every claimed burn would inflate untracked value and every subsequent UTXO burn of claimed value would drive tracked supply toward `ResourceSupplyWouldUnderflow`. Nothing near `claim_burn` checks or documents the tracking-disabled assumption the correctness now rests on.

- `crates/engine/src/runtime/working_state.rs:631` — **Burn-commitment guard is enforced at one call site, not in burn_bucket**: The cannot-burn-commitments guard and supply-decrease live in the engine's `BucketAction::Burn` arm, while `WorkingState::burn_bucket` itself destroys commitment value with no check of its own. It has one caller today, but a new engine action, template op, or cleanup path calling it would silently bypass the exact leak this PR fixes. Likewise the mint accounting hangs on `created_outputs.first()` trusting `mint_confidential`'s change-rejection invariant enforced far away. Placing the invariant in `burn_bucket`/`ResourceContainer` (and iterating all created outputs on mint) makes it structural rather than positional.

crates/engine/src/runtime/impl.rs
```diff
@@ -2860,28 +2890,52 @@ where
...
-                let (resource_lock, maybe_auth_hook, auth_caller) = self.tracker.write_with(|state_mut| {
-                    let bucket = state_mut.get_bucket(bucket_id)?;
+                let arg: BurnBucketArg = args.assert_one_arg()?;
 
-                    let resource_lock =
-                        state_mut.write_lock_substate(SubstateId::Resource(*bucket.resource_address()))?;
+                let (resource_lock, maybe_auth_hook, auth_caller, num_value_proofs) =
+                    self.tracker.write_with(|state_mut| {
+                        let bucket = state_mut.get_bucket(bucket_id)?;
+                        let num_commitments = bucket.number_of_confidential_commitments();
```
sdbondi: **TOCTOU between value-proof charge and verification**

`num_value_proofs` is counted before the resource Burn auth hook runs, but hook template code can join extra commitments into the bucket (`BucketAction::Join` has no frame-ownership check on bucket ids, which are sequential/guessable). The engine charges N×`PER_VALUE_PROOF` but then verifies N+K proofs in `spend_confidential_outputs_with_value_proofs`, so the pre-charge no longer bounds the native crypto actually run. Amplification is bounded (minting each commitment costs ~10× more than the free verification), but recomputing the count (or charging) after the hook, inside the second `write_with`, closes the window.

crates/engine/src/runtime/impl.rs
```diff
@@ -2860,28 +2890,52 @@ where
...
+                    })?;
 
                 if let Some(auth_hook) = maybe_auth_hook {
                     self.invoke_resource_access_hook(auth_hook, auth_caller, ResourceAuthAction::Burn)?;
                 }
 
+                // Charge the value proofs' native verification against the payment-funded allowance before they run.
+                if num_value_proofs > 0 {
+                    self.tracker.charge_native_execution(
+                        tari_engine_types::limits::NativeExecutionPoints::PER_VALUE_PROOF
```
sdbondi: **ElGamal value proofs undercharged (~2x a Schnorr verification)**

`PER_VALUE_PROOF` is documented in `limits.rs` as one Schnorr verification and the calibration harness doesn't measure it, but `ValueKnowledgeProof::ElgamalEncrypted` verifies two Schnorr-style equations plus 4 point decompressions — roughly double the work. This PR multiplies the charge per commitment in bucket burns and adds it to mints, so the undercharge scales linearly with N. Price the ElGamal variant separately or raise `PER_VALUE_PROOF` to cover the dearer variant.

crates/engine/src/runtime/impl.rs
```diff
@@ -2860,28 +2890,52 @@ where
...
+                            .saturating_mul(num_value_proofs as u64),
+                    )?;
+                }
+
                 self.tracker.write_with(|state| {
                     let bucket = state.take_bucket(bucket_id)?;
                     // It is invalid to burn a bucket that has locked funds (e.g. via a proof). Burning downs only
```
sdbondi: **Guaranteed-failing burns still pay the hook + proof charges**

The `has_locked_funds()` rejection runs in the second `write_with`, after the auth hook is invoked and `PER_VALUE_PROOF` × num_commitments is charged — and `number_of_confidential_commitments()` also counts locked commitments whose proofs can never be verified. Moving the cheap `has_locked_funds` check into the first `write_with` (where `get_bucket` already ran) avoids both, and would charge only for unlocked commitments.

crates/engine/src/runtime/working_state.rs
```diff
@@ -598,8 +626,31 @@ impl<TStore: StateReader> WorkingState<TStore> {
...
+                })?;
+            let value = crypto::validate_value_proof(
+                &commitment,
+                view_key,
+                output.output().viewable_balance.as_ref(),
+                value_proof,
+            )?;
+            proven_value = proven_value
+                .checked_add(value)
+                .ok_or(RuntimeError::ResourceSupplyWouldOverflow {
```
sdbondi: **ResourceSupplyWouldOverflow reports accumulators as current_supply**

This error variant is reused for running-sum overflows with fields that don't mean what they say: `current_supply` is set to the loop accumulator (`proven_value` here, `burnt_amount` at ~771, `minted_commitment_value` at ~928), not the resource's supply. An operator debugging a rejection will compare against the actual `total_supply`, see a different number, and misdiagnose state corruption. A dedicated variant (e.g. `ValueProofTotalOverflow`) or summing against the real supply keeps the message truthful.

crates/engine/src/runtime/working_state.rs
```diff
@@ -627,11 +678,64 @@ impl<TStore: StateReader> WorkingState<TStore> {
...
+        &mut self,
+        bucket: Bucket,
+        resource_lock: &LockedSubstate,
+        value_proofs: &BTreeMap<PedersenCommitmentBytes, CommitmentValueProof>,
+    ) -> Result<(), RuntimeError> {
         if bucket.is_empty() {
             return Ok(());
         }
         let resource_address = *bucket.resource_address();
+        let resource = self.get_resource(resource_lock)?;
```
sdbondi: **Duplicate resource read + unconditional decompression per burn**

`burn_bucket` re-reads the resource and re-derives `is_supply_tracking_enabled` although the Burn handler's first `write_with` already read the same resource to compute `num_value_proofs`, and `to_view_key_public_key()` (a Ristretto point decompression) runs unconditionally even when the bucket holds no commitments or tracking is off. Passing the flags (or moving the charge computation into `burn_bucket`) inspects the resource once.

crates/engine/tests/recall.rs
```diff
@@ -21,10 +25,15 @@ fn it_recalls_all_resource_types() {
 
     let (mut initial_supply, mask, _) = generate_confidential_output_statement(1000, None);
     initial_supply.output_revealed_amount = Amount::from(1000u64);
+    let commitment = commit_amount(&mask, Amount::from(1000u64)).unwrap().to_byte_type();
```
sdbondi: **Commitment+proof-map test idiom triplicated**

`recall.rs:28-32` and `resource.rs:36-38` inline the same `commit_amount(...).to_byte_type()` + `BTreeMap::from([(commitment, generate_value_proof_mask_knowledge(...))])` idiom that `confidential.rs` wraps as its local `value_proofs_for` helper. A shared helper belongs in `template_test_tooling::support::value_proof` next to `generate_value_proof_mask_knowledge`; the next commitment-burning test will otherwise copy a fourth version.

crates/engine/src/runtime/working_state.rs
```diff
@@ -587,6 +588,33 @@ impl<TStore: StateReader> WorkingState<TStore> {
...
+        view_key: Option<&RistrettoPublicKey>,
+    ) -> Result<Amount, RuntimeError> {
+        self.spend_confidential_outputs_internal(resource_address, commitments, Some((value_proofs, view_key)))
+    }
+
+    fn spend_confidential_outputs_internal<I: IntoIterator<Item = PedersenCommitmentBytes>>(
+        &mut self,
+        resource_address: ResourceAddress,
+        commitments: I,
+        value_proofs: Option<(
```
sdbondi: **Nested Option-tuple mode flag on spend_confidential_outputs_internal**

`value_proofs: Option<(&BTreeMap<...>, Option<&RistrettoPublicKey>)>` is a doubly-optional mode flag decoded via let-else/continue inside the loop — readers must work out what `Some((map, None))` vs `None` means. A plain downing loop yielding the downed outputs, with the proving wrapper folding over it, removes the mode flag and keeps a third mode from being bolted on.

sdbondi (commented):
Automated re-review of the rebased branch (Claude Code, verified findings). The previously reported ElGamal soundness issues are resolved by the #2425 fix. Migration/compatibility findings are omitted per the planned testnet reset.

**Findings on lines outside the diff:**
- `docs/developer-docs/src/content/docs/guides/stealth-resources.mdx:508` — **Docs reference removed StealthValueProof; new proof requirement undocumented**: Lines 508 and 523 still reference `StealthValueProof`, which is now `CommitmentValueProof` (the rename left no alias), and the guide doesn't mention the new value-proof requirement for confidential mint / bucket burn on supply-tracking resources — a template author following it hits a compile error and then an unexplained `MissingValueProofForCommitment` rejection.

crates/engine/src/runtime/working_state.rs
```diff
@@ -801,14 +949,19 @@ impl<TStore: StateReader> WorkingState<TStore> {
...
-            // Increase the total supply of the resource
-            if !resource_mut.increase_total_supply(resource_container.unlocked_amount()) {
+            let current_supply = resource_mut
+                .total_supply()
+                .expect("Resource supply tracking is enabled");
+            // The minted amount is the revealed funds plus the value of any minted commitments
+            let minted_amount = resource_container
+                .unlocked_amount()
+                .checked_add(minted_commitment_value)
+                .ok_or(RuntimeError::CommitmentValueSumOverflow { resource_address })?;
```
sdbondi: **Minor:** this sum is revealed funds + minted commitment value, but `CommitmentValueSumOverflow`'s message says the *confidential commitments'* proven values overflowed — misdescribing the case where the revealed portion tips it over. A message covering both ("minted value would exceed the maximum amount") or keeping `ResourceSupplyWouldOverflow` here (current_supply is accurate at this site) would be truthful.

sdbondi (commented):
Re-review of e50d8b4: the post-hook recount closes the charge/verify TOCTOU, the early + re-checked `check_bucket_is_burnable` fixes the failing-burns-still-pay ordering, and `CommitmentValueSumOverflow` fixes the misleading error payload. One minor nit inline. Still open from the previous review: ElGamal PER_VALUE_PROOF undercharge, duplicate resource read/decompression in `burn_bucket`, stale `StealthValueProof` docs, test-idiom reuse, and the nested-Option mode flag.

crates/engine/src/runtime/impl.rs
sdbondi: Fixed in e50d8b487. The commitments are now read in a `write_with` that runs after the auth hook, immediately before charging, so the charge is derived from the bucket the verification actually sees.

crates/engine/src/runtime/impl.rs
sdbondi: Fixed in e50d8b487. `has_locked_funds` is now checked in the first `write_with`, before the hook or any charge. The later check is kept because the hook itself can lock funds. Charging is also now derived from the unlocked commitments only (104d59478), so locked commitments are never paid for.

crates/engine/src/runtime/working_state.rs
sdbondi: Fixed in e50d8b487. Running-sum overflows now raise a dedicated `CommitmentValueSumOverflow { resource_address }`. `ResourceSupplyWouldOverflow` is left for the one site where the supply itself would overflow, where `current_supply` is read from the resource. That site also no longer reports `Amount::MAX` as the amount when the sum overflowed — the two conditions are now distinct.

crates/engine/src/runtime/impl.rs
sdbondi: Fixed in 104d59478. Added `PER_ELGAMAL_VALUE_PROOF` at double `PER_VALUE_PROOF`, and both mint and burn now price each proof by its variant through `crypto::value_proof_native_points`, which is documented as needing to stay in lock-step with `validate_value_proof`.

crates/engine/src/runtime/working_state.rs
sdbondi: Partly fixed in 104d59478. The view-key decompression is now done lazily — only once a proof is actually going to be verified against it — so burns with no commitments, or on resources without supply tracking, no longer pay for it. I left the second `get_resource`: it is a lookup in the working state's already-locked substates rather than IO, and passing the flags in would make `burn_bucket` depend on the caller having read the same resource, which is the kind of positional coupling an earlier review comment asked me to remove.

crates/engine/src/runtime/working_state.rs
sdbondi: Fixed in 104d59478. `spend_confidential_outputs` now returns the outputs it downs and the proving wrapper folds over them, so the mode flag is gone.

crates/engine/tests/recall.rs
sdbondi: Fixed in 104d59478. Now `support::value_proof::value_proofs_for_commitment`, used by all three tests.

sdbondi (commented):
Re-review of 104d5947: this resolves the ElGamal undercharge (`PER_ELGAMAL_VALUE_PROOF` at 2x with `value_proof_native_points` selecting per variant), the duplicate resource read + unconditional view-key decompression in `burn_bucket`, the triplicated test idiom (now `value_proofs_for_commitment` in `template_test_tooling`), and the nested-Option mode flag (`spend_confidential_outputs` now returns the downed outputs and the proving wrapper folds over them). One minor ordering note inline. Still open: the developer docs (`docs/developer-docs/src/content/docs/guides/stealth-resources.mdx:508,523`) reference the removed `StealthValueProof` and don't cover the new value-proof requirement.

crates/engine/src/runtime/impl.rs
```diff
@@ -2860,43 +2915,67 @@ where
...
+                        .flatten()
+                        .filter_map(|commitment| arg.value_proofs.get(commitment))
+                        .map(crypto::value_proof_native_points)
+                        .fold(0u64, u64::saturating_add);
+                    Ok(points)
+                })?;
+
+                // Charge the value proofs' native verification against the payment-funded allowance before they run.
+                if value_proof_points > 0 {
+                    self.tracker.charge_native_execution(value_proof_points)?;
```
sdbondi, line 2973: **Minor:** the charge lands before the post-hook `check_bucket_is_burnable` re-check in the final `write_with`, so a hook that locks funds in the bucket causes value-proof charges for verifications that never run. The transaction is rejected either way and the amount is bounded, so this is at most a diagnostic/fee-fairness nit — folding the burnable re-check into this counting `write_with` would close it.

crates/engine/src/runtime/working_state.rs
sdbondi: Fixed in 44b70ff62. You're right that two of the four sites fold revealed funds into the same sum. Renamed to `ValueSumOverflow` with a message that holds at every site — "the amounts being minted or destroyed ... sum to more than the maximum amount" — rather than keeping `ResourceSupplyWouldOverflow` here, since its `amount` field has no honest value when the sum itself is what overflowed.

crates/engine/src/runtime/impl.rs
sdbondi: Fixed in 443d19009. The burnable re-check is folded into the counting `write_with`, so it runs after the hook and before any charge. I also moved the authoritative check into `burn_bucket` itself — it is the function that destroys the value, so the invariant now holds for any caller rather than only those that check first; the call-site checks remain purely as fail-fast.

PR #2424, opened 2026-08-13
feat(mempool): validate blob references at ingress

sdbondi:
## Description

A transaction's blob list was never validated against what its instructions actually reference.

`Transaction::validate_blob_references` already implements both halves of the check — every `BlobIndex` in bounds, and every blob referenced by some instruction — but its only callers are the wallet daemon handlers, i.e. client-side before signing. Nothing enforced it on transactions arriving from the network.

This adds `BlobReferenceValidator`, which runs that check at ingress:

- `create_mempool_transaction_validator` (validator node) — also covers consensus, since `TariBlockTransactionValidator` wraps the mempool chain
- `create_structural_transaction_validator` — the context-free chain the indexer runs before forwarding to committees

It sits after `BasicValidations` and before the weight and signature validators, so a malformed blob list is rejected before signature verification hashes every blob payload.

Failures surface as `TransactionValidationError::InvalidBlobReferences`, classified `is_sender_fault() == true`: both failure modes are deterministic properties of the transaction itself, so every honest node reaches the same verdict and the sending peer is answerable for it.

## Motivation and Context

Two gaps, neither of them caught before execution:

- **Out-of-bounds index** — only failed at execution time (`TransactionErrorKind::BlobIndexOutOfBounds`), after the mempool, gossip and consensus work had already been spent on a transaction that can never succeed.
- **Unreferenced blob** — never caught at all. The bytes are gossiped, stored and retained by every node while contributing nothing to execution, so a transaction could carry arbitrary padding up to the weight cap. Weight pricing bounds the size but nothing rejects the payload.

Blob payload *integrity* was already covered — the signing domain binds `blobs.hashes()`, so tampered bytes fail signature verification. This PR closes the structural side.

Instruction coverage is complete against the current instruction set: only `PublishTemplate.binary` and `InstructionArg::Blob` carry blob indices, and both are collected by `Instruction::referenced_blob_ids`.

## How Has This Been Tested?

New unit tests in `blob_references.rs`: no blobs, all blobs referenced, out-of-bounds index, unreferenced blob. Added the new variant to the existing `is_sender_fault` exhaustiveness test.

`cargo test -p tari_ootle_transaction_validation` passes (27 tests); `cargo check -p tari_validator_node -p tari_indexer` is clean.

Read `create_mempool_transaction_validator` and `create_structural_transaction_validator` to confirm the validator is in both chains and ordered before signature verification. The unit tests cover both rejection paths directly.

## Breaking Changes

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

### DIFF
--- a/applications/tari_validator_node/src/bootstrap.rs
+++ b/applications/tari_validator_node/src/bootstrap.rs
@@ -79,6 +79,7 @@ use tari_ootle_template_provider::MemoryCacheTemplateProvider;
 use tari_ootle_transaction::{Network, Transaction};
 use tari_ootle_transaction_validation::{
     BasicValidations,
+    BlobReferenceValidator,
     EpochRangeValidator,
     PublishTemplateLimitValidator,
     SignatureLimitValidator,
@@ -522,6 +523,9 @@ pub fn create_mempool_transaction_validator<TProvider: TemplateProvider>(
     TransactionNetworkValidator::new(network)
         .and_then(TransactionDryRunValidator)
         .and_then(BasicValidations::new())
+        // Blob payloads must be exactly what the instructions reference: bad indices would only
+        // fail at execution, and unreferenced blobs would never fail at all.
+        .and_then(BlobReferenceValidator::new())
         // Cheap structural check — reject over-weight transactions before verifying signatures.
         .and_then(TransactionWeightValidator::new(max_transaction_weight))
         // Reject transactions whose aggregate stealth-transfer work exceeds the per-transaction caps before
--- a/crates/transaction_validation/src/blob_references.rs
+++ b/crates/transaction_validation/src/blob_references.rs
@@ -1,0 +1,141 @@
+// Copyright 2026 The Tari Project
+// SPDX-License-Identifier: BSD-3-Clause
+
+use log::warn;
+use tari_ootle_transaction::Transaction;
+
+use crate::{TransactionValidationError, Validator};
+
+const LOG_TARGET: &str = "tari::ootle::mempool::validators::blob_references";
+
+/// Rejects transactions whose blob list does not correspond exactly to what the instructions
+/// reference: every `BlobIndex` must be in bounds, and every blob must be referenced.
+///
+/// Out-of-bounds indices are otherwise only caught during execution, wasting the mempool, gossip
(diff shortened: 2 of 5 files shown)


applications/tari_validator_node/src/bootstrap.rs
```diff
@@ -522,6 +523,9 @@ pub fn create_mempool_transaction_validator<TProvider: TemplateProvider>(
     TransactionNetworkValidator::new(network)
         .and_then(TransactionDryRunValidator)
         .and_then(BasicValidations::new())
+        // Blob payloads must be exactly what the instructions reference: bad indices would only
+        // fail at execution, and unreferenced blobs would never fail at all.
```
sdbondi, line 527: Comment convention: CLAUDE.md asks for present-tense statements of intent/invariant and explicitly calls out the "without this guard X would happen" form as one to avoid. The second half here ("bad indices would only fail at execution, and unreferenced blobs would never fail at all") describes what happens in the validator's absence. The first clause already carries the intent and stands on its own, and the fuller rationale is in `BlobReferenceValidator`'s doc comment.

```suggestion
        // Blob payloads must be exactly what the instructions reference.
```

sdbondi (commented):
Reviewed the wiring and the underlying `validate_blob_references()` in the transaction crate. Coverage looks right: `referenced_blob_ids()` matches all `Instruction` variants exhaustively (so a new variant is a compile error), covers `PublishTemplate.binary`, `CallFunction`/`CallMethod` args and `UpdateComponentTemplate.migrate` args, and both the instruction and fee-instruction lists are chained. `InstructionArg::Blob` is flat, so there's no nesting to miss.

No false-reject risk that I can see: `Blobs` is a plain `Vec<Blob>` and pruning lives on the separate `PrunedTransaction` type, which can't reach a `Validator<Transaction>`. Both ingress points are wired (mempool and indexer), and the mempool instance at `bootstrap.rs:364` also wraps the consensus `TariBlockTransactionValidator`, so a byzantine proposer can't bypass it. Placing the validator ahead of the weight cap is fine — it's O(instructions) plus a 256-slot bitvec, negligible against the deserialization already paid.

One nit below.

PR #2426, opened 2026-08-13
feat(ootle-wasm): expose script-path witness construction for spending PayTo::Conditions outputs

chironbuilds:
## Summary

`createStealthOutputWitness`'s `pay_to_json` already accepts a full `PayTo` -- including `Conditions`, an HTLC-style hashlock/timelock MAST tree -- because `pay_to_output_authorization` handles every `PayTo` variant generically. This isn't obvious from the wasm binding's own doc comment, which only documents `"StealthPublicKey"` and `{"AccessRule": ...}`; I confirmed the `Conditions` variant works today against the published `0.38.0` build by constructing a real hashlock+timelock condition tree and inspecting the returned `auth` field (produces a real `SpendAuthorization::Script` root).

What's actually missing is the *spend* side: there was no way to build a `SpendWitness::ScriptPath` (revealing a leaf + Merkle inclusion proof + optional witness data, e.g. a hashlock preimage) from the wasm crate at all, even though the underlying crypto already exists and is already used by `tari_walletd` -- `script_path_witness_with_data` (`tari_ootle_wallet_crypto::stealth`) and `StealthInput::with_witness` (`tari_template_lib_types::stealth`) are both real, tested, unexported-from-wasm functions.

## Changes

Two purely-additive `#[wasm_bindgen]` exports, no existing export touched:

- **`buildScriptPathWitness(conditions_json, leaf_json, data)`** -- wraps `script_path_witness_with_data`. Returns `{ "witness": <SpendWitness>, "condition_root": <Hash32> }`.
- **`buildStealthInputsStatementFromInputs(inputs_json, revealed_amount_microtari)`** -- like the existing `buildStealthInputsStatement`, but takes a JSON array of `{commitment, witness}` so a caller can attach a script-path witness per input instead of the existing function's hardcoded key-path (`StealthInput::new`).

## Motivation

Came out of a conversation about wiring HTLC-based atomic-swap support into a browser-extension wallet (`tari-wallet-extension`, entirely client-side, no `tari_walletd`). Output creation already worked; this was the missing half. Related to the walletd-side proof of concept in #2343 and the `create_stealth_transfer_statement` handler merged in #2386 -- same underlying capability, exposed for a client that has no daemon to delegate to.

## Testing

- 6 new unit tests in `crates/ootle_wasm/core/src/stealth/inputs.rs`, including a cross-check that the condition root computed on the spend side (`buildScriptPathWitness`) matches the root independently computed the way output creation does (`validated_condition_root`) -- the property that actually matters for fund safety, since a mismatch there means an output could be created under one root and never spendable under the witness this produces.
- Full existing test suites for both touched crates (`ootle-wasm-core`, `tari_ootle_wallet_crypto`) still pass: 46 tests, 0 failures.
- `cargo clippy -p ootle-wasm-core -p ootle-wasm --all-targets`: clean.
- `cargo +nightly-2025-12-05 fmt --all`: applied.
- Built the actual wasm package (`bash crates/ootle_wasm/build.sh nodejs release`) and ran a real end-to-end Node.js test against the compiled binary: create an HTLC-conditioned output via `createStealthOutputWitness`, build a claim witness via the new `buildScriptPathWitness`, confirm the two independently-arrived-at condition roots match exactly, and confirm the preimage witness data survives into the resulting `StealthInputsStatement` via `buildStealthInputsStatementFromInputs`. All checks passed.

## Scope

Wasm bindings only. No change to `tari_walletd`, transaction validation, submission, or any signing/sealing path -- this only builds the witness JSON a caller (or a higher-level SDK like `@tari-project/ootle`) would still pass through the existing `addTransactionSigner`/`sealTransaction` flow, unchanged.

### DIFF
--- a/crates/ootle_wasm/core/src/stealth/inputs.rs
+++ b/crates/ootle_wasm/core/src/stealth/inputs.rs
@@ -9,10 +9,12 @@
 //! without hand-crafting JSON.
 
 use tari_crypto::{ristretto::RistrettoSecretKey, tari_utilities::ByteArray};
+use tari_ootle_wallet_crypto::stealth::script_path_witness_with_data;
 use tari_template_lib_types::{
     Amount,
+    bytes::Bytes,
     crypto::PedersenCommitmentBytes,
-    stealth::{StealthInput, StealthInputsStatement},
+    stealth::{SpendCondition, StealthInput, StealthInputsStatement},
 };
 
 use crate::{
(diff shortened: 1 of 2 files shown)


sdbondi (approved):
Thanks!

PR #2439, opened 2026-08-18
feat(engine)!: price logs by the byte, drop EmitLog, bound the burn rate

sdbondi:
Follow-ups from #2438, in three independent commits.

## Charge a log for the bytes it carries

`emit_log` was charged a flat `per_module_call_cost` whatever it carried. With
`max_logs: 256` and `max_log_size_bytes: 32 KiB`, that puts **8 MiB within reach of a transaction
paying 256 µT**. #2416 removed logs from the persisted `TransactionReceipt`, so the permanent cost
is gone, but the transient one is not: the payload is held in memory for the length of execution
and retained in every validator's `transaction_executions` record via `FinalizeResult.logs`.

`emit_log` now raises `RuntimeEvent::LogEmitted { size_bytes }`, and `FeeModule` charges
`per_byte_storage_cost × bytes / log_bytes_cost_divisor` for it.

- **Charged as the log is emitted**, not at finalisation, so the bytes shrink the compute allowance
  #2438 introduced rather than being billed after the work is already done.
- **Under `FeeSource::RuntimeCall`**, the same source as the host call that emitted it — a new
  `FeeSource` would widen `FeeReceipt::widest()` and move the receipt size bound, and therefore the
  fee, on every transaction on the network.
- **Divisor 64 on the shipped tables.** A log lives in a per-node, prunable execution record rather
  than in consensus state, so it is priced well under a persisted byte: a max-size entry is ~512 µT
  and a transaction that fills `max_logs` with them ~0.13 tTARI, against ~8.4 tTARI for the same
  bytes in a substate. An ordinary diagnostic line costs about what the host call does, so no
  shipped fee assertion moved.

## Drop the `EmitLog` instruction and the manifest log macros

A transaction-carried log gives the sender nothing a template's own `debug!` does not: the message
is a literal fixed at signing time, so it says only what the sender already knew, and with the
change above it is paid for twice — as transaction weight for the bytes on the wire, and again for
the bytes it puts in the execution record.

Removes `Instruction::EmitLog`, its weight term and execution arm, and the
`info!`/`debug!`/`warn!`/`error!` manifest macros that generated it, along with the TS binding, the
web-UI branch and the doc references. **Logging from a template is untouched.**

## Hold the exhaust burn rate to the fee estimate's ceiling

`FEE_ESTIMATE_ALLOWANCE` is derived against `MAX_EXHAUST_BURN_RATE_BPS`, so a network configured
above that rate under-states `FeeReceipt::required_fees` and every submission built from a dry run
is rejected as underpaid. The rate was a bare `u16` on `ConsensusConstants`, guarded only by a test
over the values we ship.

`ExhaustBurnRate` cannot hold a rate above the ceiling: `new` panics — a compile error in the const
context the network constants are built in — and `try_new` validates a rate known only at runtime.
A `const _` block const-evaluates every shipped network, so an out-of-range rate fails the build
rather than a test. Verified by temporarily setting a network to 20 000 bps: `error[E0080]:
evaluation panicked: exhaust burn rate is above MAX_EXHAUST_BURN_RATE_BPS`.

## Breaking changes

- A log's bytes are now charged for, and `FeeTable` gains `log_bytes_cost_divisor`.
- `Instruction::EmitLog` is gone, and a manifest using the log macros no longer parses.
- `ConsensusConstants::exhaust_burn_rate_bps: u16` is now `exhaust_burn_rate: ExhaustBurnRate`, and
  `exhaust_burn_rate()` returns that type.

## Test plan

- [x] `tari_engine --test {fees, compute_fee_budget, native_compute_budget, metering_budget, complex_fee_payment, limits}`
- [x] `tari_template_builtin --test liquidity_pool`
- [x] `tari_transaction_manifest`, `tari_ootle_transaction`, `ootle_sdk_core` (golden vectors included)
- [x] `tari_indexer`, `tari_validator_node`, `consensus_tests` compile against the new constant type

## Still open from #2438's follow-up list

Items 7 (tighten the receipt size bound) and 8 (a test for the `new_substates` asymmetry) are not
in this PR.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -40,8 +40,8 @@ invariants, the reason for a non-obvious approach, tricky edge cases. They must 
 
 - DSL parsed by `tari_transaction_manifest::parse_manifest()`
 - Parser is in `crates/transaction_manifest/src/parser.rs`, generator in `generator.rs`
-- Macros: `var!`/`arg!`/`global!`, `new_component_addr!`, `new_resource_addr!`, `create_account!`, `drop_all_proofs!`,
-  log macros
+- Macros: `var!`/`arg!`/`global!`, `new_component_addr!`, `new_resource_addr!`, `create_account!`,
+  `drop_all_proofs!`
 - `create_account!(owner_pk)` generates `Instruction::CreateAccount` - idempotent account creation
 - Variables passed to manifests are parsed via `ManifestValue::FromStr` which handles: substate IDs (`component_<hex>`,
   `resource_<hex>`, etc.), NFT IDs, Rust literals, and raw hex bytes (for public keys)
--- a/applications/tari_indexer/src/dry_run/processor.rs
+++ b/applications/tari_indexer/src/dry_run/processor.rs
@@ -104,7 +104,7 @@ impl DryRunTransactionProcessor {
 
         // Estimate the burn at the rate in effect for the current epoch.
         let current_epoch = self.epoch_manager.current_epoch().await?;
-        let burn_rate_bps = self.consensus_constants.exhaust_burn_rate(current_epoch);
+        let burn_rate_bps = self.consensus_constants.exhaust_burn_rate(current_epoch).as_bps();
 
         let mut state_store = new_memory_store();
         state_store.set_many(found_substates)?;
--- a/applications/tari_indexer/src/network_state_sync/worker.rs
+++ b/applications/tari_indexer/src/network_state_sync/worker.rs
@@ -196,7 +196,7 @@ impl NetworkWideStateSync {
         match ReadOnlyStore::new(self.store.clone()).get_tari_economics().await {
             Ok(economics) => {
                 let current_epoch = self.epoch_manager.get_current_epoch();
-                let target_burn_rate_bps = self.consensus_constants.exhaust_burn_rate(current_epoch);
+                let target_burn_rate_bps = self.consensus_constants.exhaust_burn_rate(current_epoch).as_bps();
                 self.metrics.update(&economics, target_burn_rate_bps);
             },
             Err(err) => {
--- a/applications/tari_indexer/src/rest_api/handlers/network.rs
+++ b/applications/tari_indexer/src/rest_api/handlers/network.rs
@@ -54,7 +54,9 @@ pub async fn get_economics(Extension(context): Extension<HandlerContext>) -> Han
         .await
         .map_err(ErrorResponse::anyhow)?;
 
-    let target_burn_rate_bps = ConsensusConstants::from(context.network()).exhaust_burn_rate(current_epoch);
+    let target_burn_rate_bps = ConsensusConstants::from(context.network())
+        .exhaust_burn_rate(current_epoch)
+        .as_bps();
     // Supply nets the receipt-sourced burn against claimed (both advance on the state-sync frontier), matching
     // `get_xtr_total_supply`; `total_exhaust_burned` (header) is reported alongside as a cross-check.
     let total_supply = econ
--- a/applications/tari_ootle_app_utilities/src/fee_tables.rs
+++ b/applications/tari_ootle_app_utilities/src/fee_tables.rs
@@ -37,6 +37,10 @@
 //! - **`per_byte_storage_cost`**: Cost per byte of data written to persistent storage (substates). Note: The actual
 //!   cost is reduced by a factor of 4 (cost × 0.25) to make storage more affordable.
 //!
+//! - **`log_bytes_cost_divisor`**: Divides the per-byte storage rate when charging for a log message
+//!   (`per_byte_storage_cost × message_bytes / log_bytes_cost_divisor`). A log is retained in every validator's
+//!   execution record but is not consensus state and is prunable, so it is priced below a persisted byte.
+//!
 //! ## Template Publishing Fees
 //!
 //! A published template's binary is priced by its own model instead of the flat per-byte storage rate, since it is
@@ -93,6 +97,12 @@ const TESTNET_FEE_TABLE: FeeTable = FeeTable {
     template_load_bytes_cost_divisor: 3000,
     // 1 µT per 1000 Wasmer points. Lower values make metering more aggressive.
     wasm_points_cost_divisor: 1000,
+    // A log byte costs a 64th of a persisted byte. A log is retained in each validator's record of
+    // the execution rather than in consensus state, so it is priced well under permanent storage:
+    // a max-size entry (32 KiB) is ~512 µT and a transaction that fills `max_logs` with them is
+    // ~0.13 tTARI, against ~8.4 tTARI for the same bytes in a substate. An ordinary diagnostic line
+    // costs about as much as the host call that emits it.
+    log_bytes_cost_divisor: 64,
     // First 96 KiB of a template binary are priced at the per-byte storage rate; beyond that the
     // quadratic publish premium applies. A published template carries ~25 KiB of fixed
     // `template_lib` machinery before any author code, and a minimal component template optimised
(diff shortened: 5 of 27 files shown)


sdbondi (commented):
Reviewed all three commits against the head at `18c52eb`, built `tari_consensus` locally, and checked the load-bearing claims in the description rather than taking them on trust.

**Verified sound**

- The const guard actually fires. Temporarily setting mainnet to `20_000` bps and running `cargo check -p tari_consensus` gives `error[E0080]: evaluation panicked: exhaust burn rate is above MAX_EXHAUST_BURN_RATE_BPS`, so the `const _` block at `crates/consensus/src/consensus_constants.rs:294` is doing the work the removed test used to.
- The `FeeSource` argument holds up: `FeeReceipt::widest()` (`crates/engine_types/src/fees.rs:167`) adds a breakdown entry per `FeeSource::ALL`, so a new source really would move the receipt bound. Reusing `RuntimeCall` keeps `FeeSource::ALL` at 11 and the bound unchanged.
- The compute-allowance claim holds: `wasm_point_allowance` (`crates/engine/src/runtime/tracker.rs:186`) subtracts `fee_state.total_charges()`, so charging at emit time does shrink what execution may still spend.
- Dropping `EmitLog` keeps the `#[n(4)]` slot vacant, so every other variant's CBOR index is unchanged — no golden-vector churn, and `crates/ootle_sdk_core/tests/golden_vectors.rs` contains no `EmitLog` case to break.
- Every `FeeTable` literal is accounted for: the two shipped tables, `template_test.rs:242`, and `zero_rated()` — the two `..FeeTable::zero_rated()` literals in tests inherit the new field.
- The numbers in the docs check out at `per_byte_storage_cost: 1`, divisor 64: 32 KiB → 512 µT, 256 logs → 0.131 tTARI, 8 MiB at `storage_cost_divisor: 1` → 8.39 tTARI.
- No stray `EmitLog` or manifest-log references left anywhere; `bindings/src/types/LogLevel.ts` is still reachable via `LogEntry`, so removing the `Instruction` import doesn't orphan it.

**Non-blocking**

1. `crates/engine/src/runtime/impl.rs:1092` — the byte charge is raised before `add_log` enforces `max_log_size_bytes` and `max_logs` (`crates/engine/src/runtime/working_state.rs:1791`, `:1798`). A 4 MiB message is billed ~65 000 µT for a log that is then rejected and never recorded. The transaction fails either way so nothing diverges, but the charge is then pricing bytes that were never retained — which is the justification the commit gives for the charge existing. Either clamp `size_bytes` to `ENGINE_LIMITS.max_log_size_bytes`, or say in the comment that overshoot is billed deliberately.

2. `crates/consensus/src/consensus_constants.rs:294` — the guard is a hand-maintained list of constructors. A `ConsensusConstants::stagenet()` added later and not listed silently reverts to a runtime panic, which is the failure mode the commit set out to remove. Binding each network as a `const` item the constructor returns (`pub const MAINNET: Self = ...; pub const fn mainnet() -> Self { Self::MAINNET }`) makes the evaluation unforgettable rather than opt-in.

3. `crates/engine_types/src/fees.rs:47-77` — `try_new`, `TryFrom<u16>`, `is_zero`, `ZERO` and `Display` have no caller in the workspace. Fine as API surface, but the commit message justifies `try_new` with "a rate that is only known at runtime, such as one read from configuration" and no such path exists: `ConsensusConstants` is only ever built from the four const constructors. Worth softening the claim or dropping the unused half.

4. `crates/engine_types/src/fees.rs:301` — `FeeSource::RuntimeCall` now carries two unlike prices, so a reader can no longer divide the line by `per_module_call_cost` to recover the host-call count. The reason is good and it is written down in `fees.mdx`, but the enum's own doc comment is where someone reading a receipt will look. One line there would pay for itself.

5. Pricing bounds the cost, not the footprint. `max_logs × max_log_size_bytes` is still 8 MiB per transaction, and since logs generated inside a template carry no transaction weight, `max_block_weight` doesn't constrain them — the binding limit is `max_block_execution_points / MAX_WASM_POINTS_PER_TRANSACTION` ≈ 45 transactions, so a block can still land a few hundred MiB in validators' `transaction_executions` for a few tTARI. Is tightening the two limits on the follow-up list alongside items 7 and 8?

6. `crates/engine/tests/fees.rs:83` — nothing covers the rounding the `fee_module` comment reasons about: a 63-byte message is free at the shipped divisor. The helper is already parameterised by length, so a second case pins it for the cost of one line.

**Nit**

`applications/tari_ootle_app_utilities/src/fee_tables.rs:36` — the new bullet lands directly beneath a stale one. `per_byte_storage_cost` is documented as "reduced by a factor of 4 (cost × 0.25)", but both shipped tables set `storage_cost_divisor: 1` and the inline comment at `:94` says why. Pre-existing, but you're editing that list.

Approving once `test (1..3/3)` and `integration test` finish — everything else is green and the three commits are cleanly separable.

sdbondi (commented):
Re-checked each of the six items against `59d0305`. Four are properly fixed; one fix is inert and one broke the build.

**Blocking**

1. `crates/engine_types/src/fees.rs:4` — dropping the `Display` impl orphaned `use std::fmt;`, and `lints.toml` denies `unused_imports`, so this is an error rather than a warning. The clippy job is already red on this head:

   ```
   error: unused import: `std::fmt`
   error: could not compile `tari_engine_types` (lib) due to 1 previous error
   ```

2. `crates/consensus/src/consensus_constants.rs:126` — `pub const DEVNET: Self = Self::devnet(Self::DEFAULT_DEVNET_COMMITTEE_SIZE)` never gets evaluated, so `devnet`'s body is no longer const-checked. An associated const is only evaluated when something forces it, and `DEVNET` is referenced nowhere outside its own doc comment — `devnet()` builds inline from the parameter rather than reading it, and `From<Network>` calls `Self::devnet(...)`. `MAINNET`/`ESMERALDA`/`TESTNET` are fine because `mainnet()`/`esmeralda()`/`testnet()` return them and are reachable.

   This is a regression against the `const _` block it replaced. Setting devnet's rate to `20_000` and running `cargo check -p tari_consensus`:

   - at `18c52eb` (the `const _` block): `error[E0080]: evaluation panicked: exhaust burn rate is above MAX_EXHAUST_BURN_RATE_BPS`
   - at `59d0305` (associated consts): compiles clean

   The same edit to `MAINNET` still errors at `59d0305`, so only the devnet path is uncovered — which is the one that most wants the check, being the only constructor whose body isn't already a const item. Making `From<Network>` fall back to the const rather than re-stating the default would both force evaluation and remove the duplicated `7`:

   ```rust
   Network::LocalNet => env::var("TARI_DEVNET_COMMITTEE_SIZE")
       .ok()
       .and_then(|s| s.parse().ok())
       .map_or(Self::DEVNET, Self::devnet),
   ```

**Verified fixed**

- The log charge now sits after `add_log` (`crates/engine/src/runtime/impl.rs:1103`), so a rejected entry is billed only the flat per-call cost. `size_bytes` is read before the move, and the comment says why the ordering matters.
- The rounding case is pinned (`crates/engine/tests/fees.rs:99`) — a 63-byte message leaves the `RuntimeCall` total unchanged.
- `FeeSource::RuntimeCall` now carries the two-unlike-prices note at the enum (`crates/engine_types/src/fees.rs:268`).
- The `per_byte_storage_cost` doc no longer claims a ×0.25 that neither shipped table applies.
- The unused `try_new`/`TryFrom`/`is_zero`/`ZERO`/`Display` surface is gone and the test names now describe what they cover.

**On the constants refactor**

Moving three constructor bodies into const items by hand is exactly where a transposed field hides, so I diffed the resolved values rather than reading them. Dumping `{:?}` for all six networks plus `devnet(7)` and `devnet(3)` at `18c52eb` and at `59d0305` gives byte-identical output — every field preserved, `epoch_end_spread_blocks` included.

Item 5 from the last round (log footprint vs. price) is unaddressed, which is fine — it reads as follow-up work rather than something this PR owes.

PR #2443, opened 2026-08-19
fix(walletd): settle the stealth transfer fee estimate before reporting it

sdbondi:
## The bug

A dry run prices whatever shape the caller's guessed fee produces — and that fee is an input to the
shape. Stealth input selection targets `amount + max_fee`, so the fee decides which UTXOs are spent
and whether any change is left over, and a change output is another stealth output to verify at
`PER_OUTPUT` = 6,000,000 points. The estimate could therefore describe a materially cheaper
transaction than the one built from it.

Reproduced on a local swarm, on the transaction that prompted this:

| | inputs selected | stealth outputs | `NativeExecution` | required |
|---|---|---|---|---|
| dry run at the frontend's `max_fee=1` | worth 8,000,001 | 1 (change = 0) | 8,184 | **9,354** |
| dry run at 9,354 | worth 8,986,059 | 2 (change 976,705) | 14,184 | 15,951 |
| the submission built from the estimate | worth 8,986,059 | 2 | 14,184 | **15,951** |

At `max_fee=1` the selected inputs covered the transfer exactly, so there was no change and the
statement carried a single output. At the real fee they no longer cover it, a larger pair is chosen,
change appears, and the transaction is rejected as underpaid — collecting nothing, since a fee intent
that fails leaves no checkpoint to fall back to.

The whole 6,597 shortfall is one stealth output: 6,000 for the output, 300 of exhaust burn on it, and
~297 of storage and substate creation.

## The fix

The dry-run path in `handle_stealth_transfer` rebuilds at each figure it reports until a build at
that figure pays for itself, then answers with the run that named it. The caller therefore reads a
fee that some build has been held to, rather than one derived from a shape that is about to change.

The settling test is against **what the shape was charged**, not against `required_fees`. The estimate
allowance `required_fees` carries on top is the caller's margin for the metering drift a wider
`max_fee` causes; testing against it treats a fee that already pays in full as insufficient and spends
another round chasing the padding. On the swarm that difference alone took the loop from 4 rounds to
2. The reported figure still includes the allowance.

Bounded at 5 rounds, with a warning and the highest figure reached if it does not settle — an estimate
below the cost cannot be submitted at all, so the higher figure is the safer answer.

`StealthTransferParams` gains `Clone` so a round can rebuild. Each round's lock is released as before,
which also discards that round's unconfirmed change outputs.

## Verified on a swarm

Wallet daemon rebuilt and restarted on this branch, same account and destination as the failing
transaction:

| transfer | estimate | rounds | submission |
|---|---|---|---|
| 8,000,000 | 15,978 | 4 (before the `charged` refinement) | **Accepted**, paid 15,978 |
| 500,000 | 15,909 | 2 | **Accepted**, paid 15,909 |

Both settle with an overcharge equal to the estimate allowance (25 and 23 µT), which is expected and
is not refundable on a stealth-revealed fee — the reason this iterates rather than padding the
estimate.

## Also, from review on #2441

- `enable_fees` gets its doc comment back; inserting `set_dry_run` above it had taken it.
- The dry-run credit paragraph in `reference/fees.mdx` moves below the `FEE_ESTIMATE_ALLOWANCE`
  bullets, so "That figure" refers to `required_fees()` again.
- Adds a paragraph there on why an estimate only describes the shape it was metered on, which is the
  bug class this PR fixes.

## Test plan

- [x] `cargo check` clean on `tari_ootle_walletd` and `tari_ootle_wallet_sdk`, including tests
- [x] Docs site builds
- [x] End-to-end on a local swarm: estimate → submit → committed, twice (table above)

### DIFF
--- a/applications/tari_walletd/src/handlers/accounts.rs
+++ b/applications/tari_walletd/src/handlers/accounts.rs
@@ -1283,7 +1283,7 @@ pub async fn handle_stealth_transfer(
         })
         .collect::<anyhow::Result<_>>()?;
 
-    let params = StealthTransferParams {
+    let mut params = StealthTransferParams {
         max_epoch: context.transaction_max_epoch().await?,
         fee_params: req.fee_params,
         input_selection: req.input_selection,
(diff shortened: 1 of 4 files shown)


sdbondi (commented):
Reviewed at `6925353`, read against the base and traced into the SDK/engine definitions.

The settling logic itself is sound, and worth stating plainly before the findings:

- The loop only continues while `charged > max_fee`, so `required_fees` strictly increases across rounds. That makes the cap's claim of returning "the highest figure reached" actually true, rather than merely likely.
- Returning `verified_estimate` (round *k-1*) rather than the settling round is the right choice: the caller submits at `required_{k-1}`, which is the fee round *k* was built at, and round *k* is what showed a build at that fee pays for itself.
- Testing against `total_fees_required` rather than `required_fees()` is correct — the allowance is the caller's margin, not part of the shape's cost, and testing against it would spend a round chasing padding.
- The "each round's lock is released, discarding its unconfirmed change outputs" claim in the description checks out: `stealth_outputs_release_by_lock_id` deletes the `LockedUnconfirmed` rows for the lock (`crates/wallet/storage_sqlite/src/writer.rs:144`), so a round leaves no residue.
- The extra dry runs don't pollute transaction history — `transactions_fetch_all` filters `dry_run = false` (`crates/wallet/storage_sqlite/src/reader.rs:296`).
- The docs reorder does fix the dangling reference: "That figure" follows `required_fees()` again.

## Blocking

**`charged` reads the raw wire field, defeating the fix on a legacy payload** — `applications/tari_walletd/src/handlers/accounts.rs:1356`

`total_fees_required` is additive on the wire and decodes as `0` from a peer that predates it. `required_fees()` guards exactly this, deliberately, and says so (`crates/engine_types/src/commit_result.rs:226-233`):

> Falls back to what the receipt was charged. `total_fees_required` is additive on the wire, so a payload from a peer that predates it decodes as zero — reading it raw would answer a fee estimate with the allowance alone.

There is a test pinning it: `a_missing_required_total_falls_back_to_what_was_charged`.

Against such an indexer, `charged` is `0`, so `charged <= params.max_fee` holds unconditionally. Round 2 then returns round 1's estimate — the estimate taken at the caller's guessed fee, priced on the shape that fee produced. That is precisely the bug this PR fixes, silently restored, with no warning and no round cap hit to signal it. `required_fees()` still falls back correctly, so the returned figure looks plausible; it is just the wrong shape's figure.

Take the same `.max()`:

```rust
let charged = result
    .finalize
    .total_fees_required
    .max(result.finalize.fee_receipt.total_fees_charged());
```

Better as a `FinalizeResult::charged_fees()` accessor that `required_fees()` is then defined in terms of, since there is currently no way to ask for "what this cost" without the allowance and the next caller will reach for the raw field too.

## Non-blocking

**The round bound in the description doesn't match the code** — `accounts.rs:1414`

The description says "Bounded at 5 rounds"; the constant is `4`. The constant's own doc says "how many times a dry run may *rebuild*", but `rounds` counts builds, so 4 permits 3 rebuilds — and "the third confirms it; the rest is headroom" describes exactly one spare round, not "the rest". Worth reconciling, especially as the description's first swarm row (8,000,000, 4 rounds) would have hit the cap and emitted the warning rather than settled.

**The docs claim more than the code does** — `docs/developer-docs/src/content/docs/reference/fees.mdx:382`

"The wallet daemon does that internally for its dry runs" reads as a property of the daemon. Only `handle_stealth_transfer` iterates. `handle_confidential_transfer` (`accounts.rs:1091`, dry run at `:1144`) passes `max_fee` into `ConfidentialTransferParams` input selection and still answers from a single round — same feedback, same bug class. Either narrow the sentence to the stealth transfer path or say the confidential path is still single-round.

**A later round can now fail an estimate that previously succeeded** — `accounts.rs:1317`

Re-deriving at a higher `max_fee` widens the selection target to `amount + max_fee`, so an account that could fund the guessed-fee shape but not the real one now gets `InsufficientFunds` out of `lock_inputs_for_transfer` (`crates/wallet/sdk/src/apis/stealth_transfer/api.rs:160`) instead of an estimate it would have failed to submit at anyway. That is arguably the better answer, but it surfaces as a bare insufficient-funds error with nothing tying it to the round that raised the fee. Worth wrapping with the round and fee for context.

**No test covers the settle decision**

The condition and the cap are the kind of logic that regresses silently — the blocking finding above is an instance of it. The loop is inline in the handler with `sdk` and `transaction_service` in scope, so nothing can reach it from a unit test. Lifting the decision out — `(charged, max_fee, has_candidate, rounds) -> Settle | Retry(next_fee) | GiveUp` — would make the invariant testable without touching the I/O.

## Nits

- `verified_estimate` (`accounts.rs:1313`) holds the previous round's response, which is not verified when it is assigned — it becomes so only if the next round's check passes. `candidate` or `pending_estimate` reads truer; the comment at `:1362-1363` is already carrying the explanation the name doesn't.
- The warning at `:1372` mixes an inline `{charged}` with a positional `{}` and reads as a garden path ("a max fee of X was charged Y"). Something like `"built at {} but cost {charged}"` is easier to parse in a log.

PR #2445, opened 2026-08-19
fix(walletd): price the settle decision against a figure that survives the wire

sdbondi:
Follow-ups from review on #2443, which landed after it merged.

## Blocking: the settle test read a field that decodes as zero

The loop tested `FinalizeResult::total_fees_required` raw. That field is additive on the wire and
decodes as `0` from a peer that predates it — which is exactly why `required_fees()` applies a
`.max(fee_receipt.total_fees_charged())` fallback and says so, pinned by
`a_missing_required_total_falls_back_to_what_was_charged`.

Against such an indexer the shape's cost read as zero, so `charged <= max_fee` held unconditionally
and round 2 answered with round 1's estimate: the figure taken at the caller's guessed fee, priced on
the shape that fee produced. That is the bug #2443 exists to prevent, restored silently — no warning,
no round cap hit, and `required_fees()` still falling back correctly, so the returned figure looked
plausible while describing the wrong shape.

`FinalizeResult::charged_fees()` now names "what this cost, with no allowance on top" and applies the
same fallback; `required_fees()` is defined in terms of it. There was previously no way to ask that
question without reaching for the raw field, so the next caller would have made the same mistake.

## The settle decision is now testable

`next_fee_estimate_step(built_at, charged, required, has_candidate, rounds_taken)` returns
`Settle | Retry { max_fee } | GiveUp`. The loop keeps the I/O; the conditions it turns on are a pure
function with five unit tests, including the case above — a round that covers itself with **no**
candidate must still retry, because the fee under test is the caller's guess and nothing has verified
it.

## Also

- `MAX_FEE_ESTIMATE_ROUNDS` is 5. #2443's description said 5 while the constant said 4; the intended
  value is 5, which leaves headroom past the three builds the common case needs. Its doc no longer
  says "rebuild" when the counter counts builds.
- A build failure carries the fee and round that provoked it. A later round widens the selection
  target to `amount + max_fee`, so it can exhaust an account that funded the earlier one, and that
  surfaced as a bare `InsufficientFunds` with nothing tying it to the round that raised the fee.
- `verified_estimate` is now `candidate` — it is not verified when assigned; the round after it is
  what verifies it.
- The fee reference no longer implies the daemon iterates generally: `accounts.stealth_transfer`
  does, `accounts.confidential_transfer` still answers from a single round and has the same
  feedback.

## Test plan

- [x] `cargo test -p tari_ootle_walletd --lib fee_estimate_step` — 5 passing
- [x] `cargo test -p tari_engine_types --lib commit_result` — including the extended fallback test
- [x] Docs site builds

### DIFF
--- a/applications/tari_walletd/src/handlers/accounts.rs
+++ b/applications/tari_walletd/src/handlers/accounts.rs
@@ -1315,16 +1315,23 @@ pub async fn handle_stealth_transfer(
 
         loop {
             let max_fee_this_round = params.max_fee;
-            let (lock, transfer) = sdk
+            let build = sdk
                 .stealth_transfer_api()
                 .transfer(owner_account.clone(), params.clone())
-                .await
-                .with_context(|| {
+                .await;
+            // Only an estimate names a round, and only an estimate raises the fee between builds:
+            // a later round widens the selection target to `amount + max_fee`, so it can exhaust an
+            // account that funded the earlier one, and the failure needs to say which fee did it.
+            let (lock, transfer) = if req.dry_run {
+                build.with_context(|| {
                     format!(
-                        "building the transfer at a max fee of {max_fee_this_round} (round {})",
+                        "building the transfer at a max fee of {max_fee_this_round} (fee estimate round {})",
                         rounds + 1
                     )
-                })?;
+                })?
+            } else {
+                build?
+            };
 
             let transaction = transfer.transaction;
             let main_pk = transfer.main_signer.public_key().to_byte_type();
@@ -1353,7 +1360,13 @@ pub async fn handle_stealth_transfer(
                 let result = transaction_service
                     .submit_dry_run_transaction(transaction)
                     .await
-                    .map_err(|e| anyhow::anyhow!("Dry run transaction failed: {}", e))?;
+                    .map_err(|e| {
+                        anyhow::anyhow!(
+                            "Dry run transaction failed at a max fee of {max_fee_this_round} (fee estimate round {}): \
+                             {e}",
+                            rounds + 1
+                        )
+                    })?;
 
                 let required_fees = result.finalize.required_fees();
                 // What this round's shape actually costs, without the estimate allowance that
@@ -1944,8 +1957,11 @@ mod fee_estimate_step_tests {
 
     /// The caller's guessed fee has nothing behind it, so a round that covers itself with no
     /// candidate still has to be built at the figure it reports before that figure can be answered
-    /// with. This is what stops a legacy payload — whose cost decodes as zero — settling on the
-    /// guess.
+    /// with.
+    ///
+    /// This delays a cost of zero by one round rather than rejecting it: a second round does hold a
+    /// candidate, and `0 <= built_at`, so it would settle. What keeps a zero out of here is
+    /// `FinalizeResult::charged_fees`, which falls back to what the receipt was charged.
     #[test]
     fn does_not_settle_on_the_callers_guess() {
         assert_eq!(next_fee_estimate_step(1, 0, 9_354, false, 1), FeeEstimateStep::Retry {
--- a/applications/tari_walletd/src/handlers/accounts.rs
+++ b/applications/tari_walletd/src/handlers/accounts.rs
@@ -3,7 +3,7 @@
 
 use std::{collections::HashSet, iter, path::Path, time::Duration};
 
-use anyhow::{Context, anyhow};
+use anyhow::anyhow;
 use axum_extra::headers::authorization::Bearer;
 use indexmap::{IndexMap, IndexSet};
 use log::*;
@@ -1322,10 +1322,12 @@ pub async fn handle_stealth_transfer(
             // Only an estimate names a round, and only an estimate raises the fee between builds:
             // a later round widens the selection target to `amount + max_fee`, so it can exhaust an
             // account that funded the earlier one, and the failure needs to say which fee did it.
+            // The cause is interpolated rather than added as context: `anyhow::Error` renders only
+            // its outermost layer under `{}`, which is what the caller and the log see.
             let (lock, transfer) = if req.dry_run {
-                build.with_context(|| {
-                    format!(
-                        "building the transfer at a max fee of {max_fee_this_round} (fee estimate round {})",
+                build.map_err(|e| {
+                    anyhow!(
+                        "building the transfer at a max fee of {max_fee_this_round} (fee estimate round {}): {e}",
                         rounds + 1
                     )
                 })?
@@ -1407,7 +1409,7 @@ pub async fn handle_stealth_transfer(
                     Some(lock.id()),
                 )
                 .await
-                .context("Transaction failed to submit")?;
+                .map_err(|e| anyhow!("Transaction failed to submit: {e}"))?;
 
             // Transaction submitted, we're home free, make sure to allow the lock to persist past this call.
             // The wallet will monitor the transaction and release the lock when it's finalized.
@@ -1999,7 +2001,9 @@ mod fee_estimate_step_tests {
     /// a build at what that shape reported.
     #[test]
     fn the_cap_leaves_room_for_the_common_case() {
-        assert!(MAX_FEE_ESTIMATE_ROUNDS >= 3);
+        const {
+            assert!(MAX_FEE_ESTIMATE_ROUNDS >= 3);
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


sdbondi (commented):
Reviewed at `f87d589` against `5cf814cb`, re-checking each finding from the review on #2443 rather than taking the description's word for it.

## The findings, re-verified

- **`charged` reading the raw field** — fixed properly. `charged_fees()` applies the same `.max(fee_receipt.total_fees_charged())` fallback and `required_fees()` is now defined in terms of it (`crates/engine_types/src/commit_result.rs:229`), so the two cannot drift, and the existing fallback test asserts both. This is the shape I asked for, and naming the question is what stops the next caller reaching for the raw field.
- **Round bound** — `MAX_FEE_ESTIMATE_ROUNDS` is 5 and the doc no longer says "rebuild" while the counter counts builds. The three-build walkthrough matches what the loop actually does: guess → the shape that guess produces → a build at the figure that shape reported.
- **Docs overclaiming** — `fees.mdx:382` now names `accounts.stealth_transfer` and says `accounts.confidential_transfer` still answers from a single round.
- **`verified_estimate` → `candidate`** and the reworded warning — both done.
- **Testability** — `next_fee_estimate_step` is the right seam. It is total, the settle-before-cap ordering is deliberate and `gives_up_at_the_round_cap`'s second assertion pins it, and `Settle`'s `expect` is genuinely unreachable because `has_candidate` gates that arm.
- **Build-failure context** — added, but it costs more than it buys. See below.

## Blocking

**The added context replaces the cause instead of adding to it** — `applications/tari_walletd/src/handlers/accounts.rs:1322`

`resolve_any_error` renders an unrecognised error with `e.to_string()` (`applications/tari_walletd/src/server.rs:374`), and `anyhow::Error`'s `Display` under `{}` prints only the outermost context — the chain needs `{:#}`. Checked against anyhow 1.0:

```
plain  {}: building the transfer at a max fee of 15976 (round 3)
alt  {:#}: building the transfer at a max fee of 15976 (round 3): Insufficient funds: available 8000001, required 15976
```

`StealthTransferApiError` isn't among the types `resolve_any_error` downcasts (`server.rs:349-361`), so it lands in that final branch. Before this PR the caller saw `Insufficient funds: available …, required …`. Now they see only the round and the fee.

That inverts the finding it is meant to address: the complaint was a bare `InsufficientFunds` with nothing tying it to the round that raised the fee, and the answer is a bare round with nothing saying what went wrong. The daemon's own log loses it too — `warn!(… "🌐 JSON-RPC error: {}", e)` at `server.rs:348` is the same `{}`.

Both `server.rs:348` and `:374` should be `{:#}`. That is the better fix rather than the narrower one, because the truncation is not specific to this call — `.context("Transaction failed to submit")` at `accounts.rs:1397` has been losing its cause the same way all along, and every future `.context` in a handler would too. If you would rather not touch the server, interpolate instead:

```rust
.map_err(|e| anyhow!("building the transfer at a max fee of {max_fee_this_round} (round {}): {e}", rounds + 1))?
```

## Non-blocking

**A test comment claims a guarantee the test doesn't provide** — `accounts.rs:1946-1949`

> This is what stops a legacy payload — whose cost decodes as zero — settling on the guess.

It doesn't. Hold `charged` at 0 and trace two rounds: round 1 has no candidate and retries at `required`, but round 2 *does* have one and `0 <= built_at`, so it settles on round 1's estimate regardless. The no-candidate rule buys one round's delay, not immunity. `charged_fees()` is the thing that actually prevents this, and it is worth the comment saying so — a reader who trusts this line has the wrong model of which change is load-bearing. The assertion itself is right and worth keeping; it pins that the caller's guess is never answered with directly.

**The round number is attached to the real-submission path too** — `accounts.rs:1317`

`with_context` wraps the build for both branches, so a failed non-dry-run transfer reports "(round 1)" — a number that can only ever be 1 and means nothing outside the estimation loop. Worth confining the round to the dry-run case.

**The other thing a round can fail at carries no round** — `accounts.rs:1356`

`"Dry run transaction failed: {}"` names neither the fee nor the round, though a round fails there as readily as at the build, and it does interpolate its cause. Same treatment would make the two consistent (and, incidentally, it is already immune to the `{}` truncation above precisely because it interpolates).

**`the_cap_leaves_room_for_the_common_case` doesn't exercise the function** — `accounts.rs:1985`

`assert!(MAX_FEE_ESTIMATE_ROUNDS >= 3)` compares a constant to a literal. It documents intent, which has some value, but it cannot fail in any way the other four tests wouldn't already catch. Not worth removing; worth knowing it isn't covering anything.

---

CI is still pending at the time of writing (clippy, tests, integration).

sdbondi (commented):
Reviewed `912a70d7` against `f87d5899`.

Every finding from the last review is addressed, and the two I flagged as smallest are the ones handled best: the round is now confined to the dry-run path and labelled "fee estimate round", the dry-run submit failure carries the fee, the round *and* its cause, and the comment on `does_not_settle_on_the_callers_guess` now says plainly that the no-candidate rule delays a zero by one round rather than rejecting it, and names `charged_fees` as what actually keeps a zero out. That last one is the correction I care about most — it puts the load-bearing change where a reader will find it.

The blocking finding is fixed, but the fix I suggested doesn't hold up. Detail below.

## Blocking

**`{:#}` duplicates every message, because the SDK's error enums already interpolate their sources** — `applications/tari_walletd/src/server.rs:350` and `:378`

`#[from]` implies `#[source]`, and these enums *also* render the source into their own `Display`:

```rust
// crates/wallet/sdk/src/apis/accounts.rs:612
#[error("Store error: {0}")]
StoreError(#[from] WalletStorageError),
```

So each level of the chain already contains everything below it, and `{:#}` prints every level. Reproduced against anyhow 1.0 with `StealthTransferApiError::Accounts(AccountsApiError::StoreError(…))`, which is a reachable combination:

```
{}   : Accounts API error: Store error: Missing entry for key foo
{:#} : Accounts API error: Store error: Missing entry for key foo: Store error: Missing entry for key foo: Missing entry for key foo
```

Note there is no `.context(…)` anywhere in that reproduction — `?` alone preserves `source()`, so anyhow builds the chain regardless. This is not confined to the two call sites that add context; it is every error walletd returns through these enums, on both the RPC message and the log line. That is a worse trade than the one it fixes.

The convention this codebase actually follows is to interpolate the cause at the wrap site, which is exactly what the new `map_err` at `accounts.rs:1365` does, correctly. Doing the same at the two context sites and putting `server.rs` back to `{}` / `e.to_string()` gets the cause through without touching anything else:

```rust
// accounts.rs:1326
build.map_err(|e| {
    anyhow!("building the transfer at a max fee of {max_fee_this_round} (fee estimate round {}): {e}", rounds + 1)
})?
// accounts.rs:1410
.map_err(|e| anyhow!("Transaction failed to submit: {e}"))?
```

Rendering the chain is the better end state and `{:#}` is how you get there — but only once the `#[from]` variants stop interpolating `{0}` into their own `Display`, across the wallet SDK's error enums. That is a real cleanup and not this PR's job.

## Non-blocking

**The swarm daemon has a copy of `resolve_any_error` that was not touched** — `applications/tari_swarm_daemon/src/webserver/server.rs:189`

Same `warn!("… {}", e)` and same `e.to_string()`. Moot if `server.rs` goes back to `{}`; worth a note if the chain rendering stays, since the commit title claims the behaviour generally.

---

Nothing else new in this commit. CI still pending.

sdbondi (commented):
Reviewed `8babeab4` against `912a70d7`. The blocking finding is resolved and nothing new is introduced. No further findings from me.

Checked rather than assumed:

- `server.rs` is a clean revert — `git diff 5cf814cb..8babeab4 -- applications/tari_walletd/src/server.rs` is empty, so the duplication that `{:#}` would have produced across every error walletd returns is gone, and the swarm daemon's copy of `resolve_any_error` no longer diverges. That note is moot now.
- Both wrap sites carry their cause in the message, so `{}` at `server.rs:373` renders it in full. `accounts.rs:1409` is a strict improvement on what was there before this PR: `"Transaction failed to submit"` dropped its cause, and now it doesn't.
- Dropping `anyhow::Context` from the imports is right — the only remaining `with_context` in the file is `sdk.signer_api().with_context(&main_pk)` at `:1342`, which is the signer's own inherent method, not the trait.
- Losing the source chain at those two sites breaks no downcast: walletd only ever downcasts to `HandlerError`, `JsonRpcError` and `AuthError` (`server.rs:349-360`), and neither `StealthTransferApiError` nor the transaction service's error reaches those arms.
- The comment at `:1325-1326` gives the reason the next person will need — that `anyhow::Error` renders only its outermost layer under `{}` — at the site where the temptation to reach for `.context` is.

One cosmetic thing, entirely take-it-or-leave-it: `:1366` uses the fully-qualified `anyhow::anyhow!` while its two neighbours now use the imported `anyhow!`.

CI: fmt, docs, web ui, prettier, licences and title all green; clippy, tests, integration and fuzz still pending.

PR #2446, opened 2026-08-19
fix(engine)!: refuse engine calls from the response allocation

### LINKED ISSUE #2442: Engine: template tari_alloc calling tari_engine recurses host→WASM→host and crashes the node
## Summary

A template whose `tari_alloc` calls `tari_engine` drives unbounded host→WASM→host recursion and aborts the validator process with a stack overflow. Verified: `thread has overflowed its stack / fatal runtime error: stack overflow, aborting` (SIGABRT).

This is the in-invocation half of the same shape as #2440. That PR closes the outside-invocation entry points (`tari_alloc` for the `CallInfo`, `tari_free` for the return value) — this one is still open with #2440 applied, because the recursion happens while a legitimate invocation is in flight, where engine calls are and must remain permitted.

## Mechanism

`WasmProcess::handle` writes every engine-call response into WASM memory through the template's own `tari_alloc` (`crates/engine/src/wasm/process.rs:340`):

```rust
let resp = f(env_mut.state_mut(), decoded)?;
let len = encoded_len(&resp)?;
let ptr = env_mut.alloc(&mut store, len as u32)?;   // re-enters WASM
```

So a `tari_alloc` that itself calls the engine cycles:

```
template fn --tari_engine--> handle --alloc--> tari_alloc --tari_engine--> handle --alloc--> tari_alloc --> ...
```

Every level is one host frame plus one WASM frame, and none of them return until the innermost does.

## Why the existing bounds don't apply

- **`max_call_depth: 10`** — not reached. This is all one call frame; no `CallInvoke`, no new `WasmProcess`, no runtime call-frame push.
- **Wasmer metering** — deducts per level, but the per-call ceiling is `MAX_WASM_POINTS_PER_CALL = 100_000_000` and a `tari_alloc` body costs on the order of tens of points. The permitted depth is far past what the native stack survives, so the process dies long before the meter traps.
- **Per-op count caps** — only `EmitLog` has one (`max_logs: 256`). The reproduction below uses `GenerateUniqueId`, which has no such cap. Most ops don't.

## Reproduction

Against `tari-free-engine-call` at `6e29d17` (i.e. with #2440's fix applied), as a temporary variant of the `buggy` test template:

```rust
static mut IN_INVOCATION: bool = false;

#[unsafe(no_mangle)]
pub extern "C" fn tari_alloc(size: usize) -> *mut u8 {
    // Skip the CallInfo alloc, which happens before the invocation and is refused by #2440.
    if unsafe { core::ptr::read_volatile(&raw const IN_INVOCATION) } {
        call_engine_uuid();  // tari_engine(0x07 /* GenerateUniqueId */, &[0x80], 1)
    }
    internal_alloc(size)
}

#[unsafe(no_mangle)]
pub unsafe extern "C" fn Buggy_main(_ci: *mut u8, _len: usize) -> *mut u8 {
    unsafe { core::ptr::write_volatile(&raw mut IN_INVOCATION, true) };
    call_engine_uuid();   // first engine call: its response allocation starts the cycle
    unsafe { core::ptr::write_volatile(&raw mut IN_INVOCATION, false) };
    /* return encoded () */
}
```

Executing `Buggy::main` aborts the test process with SIGABRT on stack overflow.

The flag only exists to get past the `CallInfo` allocation; it is not load-bearing to the bug. Any real template built with `#[template]` reaches the same place if its `tari_alloc` calls the engine — the macro just doesn't generate such a template today.

## Impact

Any validator executing the transaction crashes rather than rejecting it. The template is cheap to publish and the transaction is cheap to submit; there is no fee ceiling involved, since the process dies before accounting completes. Consensus-relevant: every validator in the committee executes the same transaction.

## Possible directions

Not obviously a one-liner, hence a separate issue rather than folding it into #2440:

1. **Depth counter on the host boundary.** Count nested `tari_engine_entrypoint` entries per `WasmProcess` and refuse past a small bound (1 is arguably correct — an engine call should not be reachable from inside the servicing of another). Cheap, and independent of what the recursion is made of.
2. **Take response allocation away from the template.** Have the host write responses into a region it owns rather than calling back into `tari_alloc`. Removes the cycle entirely, but changes the ABI contract.
3. **Guard the alloc specifically.** Refuse engine calls while inside a host-driven `tari_alloc`, in the same style as #2440's invocation window. Narrower than (1) but does not bound other re-entrant shapes if any exist.

(1) looks like the best value for the risk.

## Related

- #2440 — refuses engine calls from `tari_alloc`/`tari_free` when they run *outside* an invocation. Necessary, and it closes the same crash on the pre-invocation `CallInfo` allocation, but by design it cannot cover this path.

sdbondi:
Closes #2442, which has the mechanism and the impact.

## Why option 3 and not option 2

#2442 leaned to option 2 — take response allocation away from the template entirely. I designed it
out, and it does not survive contact:

- **The region cannot be sized.** `ComponentManager::get_state()` returns the *whole* component
  state, and the generated dispatcher issues it on every method call, so a host-owned region is
  sized by the largest state any template holds, on the hottest path. A template's entire linear
  memory is 2 MiB (`max_memory_pages = 32`), and `max_substate_size` is 1 MiB, so preserving today's
  semantics means handing half the memory budget to a buffer.
- **The escape hatch does not work.** A per-template `#[template(io_region_kib = N)]` was the answer
  to that, with a write-side check to stop a template writing state larger than the region that has
  to read it back — otherwise a component is written successfully and then permanently uncallable.
  But `GetState` can read *another* template's component, so the writer's region is not the
  reader's. The check has to be against a global ceiling, which puts a single fixed cap straight
  back and leaves the knob buying nothing.
- **Template-supplied buffers need host-side state.** The alternative — template passes the
  destination, host writes into it — requires the template to size the buffer before it knows the
  response length, and engine ops cannot be re-run to find out. The host has to hold the encoded
  response between two calls. Multipart is the same stash with more round trips.

So option 2 as scoped would have shipped a knob that does not knob, a new consensus check, and a
hard ABI break requiring every template to be republished — to remove a class of bugs that option 3
closes for three lines.

**The performance argument does not carry it either.** I measured the response allocation across the
engine suite: 6,341 of them, median 7 bytes, p99 692, max 1,964 — nothing near any plausible region
size. Each costs ~126 metered points, which is **~1% of a transaction's WASM points** (0.4-1.2%
across sampled transactions), or ~2% counting the matching free. Real, but not what an ABI break is
for.

What option 3 does not do is retire the class: the engine still calls template-supplied `tari_alloc`
and `tari_free`, and the invariant lives in three hand-written brackets rather than at the boundary.
Worth its own issue if it is picked up later — the sizing data above says the right shape is a small
region (4 KiB covers everything observed) with a fallback, not a region sized for the worst case.

## Implementation

`handle` writes each engine call's response through the template's `tari_alloc`, so servicing a call
runs template code while an invocation is in flight. That allocation moves into `alloc_response`,
which shuts the invocation window around it:

```rust
let was_open = env.data_mut().suspend_template_invocation();
let result = alloc_fn.call(&mut *env, len);
env.data_mut().restore_template_invocation(was_open);
take_refused_engine_call(env.data_mut())?;
```

The window is restored rather than reopened, so the bracket is correct wherever it is called from
rather than only behind the entrypoint guard.

Same mechanism #2440 used for the two entry points that already ran outside an invocation, so all
three places the engine drives template code are now closed to engine calls.

`handle` takes the `FunctionEnvMut` rather than a pre-split `(&mut WasmEnv, StoreMut)`. It needs to
mutate the environment either side of a call that re-enters WASM, and the environment must not stay
borrowed across that call — the refusal is recorded through the engine's own `&mut` to it. The
dispatch sites in `tari_engine_entrypoint` pass `&mut env`, and the guard/meter split is scoped to a
block ahead of them.

`WasmEnv::alloc` is gone; `alloc_response` is the only caller left and it needs the borrow split.

## Refusals are drained on both paths

A refusal now fails the call before the `Ok`/`Err` match rather than only on the trap path.
`tari_engine_entrypoint` can answer a refused call only with a null pointer, and a template is free
to ignore that and return normally — which is exactly what the new test template does. Checking only
the trap path let the refusal be dropped and the transaction commit.

This also picks up the case flagged in review on #2440: an engine error raised mid-invocation and
swallowed by a template no longer disappears.

## Test

`test_engine_call_in_response_alloc` adds a third `buggy` variant whose `tari_alloc` calls the engine
once an invocation is under way — a static flag skips the `CallInfo` allocation, which #2440 already
refuses before the invocation begins, so the response allocation is the one that runs. It ignores the
null it gets back and returns normally.

Verified load-bearing: with the bracket removed the test aborts the process with
`fatal runtime error: stack overflow` (SIGABRT). With it, the transaction rejects with
`EngineCallOutsideInvocation`.

`cargo test -p tari_engine` passes in full (36 suites); `cargo lints clippy` clean.

## Breaking

Consensus-affecting: a transaction whose template calls the engine from `tari_alloc` during an
invocation previously crashed the node and now rejects.


sdbondi (commented):
Checked out the branch, traced the control flow, and ran the tests. The fix is correct, and I verified the new test is load-bearing: with the `exit`/`enter` bracket removed from `alloc_response`, `test_engine_call_in_response_alloc` aborts with `fatal runtime error: stack overflow` (SIGABRT); with it, all three guard tests pass.

## What I verified

- **Coverage is complete.** Four places the engine drives WASM: `alloc_checked` (`CallInfo`, before the window), `free_checked` (return ptr, after it), `alloc_response` (new), and `func.call` itself. Grepping `.call(` in `process.rs` finds no fifth, and `with_alloc_and_mem_writer` is only used for the `CallInfo`.
- **`enter_template_invocation()` after the alloc restores rather than sets**, because `handle` is only reachable past the entrypoint guard at `crates/engine/src/wasm/process.rs:250`, which returns null when the window is closed.
- **Nesting cannot corrupt the bool.** `invoke_template` (`crates/engine/src/transaction/processor.rs:939`) builds a fresh `Store` + `WasmProcess` per cross-template call, so each `WasmEnv` sees exactly one invocation.
- **Draining `last_engine_error` on both paths is strictly additive.** `template_abi::call_engine` panics on a null response (`crates/template_abi/src/abi/mod.rs:67`), so a swallowed engine error only arises for a template that ignores the ABI. No legitimate flow is newly rejected.
- **Error precedence survives the hoist.** The engine error is still checked ahead of the panic message and the out-of-gas classification. An out-of-gas trap inside `alloc_response` surfaces as `WasmExecutionError::WasmRuntimeError`, not `RuntimeError`, so it is not stored in `last_engine_error` and `FeeIntentComputeExceeded`/`InsufficientFeesForCompute` still win in the `Err` arm.
- **Hardening not called out in the description:** `len as u32` became `u32::try_from`. The old `WasmEnv::alloc` call site truncated a `usize` from `encoded_len`, so a >=4 GiB response would have allocated a truncated buffer that `encode_into_writer` then wrote past. Unreachable in a wasm32 process, but good to have closed.

## Worth resolving before merge

**1. This implements option 3, but the issue leans to option 2.** #2442 lists three directions, and the comment there is "leaning to option 2 because that is the more correct way (removes a class of bugs and slightly reduces WASM size to boot), we can tolerate ABI breaking changes at this stage." This PR does option 3 (guard the alloc) and closes the issue. Option 3's own stated downside is that it "does not bound other re-entrant shapes if any exist" — the invariant now lives in three near-duplicate hand-written brackets (`alloc_checked`, `free_checked`, `alloc_response`) rather than at the boundary, so a fourth host->WASM call site added later has to remember. Either keep #2442 open as the follow-up for option 2, or record in the PR body why option 3 won. As written, the decision is lost when the issue auto-closes.

## Minor

**2. `handle` is `pub` (`crates/engine/src/wasm/process.rs:361`) but the bracket assumes it is not reachable outside the window.** Nothing outside `process.rs` calls it (grepped `crates/` and `applications/`). Since the signature is changing anyway, making it private — or saving and restoring the flag instead of the unconditional `enter_template_invocation()` — turns "safe because of a guard three functions away" into "safe by construction".

**3. `take_refused_engine_call` at `crates/engine/src/wasm/process.rs:512` cannot fire.** A refusal is only recorded while the window is closed, and every site that closes it (`alloc_checked`, `free_checked`, `alloc_response`) drains it before returning, including on the trap path since the drain precedes `result?`. In the new test the refusal reaches `invoke` as `last_engine_error`, not as a pending refusal. Harmless belt-and-braces, but the comment above it ("A refusal or engine error recorded during the invocation fails the call on both paths") reads as if the refusal arm is live.

**4. rustfmt: three lines now exceed `max_width = 120`** in `crates/engine/tests/templates/buggy/src/lib.rs` (lines 27, 59 and 93, at 121-126 chars). CI will not catch it because that crate declares its own `[workspace]` and root `cargo fmt --all` does not reach it, but the file was inside the limit before.

**5. Module name.** `engine_call_outside_invocation` now hosts a variant that runs *inside* an invocation — the engine just closes the window around it. The doc comment explains this well; the name is the only thing left saying otherwise.

Nothing here blocks except deciding (1).

---

sdbondi (commented):
Reviewed `bf680c1`. All five points from the previous review are addressed, and the option-2 section in the description is substantive — I reproduced the sizing data it rests on. One new minor suggestion.

## Verified on this commit

- **`suspend`/`restore` genuinely generalizes the bracket.** `alloc_response` no longer depends on the entrypoint guard at `crates/engine/src/wasm/process.rs:245` for correctness, which is what I was after.
- **`handle` is private** and nothing outside `process.rs` referenced it; the crate builds.
- **The check reorder in `invoke` is behaviourally identical today.** A refusal recorded while the window is closed is always drained by the site that closed it, including on the trap path since the drain precedes `result?`, so `refused_engine_call` is `None` by the time `invoke` looks. The new test reaches `invoke` through `last_engine_error`, not through the refusal arm. The revised comment now says exactly that rather than implying the arm is live.
- **Module rename is complete.** `grep -rn engine_call_outside_invocation` over the tree is empty, and the header comment at the top of `buggy/src/lib.rs` tracks the new name.
- **No line in `buggy/src/lib.rs` exceeds 120 any more.** The two diffs `cargo +nightly-2025-12-05 fmt --check` still reports in that crate (the two `_ABI_TEMPLATE_DEF` array literals) are both pre-existing on `development`, so this PR now introduces none.
- **`cargo test -p tari_engine`**: 4 test binaries, 37 tests, 0 failures. **`cargo clippy -p tari_engine --all-targets`**: clean — the two `too_many_arguments` warnings come from `crates/engine_types/src/resource.rs:68,99` and are pre-existing.

## On the option-2 rationale

I checked the numbers that carry the argument:

- `max_memory_pages: 32` (~2 MiB) and `max_substate_size: 1024 * 1024` at `crates/engine_types/src/limits.rs:19` and `:128`, so "half the memory budget to a buffer" is accurate.
- The generated dispatcher does issue `component_manager.get_state::<..>()` per method call (`crates/template_macros/src/template/dispatcher.rs:144`), so a host-owned region really would be sized by the largest component state on the hottest path.
- **Reproduced the size distribution.** Instrumented `handle` to record `encoded_len(&resp)` and ran the full engine suite: 6,341 samples, median 7 B, p90 43 B, p99 728 B, max 1,965 B. (I dropped 15 lines torn by concurrent test binaries appending to one file — the two apparent outliers above 2 KiB were both `39` repetitions.) That matches the median 7 / max 1,964 in the description, and confirms 4 KiB covers everything observed with room to spare.

The per-allocation point cost (~126) and the 0.4-1.2% of transaction WASM points I did not reproduce.

## New (minor)

**`#[must_use]` on `suspend_template_invocation`.** The point of this commit is that the bracket is correct wherever it is called from — but a caller that drops the returned bool leaves the window closed for the remainder of the invocation, so every later engine call in it is refused and the transaction rejects. `unused_must_use` is denied in `lints.toml:21`, so the attribute turns that into a build failure instead of a silently wrong rejection. `restore_template_invocation` taking a plain `bool` means nothing else flags it.

---

PR #2447, opened 2026-08-19
feat(walletd): settle the stealth fee estimate from the transfer's shape

sdbondi:
## The cost

A stealth transfer's fee is an input to its own shape. Input selection targets `amount + max_fee`, so
the fee decides which UTXOs are spent and whether any change is left over — and a change output is
another stealth output, another `PER_OUTPUT` of verification, another substate. #2443 resolved that
fixed point by dry-running until a build paid for itself: 2 to 4 network round trips per estimate,
each one re-pricing a build the wallet had already done locally.

## The fix

`MergedStealthTransferShape::estimate_fee` prices that shape without executing it. The wallet daemon
rebuilds locally until the price agrees with the fee the build was made at, then spends a single dry
run confirming it.

On a local swarm, an 8,000 µT transfer at `max_fee: 1`: **2.09s and two dry runs → 0.13s and one.**

The estimate sums the charges the engine takes over a shape — transaction weight, host calls, native
verification, persisted storage, substate creation, and the exhaust burn over their total, in the
order `FeeSource` accumulates them. Nothing in the shape reads the fee amount back: the amount lives
inside the statement, which weighs by input count rather than by the amounts it carries, and
`pay_fee_from_bucket` carries no `max_fee` literal. So the cost is a function of the shape alone.

Rates come from a new `FeeRates` that `FeeTable::to_rates` projects onto. `FeeTable` cannot move to
`tari_engine_types` — `fee_estimate_allowance` reads `LITERAL_BYTE_DIVISOR` from a crate downstream
of it — and a wallet must price a transfer without linking the execution engine.

## Which shapes

Only the one the estimate actually prices: a merged transfer whose statement both moves the funds and
reveals the fee, with no badge proof, no revealed input to withdraw and no recipient owed revealed
funds. Each of those calls into a template, bringing a template load and WASM execution the estimate
does not model. `StealthTransferApi::transfer` reports the shape only when the build is that one, and
every other transfer settles over the wire exactly as before.

## Why the reported figure is the fee the build was made at

A settled estimate answers with the fee its confirming run was **built at**, not the lower figure that
run turned out to need.

Input selection is branch-and-bound over the *total value* selected rather than the input count, so a
target one microtari lower can pick a different and larger set. With UTXOs `{100, 30, 25, 20}`, a
target of 76 takes the 100 alone; a target of 75 takes all three others. Only a fee some build was
actually made at describes the transaction a submission at that fee will produce. The same reasoning
is why the handoff plan's "rebuild at the exact figure over the same locked inputs" is not what
landed: the estimate and the submission are separate RPC calls with separate selections, so no lock
spans them.

`TransactionApi::submit_dry_run_transaction` takes the figure to record as the one to submit with,
and records `required_fees` when there is none — the right answer everywhere the fee does not feed
back into the shape.

## The margin is measured, not padded

The difference between the reported figure and the charge is not refundable on a stealth-revealed
fee, so the estimate measures rather than stands in. `persisted_utxo_bytes` reports what the outputs
a build generated will actually occupy — memo, authorization, viewable balance — leaving the
receipt's epoch as the only field still priced at its widest. Against real executions that is **7 µT**,
and the equivalence test holds it under 12. Reported overcharge is 30 µT, against 25 today.

`max_fee` is a ceiling rather than a price, so the settle converges on the estimate from either side:
a caller passing 1,000,000 is answered with 15,916, not their ceiling.

## The equivalence test is the load-bearing part

A second statement of consensus pricing drifts, and drift downward is the expensive direction — a
builder that settles under the real charge submits a transaction rejected for underpayment, and a fee
intent that fails leaves no checkpoint, so it collects nothing.

`crates/engine/tests/stealth_fee_estimate.rs` prices a matrix of `(1..4 inputs, 1..3 outputs)` twice,
once by the estimator and once by executing it, **revealing exactly the estimate** so the engine
either commits or rejects the transaction as underpaid. A companion case reveals half the estimate to
show the rejection has teeth. `the_runtime_call_count_matches_the_instruction_sequence` pins the
host-call constant to what the engine counts.

The exhaust burn is a consensus constant rather than part of the fee table, so
`exhaust_burn_rate_by_network` restates it where a wallet can reach it without linking the consensus
engine, pinned to `ConsensusConstants` across a sweep of epochs.

## What the engine turned out to charge

Two things the estimate had to get right, neither obvious from the fee module alone:

- A spent UTXO is *downed* — it leaves state rather than being rewritten as spent — so it costs
  verification but neither storage bytes nor a diff-summary entry.
- A merged stealth transfer emits no events at all: `PayFee::FromBucket` does not go through a vault,
  so there is no `std.vault.pay_fee` event and none of the `max_fee`-in-decimal coupling that
  `Event::charged_size_padding` exists for.

## Three defects caught before landing

Two against the swarm, one by `/code-review high`, each fixed in its own commit:

- `transactions_update` writes every column it knows about, so recording the settled fee *after* the
  run wiped the dry run's result and reset its status to `New`. The fee now travels with the run.
- The settle only ever raised the fee, so a caller passing a generous `max_fee` was answered with
  their own ceiling — unrefundably.
- The static and wire settle paths could interleave: a statically settled round that the wire
  disagreed with became the wire loop's `candidate`, which would then answer with a figure a dry run
  had just shown insufficient. The static estimate is now dropped for the rest of a request the first
  time a run disagrees with it.

## Verified on a swarm

Wallet daemon rebuilt and restarted on this branch:

| case | builds | dry runs | reported | cost |
|---|---|---|---|---|
| one recipient, `max_fee` 1 | 2 | 1 | 15,916 | 15,886 |
| one recipient, `max_fee` 1,000,000 | 2 | 1 | 15,916 | 15,886 |
| one recipient, `max_fee` 15,916 | 1 | 1 | 15,916 | 15,886 |
| two recipients, `max_fee` 1 | 2 | 1 | 22,539 | 22,508 |
| revealed payout (not covered) | 2 | 2 | 16,363 | 16,338 |
| `PreferRevealed` (not covered) | 2 | 2 | 9,687 | 9,662 |

Submitting at 15,916 is **accepted**, paid 15,916, with the same fee breakdown the dry run reported —
the same shape, priced the same way.

## Test plan

- [x] `cargo check --workspace --all-targets` clean
- [x] `cargo test -p tari_engine --test stealth_fee_estimate` (3 tests, the shape matrix)
- [x] `cargo test -p tari_engine --test {fees, compute_fee_budget, native_compute_budget, metering_budget, complex_fee_payment}`
- [x] Unit tests for `tari_engine_types`, `tari_ootle_wallet_sdk`, `tari_ootle_walletd`, `tari_ootle_app_utilities`
- [x] End-to-end on a local swarm: estimate → submit → committed (table above)
- [ ] `integration_tests` — CI

## Known limitation

The local settle demands an exact fixed point, which need not exist: selection is not monotone in the
fee, so two fees can each price to the other. Bounded at `MAX_STATIC_SETTLE_ROUNDS` rebuilds — tens of
milliseconds of range-proof generation, still inside a single round trip — after which it settles over
the wire as before. Loosening the rule would terminate on the first pass but costs overcharge, which
is the thing being optimised.


applications/tari_walletd/src/handlers/accounts.rs
```diff
@@ -1356,11 +1367,43 @@ pub async fn handle_stealth_transfer(
...
+                    if estimate != params.max_fee && static_rounds < MAX_STATIC_SETTLE_ROUNDS {
+                        params.max_fee = estimate;
+                        static_rounds += 1;
+                        lock.release();
+                        continue;
+                    }
+                    // Out of local rounds without agreement, the fee still stands if it covers the
+                    // estimate; if it does not, the rounds below settle it over the wire.
+                    settled_statically = estimate <= params.max_fee;
+                }
```
sdbondi, line 1397: This block can run after a wire round has already set `candidate`, and when it does it invalidates it.

`settle_statically` is cleared only in the `Retry` arm, and only when `settled_statically` was true — so a round whose build simply had no `statically_priced_shape` retries with the static path still live. If a later round's build *is* merged, this block runs, moves `params.max_fee` off the figure the candidate named, and rebuilds. The wire path's invariant — "`candidate` is the round that named the fee the current round was built at" — no longer holds, and the two `return`s below both act on the broken version (see the separate comments).

That sequence is reachable exactly in the generous-`max_fee` case the PR measures: round 1 at `max_fee` 1,000,000 pulls in revealed inputs, so `inputs_to_spend.revealed > 0` and the shape is `None`; the retry at `required_fees` ≈ 16k selects a merged build, and this block then settles it statically.

Dropping the static path here would lose the optimization for precisely that case, so the fix belongs on `candidate`: it should be cleared whenever this block changes the fee, since nothing named the new one.

```rust
if estimate != params.max_fee && static_rounds < MAX_STATIC_SETTLE_ROUNDS {
    params.max_fee = estimate;
    static_rounds += 1;
    // No wire round named this fee, so nothing carried over describes the build about
    // to be made at it.
    candidate = None;
    lock.release();
    continue;
}
```

---

applications/tari_walletd/src/handlers/accounts.rs
```diff
@@ -1381,9 +1424,17 @@ pub async fn handle_stealth_transfer(
...
+                    charged,
+                    required_fees,
+                    candidate.is_some() || settled_statically,
+                    rounds,
+                ) {
                     FeeEstimateStep::Settle => {
-                        return Ok(candidate.expect("BUG: Settle is only reached with a candidate"));
+                        // Without a candidate the fee was settled locally, and this round is the one
+                        // that confirmed it, so its own response is the answer.
+                        return Ok(candidate.unwrap_or(response));
```
sdbondi: `unwrap_or` prefers the candidate, but a statically settled round is the one that confirmed its own fee, so its response is the answer even when a candidate exists.

The comment reads "without a candidate the fee was settled locally", which assumes the two are mutually exclusive. They are not — see the static block above: a round settled statically can follow a wire round that set `candidate`. Then this returns round 1's transaction id, whose stored `final_fee` is round 1's `required_fees`, describing a build made at a different `max_fee` with a different shape. The statically settled run that was actually confirmed gets thrown away, and since selection targets `amount + max_fee`, submitting at the returned figure can pick a UTXO set nothing dry-ran.

Clearing `candidate` in the static block fixes the reachable path, but the condition here is worth making explicit rather than leaning on `candidate` being `None`:

```rust
FeeEstimateStep::Settle => {
    // A statically settled fee was confirmed by this very round, so this response names
    // it. Otherwise the candidate is the round that named the fee this one was built at.
    if settled_statically {
        return Ok(response);
    }
    return Ok(candidate.expect("BUG: Settle is only reached with a candidate or a static settle"));
}
```

---

applications/tari_walletd/src/handlers/accounts.rs
```diff
@@ -1392,11 +1443,25 @@ pub async fn handle_stealth_transfer(
                              at {} but cost {charged})",
                             params.max_fee
                         );
+                        // Reachable only after `MAX_FEE_ESTIMATE_ROUNDS` rounds, by which point a
+                        // retry has dropped the static path, so this response records
+                        // `required_fees` — a figure above what the round cost — rather than a
+                        // settled fee the round disagreed with.
                         return Ok(response);
```
sdbondi: The guarantee this comment rests on doesn't hold, and the failure mode is the expensive direction — a reported fee below the charge.

A retry drops the static path only when `settled_statically` was true for that retry. All four preceding retries can have `settled_statically == false` (no `statically_priced_shape` on that build, or the static rounds exhausted with `estimate > max_fee`) while `settle_statically` stays true, so the static path can still be live in round 5.

If round 5 is settled statically, `verified` is true, so reaching `GiveUp` at all requires `charged > built_at`. But `settled_fee = Some(params.max_fee)` was already written by `submit_dry_run_transaction_settled_at`, so the row this response points at records `params.max_fee` while the run cost more. A submission at the reported figure is rejected as underpaid, and the revealed fee is unrecoverable.

Getting here means the estimator priced under the engine, which is what `stealth_fee_estimate.rs` exists to prevent — so this is the drift backstop rather than an everyday path. That's an argument for making it loud, not for leaving it: as written, drift that slips past the equivalence test surfaces as a silently under-reported fee. `GiveUp` shouldn't return a figure a run has just shown insufficient — re-record `required_fees` before returning, or fail the request outright, and log that the static estimate came in under the charge.

---

sdbondi (commented):
Reviewed the pricing model line-by-line against the engine rather than against the description.

**What checks out**

- Charge order and rounding match `FeeModule`: `FeeRates::storage_cost` multiplies then divides, `execution_cost` divides then multiplies — the same way `charge_state_fees` / `charge_finalization_fees` do it — and the burn is taken last over the accumulated total.
- `initial_cost` is `0` at every production construction site (`transaction_executor.rs:105`), so omitting `FeeSource::Initial` is right.
- `num_outputs + 1` creates matches `count_newly_created_substates().saturating_add(1)`, and downing spent UTXOs justifies inputs contributing verification but no storage bytes.
- `persisted_utxo_bytes` reconstructs the stored form — commitment moved into the address, viewable balance standing in for its proof at equal fixed width — rather than measuring the statement output, which is the right thing to measure.
- All `submit_dry_run_transaction` call sites are updated; the handle keeps its old arity by delegating with `None`, so the other five callers are untouched.
- `exhaust_burn_rate_by_network` is pinned to `ConsensusConstants` across a network × epoch sweep, and the estimator/execution equivalence matrix in `crates/engine/tests/stealth_fee_estimate.rs` — with a companion case proving the rejection has teeth — is the right load-bearing choice for a second statement of consensus pricing.

**What needs changing**

Three comments below, all in the walletd settle loop and all from one root: `settle_statically` is cleared only by a `Retry` taken while `settled_statically` was true, so the static and wire paths are **not** mutually exclusive within a request the way the comments assert. A round can be settled statically after earlier wire rounds — reachable in the generous-`max_fee` case, where round 1 pulls in revealed inputs and has no shape, and the retry selects a merged build.

From that: `Settle` can answer with a stale `candidate` naming a fee no confirmed build was made at, and `GiveUp` can answer with a fee below what the run was charged. The first is the reachable one; the second is a drift backstop that currently fails silently in the direction the design says it must not.

---

applications/tari_walletd/src/handlers/accounts.rs
sdbondi: Fixed in 5eb65d44b — `candidate = None` in the static block, as suggested. You're right that dropping the static path here would lose the optimisation for exactly the case worth optimising.

The reachable sequence is slightly different from the one described: with `PreferRevealed`, `inputs_to_spend.revealed` is `min(revealed_balance, spend_amount)`, so it is positive at every `max_fee` once the account holds any revealed funds — it doesn't flip between rounds. What does flip is the merge: a wider `max_fee` selects more inputs, and past `merged_statement_fits_credit` the build falls back to a separate fee statement and loses its shape, while the retry at a lower figure gets it back. Same hole, same fix.

applications/tari_walletd/src/handlers/accounts.rs
sdbondi: Fixed in 5eb65d44b, with the explicit condition rather than leaning on `candidate` being `None`:

```rust
if settled_statically {
    return Ok(response);
}
return Ok(candidate.expect("BUG: Settle needs a candidate or a static settle"));
```

The `expect` is sound now: `verified` is `candidate.is_some() || settled_statically`, so reaching `Settle` with `settled_statically` false requires a candidate.

applications/tari_walletd/src/handlers/accounts.rs
sdbondi: Fixed in 5eb65d44b, made loud and structural rather than defensive.

A statically settled round the run disagrees with is diverted before the step function sees it — logged as the estimator having priced under the engine, static path dropped for the rest of the request, `candidate` cleared, fee set to `required_fees`:

```rust
if settled_statically && charged > params.max_fee {
    warn!(... "Static stealth fee estimate came in under the charge" ...);
    settle_statically = false;
    candidate = None;
    params.max_fee = required_fees;
    continue;
}
```

So `settled_statically` now implies `charged <= params.max_fee`, which with `verified` implies `Settle`. `Retry` and `GiveUp` are unreachable with a statically settled round, and neither can answer with its row. That also removes the special-casing the `Retry` arm had grown.

I went with diverting rather than re-recording `required_fees`: a second write would go through `transactions_update`, which sets every column and would wipe the run's result and status — the bug fixed in 9896b3942.

applications/tari_ootle_app_utilities/src/fee_tables.rs
```diff
@@ -181,25 +182,70 @@ pub const fn get_fee_table_by_network(network: Network) -> &'static FeeTable {
...
+            let consensus = tari_consensus::consensus_constants::ConsensusConstants::from(network);
+            for epoch in [0, 1, 2, 100, 10_000, 1_000_000, u64::MAX] {
+                assert_eq!(
+                    exhaust_burn_rate_by_network(network),
+                    consensus.exhaust_burn_rate(tari_ootle_common_types::Epoch(epoch)),
+                    "{network} at epoch {epoch}"
+                );
+            }
+        }
+    }
```
sdbondi, line 244: Nit, on the comment rather than the code: this is a sample of seven epochs, not a sweep, and the name and doc both claim otherwise.

`ConsensusConstants::exhaust_burn_rate` is `pub fn exhaust_burn_rate(&self, _epoch: Epoch) -> ExhaustBurnRate` (`consensus_constants.rs:297`) — the parameter is unused and the body returns a plain field. So today the loop asserts the identical thing seven times, and if the rate ever does start varying by epoch it is caught only when it happens to differ at one of these seven points.

The comment this replaced named the actual remedy:

> The rate does not vary by epoch today; if it starts to, this restatement needs an epoch of its own rather than a looser assertion here.

That is the load-bearing observation, and it holds no matter how densely the test samples: `exhaust_burn_rate_by_network` takes only a `Network`, so an epoch-varying consensus rate breaks the restatement's signature, not just its value. Worth keeping.

Keep the extra epochs — they're free and they do catch a `match epoch` whose arms split at the sampled points — but I'd re-title to something like `the_burn_rate_matches_consensus_across_epochs`, say sampled rather than swept, and restore the note about the restatement needing an epoch of its own.

Also, the doc on `exhaust_burn_rate_by_network` (line 189) still points at `the_burn_rate_matches_consensus`, which no longer exists under that name.

---

sdbondi (commented):
Reviewed `95eef80` (`keep the static and wire settle paths from sharing a request`) on its own.

LGTM

One nit on the epoch sweep below.

applications/tari_ootle_app_utilities/src/fee_tables.rs
sdbondi: Fixed in 403c62261 — all three points.

Renamed to `the_burn_rate_matches_consensus_across_epochs`, the doc now says sampled rather than swept ("enough to catch a rate that splits at one of them, not a proof that none exists"), and the note you quoted is restored, with the reason made explicit: `exhaust_burn_rate_by_network` takes only a `Network`, so an epoch-varying rate breaks the restatement's signature rather than just its value.

The stale reference on `exhaust_burn_rate_by_network` is updated to the new name.

PR #2454, opened 2026-08-20
feat(engine): render dispatcher decode panics from the template definition

sdbondi:
## Description

Macro-generated dispatch code panicked with the full text of a decode failure:

```
failed to decode argument at position 2 (Amount) for function 'withdraw': {}
```

That is one unique string per function and per argument. Every byte is replicated to each validator and paid for at publish time — and every byte of it except the underlying decode error is already known to the engine from the `FunctionDef` it is invoking.

The dispatcher now panics with a compact marker and the engine renders the message from its own copy of the definition:

- template emits `"\u{1}A2: {}"` (5 bytes) instead of the 74-byte string;
- `WasmProcess::invoke` expands it at the one site that pops the panic message — `func_def` was already a parameter, so there is no new plumbing;
- whether the argument is a tuple is derived from the definition's `Type`, so the "tuple argument" wording needs no payload.

### Sizes

Measured on published (`wasm-opt`'d) binaries:

| Template | Before | After | Δ |
|---|---|---|---|
| `account` (21 fns) | 275,362 | 268,671 | **−6,691 (−2.4%)** |
| `state` (5 fns) | 79,433 | 79,120 | −313 (−0.4%) |
| `hello_world` | 71,779 | 71,567 | −212 (−0.3%) |

The saving exceeds the 4,567 bytes of string text `account` carried, because the format arguments those strings interpolated go with them. It scales with function × argument count, so it pays most on exactly the templates that are large enough to attract the publish premium.

### Compatibility

Non-breaking in both directions:

- A message carrying no marker — a template author's own `panic!`, or any template published before this encoding — is passed through untouched.
- A marker whose index does not fit the function being invoked is also passed through, so a forged marker degrades to the literal text rather than a misleading message.
- The reject text is local diagnostic data, never consensus state: consensus carries `Decision::Abort(AbortReason)` (a plain enum), `Decision::Abort(_)` compares equal regardless of reason, and `RejectReason` — the variant holding the `String` — does not appear in `consensus_types` at all. Two validators rendering the message differently cannot disagree on a transaction.

Forward compatibility is a release-ordering matter rather than a code one: validators get the engine before `tari_template_macros` 0.22.1 is published, so no template can emit a marker before engines can render it. Operators who lag behind that release see a decode failure as `\u{1}A2: <cbor error>` — the tag, the argument position and the underlying error, without the function name or argument type. That is a less descriptive message on a transaction that fails identically either way, and it is documented on the module.

Gating the ABI/template version instead would make an old engine reject the template outright rather than report it tersely, which is a worse failure than a terse message.

During a rolling upgrade, mixed-version validators render different text for the same failure. That is cosmetic: the text never enters consensus, per the reasoning above.

## Note on message wording

`Type`'s `Display` rendered tuple members with `{:?}`, which leaked `Other { name: "X" }` into any message naming a tuple type. It now renders them with `Display`.

One consequence: types in these messages are now spelled in ABI terms. A tuple argument reads `Tuple<String,U32>` where the macro used to emit the Rust token stream `(String, u32)`. Same facts, different spelling. This also updates an existing assertion in `test_tuples`.

## Robustness of the parse

The panic message is chosen by the executing template, so it reaches `expand_panic_message` as attacker-controlled input at a consensus boundary — a panic there would take down the executing validator.

What guarantees the input's shape before it gets there: `on_panic_handler` reads it through `WasmEnv::with_memory_slice`, which bounds-checks the WASM pointer and length (`checked_add` + `slice.get(start..end)`, both returning `Err`, never panicking), caps it at `ENGINE_LIMITS.max_panic_message_size` (32 KiB) truncated to a char boundary, and runs it through `from_utf8_lossy`. So the expander always sees valid UTF-8 of bounded length.

The parse itself uses only `strip_prefix` / `strip_suffix` / `split_once` and `str::parse` — no byte indexing, no `split_at`, no slicing — so char boundaries are safe by construction and integer overflow lands on `Err`. `render` bounds-checks the index with `arguments.get(index)`.

`fuzz/fuzz_targets/panic_diagnostic_expand.rs` covers the sink, asserting more than "does not panic": marker-less messages come back byte-for-byte, output is either verbatim or a rendered diagnostic, an out-of-range index never renders, and output is bounded by input + the definition's own text so nothing can amplify allocation. **1.66M executions, no findings** (256 edges, 94 corpus entries). Two unit tests cover the same sink on every build without cargo-fuzz.

One behaviour is accepted rather than prevented: a template can deliberately panic with a marker and have the engine render an authentic-looking decode failure. This grants nothing new — a template could always `panic!` with the full text verbatim — and the consequence is a misleading diagnostic on a transaction that rejects either way. The control-character prefix is what makes *accidental* collision impossible.

## Test plan

- 7 new unit tests in `tari_template_abi::diagnostics` — marker round-trip, pass-through for unmarked messages, out-of-range index, unknown tag, and all three renderings.
- `errors::invalid_args`, which asserts `"Panic! failed to decode argument at position 0 (Amount) for function 'please_pass_invalid_args'"`, **passes unmodified**: the message is byte-identical, produced by a template that no longer contains it.
- `errors::panic` still passes, covering pass-through of an author's own panic text.
- New `errors::invalid_tuple_arg` covers the tuple wording end to end.
- 17 unit tests in `tari_template_abi::diagnostics`, including the crafted-shape table and the deterministic hammer described above.
- `fuzz/run_all.sh panic_diagnostic_expand` — 1.66M executions, no artifacts.
- `no_std` template builds; `cargo clippy` clean on `tari_template_abi`, `tari_template_macros` and `tari_engine`.

Builtin templates are compiled from source by `template_builtin/build.rs`, so they pick up the new dispatcher automatically.

## Not included

`tari_template_lib`'s own `expect("failed to decode Bucket")`-style strings in `vault.rs` / `engine.rs` (~100 bytes in `account`) are left alone — they are not dispatcher output and the engine has no context from which to reconstruct them.

## Versioning

`tari_template_abi` 0.19.0 → **0.19.1** and `tari_template_macros` 0.22.0 → **0.22.1**, both patches. `crate_versioning.py impact tari_template_abi` confirms a patch: the change is additive (`pub mod diagnostics`), nothing is removed or re-signed, and dependents pick it up through their `^0.19` pins without republishing.

`tari_template_macros` must move because its dispatcher changed and 0.22.0 is already live on crates.io: `publish_crates.py` skips published versions, so the new macros would never ship and template authors would keep embedding the full strings. `tari_engine` needs nothing — its published max is 0.39.0 against the workspace's unpublished 0.39.1, so the engine half ships with the pending tier-3 release.

The workspace pin moves from `0.19` to `0.19.1` in the same commit. `tari_template_macros` and `tari_engine` now *require* `diagnostics`, and `^0.19` would let a fresh resolve pick 0.19.0, which does not have it. Still a compatible range for everyone else.


Cargo.toml
```diff
@@ -142,7 +142,7 @@ tari_ootle_template_metadata = { path = "crates/template_metadata", version = "0
 tari_ootle_template_provider = { path = "crates/template_provider" }
 tari_template_lib = { version = "0.31", path = "crates/template_lib" }
 tari_template_lib_types = { version = "0.31", path = "crates/template_lib_types", default-features = false }
-tari_template_abi = { version = "0.19", path = "crates/template_abi", default-features = false }
+tari_template_abi = { version = "0.19.1", path = "crates/template_abi", default-features = false }
 tari_template_macros = { version = "0.22", path = "crates/template_macros" }
```
sdbondi: `tari_template_macros` changed materially in this PR (`dispatcher.rs` now emits `\u{1}` markers instead of full messages), but its version stays at 0.22.0 — which is already published on crates.io. `scripts/publish_crates.py` skips already-published versions ("already published, skipping"), so on the next release the changed macros would silently never ship: external template authors would keep the old macros that embed full panic strings, and the size reduction this PR exists for wouldn't reach published templates. The workspace also now pins `tari_template_abi` 0.19.1 while the published 0.22.0 macros were built against 0.19.0, so workspace source and the published crate diverge.

Per `scripts/crate_versioning.py`, the macros crate needs a bump to 0.22.1 in this PR (and this pin updated to match).

---

crates/template_macros/src/template/dispatcher.rs
```diff
@@ -107,10 +107,10 @@ fn get_function_block(template_ident: &Ident, ast: &FunctionAst, stateless: bool
...
     let mut args: Vec<Expr> = vec![];
     let mut stmts = vec![];
-    let func_name = &ast.name;
-
-    let error_failed_decode_component =
-        format!("failed to decode component instance for function '{}': {{}}", func_name);
+    // The engine renders these markers into their full text from the template definition, so the
+    // published binary carries neither the function name nor the argument type. See
+    // `tari_template_abi::diagnostics`.
+    let error_failed_decode_component = PanicDiagnostic::ComponentDecode.panic_format_string();
```
sdbondi, line 113: Templates built with the new macros bake the raw `\u{1}` marker into the binary as their panic text, and expansion to readable text happens only in engines that have `expand_panic_message`. Nothing gates a new-macro template from running on an older engine: `version::LATEST_TEMPLATE_VERSION` / `MINIMUM_SUPPORTED_WASM_ABI_VERSION` weren't bumped, so an engine ≤ 0.39.0 accepts the template and a decode failure surfaces as the reject reason `\u{1}A1: <cbor error>` — a control character with no function name or argument type, strictly worse than the message the old macros embedded. During a rolling upgrade, validators on different engine versions would also derive different reject-reason strings for the same failing transaction.

The PR's backward-compat story covers old templates on the new engine, but not this direction. Worth either bumping the ABI/template version gate or accepting and documenting the degraded message on old engines.

---

sdbondi (commented):
Reviewed the diff against its base. The core design holds up well under verification: marker indices align 1:1 with `FunctionDef::arguments` (receiver included), old templates pass through the new engine unchanged, malformed or crafted markers (`\u{1}` alone, `\u{1}C4`, out-of-range index) degrade gracefully to literal passthrough, truncation at `max_panic_message_size` preserves the leading marker, the migration path reuses the correct `ComponentDecode` marker, and the `Type::Tuple` Display change has no consensus/hashing consumers. Unit tests pass and `tari_engine` type-checks; the `--no-default-features` compile failure predates this PR.

Two findings, inline: a crate-versioning gap (`tari_template_macros` changed but stays at the already-published 0.22.0) and a forward-compat gap (new-macro templates on pre-`expand_panic_message` engines surface the raw marker, with no version gate preventing it).

---

Cargo.toml
sdbondi: Confirmed and fixed in 22228a4 — `tari_template_macros` 0.22.0 is indeed live on crates.io (checked the registry), so the new dispatcher would have been skipped at publish time and template authors would have kept embedding the full strings this PR exists to remove.

Bumped to 0.22.1 with the workspace pin to match. `tari_engine` needs nothing: its published max is 0.39.0 against the workspace's unpublished 0.39.1, so the engine half ships with the pending tier-3 release.

crates/template_macros/src/template/dispatcher.rs
sdbondi: Valid, and fixed in 22228a4 by making the marker legible rather than by gating the version.

Bumping the ABI/template version would have been the wrong lever: it would make old engines reject the template outright rather than merely report it badly, which is a far worse failure than a poor error message. And the window is not hypothetical — published `tari_template_lib` 0.31.0 pins `^0.22`, so every fresh template build picks up the new macros the moment 0.22.1 lands, with no author action, while validators upgrade on their own schedule.

So the marker body now says what failed in words: `\u{1}arg 2 decode: <cbor error>` and `\u{1}state decode: <cbor error>`. An engine without `expand_panic_message` surfaces a readable sentence missing only the function name and argument type — the two things an engine supplies from its own `FunctionDef`. The leading control character stays, so an author's own panic text still cannot be mistaken for a marker.

It costs 48 bytes on `account`, not the ~560 the per-site arithmetic suggests: only a handful of distinct literals exist across all 56 sites and they get merged. Net saving there is 6,643 bytes (-2.4%) against the pre-PR baseline. Added a unit test that pins exactly what an un-upgraded engine shows.

On the rolling-upgrade divergence: worth stating that it is cosmetic only. Consensus carries `Decision::Abort(AbortReason)`, a plain enum; `RejectReason` — the variant holding the `String` — never enters `consensus_types`, and `Decision::Abort(_)` compares equal regardless of reason. Mixed-version validators cannot disagree on a transaction, only on the text they log for it.

sdbondi: Reverting the legible-marker part of this — the reasoning above doesn't hold up. `\u{1}state decode` is no more self-explanatory than `\u{1}C` to anyone reading it cold, so those 48 bytes bought legibility that wasn't really there. Markers are back to `\u{1}A{index}` / `\u{1}C` in 7e2f662, restoring the full 6,691-byte saving on `account`.

What actually closes this finding is release ordering: validators get the engine before `tari_template_macros` 0.22.1 publishes, so no template can emit a marker before engines can render it. Operators lagging that release see `\u{1}A2: <cbor error>` — position and underlying error, minus the function name and argument type — on a transaction that fails identically either way. The module doc now records that, so someone debugging a terse reject reason from an older validator can find it.

PR #2458, opened 2026-08-21
fix(state-store): scope block diff queries to the queried branch

sdbondi:
## Description

`block_diffs_get_last_change_for_substate` and `block_diffs_get_change_for_versioned_substate` filtered candidate diff entries with `applicable_blocks.contains(block_id)` — the query's own argument, which is trivially a member of its own pending chain — instead of `key.block_id`, the block that recorded the entry. The branch filter was a no-op: the `(SubstateId, BlockId, Version, Seq)` index scan accepted diffs from every stored block, including forked-out siblings.

`PendingSubstateStore::{get, get_latest_version, get_latest_change_from_store}` are built on these queries, so pending substate state from an abandoned branch leaked into the evaluation of blocks on the surviving branch.

Both queries now match on the entry's own block id. They also order two changes of the same substate by `(version, is_down)` rather than by `version` alone, so a DOWN supersedes the UP of the same version instead of the winner depending on block id iteration order.

## Motivation and Context

This is the root cause of the flaky `consensus_tests::catch_up_rewind::catch_up_rewind_below_leaf_recovers` ([failing job](https://github.com/tari-project/tari-ootle/actions/runs/32454421433/job/96689191530)). In that run:

- h2 block `c2eece80` (justify h1) carried all 5 test transactions as `LocalOnly Commit`; its block diff, downing the inputs, was persisted on vote.
- It got no QC. A TC formed and h3 `a2b9792d` was built on a **dummy** h2 block, so `c2eece80` was forked out and the transactions were correctly re-proposed.
- Replicas evaluating h3 nonetheless hit `Lock failure: Substate component_00d248…:0 is DOWN` — the down from the forked-out block — and voted no (`Decision disagreement. Local: Abort(LockInputsOutputsFailed), Remote: Commit`).
- h4 resolved the same leaked DOWN into `RejectReason::SubstateNotFound` and proposed all 5 as `Abort(OneOrMoreInputsNotFound)`, which was certified at h5 and committed.

The test's `has_committed_substates` check could then never pass, and it panicked at the height budget with the misleading "catch-up view-rewind deadlock" message even though the target had re-joined consensus (HighPC 73 vs network height 71). The tight 2s `pacemaker_block_time` makes that early fork likely on a loaded CI runner.

Beyond the test, any timeout fork on a live network could spuriously and permanently abort transactions on the surviving branch.

Introduced in 45b633b9ae (#1340).

## How Has This Been Tested?

Two new tests in `crates/state_store_rocksdb/tests/block_diffs.rs`:

- `block_diffs_are_scoped_to_the_queried_branch` — writes a diff into a sibling of `chain[9]` and asserts both queries do not see it from `chain[9]`, but do from the fork. Fails on `development`, passes here.
- `block_diffs_last_change_prefers_the_down_of_a_version` — an UP of v0 in one branch block and a DOWN of v0 in its child; both queries must return the DOWN.

Full `cargo test -p tari_state_store_rocksdb` suite passes locally.

Revert the two `applicable_blocks.contains(&key.block_id)` lines to `contains(block_id)` and run `cargo test -p tari_state_store_rocksdb --test block_diffs`; the branch-scoping test fails.

## Breaking Changes

- [x] None
- [ ] Other

### DIFF
--- a/crates/state_store_rocksdb/src/reader.rs
+++ b/crates/state_store_rocksdb/src/reader.rs
@@ -1096,7 +1096,13 @@ impl<'tx, TAddr: NodeAddressable + Serialize + DeserializeOwned + 'tx, R: RocksR
         let mut max_change = None::<BlockDiffKey>;
         for result in iter {
             let key = result?;
-            if max_change.as_ref().is_none_or(|c| c.version < key.version) && applicable_blocks.contains(block_id) {
+            if !applicable_blocks.contains(&key.block_id) {
+                continue;
+            }
+            if max_change
+                .as_ref()
+                .is_none_or(|c| block_diff_change_order(c) < block_diff_change_order(&key))
+            {
                 max_change = Some(key);
             }
         }
@@ -1130,18 +1136,27 @@ impl<'tx, TAddr: NodeAddressable + Serialize + DeserializeOwned + 'tx, R: RocksR
         let query = self.db().cf(block_diff::BySubstateIdQuery)?;
         let iter = query.query_prefix_range_key_iterator(Ordering::default(), versioned.substate_id());
 
+        let mut last_change = None::<BlockDiffKey>;
         for result in iter {
             let key = result?;
-            if versioned.version() == key.version && applicable_blocks.contains(block_id) {
-                let change = cf.get(&key, OPERATION)?;
-                return Ok(change);
+            if versioned.version() != key.version || !applicable_blocks.contains(&key.block_id) {
+                continue;
+            }
+            if last_change
+                .as_ref()
+                .is_none_or(|c| block_diff_change_order(c) < block_diff_change_order(&key))
+            {
+                last_change = Some(key);
             }
         }
 
-        Err(StorageError::NotFound {
+        let key = last_change.ok_or_else(|| StorageError::NotFound {
             item: "SubstateChange",
             key: format!("{versioned} in {block_id}"),
-        })
+        })?;
+
+        let change = cf.get(&key, OPERATION)?;
+        Ok(change)
     }
 
     fn proposal_certificates_get(&self, epoch: Epoch, qc_id: &PcId) -> Result<ProposalCertificate, StorageError> {
@@ -2081,4 +2096,10 @@ impl<'tx, TAddr: NodeAddressable + Serialize + DeserializeOwned + 'tx, R: RocksR
 
         Ok(count)
     }
+}
+
+/// Orders two changes for the same substate within a branch. A substate version is only ever DOWNed after it is UPed,
+/// so a DOWN supersedes the UP of the same version.
+fn block_diff_change_order(key: &BlockDiffKey) -> (u32, bool) {
+    (key.version, !key.is_up)
 }
--- a/crates/state_store_rocksdb/src/reader.rs
+++ b/crates/state_store_rocksdb/src/reader.rs
@@ -349,6 +349,44 @@ impl<'a, TAddr: NodeAddressable + Serialize + DeserializeOwned + 'a, R: RocksRea
         Ok(blocks)
     }
 
+    /// Returns the key of the last change to `substate_id` on the branch ending at `end_block`, or `None` if the
+    /// branch contains no change for it. If `version` is given, only changes to that version are considered.
+    ///
+    /// Changes recorded by blocks that are not in the branch (forked-out siblings, other subtrees) are never
+    /// considered. A substate version is only ever DOWNed after it is UPed, so a DOWN supersedes the UP of the same
+    /// version.
+    fn block_diffs_select_last_change(
+        &self,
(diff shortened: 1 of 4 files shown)


crates/state_store_rocksdb/src/reader.rs
```diff
@@ -1130,18 +1136,27 @@ impl<'tx, TAddr: NodeAddressable + Serialize + DeserializeOwned + 'tx, R: RocksR
         let query = self.db().cf(block_diff::BySubstateIdQuery)?;
         let iter = query.query_prefix_range_key_iterator(Ordering::default(), versioned.substate_id());
 
+        let mut last_change = None::<BlockDiffKey>;
         for result in iter {
```
sdbondi: `block_diffs_get_change_for_versioned_substate` lost its early return and now scans the entire `BySubstateIdQuery` prefix for the substate even after finding the winning entry.

Per input-lock check on the consensus hot path (`PendingSubstateStore::get` / `assert_is_up` / `assert_is_down`), the loop visits every stored diff entry for the substate across all blocks — including not-yet-GC'd forked siblings — even after finding a DOWN at the queried version in an applicable block, which nothing can supersede (with version fixed, `block_diff_change_order` reduces to `!is_up`). Adding `if !key.is_up { last_change = Some(key); break; }` restores near-first-hit cost.

crates/state_store_rocksdb/src/reader.rs
```diff
@@ -1130,18 +1136,27 @@ impl<'tx, TAddr: NodeAddressable + Serialize + DeserializeOwned + 'tx, R: RocksR
         let query = self.db().cf(block_diff::BySubstateIdQuery)?;
         let iter = query.query_prefix_range_key_iterator(Ordering::default(), versioned.substate_id());
 
+        let mut last_change = None::<BlockDiffKey>;
```
sdbondi: `block_diffs_get_change_for_versioned_substate` duplicates the entire scan/filter/pick-max/NotFound/`cf.get` body of `block_diffs_get_last_change_for_substate`, differing only in the version filter.

Two ~25-line copies of the branch-scoping + `block_diff_change_order` selection logic must be kept in sync — the very bug this PR fixes existed identically in both, showing the cost is real. A future fix to the ordering or branch filter applied to one silently leaves the other divergent. Factor a shared helper taking `Option<u32>` for the version filter.

crates/state_store_rocksdb/tests/block_diffs.rs
```diff
@@ -97,3 +98,93 @@ fn block_diffs_operations(db: impl StateStore) {
...
+    .unwrap();
+
+    // The changes only exist on the forked-out branch, so they must not be visible from chain[9].
+    assert!(
+        tx.block_diffs_get_last_change_for_substate(chain[9].id(), &substate_id)
+            .optional()
+            .unwrap()
+            .is_none()
+    );
+    assert!(
```
sdbondi, line 139: The branch-scoping test asserts `block_diffs_get_change_for_versioned_substate` is invisible from `chain[9]` but never asserts it is visible from the fork; only the last-change query gets the positive assertion.

A regression in the versioned query's fork-side visibility (e.g. an over-aggressive filter specific to that query) passes this test; the sibling test only exercises the versioned query on the main chain with no fork present, so the fork-positive direction is untested.

sdbondi (commented):
One finding outside the diff:

**crates/storage/src/state_store/mod.rs:207** — The selection contract this PR establishes (branch-scoped, max by version with DOWN superseding UP) is stated only in a private helper comment in the rocksdb backend; the trait methods and BlockDiff wrappers are undocumented.

Callers of `BlockDiff::get_last_change_for_substate` / `get_for_versioned_substate` (crates/storage/src/consensus_models/block_diff.rs:82,90) and any future backend have no stated contract to rely on or implement; the DOWN-supersedes-UP invariant can be unknowingly violated. Low severity — doc-only.

crates/state_store_rocksdb/src/reader.rs
sdbondi: Fixed in 2c7892cef: with the version fixed, the scan returns on the first DOWN found in an applicable block.

crates/state_store_rocksdb/src/reader.rs
sdbondi: Fixed in 2c7892cef: both queries now delegate to `block_diffs_select_last_change(end_block, substate_id, version: Option<u32>)`; each trait method is left with the existence check, the NotFound mapping and the `cf.get`.

crates/state_store_rocksdb/tests/block_diffs.rs
sdbondi: Fixed in 2c7892cef: the test now also asserts the versioned query returns the DOWN of v0 when queried from the fork.

PR #2460, opened 2026-08-21
fix(state-store): give the substate lock substate-id index its table prefix

sdbondi:
Follow-up to #2459, which noted this as latent fragility and deferred it because fixing it properly needs a data migration.

## Problem

`SubstateIdIndex` shares the substates column family with the substate and lock tables, where each table is namespaced by a leading `KeyPrefix` byte. It declared `type Prefix = ()`, so its keys were written unprefixed and sat in the low keyspace of that family, namespaced only by the leading borsh `SubstateId` discriminant. A `SubstateIdIndexPrefix` was declared for it but never used.

Nothing collides today — `SubstateId` discriminants are 0..=9 and the other tables in the family use 21 and 32..=38 — so this is latent fragility rather than a live bug. But a new `SubstateId` variant reaching the prefix range would start overlapping another table, and the failure would be silent key overlap in the lock index.

## Fix

The index now encodes its prefix. Because the key encoding changes, existing entries need rewriting, hence migration `v1`.

The migration scans the column family one leading byte at a time, below the lowest table prefix, and rewrites each key it finds through the current index. Scanning a single leading byte per pass keeps each scan consistent with the family's one byte prefix extractor. Rewritten keys lead with a byte above the scanned range, so they land outside every scan and are never revisited — which makes the migration idempotent and safe to interrupt.

This is the first migration since the version marker was reset to 0, so it also wires up the version step loop that `migrations/mod.rs` already documented, and the module docs are updated for adding the next one. A fresh database is stamped at `CURRENT_VERSION` with the genesis state and skips the migration, which is correct — it has no legacy rows.

Migration timing is logged, both per run and overall.

## Testing

Three tests in `migrations/v1.rs`, run against a real store opened with the production options (so the prefix extractor is in play):

- `it_migrates_every_entry` — 256 entries spanning every legacy leading byte all survive the rewrite, confirming that deleting each entry as the scan yields it skips none.
- `it_prefixes_legacy_entries_and_leaves_the_rest_alone` — legacy entries are rewritten and an already-prefixed entry is neither rescanned nor prefixed twice.
- `it_is_idempotent` — a second run is a no-op.

`cargo test -p tari_validator_node --lib migrations` — 3 passed. `cargo test -p tari_state_store_rocksdb` — 11 passed.

### DIFF
--- a/applications/tari_validator_node/src/migrations/mod.rs
+++ b/applications/tari_validator_node/src/migrations/mod.rs
@@ -9,36 +9,34 @@
 //! stored version and [`CURRENT_VERSION`]. A fresh database skips migrations entirely - it is stamped
 //! directly with `CURRENT_VERSION` once the genesis state is laid down.
 //!
-//! There are currently no migrations: the previous ones were folded into the genesis state on a
-//! testnet reset (see [`CURRENT_VERSION`]).
-//!
 //! # Adding a migration
 //!
-//! To take the schema from version 0 to 1:
+//! To take the schema from version `N` to `N + 1`:
 //!
-//! 1. Add `v1.rs` with `pub fn migrate(...) -> ...` performing the upgrade, and declare it here with `mod v1;`.
-//! 2. Bump [`CURRENT_VERSION`] to `1`.
-//! 3. Apply it to already-bootstrapped databases by stepping the stored version up to `CURRENT_VERSION`, persisting the
-//!    new version as you go - replace the `Some(version)` arm in [`migrate`] with:
+//! 1. Add `v{N+1}.rs` with `pub fn migrate(...) -> ...` performing the upgrade, and declare it here with `mod v{N+1};`.
+//! 2. Bump [`CURRENT_VERSION`] to `N + 1`.
+//! 3. Add the arm that applies it to the step loop in [`migrate`], keyed by the version it upgrades *from*:
 //!
 //! ```ignore
-//! Some(mut version) => {
-//!     while version < CURRENT_VERSION {
-//!         match version {
-//!             0 => v1::migrate(tx, network, consensus_constants.num_preshards)?,
-//!             other => unreachable!("no migration defined for database version {other}"),
-//!         }
-//!         version += 1;
-//!         tx.db().cf(DatabaseMigrationVersion)?.put(&ByteColumn, &version, OPERATION)?;
-//!     }
-//! },
+//! match version {
+//!     0 => v1::migrate(tx)?,
+//!     1 => v2::migrate(tx, network)?,
+//!     other => unreachable!("no migration defined for database version {other}"),
+//! }
 //! ```
+//!
+//! A migration must be able to run against a database at any earlier supported version, so it may not assume the
+//! current schema of anything it does not itself write.
 //!
 //! IMPORTANT: a migration that creates or mutates substates must write them to the per-shard state
 //! tree (JMT), not only the substate store - otherwise they have no inclusion proof and verified
 //! reads of them fail. Mirror [`crate::genesis_state::create_genesis_state`], which commits each
 //! substate to both the store and the state tree. (Note that, unlike genesis, adding state-tree
 //! entries to a live chain shifts its state root, so such a migration is itself consensus-affecting.)
+
+mod v1;
+
+use std::time::Instant;
 
 use log::*;
 use tari_consensus::consensus_constants::ConsensusConstants;
@@ -57,10 +55,9 @@ const LOG_TARGET: &str = "tari::validator::migrations";
 /// The on-disk state schema version stamped onto a freshly bootstrapped database.
 ///
 /// Bump this and apply the upgrade in [`migrate`] whenever the persisted state must change for
-/// already-running nodes. It was reset to 0 with the genesis-in-state-tree testnet reset: the former
-/// v1 (token symbol) and v2 (faucet claim resource) migrations are now part of the genesis state, so
-/// they no longer need to run.
-const CURRENT_VERSION: u64 = 0;
+/// already-running nodes. It was reset to 0 with the genesis-in-state-tree testnet reset, and is at 1
+/// for the substate lock index prefix ([`v1`]).
+const CURRENT_VERSION: u64 = 1;
 
 pub fn migrate<TAddr: NodeAddressable + 'static>(
     tx: &mut RocksDbStateStoreWriteTransaction<'_, TAddr>,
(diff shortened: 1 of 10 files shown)


sdbondi (commented):
Three findings on files outside this diff that interact with the change:

**`utilities/validator_rollback/src/storage/write.rs:330` — rollback tool orphans unmigrated lock-index rows.** `substate_locks_remove_any_by_block_id` now deletes `[35]`-prefixed `SubstateIdIndex` keys, but the rollback tool never runs (or version-checks against) migrations. Run against a v0 database, it deletes the lock model and block-id index rows but misses every legacy unprefixed substate-id index row. When the upgraded validator then starts, `v1::migrate` re-prefixes those orphans into live index entries for blocks the rollback deleted — and since `substate_locks_has_any_write_locks_for_substates` does no block filtering, a resurrected Write entry reports a phantom conflicting write lock. The tool should refuse to run (or handle both key shapes) when `DatabaseMigrationVersion` < `CURRENT_VERSION`.

**`crates/state_store_rocksdb/src/column_families/transaction_pool_state_update.rs:125` — the identical latent bug one file over.** `TransactionPoolStateUpdateDebugHistoryCf` shares the DIAGNOSTICS column family with the prefixed `DiagnosticsNoVoteCf` but declares `type Prefix = ()`, while its `prefixed!` prefix (KeyPrefix 27, declared directly above) is used nowhere. Its keys lead with big-endian `Epoch` bytes, squatting in the low keyspace of DIAGNOSTICS exactly as the lock index squatted in SUBSTATES, and every transaction-pool update with `debugging_data` on writes more unprefixed rows. Fixing it in this PR is one line plus reusing the migration infrastructure being built here (diagnostics-only data could simply be dropped in the migration rather than rewritten).

**`utilities/db_inspector/src/webserver/server.rs:213` — inspector shows an empty index on a not-yet-migrated DB.** The inspector opens the store read-only (so migrations can never run) and registers the prefixed `SubstateIdIndex` reader. On a v0 database the page scans `PrefixRange([35])` and shows zero entries while the real rows sit unprefixed, visible in no registered table — easy to misread as a corrupt/empty index. Surfacing the DB's `DatabaseMigrationVersion` vs `CURRENT_VERSION` mismatch would prevent the misread.

The core of the change checks out: the `0..21` legacy sweep bound is correct for the SUBSTATES family (lowest live prefix is `ForeignSubstatePledges` = 21, legacy borsh `SubstateId` discriminants are 0–9), legacy and new codecs are identical apart from the prefix byte, no other unprefixed table shares the family, and the delete-while-iterating pattern matches what already ships in `writer.rs`.

applications/tari_validator_node/src/migrations/mod.rs
```diff
@@ -77,15 +74,36 @@ pub fn migrate<TAddr: NodeAddressable + 'static>(
     };
 
     match maybe_version {
-        // An already-bootstrapped database. No migrations are currently defined; when one is needed,
-        // step `version` up to `CURRENT_VERSION` here, applying each upgrade and persisting the new
-        // version as it goes.
-        Some(version) => {
+        // An already-bootstrapped database: step it up to `CURRENT_VERSION`, applying each upgrade and
+        // persisting the new version as it goes.
+        Some(version) if version >= CURRENT_VERSION => {
```
sdbondi, line 76: A database stamped with a version **newer** than `CURRENT_VERSION` (binary downgrade) is silently accepted here with only a debug-level log. Once a future v2 migration re-keys another table, an operator who upgrades (DB stamped 2) and then rolls back to this binary gets a node that starts normally against a schema it doesn't understand — empty or garbled indexes with no diagnostic. Now that schema versions can actually diverge, `version > CURRENT_VERSION` should bail ("database was written by a newer version") so this fails fast instead.

applications/tari_validator_node/src/migrations/v1.rs
```diff
@@ -0,0 +1,195 @@
...
+    use tari_state_store_rocksdb::{DatabaseOptions, RocksDbStateStore, column_families::substate_locks};
+    use tari_template_lib::types::{ComponentAddress, ObjectKey};
+
+    use super::*;
+
+    fn lock_key(seed: u8) -> substate_locks::SubstateLockKey {
+        substate_locks::SubstateLockKey {
+            block_id: BlockId::new(FixedHash::from([seed; FixedHash::byte_size()])),
+            block_height: NodeHeight(u64::from(seed)),
+            substate_id: ComponentAddress::from_array([seed; ObjectKey::LENGTH]).into(),
```
sdbondi, line 69: All test keys are `ComponentAddress` substates (borsh leading byte 0), so the multi-leading-byte sweep over `0..21` and `LEGACY_UNPREFIXED_UPPER_BOUND` are never exercised — iterations for bytes `1..=20` scan empty ranges. If the bound or per-byte loop regressed (bound computed as 1, or the loop collapsed to a single prefix-0 scan), every test would still pass while production Vault/Resource/NonFungible lock rows (discriminants 1–9) silently stayed unprefixed. Varying the `SubstateId` variant across test keys would cover the multi-range scan with the existing fixtures.

utilities/validator_rollback/src/apply.rs
```diff
@@ -106,6 +111,39 @@ pub struct ApplyOutcome {
 
 /// Run the rollback flow with explicit options. Returns a summary of what the audit
 /// footer recorded so callers can assert on impact.
+/// Refuses to operate on a database whose schema version is not the one this build encodes.
+///
+/// The rollback primitives write through the current key encodings. Against an older database those encodings do not
+/// match what is on disk, so a cascade silently misses the rows a pending migration has yet to rewrite - leaving
+/// orphans that the migration then promotes into live entries referring to rolled-back blocks. Against a newer
+/// database the encodings are simply unknown to this build.
+fn check_schema_version(store: &RocksDbStateStore<PeerAddress>) -> anyhow::Result<()> {
```
sdbondi, line 118: `run_with_options`'s doc comment ("Run the rollback flow with explicit options. Returns a summary of what the audit footer recorded...") is now fused onto `check_schema_version` — the new fn and its doc were inserted between that comment and `run_with_options`, so this function's docs open with a description of a different function and `run_with_options` is left undocumented.

sdbondi (commented):
Re-reviewed 7844d94. The rollback-tool schema-version guard and the debug-history prefix (v2 migration + shared `rewrite_unprefixed_rows`) both check out: the v2 scan bound of `DiagnosticsNoVotes` (11) is the lowest prefix in the diagnostics family and legacy keys lead with a big-endian `Epoch` high byte of 0, the fresh-bootstrap path stamps schema version 2, and `check_schema_version` covers older/newer/unstamped databases with actionable messages.

Still open from the previous review: the validator's own `version >= CURRENT_VERSION` arm in `migrations/mod.rs` (the newer-version bail only landed in the rollback tool), and the v1 test keys all sharing leading byte 0.

PR #2463, opened 2026-08-21
fix(state-store): bound query range scans to their logical table

sdbondi:
## Description

Two independent defects in `tari_state_store_rocksdb`, both on the boundary between logical tables that share a physical column family.

### 1. Range query scans were not bounded to their logical table

Logical tables share a physical column family and are separated only by a leading key prefix byte. Two of the four start/end range query methods did not bound the scan to that prefix:

- `query_end_range_iterator` built `..key`, so the scan started at the beginning of the *physical* column family.
- `query_start_range_iterator` built `key..`, so it ran past the end of the table into the next prefix.

`PrefixCodec` hard-validates the prefix byte, so either scan fails with `MalformedData` as soon as a neighbouring table holds a single row.

**Live impact: epoch GC never made progress.** `cleanup::foreign_proposals_for_epoch` scans `foreign_proposal::EpochIndex` (prefix 18), which shares its column family with `ForeignProposalCf` (prefix 17). One stored foreign proposal was enough to fail the scan. The failure rolls back the whole GC write transaction — discarding the block, QC and finalized-transaction pruning already done in it — and the error is only logged at the `tx.epoch_cleanup(..)` call site in `hotstuff::epoch_gc`. The database therefore grew without bound.

`query_start_range_iterator` has no callers today, so its missing upper bound was latent.

The fix introduces `TableBounds`, which closes any bound the caller leaves open at the extent of the column family's key prefix, and uses it for all four start/end range query methods. That also lets `query_start_range_key_iterator` drop its two-branch `Box<dyn Iterator>` return in favour of `impl Iterator`.

### 2. Wrong iterator family when counting remaining missing transactions

`parked_block_remove_missing_transaction` called `prefix_range_key_iterator` — the plain-`Cf` method — on a `CfContext<_, ByBlockIdQuery>`. Via the blanket `impl<T: QueryCf> Cf for T` that decodes stored index keys of `[prefix][block_id][transaction_id]` with the *query's* `BlockIdCodec`, leaving 32 trailing bytes, which errors. The result was tested with `is_some()`, and `Some(Err(_))` is also `is_some()`.

To be clear about severity: the two outcomes happened to coincide, because the decode error can only occur when a row exists — which is exactly when the caller wants "still missing". Blocks did unpark correctly. This is a latent defect, not a live bug: it silently swallowed a storage error and would have produced a wrong answer under any change to the index key shape.

Fixed by using `query_prefix_range_key_iterator` and propagating the result with `.transpose()?`.

This is the only site in the crate that invoked a plain-`Cf` key iterator on a `QueryCf` type. `reader.rs` has one `prefix_range_value_iterator` on a query type, which is fine because it decodes only the value.

## Motivation and Context

Found while auditing every iteration site in `state_store_rocksdb` for mutate-during-iteration hazards. Neither defect is related to that question; both were found on the files being read.

## How Has This Been Tested?

- `epoch_cleanup_prunes_foreign_proposals` (`tests/foreign_proposals.rs`) — end-to-end regression test for issue 1. Verified it fails on the pre-fix code with `Invalid prefix byte. Expected 18, got 17`, and passes after.
- `end_range_iterator_does_not_read_the_preceding_prefix` / `start_range_iterator_does_not_read_the_following_prefix` (`src/tests.rs`) — unit tests over a pair of prefixed column families sharing one physical CF. Both fail on the pre-fix code.
- `tests/missing_transactions.rs` extended to two missing transactions, asserting the first removal leaves the block parked and the second unparks it. This passes before and after, which is what established that issue 2 was latent rather than live.
- `cargo test -p tari_state_store_rocksdb` — all green.
- `cargo lints clippy --all-features --all-targets` — clean.

Revert the `TableBounds::for_cf::<TQuery::Cf>(None, Some(key))` argument in `query_end_range_iterator` back to `..key` and run `cargo test -p tari_state_store_rocksdb --test foreign_proposals`; `epoch_cleanup_prunes_foreign_proposals` fails with a prefix validation error.

## Breaking Changes

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/crates/state_store_rocksdb/src/cf_api.rs
+++ b/crates/state_store_rocksdb/src/cf_api.rs
@@ -586,18 +586,19 @@ impl<'db, TQuery: QueryCf, DB: RocksReader> CfContext<'db, DB, TQuery> {
         >(ordering, TableBounds::for_cf::<TQuery::Cf>(None, Some(key)))
     }
 
+    /// Returns the last key/value in the logical table.
     pub fn query_last(&self, operation: &'static str) -> Result<QueryCfKv<TQuery>, RocksDbStorageError> {
-        let mut iter = self.db.iterator_cf(self.handle, IteratorMode::End);
-        let result = iter.next().ok_or_else(|| RocksDbStorageError::QueryError {
+        let mut iter = self.range_iterator_with_codecs::<
+            PrefixedCodec<TQuery::Cf>,
+            <TQuery::Cf as Cf>::Key,
+            <TQuery::Cf as Cf>::ValueCodec,
+            <TQuery::Cf as Cf>::Value,
+        >(Ordering::Descending, TableBounds::for_cf::<TQuery::Cf>(None, None));
+
+        iter.next().transpose()?.ok_or_else(|| RocksDbStorageError::QueryError {
             operation,
             details: format!("No values in TQuery {}", TQuery::name()),
-        })?;
-        let key_codec = TQuery::make_cf_key_codec();
-        let value_codec = TQuery::make_cf_value_codec();
-        let (key, value) = result.map_err(|e| RocksDbStorageError::RocksDbError { operation, source: e })?;
-        let key = key_codec.decode_exact(&key)?;
-        let value = value_codec.decode_exact(&value)?;
-        Ok((key, value))
+        })
     }
 }
 
--- a/crates/state_store_rocksdb/src/cf_api.rs
+++ b/crates/state_store_rocksdb/src/cf_api.rs
@@ -484,40 +484,25 @@ impl<'db, TQuery: QueryCf, DB: RocksReader> CfContext<'db, DB, TQuery> {
         })
     }
 
+    /// Returns an iterator over the keys in the column family that are greater than or equal to the start key.
     pub fn query_start_range_key_iterator(
         &self,
         ordering: Ordering,
         start_key: &TQuery::Key,
-    ) -> Box<dyn Iterator<Item = Result<<TQuery::Cf as Cf>::Key, RocksDbStorageError>> + 'db> {
+    ) -> impl Iterator<Item = Result<<TQuery::Cf as Cf>::Key, RocksDbStorageError>> + 'db {
         let key = self.encode_key(start_key);
-        // If this is 0xFF then we'll read to the end (no end bound)
-        match TQuery::Cf::key_prefix().and_then(|p| p.checked_add(1)) {
-            Some(next_prefix) => {
-                let end = EncodeVec::new_from_array([next_prefix]);
-                let iter = self
-                    .range_iterator_with_codecs::<PrefixedCodec<TQuery::Cf>, <TQuery::Cf as Cf>::Key, UnitCodec, ()>(
-                        ordering,
-                        key..end,
-                    );
-                Box::new(iter.map(|res| {
-                    let (k, _) = res?;
-                    Ok::<_, RocksDbStorageError>(k)
-                }))
-            },
-            None => {
-                let iter = self
-                    .range_iterator_with_codecs::<PrefixedCodec<TQuery::Cf>, <TQuery::Cf as Cf>::Key, UnitCodec, ()>(
-                        ordering,
-                        key..,
-                    );
-                Box::new(iter.map(|res| {
-                    let (k, _) = res?;
-                    Ok::<_, RocksDbStorageError>(k)
-                }))
-            },
-        }
+        let iter = self
+            .range_iterator_with_codecs::<PrefixedCodec<TQuery::Cf>, <TQuery::Cf as Cf>::Key, UnitCodec, ()>(
+                ordering,
+                TableBounds::for_cf::<TQuery::Cf>(Some(key), None),
+            );
+        iter.map(|res| {
+            let (k, _) = res?;
+            Ok::<_, RocksDbStorageError>(k)
+        })
     }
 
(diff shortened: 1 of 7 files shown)


crates/state_store_rocksdb/src/cf_api.rs
```diff
@@ -601,7 +583,7 @@ impl<'db, TQuery: QueryCf, DB: RocksReader> CfContext<'db, DB, TQuery> {
             <TQuery::Cf as Cf>::Key,
             <TQuery::Cf as Cf>::ValueCodec,
             <TQuery::Cf as Cf>::Value,
-        >(ordering, ..key)
+        >(ordering, TableBounds::for_cf::<TQuery::Cf>(None, Some(key)))
     }
 
     pub fn query_last(&self, operation: &'static str) -> Result<QueryCfKv<TQuery>, RocksDbStorageError> {
```
sdbondi, line 590: `query_last` is the one method left in this `impl` block that still scans the whole *physical* column family: `iterator_cf(self.handle, IteratorMode::End)` sets no bounds at all, and the key is then decoded with `TQuery::make_cf_key_codec()`, i.e. the same `PrefixCodec` that hard-validates the prefix byte.

So it has exactly the failure mode this PR fixes elsewhere: `foreign_proposal::ByEpochQuery::query_last()` would land on the last key in `FOREIGN_PROPOSALS`, which belongs to `UnconfirmedIndex` (prefix 20), and fail with `Invalid prefix byte. Expected 18, got 20` as soon as a single unconfirmed row exists. It is latent today because nothing calls it, which is the same position `query_start_range_iterator` was in.

Since `TableBounds` now exists, `TableBounds::for_cf::<TQuery::Cf>(None, None)` + `range_opts` would close this too, and would keep the next caller of `query_last` from re-discovering the bug.

crates/state_store_rocksdb/src/tests.rs
```diff
@@ -175,3 +238,47 @@ fn it_iterates_over_epoch_heights_ordering() {
...
+    let upper = ctx.cf(UpperEpochBlockCf).unwrap();
+    for i in 0..5u64 {
+        lower.put(&(Epoch(i), block_id_from_seed(i)), &(), OP).unwrap();
+        upper.put(&(Epoch(i), block_id_from_seed(i)), &(), OP).unwrap();
+    }
+}
+
+#[test]
+fn end_range_iterator_does_not_read_the_preceding_prefix() {
+    let (db, _tmp) = create_rocksdb([SHARED_CF]);
```
sdbondi, line 255: These two tests go through `src/tests.rs::create_rocksdb`, which builds `rocksdb::Options::default()`. Production CFs are opened with `build_default_store_opts` (`store.rs`), which additionally sets `SliceTransform::create_fixed_prefix(1)` and a memtable prefix bloom on every CF.

That 1-byte extractor is precisely why `range_opts` has to special-case `set_total_order_seek(true)` for `IteratorMode::End`, so the config the bound logic is most sensitive to is the one the unit tests do not run under. A regression in the interaction between `TableBounds` and prefix seek would pass here.

Related: both new tests only cover `Ordering::Ascending` and only the key/value variants. The variants actually used in production are the *key* iterators (`reader.rs:837`, `reader.rs:991`, `reader.rs:994`), and `query_end_range_key_iterator` is used descending. Adding a descending case, or seeding through `tests/helpers.rs::create_rocksdb` so the real options apply, would cover both gaps cheaply.

(The `tests/foreign_proposals.rs` regression test does use the real store options, so this is coverage depth rather than a hole.)

sdbondi (commented):
Reviewed the diff for correctness. Both fixes look right to me:

- `TableBounds` closes the open bound at `[prefix]` / `[prefix+1]`, which matches what `query_start_range_key_iterator` already did by hand, and the `checked_add` -> `None` case for a `0xFF` prefix is correct since nothing can sort after that table.
- `query_prefix_range_key_iterator` + `.transpose()?` in `parked_block_remove_missing_transaction` decodes the index key with `MissingTransactionBlockIdIndex`'s `(BlockIdCodec, TransactionIdCodec)` rather than the query's `BlockIdCodec`, and now surfaces the error instead of folding it into `is_some()`.

Two minor points below, both low severity.

crates/state_store_rocksdb/src/cf_api.rs
sdbondi: Good catch — fixed in 9590ea9. `query_last` now goes through `range_iterator_with_codecs` with `TableBounds::for_cf::<TQuery::Cf>(None, None)`, so it gets both the prefix bounds and the `total_order_seek` that `range_opts` applies for reverse iteration.

Added `query_last_does_not_read_the_following_prefix` to `src/tests.rs`; verified it fails against the previous `query_last` (with the rest of the PR in place) and passes after.

crates/state_store_rocksdb/src/tests.rs
sdbondi: Both addressed in 9590ea9.

`src/tests.rs::create_rocksdb` now opens the column families with `store::build_default_store_opts()` (made `pub(crate)`) instead of `Options::default()`, so every unit test in the crate — not just the new ones — runs with the 1-byte prefix extractor and the memtable prefix bloom. All existing tests still pass under it, including the descending `query_prefix_range_key_iterator` case in `it_iterates_over_epoch_heights_ordering`.

The two range tests now also assert the descending case via the key-only variants (`query_end_range_key_iterator` descending, `query_start_range_key_iterator` descending), which is the combination `reader.rs:991`/`:994` actually uses.

PR #2468, opened 2026-08-22
fix(engine): scope address allocations to the call frame that owns them

chironbuilds:
## Summary
Buckets and proofs passed across call frames must be in the callee's `CallScope`, but pending
address allocations were tracked in one global transaction-level map, keyed by small sequential
integers, and validated only for *existence*, never *ownership* - the gap was explicitly flagged
in-source: `crates/engine/src/runtime/working_state.rs` (`// TODO: should we scope address
allocations?`).

## Why
- `GetAddress(id)` leaked any pending allocation's derived address to any call frame that could
  guess/enumerate the id.
- `ComponentAction::Create` (and the equivalent resource path) consumed whatever id the caller
  named, with no check that the caller was the frame that created it.

Combined, a template a victim contract calls could brute-force `GetAddress(0..k)`, obtain the
victim's expected future component/resource address, and create its own substate at that address
first - griefing the victim's later `use_allocated_address` call, and if anyone deposits funds to
the anticipated address in the meantime, those funds land in the attacker's substate instead.

## Change
- `crates/engine/src/runtime/scope.rs`: `CallScope` gains an `address_allocation_scope` set with
  `add_address_allocation_to_scope`/`is_address_allocation_in_scope`/
  `remove_address_allocation_from_scope`, mirroring the existing `bucket_scope` mechanism exactly.
  Allocations carry across a call frame boundary via `include_refs_in_scope` (populating a new
  child frame from its args, same as buckets/proofs) and `update_from_child_scope` (propagating
  back to the parent on return, same as buckets).
- `crates/engine/src/runtime/working_state.rs`: `new_address_allocation` adds the new id to the
  current frame's scope; `use_allocated_address` and `get_allocated_address` - the two engine
  choke points `GetAddress` and `Create` both go through - now require scope membership, erroring
  with the existing `AddressAllocationNotInScope` variant otherwise. `check_all_substates_in_scope`
  gets the same check on the arg-passing path, closing the TODO.
- `crates/engine_types/src/indexed_value.rs`: fixes a pre-existing, unrelated bug this change
  exposed - `IndexedWellKnownTypes`'s `FromIterator` impl (used to merge a multi-arg call's
  per-arg scopes into one combined `arg_scope` before pushing a new call frame) silently dropped
  several fields, including both address allocation lists. Nothing depended on allocations
  surviving that merge before this change made it load-bearing, so it was latent; without this
  fix the new scope enforcement breaks the legitimate "allocate as one transaction instruction,
  consume in a later `call_function`" pattern that the existing `address_allocation.rs` test suite
  already covers.

## Testing
- `cargo check -p tari_engine -p tari_engine_types` passes.
- `cargo +nightly-2025-12-05 fmt -p tari_engine -p tari_engine_types -- --check` passes.
- `cargo test -p tari_engine --test address_allocation`: all 6 tests pass, including the two
  cross-instruction-boundary tests that regressed before the `FromIterator` fix was added.
- `cargo test -p tari_engine` (full suite, reduced `--jobs 2` to avoid unrelated Windows
  build-parallelism flakiness I hit at full parallelism): all 37 test binaries pass, 0 failures.

### DIFF
--- a/crates/engine/src/runtime/impl.rs
+++ b/crates/engine/src/runtime/impl.rs
@@ -3681,7 +3681,7 @@ where
         self.invoke_modules_on_runtime_call("builtin_template_invoke")?;
 
         let address = match action {
-            BuiltinTemplateAction::GetTemplateAddress { bultin } => match bultin {
+            BuiltinTemplateAction::GetTemplateAddress { builtin: bultin } => match bultin {
                 BuiltinTemplate::Account => ACCOUNT_TEMPLATE_ADDRESS,
                 BuiltinTemplate::AccountNft => NFT_FAUCET_TEMPLATE_ADDRESS,
             },
--- a/crates/engine/src/runtime/scope.rs
+++ b/crates/engine/src/runtime/scope.rs
@@ -6,7 +6,7 @@ use std::fmt::Display;
 use indexmap::IndexSet;
 use tari_engine_types::{indexed_value::IndexedWellKnownTypes, lock::LockId, substate::SubstateId};
 use tari_template_lib::{
-    models::{BucketId, ProofId},
+    models::{AddressAllocationId, BucketId, ProofId},
     types::{
         EntityId,
         TemplateAddress,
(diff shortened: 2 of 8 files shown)


sdbondi (approved):
ACK - Thank you!

- added a test to check behaviour
- fixed typo on ABI fieldname (not breaking because field names are not on the wire)
- ran tests locally

PR #2475, opened 2026-08-24
fix(bindings): avoid shortening small strings

### LINKED ISSUE #1872: NFT metadata values display incorrect truncation with overlapping text in Vault WebUI
**Describe the bug**
When viewing NFT metadata in the Vault WebUI, metadata field values are displayed with incorrect truncation. The UI appears to show both a truncated version and parts of the full value overlapping.

**To Reproduce**
1. Go to the Vault WebUI
2. Navigate to an NFT collection with metadata fields (e.g., `name` field with value `Rick Astley`)
3. View the NFT metadata entries
4. Observe the metadata value display

**Expected behavior**
Long values should be truncated with `...` at the end, e.g. `Rick Ast...`

**Screenshots**
Image

**Desktop (please complete the following information)**
- OS & Version: [e.g. Linux, Windows 11, macOS 14]
- Browser & Version: [e.g. Chrome 120, Firefox 121]

**Smartphone**
(Not applicable / Not tested)

**Additional context**
Example:

| Field | Original Value | Displayed |
|-------|----------------|-----------|
| name  | Rick Astley    | Rick Ast...k Astley |

The full value is only correctly visible on hover.

xt1085:
## Summary

- return short strings unchanged when a head-and-tail abbreviation would not reduce their display length
- prevent overlapping output such as `Rick Ast...k Astley` in NFT metadata cards
- add regression coverage for short values, empty strings, boundary lengths, long text, and Substate IDs

## Why

`shortenString` always concatenated the first and last characters around `...`. For values shorter than the generated abbreviation, the slices overlap and duplicate content.

This restores the same length guard previously introduced for the wallet WebUI helper in #823, while preserving the existing head-and-tail format for long values.

## Tests

- `pnpm --filter @tari-project/ootle-ts-bindings exec vitest run test/helpers.test.ts` — 12 passed
- `pnpm --recursive run build` — passed, including all WebUI TypeScript and Vite builds
- Prettier check on both changed files — passed
- `git diff --check` — passed
- regression proof: the new helper suite has 5 behavioral failures on the original implementation and 12/12 passes with this change

The complete existing bindings suite reports 24 passed and 1 pre-existing failure in `ootleAddress.test.ts` (`decodes a known address correctly`, invalid checksum). The unchanged baseline reports the same failure with 12 passed and 1 failed. This patch does not touch address encoding or that fixture.

## Risk

Low. The exported helper changes only for strings shorter than `start + end + 3`; long addresses, hashes, and Substate IDs keep their existing output. The threshold and behavior match the repository's existing wallet helper.

Fixes #1872

### DIFF
--- a/bindings/package.json
+++ b/bindings/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tari-project/ootle-ts-bindings",
-  "version": "1.50.0",
+  "version": "1.50.1",
   "private": false,
   "publishConfig": {
     "access": "public"
--- a/bindings/src/helpers/helpers.ts
+++ b/bindings/src/helpers/helpers.ts
@@ -73,6 +73,10 @@ export function shortenSubstateId(
 }
 
 export function shortenString(string: string, start: number = 8, end: number = 8) {
+  // The number 3 is from the characters for ellipsis
+  if (string.length < start + end + 3) {
+    return string;
+  }
   return string.substring(0, start) + "..." + string.slice(-end);
 }
 
--- a/bindings/test/helpers.test.ts
+++ b/bindings/test/helpers.test.ts
@@ -1,0 +1,79 @@
+//   Copyright 2026 The Tari Project
+//   SPDX-License-Identifier: BSD-3-Clause
+
+import { describe, expect, it } from "vitest";
+import { shortenString, shortenSubstateId } from "../src";
+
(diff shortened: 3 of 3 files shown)


sdbondi (approved):
LGTM, I've bumped the bindings version

PR #2480, opened 2026-08-25
feat(p2p)!: sync state for many shards over a single stream

sdbondi:
## Problem

`sync_state` serves exactly one shard per call, so both consumers loop shards client-side — the indexer over every shard group, the validator over its intersecting shard group.

With `NumPreshards::P256` that is **257 requests per sync round**, and they are strictly serialized:

- `RpcMultiPool` holds one session per peer,
- the RPC client worker processes requests one at a time (`networking/rpc_framework/src/client/mod.rs:536`),
- the server completes one request per substream before reading the next (`networking/rpc_framework/src/server/mod.rs:743`).

An indexer round therefore costs 257 round-trips **even when every shard is empty** — roughly 21s at 80ms RTT, against a 30s work interval. A validator joining at an epoch boundary pays the same cost.

## Approach

State versions are assigned per shard and only advance when that shard changes (`crates/consensus/src/hotstuff/substate_store/sharded_state_tree.rs:90`), so a shard *range* cannot share a single start version. The request instead carries a list of `(shard, start_state_version)` cursors, and the responder streams each shard contiguously, closing it off with its own completion marker.

Verification granularity is unchanged — checkpoints already carry per-shard tree roots (`shard_tree_summary`), so the validator still verifies each shard against its own root.

No storage change was needed: the RocksDB state transition CF is already keyed `(Shard, Version)` shard-major, and `state_transitions_get_starting_at` already opens a range that scans forward across shards. The responder just loops the existing per-shard query.

## Wire format

```proto
message ShardCursor { uint32 shard = 1; uint64 start_state_version = 2; }
// SyncStateRequest: reserved 1, 2;  repeated ShardCursor cursors = 5;
// SubstateBatch:    uint32 shard = 5;
// SyncComplete:     uint32 shard = 3;  bool is_final = 4;   // one per cursor
```

Cursors must be non-empty, at most `num_preshards + 1`, in range, strictly ascending (which rules out duplicates), and start above version 0. Ascending order is what lets the responder stream each shard contiguously, and hence lets a consumer finalise a shard the moment its marker arrives.

## Indexer

Now issues **at most two streams per shard group** rather than one per shard. Two rather than one because a shard being synced for the first time wants only the head state, expressed by the `UP_ONLY` filter, and filters apply to the whole request; those shards are streamed separately from the ones being caught up incrementally.

It collapses to one stream only once every shard in the group has accumulated some state. A shard whose tree never advanced reports `synced_to = 0`, so the indexer reads `prev_version == 0` next round and puts it back in the from-scratch bucket — on a sparse network a group stays at two streams indefinitely. Harmless, since an empty shard streams nothing either way, but it is two rather than one.

Progress is still recorded per shard as markers arrive, so an interrupted stream keeps everything already committed and resumes from the recorded cursors next round. Stream buffers are now scoped to a single stream rather than shared across shards, which also removes a latent cross-shard carry-over if a stream ended mid-version.

A truncated stream (no final marker) is now an error rather than silent success, matching what the validator client already did. The indexer's supervisor restarts after a `work_interval` cooldown, the same path any RPC error already took.

## Validator

Passes a single-element cursor list, leaving its sync behaviour unchanged, plus guards rejecting a batch or marker for a shard it did not request. Converting it to a genuinely ranged sync — ranged peer failover, per-shard verification across a set — follows in a separate PR. Keeping them apart means the transport is validated in CI before the ranged validator logic lands on top of it.

## A note on rejected hardening

Rejecting shards outside the responder's own shard group looked worthwhile but does not survive contact: a validator syncing at an epoch boundary queries *previous*-epoch committees for shards those peers may no longer hold in the current epoch, so the check would reject legitimate requests. The max-shard bound stays. Requests for unheld shards remain benign — `send_complete` falls back to `start - 1`, so no false progress is recorded.

## Breaking change

`SyncStateRequest` replaces its `shard` and `start_state_version` fields with a `cursors` list, and `SubstateBatch`/`SyncComplete` carry the shard they belong to. The RPC protocol name is unversioned, so **validators and indexers must be upgraded together**.

## Testing

- Workspace `cargo check --all-targets` and `cargo clippy --all-targets` clean.
- `cargo +nightly-2025-12-05 fmt --check` clean.
- 88 unit tests pass across the touched crates, including 7 covering cursor validation and 10 covering the stream-ordering invariants (`StreamOrder`).
- `state_sync.feature` and `indexer.feature` exercise both changed paths end-to-end — CI is the real gate here.

### DIFF
--- a/applications/tari_indexer/src/network_state_sync/worker.rs
+++ b/applications/tari_indexer/src/network_state_sync/worker.rs
@@ -533,7 +533,7 @@ impl NetworkWideStateSync {
         shard_group: ShardGroup,
         session: &mut ValidatorRpcSession,
     ) -> Result<(), NetworkStateSyncError> {
-        let requested = cursors.iter().map(|c| Shard::from(c.shard)).collect::<HashSet<_>>();
+        let mut order = StreamOrder::new(&cursors);
 
         info!(
             target: LOG_TARGET,
@@ -569,11 +569,9 @@ impl NetworkWideStateSync {
                 Some(rpc::sync_state_response::Response::Batch(batch)) => batch,
                 Some(rpc::sync_state_response::Response::Complete(complete)) => {
                     let shard = Shard::from(complete.shard);
-                    if !requested.contains(&shard) {
-                        return Err(NetworkStateSyncError::InvalidStateUpdate {
-                            details: format!("Received completion marker for unrequested shard {shard}"),
-                        });
-                    }
+                    order
+                        .accept_marker(shard)
+                        .map_err(|details| NetworkStateSyncError::InvalidStateUpdate { details })?;
                     // Terminal watermark: advance recorded progress to the version the producer is
                     // synced to. This covers trailing versions that streamed no updates because their
                     // substates are all filtered out for our subscription - without it we could never
@@ -617,18 +615,16 @@ impl NetworkWideStateSync {
             };
 
             let shard = Shard::from(batch.shard);
-            if !requested.contains(&shard) {
-                return Err(NetworkStateSyncError::InvalidStateUpdate {
-                    details: format!("Received batch for unrequested shard {shard}"),
-                });
-            }
+            let state_version = StateVersion::new(batch.state_version);
+            order
+                .accept_batch(shard, state_version, batch.has_more)
+                .map_err(|details| NetworkStateSyncError::InvalidStateUpdate { details })?;
             let msg_epoch = batch
                 .epoch
                 .map(Epoch::from)
                 .ok_or_else(|| NetworkStateSyncError::InvalidStateUpdate {
                     details: "Received state update without epoch".to_string(),
                 })?;
-            let state_version = StateVersion::new(batch.state_version);
 
             for update in batch.updates {
                 let update =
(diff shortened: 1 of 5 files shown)


sdbondi (commented):
Reviewed at `9d6813c7e`. Nothing blocking — the transport change is coherent and the per-shard framing is correct. Comments below are hardening and nits.

### Verified sound

- `state_transitions_get_starting_at` already stops when the range scan runs onto the next shard (`crates/state_store_rocksdb/src/reader.rs:1742-1748`), so looping the existing per-shard query per cursor genuinely cannot bleed transitions across shard boundaries. The "no storage change needed" claim holds.
- Cursor bound is exactly right, not approximately: the indexer's worst-case request is `Shard::global()` plus a 256-shard group = 257, and `max_cursors` is 257 (`state_sync_task.rs:37`). `all_shard_groups_iter` always starts groups at shard 1 (`crates/common_types/src/num_preshards.rs:51`), so chaining `Shard::global()` can never produce the duplicate that `validate_all` would reject.
- `into_chunks` emits zero chunks for a version with no surviving updates, so `send_batches` sends nothing and the indexer's `has_more`-driven flush can never straddle two shards — the per-stream buffers in `stream_shard_state` are safe.
- Both directions of the breaking change fail loudly rather than silently: old server → new client gets `batch.shard == 0` (rejected as unrequested, unless global was requested) and then no final marker; new server → old client sends empty `cursors`, rejected outright. Given the protocol name is unversioned, loud is the right outcome.
- No new DoS surface: the RPC server already completes one request per substream before reading the next, so 257 cursors in one request costs the same order as the 257 sequential requests it replaces.
- Diff is confined to the five files the change needs — no lockfiles or generated artifacts swept in.

### Non-blocking

**1. The indexer records batch progress with no monotonicity guard, and the round's economic totals are read-modify-write.**

`worker.rs:686` inserts `(shard, state_version)` unconditionally, and `add_state_sync_progress` is a plain `IndexMap::insert` (`sync_plan.rs:34-38`). The new `requested` set checks *which* shard a batch belongs to but not *when* it may arrive, and nothing enforces the contiguity the doc comment on `stream_shard_state` says the caller relies on. A peer that replays a shard's batches after that shard's completion marker rewinds `last_state_versions` for it, and next round those versions are streamed and applied again — `TariAccumulatedClaimed` / `TariAccumulatedFees` / `TariAccumulatedReceiptExhaustBurn` (`worker.rs:721-729`) are summed into the previous value, so a replay double-counts them.

This is pre-existing for in-order replays within a single-shard stream, but the multi-shard stream makes "after the marker" newly expressible, and the validator client already has the guard the indexer lacks (`crates/rpc_state_sync/src/state_sync.rs:288-296` rejects a regressing `state_version`). Since this PR is adding the sibling shard guards anyway, it's the natural place: track completed shards and reject a batch for one, and reject a `state_version` below the last seen for that shard.

**2. "Exactly one stream in the steady state" holds only once every shard has state.**

A shard that has never accumulated any state gets `synced_to = start_state_version - 1 = 0` from `send_complete` (`state_sync_task.rs:211`), because `state_tree_versions_get_latest` returns `None` for a shard whose tree never advanced. The indexer then reads `prev_version == 0` next round and puts it back in the `from_scratch` bucket. So a sparse network stays at two streams per shard group indefinitely. Two versus 257 is still the win the PR is after — just worth correcting in the description rather than the code.

### Nits

- `state_sync_task.rs:111` — `run()` returns `Ok(())` for an empty cursor list, producing a stream with zero messages. Unreachable through `validate_all`, but `StateSyncTask::new` is `pub` and the client contract is now "a final marker always arrives", so a `debug_assert!` or an explicit error would keep the two in step.
- `rpc_impl.rs:498-499` — two `.expect()`s inside `debug!` args just to log the first/last shard. `Displayable` is already imported in this file; `cursors.first().map(|c| c.shard).display()` says the same thing without them.
- `crates/p2p/proto/rpc.proto:241` — `reserved 1, 2;` reserves the numbers but not the names. Adding `reserved "start_state_version", "shard";` stops either name being reintroduced in `SyncStateRequest` with different semantics.
- `worker.rs:598-603` — the marker path opens its own write tx per advancing shard. On a first-round from-scratch sync that is up to 257 sequential transactions, each re-serialising the whole `SyncProgress` map, inside the loop this PR exists to speed up. Keeping the write per marker is the right call for interrupt-resumability, so this is only worth revisiting if it shows up in practice.

sdbondi (commented):
Re-reviewed at `74f4e0568`. Both non-blocking findings and all four nits are resolved. CI is green on this head, including the 49m integration test.

I re-derived the `StreamOrder` invariants against the responder's actual emission order rather than taking them on trust:

- `fetch_next_batch` advances `current_state_version = transitions.state_version + 1` and `state_transitions_get_starting_at` only yields versions at or above the requested one for that shard, so a well-behaved peer always satisfies the strict-increase check. No false rejections.
- `send_batches` computes `has_more: i < num_chunks - 1`, so a single-chunk version is emitted with `has_more == false` and `pending_chunk` is never left set when the marker follows.
- Seeding `committed_versions` at `start_state_version - 1` makes a from-scratch cursor (`start = 1`) reject a v0 batch, which is correct — genesis is committed at v0 and never streamed.
- A shard whose versions are entirely filtered out emits no batch at all, just a marker; `accept_marker` handles that path.

`it_rejects_an_interleaved_shard_mid_version` is the one that earns its keep — that is the case that would have folded two shards' updates into a single commit given the per-stream buffers.

The empty-cursor path now sends `bad_request` and returns `Err` rather than an empty stream, which keeps the responder in step with the client's "a final marker always arrives" contract. Leaving the per-marker write transaction alone is the right call — it is what makes an interrupted stream resumable, and the cost only lands on a first-round sync.

Nothing outstanding from my side. I cannot submit an approval on your own PR, so this needs a second reviewer to formally clear it.

PR #2481, opened 2026-08-25
docs: add template testing tutorial

### LINKED ISSUE #2247: Docs: More detailed instructions on testing
Currently, the documentation just says to check the finished example template tests to get an idea of how to test things. It would be good to have a more detailed testing tutorial covering things like:

1. Basic test setup, including creating accounts
2. Overriding the epoch/epoch hash
3. Examining the contract state directly
4. Running assertions against the error message

xt1085:
## Summary

- add a dedicated template-testing guide covering account setup, epoch and epoch-hash overrides, direct component-state inspection, and rejected-transaction error assertions
- back the guide's Rust snippets with executable engine tests and a drift check
- link the guide from the developer-docs sidebar and the guessing-game tutorial

## Testing

- `cargo test -p tari_engine --test testing_tutorial --test guessing_game`
- `cargo test -p tari_engine --test test consensus::`
- `cargo ci-clippy`
- `cargo +nightly fmt --all` (no source changes)
- `npm ci` and `npm run build` in `docs/developer-docs`
- rendered HTML checks for headings, links, captions, substantive content, and raw MDX leakage
- negative control: changing the documented state path from `$.2` to `$.3` makes the snippet-drift test fail

The full workspace test suite was not run; the targeted engine tests, full-workspace Clippy gate, and documentation build cover the changed scope.

## Scope

Documentation and tests only. No protocol, consensus, execution, networking, wallet, dependency, or public-API changes.

Closes #2247

### DIFF
--- a/crates/engine/tests/testing_tutorial.rs
+++ b/crates/engine/tests/testing_tutorial.rs
@@ -1,0 +1,208 @@
+//   Copyright 2026 The Tari Project
+//   SPDX-License-Identifier: BSD-3-Clause
+
+//! Worked examples for the "Testing your template" guide
+//! (`docs/developer-docs/src/content/docs/guides/testing-templates.mdx`).
+//!
+//! The guide embeds the code below verbatim, so these tests are what keeps it honest:
+//! if an API in the guide changes, one of these tests stops compiling or stops passing.
+//! [`guide_snippets_match_their_source`] checks that the page and its sources have not drifted
+//! apart.
+
+use std::{fs, path::Path};
+
+use tari_ootle_transaction::args;
+use tari_template_lib::{
+    prelude::TARI_TOKEN,
+    types::{Amount, NonFungibleId},
+};
+use tari_template_test_tooling::{TemplateTest, support::assert_error::assert_reject_reason};
+
+const CRATE_PATH: &str = env!("CARGO_MANIFEST_DIR");
+
+#[test]
+fn create_accounts_and_call_a_component() {
+    let mut test = TemplateTest::new(CRATE_PATH, ["tests/templates/guessing_game"]);
+    let template = test.get_template_address("GuessingGame");
+
+    // Creates a fresh key pair, publishes an account component for it and funds it from the
+    // built-in XTR faucet.
+    let (player, _player_proof, player_key) = test.create_funded_account();
+
+    let vaults = test.read_only_state_store().get_vaults_for_account(player).unwrap();
+    assert_eq!(
+        vaults[&TARI_TOKEN].balance(),
+        Amount::from(TemplateTest::FUNDED_ACCOUNT_INITIAL_BALANCE)
+    );
+
+    // `start_game` is owner-restricted, so the transaction is sealed by the key that owns the
+    // component - the harness's own key, which created it in the same transaction.
+    test.execute_expect_success(
+        test.transaction()
+            .allocate_component_address("game")
+            .call_function(template, "new", args![Workspace("game")])
+            .call_method("game", "start_game", args![NonFungibleId::from_string("💎")])
+            .build_and_seal(test.secret_key()),
+        vec![],
+    );
+    let (game, _) = test
+        .read_only_state_store()
+        .get_components_by_template_address(template)
+        .unwrap()
+        .remove(0);
+
+    // The player seals this one with their own key. `guess` is declared `rule![allow_all]`, so no
+    // ownership proof is required: with an empty `proofs` argument the harness derives the proofs
+    // from the transaction signers.
+    test.execute_expect_success(
+        test.transaction()
+            .call_method(game, "guess", args![5u8, player])
+            .build_and_seal(&player_key),
+        vec![],
+    );
+}
+
+#[test]
+fn inspect_component_state_directly() {
+    let mut test = TemplateTest::new(CRATE_PATH, ["tests/templates/guessing_game"]);
+    let template = test.get_template_address("GuessingGame");
+
+    test.execute_expect_success(
+        test.transaction()
+            .allocate_component_address("game")
+            .call_function(template, "new", args![Workspace("game")])
+            .call_method("game", "start_game", args![NonFungibleId::from_string("💎")])
+            .build_and_seal(test.secret_key()),
+        vec![],
+    );
+    let (game, _) = test
+        .read_only_state_store()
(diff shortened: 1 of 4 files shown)


sdbondi (commented):
Thank you for the PR! This fills an important gap in the current docs.

Docs + tests only, reviewed statically — I didn't run cargo (the engine suite is WASM/JIT-heavy, CI covers it). I did re-implement `guide_snippets_match_their_source` independently and ran it against the branch: all 6 titled `rust` blocks match their sources.

Verified along the way: every API the guide names exists with the stated semantics; the documented defaults are real (epoch `0` + zero hash, fees off, auto-proofs on — `template_test.rs:255-265`); `$.2` really is `round_number` given the field order `prize_vault, guesses, round_number`; `FAUCET_AMOUNT == FUNDED_ACCOUNT_INITIAL_BALANCE == 1_000_000_000`; `TARI_TOKEN` is the faucet's resource; `0.39` matches the published max version; and the sidebar slug, `basePath` import and `#speeding-up-template-tests` anchor all resolve.

Nice piece of work — the drift test in particular is the right idea, and the "two things make this a real test rather than a green tick" framing is worth keeping.

## 1. "Signers are not proofs" explains the wrong model

This is the one substantive item. The section frames signer vs proof as two ways of *identifying who submitted the transaction*, but the actual model is: the `proofs` argument **is the transaction's initial authorization scope** — `AuthParams::initial_ownership_proofs`, handed to `AuthorizationScope::new` at `crates/engine/src/transaction/processor.rs:159`. A badge in that scope is a `NonFungibleAddress` — `check_requirement` (`runtime/tracker_auth.rs`) matches virtual badges or held `ProofId`s of any resource without caring which. But on the network the engine only ever seeds that scope with **signer** badges: `transaction_executor.rs:130` builds it from `signers_iter()`, and no other production path constructs `AuthParams`. A non-signer NFT can never appear in a real transaction's initial scope; the harness's `proofs` vec is the only thing that can put one there.

Two things follow that the current text sets a reader up to get wrong:

**A non-empty `proofs` replaces the signer badges, it doesn't add to them.** `template_test.rs:739` only auto-derives when `proofs.is_empty()`. So the guide's advice — "pass proofs explicitly only when the transaction genuinely presents a badge its signer does not hold — an admin badge held by another account, for instance" — yields a scope that is doubly unlike the network's: the signer's own badge is gone, and the admin badge is one the engine would never have injected in the first place. On-chain, holding a badge is exercised by creating a real `Proof` from the vault that holds it (`create_proof_for_resource`, then `authorize_with`) — which is also the only way it crosses a template boundary. The suppression is genuinely useful, but for the opposite purpose to the one the guide gives it: it's how you execute a transaction with a badge deliberately absent and assert the method is denied.

**Badges only reach the top-level call frame.** `working_state.rs:1512` copies the auth scope in only when `call_frame_depth() == 0`; every nested frame starts from `CallScope::new()` → `AuthorizationScope::empty()`. A method that passes when called directly can be denied when reached through another template's `ComponentManager::invoke`. `access_rules.rs:679` already says this in a comment — *"Even though I'm the owner of the resource, the scope does not carry over when cross-template calls are made"* — and it's a bug class this guide is otherwise well placed to teach.

Suggested replacement for the subsection (retitled, e.g. *Proofs are the transaction's initial auth scope*):

```mdx
The `proofs` argument to `execute_expect_success` / `execute_expect_failure` is the transaction's
**initial authorization scope**: the set of badges present when execution starts, which
<a href={basePath("/guides/authorization-and-access/")}>access rules</a> are checked against. A badge is
a non-fungible address, and the engine seeds this scope with one per transaction signer — derived from
the signer's public key — and with nothing else. No other NFT ever lands there on the network; a badge
you hold in a vault is presented by creating a `Proof` from it instead.

`TemplateTest` mirrors that: leave `proofs` empty and it derives the same signer set, which is what you
want whenever authorization is not what you are testing.

Passing a non-empty `proofs` *replaces* that automatic set rather than adding to it, and
`test.disable_auto_add_proofs_from_signers()` turns the derivation off entirely. That is the point of
the argument in a test: it lets you execute a transaction with a badge deliberately missing and assert
that the method is denied. Reach for it to take something away, not to add something — a list that
injects a badge the signer does not hold builds a scope the network cannot produce, and if you do pass
a list, include the signer's own proof (`test.owner_proof()`, or the proof returned by
`create_funded_account`) so you do not silently drop it.

One boundary is worth knowing: these badges are in scope for the top-level call only. A component
reached through another template's `ComponentManager::invoke` starts with an empty badge set, so a
method that passes when called directly can be denied when called across templates. Authorizing across
that boundary needs a real `Proof` — `create_proof_for_resource` on the account, then `authorize_with`
in the callee. `it_permits_cross_template_calls_using_proofs` in `crates/engine/tests/access_rules.rs`
shows both halves.
```

If you want that last paragraph snippet-backed like the rest of the page, citing the existing `access_rules.rs` test in a `title=` block is cheaper than adding a fourth harness build to `testing_tutorial.rs`.

## 2. `crates/engine/tests/test.rs` is now a doc source with nothing saying so

Two blocks in the guide are copied out of `mod consensus`, so editing or reformatting those tests fails `testing_tutorial::guide_snippets_match_their_source` — a test in a different file, with a message pointing at an `.mdx` page. A one-line comment on `mod consensus` stating the invariant would save the next person the hunt.

## Minor

- `tari_template_test_tooling = "0.39"` is the only hard version pin in the docs (`reference/reducing-template-size.mdx:137` uses `version = "*"`), and the drift test only covers `rust` blocks, so it goes stale at 0.40 with nothing to catch it.
- The drift test checks `rust` blocks carrying `title=`; an untitled block silently escapes it. `checked > 0` doesn't catch that. Worth a note in the guide's own contributor-facing comment, if not a stricter assertion.
- Prose says "built-in XTR faucet" while the snippet asserts on `TARI_TOKEN`; XTR is the deprecated spelling user-facing.
- `guides/transaction-overview.mdx:490` still sends readers to the guessing-game guide "for a worked example of testing a complete template", and the guessing-game guide's own "Next Steps" skips past the new page.
- Three tests, three `TemplateTest::new` builds, and tests 2 and 3 repeat the same 12-line allocate/`new`/`start_game` block that isn't part of their snippets — a local helper would be free of doc consequences. Non-blocking; consistent with the rest of the engine tests as-is.

sdbondi (approved):
Thanks

PR #2482, opened 2026-08-26
feat(engine)!: raise the per-transaction WASM metering budget to 250M points

sdbondi:
## Why

The per-transaction WASM metering ceiling was 100M points — about 12ms of validator CPU at the calibrated ~8.4M points/ms. That put non-trivial cryptography out of reach of templates entirely, and it is 24x tighter than the native ceiling a single transaction already enjoys (`MAX_NATIVE_POINTS_PER_TRANSACTION`, 2.4e9 points ≈ 286ms of stealth/confidential verification). Both are real CPU on every replica, both are priced the same, and both are bounded at the block level by `max_block_validation_execution_points` — the asymmetry had no basis.

This raises the ceiling to 250M (~30ms) and adds the measurement infrastructure that justifies the number, so the budget and what templates can actually do with it stay tied together.

## What the measurement says

A Groth16/BN254 verifier implemented entirely in WASM, with no engine support beyond the meter (`crates/engine/tests/templates/zk_verifier`, arkworks 0.5, 417 KiB binary):

| public inputs | in-WASM | native | ratio |
|---|---|---|---|
| 1 | 96.4M pts / 11.5 ms | 7.86M / 0.93 ms | 12.3x |
| 4 | 112.4M pts / 13.4 ms | 9.86M / 1.17 ms | 11.5x |
| 16 | 176.2M pts / 21.0 ms | 18.0M / 2.14 ms | 10.0x |
| 64 | 431.3M pts / 51.4 ms | 49.0M / 5.83 ms | 8.9x |

At the old ceiling nothing fitted except the single-input case, at 96% of budget with nothing left for contract logic. At 250M a verification with a sixteen-element statement fits at 70% of budget, and two verifies fit in one transaction (measured, exactly linear — no amortisation from batching). Sixty-four public inputs does not fit, which is the deliberate edge: cost is linear at ~5.3M points each, so a statement that wide belongs behind a hash, as circuits conventionally put it.

Other figures from the breakdown, which `cargo run -p tari_engine --release --example zk_points_calibrate` prints in full:

- Subgroup checks are the entire cost of proof deserialization: 8.67M points against 15.6k unchecked. Not optional, but not the problem either — the pairings are.
- Compression costs 1.5x; supplying an unprepared verifying key per call costs 1.4x over a stored prepared one.
- The per-public-input marginal is 5.32M points (0.63 ms) in WASM against 645k native.

The ~10x penalty is what a native verification op would recover, and is the case for one — but that is a separate change. This makes the WASM path viable in the meantime.

## Budget arithmetic

The block budget absorbs the raise comfortably. At 8.4M points/ms:

| | before | after | % of 10s block |
|---|---|---|---|
| `MAX_WASM_POINTS_PER_TRANSACTION` | 100M (11.9 ms) | 250M (29.8 ms) | — |
| `max_block_execution_points` (propose) | 4.5e9 (536 ms) | unchanged | 5.4% |
| `max_block_validation_execution_points` | 7.1e9 (845 ms) | 7.25e9 (863 ms) | 8.6% |

The ceiling is sized on two bounds, not one. From below, one verification plus the contract logic around it. From above, the share of a block a single transaction may claim — a block still admits at least 18 max-compute transactions, fixed by `MIN_MAX_COMPUTE_TRANSACTIONS_PER_BLOCK` and asserted by `a_block_admits_enough_max_compute_transactions`, so a leader packing heavy transactions cannot starve a block of everything else. Raising it further trades that granularity away and buys nothing: PLONK/FRI do not fit in WASM at any ceiling the block budget could support.

The validation budget moves only to keep the invariant asserted in `consensus_constants.rs` — `validation >= propose + MAX_WASM_POINTS_PER_TRANSACTION + MAX_NATIVE_POINTS_PER_TRANSACTION` — with margin. The proposal budget is unchanged, so a block packs fewer heavy transactions rather than doing more work.

`FREE_COMPUTE_GRACE_POINTS` is deliberately left at 32M. It is the transaction's unpaid compute surface, and nothing this expensive should be fundable before a fee is paid — there is a test for that.

## Tests

`crates/engine/tests/zk_verify.rs`, 7 tests:

- a valid proof verifies and leaves an explicit headroom margin for contract logic
- the flow a contract actually takes: verifying key in component state, caller supplies only proof and statement
- caller-supplied verifying keys, both encodings
- public inputs the proof does not attest to are rejected, at the same cost as acceptance
- cost grows with public input count and stays inside the budget to 16 inputs
- stacked verifies exhaust the transaction-wide budget rather than each getting a fresh per-call one
- a verify cannot be funded by the fee intent's compute credit

Cost bounds are stated as a band rather than an exact point count so toolchain and arkworks drift do not break them, with a floor so the assertions cannot pass on a template that stopped doing the work. Verified by mutation: inverting the template's verdict fails 4 of the 7.

`metering_budget.rs` now sizes its calls off `MAX_WASM_POINTS_PER_TRANSACTION` instead of a hardcoded 65M, and reads `wasm_execution_points` directly rather than reconstructing the count from the fee breakdown.

## Also here

`TemplateTest::last_execution_points()` — `ExecuteResult` carries the points, but `call_function`/`call_method` return only the decoded value and drop the result, so tests using them could not see what a call cost. Recorded in `try_execute` (the single point every execution path funnels through) and exposed alongside `get_previous_output_address`, returning both halves and a `total()`.

## Versioning

Workspace 0.39.2 -> 0.40.0, minor rather than patch: `max_block_validation_execution_points` is a consensus rule requiring network-wide uniformity, so two validators resolving `^0.39` to different patch releases would disagree on block validity. `tari_template_test_tooling` also gains public API. Tier-3 rollup plus minor bumps on the wallet-tier crates that re-expose the changed types, per `scripts/crate_versioning.py impact tari_engine_types --breaking`.

One pin that workflow does not cover: `crates/ootle_wasm/wasm` pins `ootle-wasm-core` directly rather than through `[workspace.dependencies]`, so it is not in the tier-3 pin list and broke resolution until fixed.

## Notes for review

- Template arguments are `Bytes`, not `Vec<u8>`. A `Vec<u8>` argument encodes as a CBOR array of integers and costs ~130 points per byte to decode — for the 34 KiB prepared verifying key that was more than a tenth of the verification. Worth remembering for any future native `ZkVerifyArg`.
- arkworks is a dev-dependency of `tari_engine` only; nothing ships in a published artifact.
- `publish_crates.py --dry-run` has not been run — that belongs before an actual release.
- Measurements are from one x86 machine. The points are deterministic and so reproducible, but the millisecond and native columns are not.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -230,6 +230,209 @@ dependencies = [
 ]
 
 [[package]]
+name = "ark-bn254"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d69eab57e8d2663efa5c63135b2af4f396d66424f88954c21104125ab6b3e6bc"
+dependencies = [
+ "ark-ec",
+ "ark-ff",
+ "ark-std",
+]
+
+[[package]]
+name = "ark-crypto-primitives"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1e0c292754729c8a190e50414fd1a37093c786c709899f29c9f7daccecfa855e"
+dependencies = [
+ "ahash",
+ "ark-crypto-primitives-macros",
+ "ark-ec",
+ "ark-ff",
+ "ark-relations",
+ "ark-serialize",
+ "ark-snark",
+ "ark-std",
+ "blake2 0.10.6",
+ "derivative",
+ "digest 0.10.7",
+ "fnv",
+ "merlin",
+ "rayon",
+ "sha2 0.10.9",
+]
+
+[[package]]
+name = "ark-crypto-primitives-macros"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e7e89fe77d1f0f4fe5b96dfc940923d88d17b6a773808124f21e764dfb063c6a"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.117",
+]
+
+[[package]]
+name = "ark-ec"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "43d68f2d516162846c1238e755a7c4d131b892b70cc70c471a8e3ca3ed818fce"
+dependencies = [
+ "ahash",
+ "ark-ff",
+ "ark-poly",
+ "ark-serialize",
+ "ark-std",
+ "educe",
+ "fnv",
+ "hashbrown 0.15.5",
+ "itertools 0.13.0",
+ "num-bigint",
+ "num-integer",
+ "num-traits",
+ "rayon",
+ "zeroize",
+]
+
+[[package]]
+name = "ark-ff"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a177aba0ed1e0fbb62aa9f6d0502e9b46dad8c2eab04c14258a1212d2557ea70"
+dependencies = [
+ "ark-ff-asm",
+ "ark-ff-macros",
+ "ark-serialize",
+ "ark-std",
+ "arrayvec",
+ "digest 0.10.7",
+ "educe",
+ "itertools 0.13.0",
+ "num-bigint",
+ "num-traits",
+ "paste",
+ "rayon",
+ "zeroize",
+]
+
+[[package]]
+name = "ark-ff-asm"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "62945a2f7e6de02a31fe400aa489f0e0f5b2502e69f95f853adb82a96c7a6b60"
+dependencies = [
+ "quote",
+ "syn 2.0.117",
+]
+
+[[package]]
+name = "ark-ff-macros"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "09be120733ee33f7693ceaa202ca41accd5653b779563608f1234f78ae07c4b3"
+dependencies = [
+ "num-bigint",
+ "num-traits",
+ "proc-macro2",
+ "quote",
+ "syn 2.0.117",
+]
+
+[[package]]
+name = "ark-groth16"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "88f1d0f3a534bb54188b8dcc104307db6c56cdae574ddc3212aec0625740fc7e"
+dependencies = [
+ "ark-crypto-primitives",
+ "ark-ec",
+ "ark-ff",
+ "ark-poly",
+ "ark-relations",
+ "ark-serialize",
+ "ark-std",
+ "rayon",
+]
+
+[[package]]
+name = "ark-poly"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "579305839da207f02b89cd1679e50e67b4331e2f9294a57693e5051b7703fe27"
+dependencies = [
+ "ahash",
+ "ark-ff",
+ "ark-serialize",
+ "ark-std",
+ "educe",
+ "fnv",
+ "hashbrown 0.15.5",
+ "rayon",
+]
+
+[[package]]
+name = "ark-relations"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ec46ddc93e7af44bcab5230937635b06fb5744464dd6a7e7b083e80ebd274384"
+dependencies = [
+ "ark-ff",
+ "ark-std",
+ "tracing",
+ "tracing-subscriber 0.2.25",
+]
+
+[[package]]
+name = "ark-serialize"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3f4d068aaf107ebcd7dfb52bc748f8030e0fc930ac8e360146ca54c1203088f7"
+dependencies = [
+ "ark-serialize-derive",
+ "ark-std",
+ "arrayvec",
+ "digest 0.10.7",
+ "num-bigint",
+ "rayon",
+]
+
+[[package]]
+name = "ark-serialize-derive"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "213888f660fddcca0d257e88e54ac05bca01885f258ccdf695bafd77031bb69d"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn 2.0.117",
+]
+
+[[package]]
+name = "ark-snark"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d368e2848c2d4c129ce7679a7d0d2d612b6a274d3ea6a13bad4445d61b381b88"
+dependencies = [
+ "ark-ff",
+ "ark-relations",
+ "ark-serialize",
+ "ark-std",
+]
+
+[[package]]
+name = "ark-std"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "246a225cc6131e9ee4f24619af0f19d67761fff15d7ccc22e42b80846e69449a"
+dependencies = [
+ "num-traits",
+ "rand 0.8.6",
+ "rayon",
+]
+
+[[package]]
 name = "arraydeque"
 version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1815,7 +2018,7 @@ dependencies = [
  "tonic 0.12.3",
  "tracing",
  "tracing-core",
- "tracing-subscriber",
+ "tracing-subscriber 0.3.22",
 ]
 
 [[package]]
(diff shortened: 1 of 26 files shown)


sdbondi (commented):
Reviewed at `390cf14a5` against `upstream/development` (`8df86c707`). CI is fully green.

The budget arithmetic checks out: `validation_budgets_always_admit_honest_proposals` still holds with the same 100M margin it had before (7.25e9 vs the 7.15e9 the invariant requires), `on_propose.rs:530` accumulates `total_execution_points()` so the propose budget is the real bound, and `max_block_execution_points` is unchanged so the raise costs block throughput rather than block time. The ark-* crates are genuinely dev-only (`crates/engine/Cargo.toml`, under `[dev-dependencies]`), and `crates/ootle_wasm/wasm/Cargo.toml:16` was the only remaining path+version pin outside `[workspace.dependencies]` in the tree, so that fix is complete. Deriving the test bounds from the constants (`metering_budget.rs`, `tests/max_compute.rs`) is the right direction.

## Blocking

### 1. `ENGINE_FINGERPRINT` must be bumped — stale disk cache diverges consensus

`crates/engine/src/wasm/cache.rs:60`

`MAX_WASM_POINTS_PER_CALL` is the `initial_limit` handed to the metering middleware in `crates/engine/src/wasm/module.rs:179`. wasmer bakes that value into the compiled module: `Metering::transform_module_info` appends a `wasmer_metering_remaining_points` global with `GlobalInit::I64Const(self.initial_limit)`. It is therefore part of the serialized artifact the disk cache stores.

`WasmProcess::metering_allowance` reads it straight back out:

```rust
// crates/engine/src/wasm/process.rs:189
let per_call_cap = match get_remaining_points(store, &self.instance) { ... };
...
points_before: per_call_cap.min(budget_remaining)
```

So on a validator that upgrades with a warm `wasm_cache_dir`, `{addr}_v2.bin` still deserializes to a module whose per-call ceiling is 100M, while a peer with a cold cache compiles one at 250M. A transaction consuming 150M points traps out-of-gas on the first and succeeds on the second — different `ExecuteResult`, different vote. `DiskCachedWasmTemplateProvider` is on the validator's consensus execution path (`applications/tari_validator_node/src/bootstrap.rs:317`, `src/consensus/spec.rs:30`), not just the walletd/indexer dry-run path.

This is the first change to `MAX_WASM_POINTS_PER_CALL` since the `v1` -> `v2` bump in #2228, so it has not been caught incidentally before.

Bump to `"v3"`, and add `limits::MAX_WASM_POINTS_PER_CALL` to the explicit bump-triggers list in that const's doc comment — "middleware list" reads as *which* middlewares, not their parameters, which is how this was missed.

### 2. The shipped reference doc still states the old ceiling

`docs/developer-docs/src/content/docs/reference/fees.mdx:244`

```
| `MAX_WASM_POINTS_PER_TRANSACTION` | 100,000,000 | out-of-gas, hard cap |
```

That table is the user-facing statement of the three stacked ceilings, and the row directly under it (`MAX_NATIVE_POINTS_PER_TRANSACTION`, 2.4e9) is the one this PR's asymmetry argument rests on — leaving 100M there makes the published contrast wrong in the opposite direction.

## Non-blocking

### 3. `ExecutionPoints` was inserted inside `TemplateTest`'s doc comment

`crates/template_test_tooling/src/template_test.rs:91-111`

The "Test harness for Tari Ootle templates / # Quick start" block at :91 now documents `ExecutionPoints`, and `TemplateTest` at :126 has no doc comment at all. Rustdoc renders `ExecutionPoints` with a `TemplateTest::my_crate()` example. Move the new type above the `TemplateTest` doc block (or below the struct).

### 4. Only one of the six crates the versioning script names was bumped

`scripts/crate_versioning.py impact tari_engine_types --breaking` lists six independently-versioned crates that re-expose tier-3 types and need a minor bump when the tier-3 cohort moves. Tier-3 moved 0.39 -> 0.40 here, which breaks their `^0.39` pins, but only `tari_ootle_wallet_crypto` (0.41.0 -> 0.42.0) changed. Unchanged at head: `tari_indexer_client` 0.40.0, `ootle-rs` 0.21.0, `tari_ootle_wallet_sdk` 0.41.0, `tari_ootle_wallet_storage_sqlite` 0.41.0, `tari_ootle_walletd_client` 0.41.0 — and their `[workspace.dependencies]` pins are still 0.40/0.41 to match.

#2483 did bump exactly that set for the tier-3 move it carried, so the omission looks unintentional. The PR description says the bumps were done "per `scripts/crate_versioning.py impact tari_engine_types --breaking`" — either finish the set or record why these five are excluded, because the description currently asserts something the diff does not do.

### 5. The `MIN_MAX_COMPUTE_TRANSACTIONS_PER_BLOCK` doc overstates what the test proves

`crates/engine_types/src/limits.rs:53-57`

> however heavy a block's transactions are, `max_block_execution_points` must admit at least this many of them

The block budget accumulates `execution.result().total_execution_points()` — WASM *and* native (`crates/consensus/src/hotstuff/on_propose.rs:530`). A transaction at `MAX_NATIVE_POINTS_PER_TRANSACTION` is 2.4e9 on its own, so a block admits one of those, not 18. `a_block_admits_enough_max_compute_transactions` divides by the WASM ceiling only, so what it actually fixes is "a block admits 18 WASM-max transactions". That is the property you want here, but the doc claims the stronger, false one. Worth narrowing to "max-WASM-compute", since the whole point of the constant is to be the thing a future reader checks a raise against.

### 6. The floor equals today's ratio exactly

`4_500_000_000 / 250_000_000 == 18`, so the assertion passes with zero slack. That is deliberate per the description, but it means the floor is defined as "whatever the current ratio is" rather than as an independently-derived minimum — it can only ever fail on a future raise, never warn about drift from the other side. A sentence on where 18 comes from beyond the current arithmetic would make it re-derivable.

## Nits

- `utilities/transaction_generator/tests/max_compute.rs:116` — `test.disable_fees()` is a no-op: `TemplateTest::new` sets `enable_fees: false` and nothing enables it in that test. Fine if it is there to make the comment above it load-bearing, but it reads as a fix for a problem that is not present.
- `crates/engine/examples/metering_recost.rs:30` — comment still says "the 100M per-call metering cap". Still true as a bound, but it is a stale number in a file whose job is calibration.
- `crates/template_test_tooling/src/template_test.rs:105-124` — `ExecutionPoints` and `ExecutionPoints::total()` restate `ExecuteResult::{wasm,native}_execution_points` and `ExecuteResult::total_execution_points()` (`crates/engine_types/src/commit_result.rs:63-83`) verbatim. Not wrong — `ExecuteResult` is dropped by the convenience callers, which is the whole reason for the accessor — but a `// mirrors ExecuteResult` pointer would stop the two `total()` definitions drifting.
- `crates/engine/tests/zk_verify.rs:105-110` — the `MIN_CONTRACT_HEADROOM_POINTS` assertion is implied by `MAX_EXPECTED_POINTS` (250M - 160M = 90M >= 60M), so it can never fire independently. It documents intent well, so this is not a request to remove it, just a note that it is not load-bearing as written.

sdbondi (commented):
Re-reviewed at `287206e07`. All five findings verified resolved by reading the code, not by taking the commit messages at their word. CI is 22/22 green.

**1. `ENGINE_FINGERPRINT` — fixed.** `"v2"` -> `"v3"` (`crates/engine/src/wasm/cache.rs:63`), and the bump-triggers list now names `MAX_WASM_POINTS_PER_CALL` explicitly with the mechanism spelled out. That second half is the part that matters: "middleware list" reads as *which* middlewares, and the middleware's *parameter* being baked into the artifact is not something the next person will infer. Stale `_v2.bin` files become orphans under the new suffix and recompile from source, so an upgrading node with a warm cache dir converges on the 250M ceiling.

**2. `fees.mdx:244` — fixed.** 250,000,000, sitting correctly above the 2.4e9 native row.

**3. `ExecutionPoints` placement — fixed.** Moved below the `TemplateTest` struct, so the "Test harness for Tari Ootle templates / # Quick start" block reattaches to `TemplateTest`. Confirmed nothing else was displaced in the move.

**4. Version bumps — complete.** All five remaining crates from the script's set are bumped and their `[workspace.dependencies]` pins match: `tari_indexer_client` 0.41.0, `ootle-rs` 0.22.0, `tari_ootle_wallet_sdk` / `tari_ootle_wallet_storage_sqlite` / `tari_ootle_walletd_client` 0.42.0. The `Cargo.lock` delta is exactly those five `[[package]] version` lines and nothing else. `ootle-wasm-core = "0.40"` and `ootle_ledger_client = "0.6.0"` remain the only path+version pins outside `[workspace.dependencies]`, both correct.

**5. `MIN_MAX_COMPUTE_TRANSACTIONS_PER_BLOCK` doc — fixed.** Narrowed to the WASM ceiling, with the native path explicitly scoped out to `MAX_NATIVE_POINTS_PER_TRANSACTION` / `STEALTH_LIMITS` / `CONFIDENTIAL_LIMITS` and the "far fewer than this many" caveat stated. All three rustdoc links resolve in-module; `docs build` passes.

Nothing blocking left from my side. I can't press Approve on your own PR, so this comment is the sign-off — it needs a second reviewer's approval to merge.

Two nits left open, both your call and neither worth another round-trip:

- `utilities/transaction_generator/tests/max_compute.rs:116` — `test.disable_fees()` is a no-op; fees default off and nothing enables them in that test.
- `crates/engine/examples/metering_recost.rs:30` — comment still says "the 100M per-call metering cap". Still true as a bound, stale as a number, in a file whose job is calibration.

And one thought on `MIN_MAX_COMPUTE_TRANSACTIONS_PER_BLOCK` that the doc fix doesn't reach, offered as a note rather than a request: 18 is `4_500_000_000 / 250_000_000` exactly. Had the constant existed before this PR it would have read 45, and this PR would have edited it down to 18 to pass. A bound rewritten by the change it guards records the status quo rather than constraining it — the next person raising the ceiling gets a failing test that can't tell them whether 18 was load-bearing, and editing the constant is the cheapest way past it. The guard still earns its place (it makes the coupling non-silent, and zero slack is right for that job), but a sentence tying 18 to a property — mempool drain rate, or the share of a block a single leader may monopolise — would make it re-derivable instead of historical.

PR #2487, opened 2026-08-27
feat(consensus): make the substate schema activation schedule per network

sdbondi:
## Motivation

The substate schema activation schedule was a single hardcoded table shared by every network. Networks run at independent epochs, so one epoch value cannot express when a schema goes live — the epoch a new schema activates on esmeralda says nothing about when it should activate on igor.

The table also existed in two independently maintained copies, one gating consensus and one gating substate hashing, which could disagree.

## What changed

**Per-network schedule.** `ProtocolVersion::activations` matches exhaustively on `Network`, so a new network must state its own schedule rather than silently inherit another's. `at`, `newest_scheduled_activation` and `hash_substate` take a `Network`, threaded alongside the `Epoch` that was already plumbed to every hashing site.

A few types that carry the epoch now carry the network with it:

- `SubstateUpdateBatch` — beside its `epoch`; `Block::network()` supplies it on the consensus path
- `WorkingState`, via `StateTracker` and `TransactionProcessor` — follows the existing `burn_rate_bps` path, for `DiffSummary::from_diff`
- `RpcStateSyncClientProtocol`, `CachedSubstateManager`, `NetworkWideStateSync`, `SubstateManager`, `DryRunTransactionProcessor` — each re-derives hashes or verifies substate proofs and would otherwise diverge

Applications pass `config.network`; tests and the bench use `Network::LocalNet`.

**Consolidated `ProtocolVersion`.** Removed the duplicate in `common_types`; it now lives only in `engine_types` and is re-exported from `tari_ootle_common_types`, next to the `Epoch` re-export that was already there.

**Startup guard against a late deploy.** `ProtocolVersion::check_activation_schedule` plus `MetadataKey::ProtocolNewestActivationEpoch` refuse to start a validator node whose binary introduces an activation at an epoch the node has already run past.

This matters because `hash_substate` selects the schema from the epoch a substate was *created* at. An activation landing on epochs that already hold committed substates silently re-hashes them: state sync recomputes roots that no longer match the quorum-signed ones, substate proofs stop verifying, and upgraded nodes diverge from the rest rather than stopping. None of that surfaces as an error at the point it goes wrong.

The comparison is against the schedule the node last started with, not the epoch alone, so restarting after a correctly scheduled activation is not a finding, and a node with no epoch history always passes. `--allow-past-protocol-activation` (or `validator_node.allow_past_protocol_activation`) overrides it for a node whose state is being discarded.

**Removed the `MAX_SUPPORTED` guard** in the hotstuff worker. It compared `ProtocolVersion::at(epoch)` against `MAX_SUPPORTED` — both read from the same binary's table — and `far_future_is_max_supported` pinned the two together, making the condition unsatisfiable. It read as protection against an operator who had not upgraded, which no in-binary check can provide: such a node knows nothing of the new schema, computes the old hashes and carries on. The activation epoch has to sit beyond the window in which every validator and indexer is upgraded, and that is enforced operationally.

## Notes for reviewers

- No behaviour changes for any existing network: every schedule is still `[(Epoch(0), V0)]`, so all hashes are unchanged.
- When a real activation is added, `LocalNet` wants a low but non-zero epoch. At `Epoch(0)` the previous schema would never exist locally and no test would cross a boundary.

## Test plan

- `cargo check --workspace --all-targets` clean
- `cargo test -p tari_engine_types protocol_version` — 8 tests, covering per-network resolution either side of a boundary and the four window shapes the startup guard distinguishes
- Existing substate-hash and epoch-binding tests unchanged


sdbondi (commented):
Reviewed at `c2a7661`. The per-network threading is thorough and the consolidation is right. One blocking hole in the startup guard.

## Blocking

**The startup guard only inspects the newest scheduled activation, so it misses the ordinary skipped-release deploy** — `crates/engine_types/src/protocol_version.rs:79` / `:92`

`check_activation_schedule` passes only `newest_scheduled_activation(...)` into `check_schedule`, and `check_schedule` returns `Ok` early whenever that one epoch is ahead of `last_known_epoch`. An older activation that the node has already run past is never looked at.

Concrete failure, using the schedule shapes this is designed for:

- v1 binary: `[(0, V0)]`. Node runs to epoch 1100, `recorded = None`.
- v2 binary schedules an activation at 1000. Operator never deploys it.
- v3 binary schedules `[(0, V0), (1000, V1), (1200, V2)]`. Operator upgrades v1 → v3 at epoch 1100.

`newest = 1200`, `recorded = None`, `last_known_epoch = 1100`. `1200 <= 1100` is false, so the node starts, records 1200, and proceeds — while every substate it holds from epochs 1000–1100 was hashed under V0 and the rest of the network hashed them under V1. That is exactly the silent divergence the doc comment on `:60-72` says this catches, reached by skipping a release rather than by any exotic schedule edit.

The same gap covers the case where a new binary inserts an activation behind an unchanged newest entry: `newest == recorded` short-circuits at `:92` before anything else is examined.

The rule that closes it is "the newest activation at or before the epoch the node has reached must be one the previous binary already knew about", which needs the whole schedule rather than its last element:

```rust
pub fn check_activation_schedule(
    network: Network,
    recorded: Option<Epoch>,
    last_known_epoch: Option<Epoch>,
) -> Result<Option<Epoch>, PastActivationError> {
    let scheduled = &Self::activations(network)[1..];
    if let Some(last_known_epoch) = last_known_epoch &&
        let Some((activation, _)) = scheduled.iter().rev().find(|(at, _)| *at <= last_known_epoch) &&
        recorded.is_none_or(|r| *activation > r)
    {
        return Err(PastActivationError { activation: *activation, last_known_epoch });
    }
    Ok(scheduled.last().map(|(at, _)| *at))
}
```

All six existing `check_schedule` cases still hold under this (I walked each), and it additionally rejects both the skipped-release path above and a mid-schedule insert. Worth a test for each.

## Non-blocking

**The indexer gets no equivalent guard** — `applications/tari_indexer/src/bootstrap.rs:252,276,319`

The indexer re-derives value hashes in three places the PR touches (`batch_insert_substate_transitions`, `CachedSubstateManager`'s proof verification, `DiffSummary` in the dry-run processor). A late deploy there produces the same wrong hashes; it surfaces as proof-verification failures rather than consensus divergence, which is less bad but still not diagnosed at the point it goes wrong. Reasonable to leave for a follow-up, but the PR description reads as though the deployment risk is covered and it is only covered for the VN.

**A rollback below an already-activated schema passes silently** — `crates/engine_types/src/protocol_version.rs:73`, `applications/tari_validator_node/src/lib.rs:124`

Downgrading to a binary whose schedule has fewer entries gives `newest = None`, so no error, and `if newest != recorded && let Some(newest) = newest` then declines to clear the stale `recorded`. The node happily hashes new substates under the superseded schema. Not the failure mode this PR set out to catch, but the doc on `check_activation_schedule` doesn't scope itself to forward deploys either — either handle it or say it's out of scope.

**Startup refusal exits with a generic error** — `applications/tari_validator_node/src/lib.rs:148`

The two db-setup failures immediately below map to `ExitError::new(ExitCode::DatabaseError, ..)`; this one returns a bare `anyhow::Error`, so an operator hitting a specific, documented, action-required condition gets an unclassified exit code. A dedicated `ExitCode` makes it scriptable.

**`NETWORKS` in the tests is hand-maintained** — `crates/engine_types/src/protocol_version.rs:129`

Adding a `Network` variant breaks the exhaustive match in `activations` (good), but the new network then silently has zero test coverage in all six tests below. Deriving the list from the enum, or asserting `NETWORKS.len()` against something the compiler checks, keeps them in step.

## Nits

- `newest_scheduled_activation` slices `[1..]` (`protocol_version.rs:54`), which panics on an empty schedule. The "index 0 is genesis" invariant is enforced only by `genesis_is_never_a_scheduled_activation`; `.get(1..).unwrap_or_default()` makes it total for free.
- `crates/consensus_tests/src/state_tree.rs:10` and `support/validator/builder.rs:26` import `tari_ootle_transaction::Network`; everything else added in this PR uses `ootle_network::Network`.

## Verified sound

- `block.network()` is safe to feed hash-schema selection at `on_ready_to_vote_on_local_block.rs:549` and `block.rs:681` — `check_network` (`validations/common.rs:47`) rejects a mismatched header before a block reaches either path, and `on_propose.rs:613` uses `config.network`, so proposer and voter resolve the same schedule.
- Every `hash_substate` / `to_value_hash` / `to_tree_change` / `verify_substate_value_proof*` / `SubstateRecord::new` / `SubstateUpdateBatch::new` call site is threaded. The one remaining unthreaded `hash_substate` is in `rpc_state_sync/src/manager_old.rs`, which `lib.rs:9` has commented out.
- The `MAX_SUPPORTED` argument holds: it was a hand-maintained constant read from the same binary as `ACTIVATIONS`, so no operator could ever be on the failing side of the comparison. No references to `MAX_SUPPORTED` or `UnsupportedProtocolVersion` remain anywhere.
- `SubstateUpdateBatch` carries no serde/minicbor/borsh derives, so the added `network` field is not wire- or disk-visible.
- The `Cargo.lock` delta is exactly the eight new `ootle-network` edges — nothing swept in. `ootle-network` is tier 1 and `tari_engine_types` tier 3 in `publish_crates.py`, so the new dependency doesn't disturb publish order.

sdbondi (commented):
Re-reviewed at `242bfc1`. The blocking finding is resolved, and the fix is better than what I proposed. One CI regression from the refactor.

## Blocking finding: resolved

`check_schedule` (`crates/engine_types/src/protocol_version.rs:76`) now compares the whole schedule against the whole recorded schedule, with the "newest activation at or before `last_known_epoch` that the node never ran under" rule. I walked each case against the new code rather than the tests:

- v1 → v3 skipped release at epoch 1100, schedule `[1000, 1200]`, recorded `[]` → `Passed { 1000, 1100 }` ✓ (the case that motivated the finding)
- Mid-insert behind an unchanged newest entry, `[1000, 1200]` vs recorded `[1200]` → `Passed { 1000, 1100 }` ✓
- Moving an already-fired activation, `[900]` vs recorded `[1000]` at 1100 → `Passed { 900, 1100 }` ✓ — I hadn't asked for this one, it falls out of the set comparison
- Normal upgrade ahead of the fork, `[1000]` vs `[]` at 900 → `Ok` ✓; restart after honouring it, `[1000]` vs `[1000]` at 1100 → `Ok` ✓
- Fresh node, `last_known_epoch = None` → `Ok` ✓

The `RolledBack` arm closes the non-blocking rollback item, and correctly distinguishes dropping an activation the node has run under (rejected) from rescheduling one that has not fired yet (allowed) — `[]` vs `[1200]` at 1100 is `Ok`, which is the right call since aborting an unfired fork is a legitimate operation. Storing the full schedule under a renamed metadata key rather than reusing `protocol.newest_activation_epoch` avoids reading the old single-epoch value as a schedule; the old key never shipped, so nothing to migrate.

`cargo test -p tari_engine_types protocol_version` — 12/12 pass locally.

## Blocking (CI)

**`machete` fails: `ootle-network` is now unused in the validator node** — `applications/tari_validator_node/Cargo.toml:25`

Moving the guard into `tari_ootle_app_utilities::protocol_activation` removed the only `use ootle_network::Network` from that crate (`src/lib.rs:46` at the previous head), so the dependency added in the first commit is now dead. Confirmed locally:

```
$ cargo machete applications/tari_validator_node
tari_validator_node -- applications/tari_validator_node/Cargo.toml:
	ootle-network
```

Drop the line. `applications/tari_ootle_app_utilities` and `applications/tari_indexer` both still use it.

## Other findings: resolved, with what I checked

- **Indexer guard** (`applications/tari_indexer/src/bootstrap.rs:210`) — placement is right: it runs before `spawn_service::<IndexerEpochManagerSpec>` at `:220`, so it reads the previous run's epoch rather than this one's. I confirmed the indexer's epoch manager is handed the same `global_db` (`:232`), so `EpochManagerCurrentEpoch` is actually populated there and the guard isn't a silent no-op on the indexer.
- **Exit code** — `ExitCode::DbInconsistentState` (115) exists in `tari_common` and is a reasonable fit; the db-access failures separately map to `DatabaseError` via the `db_error` closure, so the two are distinguishable from a supervisor.
- **`NETWORKS` drift** — deriving from `TryFrom<u8>` over the full byte range works because every variant carries a distinct discriminant (`0x00/0x01/0x02/0x10/0x24/0x26`) and `try_from` matches on those bytes, not on indices. A new variant is picked up without anyone editing the test.
- **`[1..]` panic** — now `.get(1..).unwrap_or_default()`, total.

## Still open (nit, unchanged)

`crates/consensus_tests/src/state_tree.rs:10` and `support/validator/builder.rs:26` still import `tari_ootle_transaction::Network` while the rest of the PR uses `ootle_network::Network`.

Once machete is green I have nothing else blocking.

sdbondi (commented):
Sign-off at `8c29dec`. Nothing outstanding from my side.

`8c29dec` is exactly the one-line dependency drop plus the matching `Cargo.lock` edge — no other change swept in. `cargo machete` is clean across the workspace locally, and the `machete` job now passes.

All 22 checks green at this head, including `clippy`, all three `test` shards, and the 55-minute `integration test`.

Recap of what I verified across the three heads:

- The startup guard's schedule comparison (`crates/engine_types/src/protocol_version.rs:76`), re-derived by hand for the skipped-release, mid-insert, moved-activation, normal-upgrade, restart and fresh-node cases, plus both `RolledBack` directions. `cargo test -p tari_engine_types protocol_version` — 12/12.
- Every `hash_substate` / `to_value_hash` / `to_tree_change` / `verify_substate_value_proof*` / `SubstateRecord::new` / `SubstateUpdateBatch::new` call site is network-threaded; the sole remaining unthreaded one is in `rpc_state_sync/src/manager_old.rs`, which `lib.rs:9` has commented out.
- `block.network()` is safe as a schema selector — `check_network` (`validations/common.rs:47`) rejects a mismatched header before a block reaches either the propose or the vote path, so proposer and voter resolve the same schedule.
- The `MAX_SUPPORTED` removal argument holds; no references to it or to `UnsupportedProtocolVersion` remain.
- `SubstateUpdateBatch` carries no serde/minicbor/borsh derives, so its new `network` field is neither wire- nor disk-visible.
- The indexer guard has real data to work with: its epoch manager is handed the same `global_db`, and the guard runs before that service spawns.
- `ootle-network` is tier 1 and `tari_engine_types` tier 3 in `publish_crates.py`, so the new dependency doesn't disturb publish order.

One cosmetic nit left, take it or leave it: `crates/consensus_tests/src/state_tree.rs:10` and `support/validator/builder.rs:26` import `tari_ootle_transaction::Network` where the rest of the PR uses `ootle_network::Network`.

(Posted as a comment — GitHub won't let me approve my own PR, so this still needs a second pair of eyes to satisfy the review requirement.)

PR #2493, opened 2026-08-31
feat(consensus)!: add a protocol version to the block header and vote signature

sdbondi:
## Description

Adds `protocol_version` to `BlockHeader`, resolved from the network's activation schedule at the block's epoch, and makes both the block hash preimage and the proposal vote signature message select their shape from it. Adds `ProtocolVersion::V1`, which is the version those two preimages change under.

Supersedes #2492, which is folded in here: `tari_sidechain` 5.7.0-pre.5 ships both base-layer changes (tari-project/tari#7983 and tari-project/tari#7982), so the two Ootle halves no longer compile apart.

## Motivation and Context

The base layer verifies a commit proof against the block ID a committee signed. Until now it could only do that by already knowing our activation schedule — an epoch → version table that lives here and moves every fork. A base-layer node with a stale copy silently hashes a post-fork block under the superseded schema and rejects a valid proof.

Making blocks and certificates self-describing moves that decision into the data. `tari_sidechain` reads the schema off the block and never needs the table. On this side, `check_protocol_version` gives a node whose schedule disagrees an explicit error naming both versions, and a deduped alarm telling the operator to upgrade, rather than an opaque signature failure.

Separately, a proposal vote signature covered only `(block_id, decision)`, which is not enough to attribute a vote to a view: two conflicting signatures from one validator were indistinguishable from two honest votes at different heights. From V1 the vote also signs the version, epoch and height, so equivocation is provable from the signatures alone — matching what timeout votes already do.

`SignedMessage` no longer requires `ToSignatureMessage`. A proposal vote's preimage depends on the protocol version, which the vote does not carry, so the vote is paired with its message at the point of verification instead of producing one on its own. Timeout votes, whose preimage is self-contained, are unaffected.

## Which networks move

| Network | Genesis | Basis |
| --- | --- | --- |
| Esmeralda | `V0` | The only network with an Ootle deployment carrying block history. Its genesis version is chain history and cannot move; it takes a scheduled activation when an epoch is decided. |
| Mainnet, Stagenet, Nextnet | `V1` | Never launched. |
| Igor, LocalNet | `V1` | Reset before use. |

The five networks launched directly at `V1` need no activation to coordinate, and a later reset can collapse the table back to a single genesis entry at whatever version is newest then.

**Esmeralda is a rolling upgrade with no reset and no restart coordination**, because it stays on `V0` until an activation is scheduled:

- **Hashing** — `V0` keeps the existing `BlockHeaderHashFieldsV1` preimage and the existing vote message, neither of which carries a version.
- **RocksDB** — `#[cbor(default)]` on the header; epoch checkpoints reach `tari_sidechain` types through `tari_bor::adapters::serde_bridge`, whose new fields are `#[serde(default)]`. Existing rows load and hash unchanged, and an older binary reading a new row ignores the added fields, so a rollback is safe too.
- **p2p** — the new fields are proto3 scalars, so a peer that does not send them reads as `V0`. Old and new nodes produce the same signed bytes and interoperate in either direction.

The other five **do** move at genesis, which is a block-layer fork for any node that already has history on them. `check_activation_schedule` now covers this: it records the genesis version alongside the activation schedule and refuses to start a node with epoch history whose binary launches its network at a different version. Substate hashes are unaffected either way — `SubstateHashMessage::new` maps `V0` and `V1` to the same schema — so this is never state corruption.

When an activation *is* scheduled for esmeralda, every validator must be on the new binary before that epoch, and base nodes and wallets must be on `tari_sidechain` 5.7.0-pre.5+ so that a proof hashed under `BlockHeaderHashFieldsV2` verifies on L1.

## How Has This Been Tested?

`cargo lints clippy --all-targets` clean, and `cargo test` on `tari_ootle_storage`, `tari_consensus`, `tari_consensus_types`, `tari_engine_types`, `tari_state_store_rocksdb` and `consensus_tests` (67 tests, including the `block.json` / `block_with_dummies.json` fixtures, which are unmodified and still hash to the same IDs even though localnet now starts at `V1` — they hash from the stored header field, which is the self-describing property doing its job).

Exercised on a localnet with an activation scheduled mid-run, crossing the boundary with the network live.

Coverage worth pointing at:
- `a_header_encoded_without_a_protocol_version_decodes_as_v0` — strips the trailing element from a real encoding and asserts the decoded header has the same ID and hash.
- `it_hashes_the_header_identically_to_sidechain_header` and `a_vote_signed_here_verifies_in_the_sidechain_crate`, both run for `V0` and `V1`, so the schema selection here and in `tari_sidechain` cannot drift apart.
- `moving_the_genesis_version_under_a_node_with_history_is_rejected`, and `the_live_network_stays_at_its_launched_version` so nobody moves esmeralda without noticing.

Run the tests above. The reviewable claims are that nothing about a `V0` block or vote changes — the untouched fixtures on both sides are the guard — and that the only network with history stays on `V0`.

## Breaking Changes

- [ ] None
- [ ] Requires hard fork
- [x] Other - Please specify

`BlockHeader` and `Block` constructors take a `ProtocolVersion`, and `QuorumCertificate` gains three wire fields. Esmeralda is unaffected: existing blocks decode and hash as `V0`. The five networks launched at `V1` require a reset for any node that already has history on them, which the startup guard enforces rather than leaving to chance.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6879,9 +6879,9 @@ dependencies = [
 
 [[package]]
 name = "minotari_app_grpc"
-version = "5.6.0"
+version = "5.7.0-pre.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bf43c7778bc46555dd2ad90f6b72d4c351676e35937aded1ae354d7abf9c6d4e"
+checksum = "cc43c88c2c11e75ce58ee9cc5d440d07fa16d387c27616a9bbd350b0fc8f8b9d"
 dependencies = [
  "argon2 0.6.0-rc.8",
  "base64 0.22.1",
@@ -6911,9 +6911,9 @@ dependencies = [
 
 [[package]]
 name = "minotari_app_utilities"
-version = "5.6.0"
+version = "5.7.0-pre.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "15ae6f92c622a8d9db7e233ac1dfbb77dc56944b2911202c4b50aa427a781516"
+checksum = "167bccda762ef8603f274049044688f92ed8819da6c44378527df79c80ccafd4"
 dependencies = [
  "clap 4.5.54",
  "dialoguer 0.10.4",
@@ -6935,9 +6935,9 @@ dependencies = [
 
 [[package]]
 name = "minotari_console_wallet"
-version = "5.6.0"
+version = "5.7.0-pre.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "91f7ce08ace070b0869a3b057bee5c819bb66422f2692177868e12a4828911b7"
+checksum = "a1a3899e395ce5294b97797f51823534a9e187965efb82e613ea15371325392a"
 dependencies = [
  "anyhow",
  "chrono",
@@ -6992,9 +6992,9 @@ dependencies = [
 
 [[package]]
 name = "minotari_ledger_wallet_common"
-version = "5.6.0"
+version = "5.7.0-pre.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bf77dbb08e7a1171ab2c1dcede5c96be82cee38a51a07cb0919278a77f64e878"
+checksum = "25cf92dd63d46fbf6a7f54f378043616c8a09ebd40b0038bae3965ed5068016b"
 dependencies = [
  "borsh",
  "bs58",
@@ -7003,9 +7003,9 @@ dependencies = [
 
 [[package]]
 name = "minotari_ledger_wallet_comms"
-version = "5.6.0"
+version = "5.7.0-pre.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7ddd26c4e76ee973b63953974f6aeff5d4780c12311514f4b488f4e86c2a78cf"
+checksum = "2c5e0876b98c67333b0c9a4fca6ee5fa337486dd25ef521c9f65a2027232524d"
 dependencies = [
  "borsh",
  "dialoguer 0.11.0",
@@ -7026,9 +7026,9 @@ dependencies = [
 
 [[package]]
 name = "minotari_node"
-version = "5.6.0"
+version = "5.7.0-pre.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6c0d86939d518eb2c2d21a6570d2e15e89cb2b6e3c0a91bda8dd35b7d0844d2c"
+checksum = "2a57b6a4fc39b9b9855beb6a176d7e561d5f4f71c08783211ddda834063a6312"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -7087,18 +7087,18 @@ dependencies = [
 
 [[package]]
 name = "minotari_node_grpc_client"
-version = "5.6.0"
+version = "5.7.0-pre.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "041d7aeea38f5c17046a28468585e84bb80e611d1c035964b306c10310b3faf0"
+checksum = "9fed967d1e1d772bed299a3b348ac20964b9191137036de73ae2c67c8ff903e8"
 dependencies = [
  "minotari_app_grpc",
 ]
 
 [[package]]
 name = "minotari_node_wallet_client"
-version = "5.6.0"
+version = "5.7.0-pre.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "87af2d0dc032a527c9b60d431a8cf01755df70f42ba826cbd61aa937b878c2cd"
+checksum = "730648bda7111453c42327c5be8df8737fcfb733414705dc7d1d1be6c4791cfd"
 dependencies = [
  "anyhow",
  "async-trait",
@@ -7115,9 +7115,9 @@ dependencies = [
 
 [[package]]
 name = "minotari_wallet"
-version = "5.6.0"
+version = "5.7.0-pre.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9abfeb0fe0cc3e6aed375b048d4ffde266967e6784ee8e06fa86d908763177d3"
+checksum = "fdd1c85dd03800c78f39462641dc5e338b49d045dc13d9405c794e7146c8be00"
 dependencies = [
  "anyhow",
  "argon2 0.6.0-rc.8",
@@ -7174,9 +7174,9 @@ dependencies = [
 
 [[package]]
 name = "minotari_wallet_grpc_client"
-version = "5.6.0"
+version = "5.7.0-pre.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f88ab95235fc67b59ad8edf24759f3b746de9a90e76978dfe55af87641b93ec4"
+checksum = "0e9adb61de06eed8aeff1a697c1856b317ff12ee623e48e86fd24feecac5a5c7"
 dependencies = [
  "minotari_app_grpc",
  "tari_common",
@@ -9356,9 +9356,9 @@ dependencies = [
 
 [[package]]
 name = "randomx-rs"
-version = "1.4.1"
+version = "1.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d108ec2b3f83ea06f8290685fa4a20c8f0803b1de43c2301639d7a0592474554"
+checksum = "34837f0cf043a56a66e7a07f835969714a27fc9b561d4743620c54a165a7fc8f"
 dependencies = [
  "bitflags 1.3.2",
  "cmake",
@@ -11239,9 +11239,9 @@ dependencies = [
 
 [[package]]
 name = "tari_common"
-version = "5.6.0"
+version = "5.7.0-pre.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "891495ada8d868186ef8d60086a0395644da58f24aceb0fb1aa9bf202df5f24f"
+checksum = "5bd1ecf6675e475d82cd01776003f288666c783d3683ce6209a460a618943317"
 dependencies = [
  "anyhow",
  "cargo_toml 0.20.5",
@@ -11265,9 +11265,9 @@ dependencies = [
 
 [[package]]
 name = "tari_common_sqlite"
-version = "5.6.0"
+version = "5.7.0-pre.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "03189effa6f218e3c7bedc4d4e78b85fbd5369b9abb82bd96741f2ae7f1254f5"
+checksum = "ab7eca6b184d142cdf0a458ec20bd5e0b4fdbc3fe0efc9f4dd9d06aa973a7cf5"
 dependencies = [
  "diesel",
  "diesel_migrations",
@@ -11283,9 +11283,9 @@ dependencies = [
 
 [[package]]
 name = "tari_common_types"
-version = "5.6.0"
+version = "5.7.0-pre.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "28c3584f28cc06ad4725623491a21f3b9afff0a7f9c6b96dfcdf20bc773355cf"
+checksum = "5b8e0e188ba9a471c8e3bbe64157805b67fc38df13380c00e5230e4dc9240fe2"
 dependencies = [
  "argon2 0.6.0-rc.8",
  "base64 0.22.1",
@@ -11321,9 +11321,9 @@ dependencies = [
 
 [[package]]
 name = "tari_comms"
-version = "5.6.0"
+version = "5.7.0-pre.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "83b11bec4ea54b75d39556fd6c3a3095b1bec1b6574e68e6ecd9cfd17f998300"
+checksum = "24e254c32979cc1cd900f60f197bcaed172d912166cd8164b6e92f50ecf9aacc"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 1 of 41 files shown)


sdbondi (commented):
Reviewed at `a2f86a3` against `development`, reading the changed files at the PR head plus `tari_sidechain` 5.7.0-pre.5 and `tari_bor`'s serde bridge.

## Blocking

**`protocol_version` is typed as `number` in the bindings but serialises as a string.**
`bindings/src/types/BlockHeader.ts:86` / `crates/storage/src/consensus_models/block_header.rs:107`

`ProtocolVersion` derives `Serialize` with no representation attribute (`crates/engine_types/src/protocol_version.rs:12-24`), and `#[repr(u32)]` does not reach serde. A fieldless enum serialises as its variant name, so JSON-RPC emits `"protocol_version": "V0"`, not `0`. The sibling precedent is `QuorumDecision`, also a fieldless foreign enum, which is `ts(type = "string")` — `bindings/src/types/ProposalCertificate.ts:14`. Any TS consumer that compares or does arithmetic on `header.protocol_version` gets a string back.

Two ways out, and I'd prefer the second:
- declare it honestly: `ts(type = "\"V0\" | \"V1\"")`, or drop the override and let ts_rs generate `ProtocolVersion.ts`;
- or make serde agree with the other two encodings — minicbor writes a `u32` (`protocol_version.rs:187-209`) and proto3 writes a `uint32` — by putting `#[serde(into = "u32", try_from = "u32")]` on the enum and keeping `number`. That also gets the "unknown version names itself in the error" property on the serde path, which the minicbor impl already documents as the reason it encodes numerically.

Worth deciding now either way: nothing persists a `ProtocolVersion` through serde today (`protocol_activation.rs:35` stores `Vec<Epoch>`), so the representation is still free to change. Once a header with the field is written to a JSON-facing store it isn't.

## Non-blocking

**A laggard's stall is indistinguishable from ordinary proposal noise.** `crates/consensus/src/hotstuff/worker.rs:1441-1448`

`InvalidProtocolVersion` falls into the generic `warn!("⚠️ Proposal validation error: {err}")` arm and returns `Ok(())`. It is the one error in that set — along with `InvalidEpochHash` directly above it — that means *this node* is wrong and is now permanently stalled, not that a peer sent something bad. `InvalidEpochHash` gets a deduped 🚨 alarm that names the cause and the operator's options; this gets one line per rejected proposal, forever, with no "upgrade this binary" anywhere. The PR sells `check_protocol_version` as giving a laggard "an explicit error naming both versions instead of an opaque signature failure" — the error exists, but at the point the operator sees it, it reads like every other rejected proposal. Same alarm treatment would land the feature.

**`calculate_hash` gives up exhaustiveness where the schema decision actually lives.** `crates/storage/src/consensus_models/block_header.rs:352`

`match self.protocol_version.as_u32() { 0 => …, protocol_version => … }`. `tari_sidechain` has to use a catch-all because it receives a raw `u32` off the wire; here it's a closed enum. `substate_hasher.rs:32` matches the variants (`V0 | V1`), so adding a `V2` that changes the header preimage compile-errors there and forces a decision, but silently keeps hashing under `BlockHeaderHashFieldsV2` here. Matching `ProtocolVersion::V0 => … , ProtocolVersion::V1 => …` keeps the comment's "must stay identical to `SidechainBlockHeader::calculate_hash`" honest by making divergence a build failure rather than a runtime one.

**Commit proofs are self-describing for the header but not for the certificates.** `crates/consensus/src/hotstuff/commit_proofs.rs:225` vs `:260`

The header takes its version from the block (`header.protocol_version()`), but every QC element takes it from `ProtocolVersion::at(network, qc.epoch())` — the activation table the PR is trying to remove from the verification path. The value is right (`qc.epoch()` is the justified block's epoch, which is what selects the message, and it handles a proof straddling an activation correctly), but a stale table here produces a proof that L1 rejects, which is the failure mode the motivation section says this removes. `generate_block_commit_proof` already walks the justified blocks, so the version could come off the stored header. Carrying it on `ProposalCertificate` would be the fuller fix — follow-up, not this PR.

**The sign/verify asymmetry is correct but implicit.** `on_receive_local_proposal.rs:821` signs with `block.header().protocol_version()`; `proposal_collector.rs:146` and `validations/common.rs:306` verify with `ProtocolVersion::at(network, epoch)`. These agree only because `check_protocol_version` pins the header's version to `at(network, header.epoch())` before any vote is cast, and that guarantee lives in a different file. A line at `common.rs:306` saying why the schedule lookup is equivalent to the header field would save the next reader the trip.

## Nits

`crates/consensus_types/src/certificates/proposal_vote.rs:22` — "this is not validated explicitly and is mainly used to determine message age and ordering" is now understated: from V1 `block_height` is part of the signed preimage. (It is still not checked against the block, and it doesn't need to be — `VoteStoreInner` keys votes by `(epoch, height)` with `block_id` as the aggregation key, so a vote carrying a wrong height lands in its own bucket and can never be folded into an honest quorum. The comment just no longer describes what the field does.)

`crates/storage/src/consensus_models/block_header.rs:239` — `zero_block` uses `ProtocolVersion::at(network, Epoch::zero())`, which is `V0` by construction for every network (`protocol_version.rs:249` asserts `activations()[0] == (Epoch(0), V0)`). The zero block's ID is hardcoded to zero and never hashed, so the field is inert; `ProtocolVersion::V0` reads more plainly than a lookup that can only return one answer.

## What I checked and found sound

- **v0 byte-identity holds end to end.** `BlockHeaderHashFields::V1` is still borsh discriminant 0 in 5.7.0-pre.5 (5.6.0's enum had exactly one variant, and V2 is appended), and `ProposalVoteMessage`'s hand-written `BorshSerialize` writes no discriminant, so a V0 vote preimage is byte-for-byte the old `ProposalCertificateSignatureFields`. The untouched `block.json` fixtures are a real guard, not a nominal one.
- **The rolling-upgrade and rollback claims check out.** `tari_bor`'s serde bridge writes structs as string-keyed CBOR maps (`serde_codec.rs:365,485`) and `deserialize_ignored_any` skips unknown keys (`serde_codec.rs:778`), so an old binary reads a new commit proof and a new binary reads an old one — the `#[serde(default)]` on the three new `QuorumCertificate` fields does the rest. `#[cbor(default)]` on a trailing minicbor array element matches existing practice across the repo, and the strip-the-last-element test pins it.
- **A QC's signed tuple can't disagree with itself.** `create_proposal_certificate` takes epoch/height from the stored `Block`, and `VoteStoreInner` buckets by `(vote.epoch, vote.block_height)` keyed on `block_id`, so every signature folded into one certificate signed the same `(version, block_id, decision, epoch, height)`.
- **The genesis exemption in `check_quorum_certificate_signatures` still short-circuits** before any message is built, so threading `network` through doesn't touch it.
- **A tampered version can't ride someone else's block ID.** `conversions/consensus.rs:511,532` rebuild the header through `create`/`dummy_block`, which recompute the ID from the wire fields on both paths; proto3 defaults the new scalar to 0/V0 for a peer that doesn't send it.
- Dep bump and `Cargo.lock` churn are the folded-in #2492; nothing unrelated swept in.

sdbondi (commented):
Re-reviewed `fc3ffb9` → `e88559c`. All six items from the previous round are genuinely fixed — details below. One thing came in with the fix commit that shouldn't have.

## Blocking

**A debugging scratch tool got committed.** `crates/state_store_rocksdb/examples/dump_substate.rs` (new file), plus the `ootle-network` and `serde_json` dev-deps added for it in `crates/state_store_rocksdb/Cargo.toml:34-35` and the resulting `Cargo.lock` churn.

Nothing about it belongs to this PR: it hardcodes `Shard::from_u32(1)` and `rec.version == 0`, indexes `args[1..3]` without a bounds check or a usage message, and dumps substate values as JSON to stdout. It's a one-off state-inspection tool, not an example of how to use the crate.

CI already agrees on the new head — `file licenses` and `fmt` both went from pass to fail, and both point here: the file has no `// Copyright` / `SPDX-License-Identifier` header (`scripts/file_license_check.sh`), and line 20 is 124 characters against `rustfmt.toml`'s `max_width = 120`.

Drop the file and the two dev-deps. If it's worth keeping, it's worth keeping in its own PR with a header, argument handling, and the shard/version filters as arguments.

## Verified fixed

- **serde representation.** `protocol_version.rs:18` now carries `#[serde(into = "u32", try_from = "u32")]` with `From<ProtocolVersion> for u32` alongside the existing `TryFrom<u32>`, whose `UnknownProtocolVersionError` is a `thiserror::Error` and so satisfies serde's `Display` bound. JSON now emits `0`, which makes `bindings/src/types/BlockHeader.ts:86`'s `number` correct, and all four encodings (borsh discriminant, minicbor `u32`, proto3 `uint32`, serde) agree. `BlockDetails.tsx:270` renders it as `V{…}`, which reads right for a number and would have read as `VV0` under the old string encoding. The header comment explaining the four-way agreement is the right place for it.
- **Laggard alarm.** `worker.rs:1397-1416` + the match arm at `:1463`. Deduped on `(epoch, expected, block)` in `WorkerState`, mirrors `InvalidEpochHash`, and the message names both versions and says what the operator has to do. Returns `Ok(())` so the node keeps running, same as the sibling.
- **Exhaustive match.** `block_header.rs:352` matches `ProtocolVersion::V0` / `ProtocolVersion::V1` on the enum. A future `V2` now fails the build here as well as in `substate_hasher.rs`, which is what the "must stay identical to `SidechainBlockHeader::calculate_hash`" comment needs to be enforceable.
- **Certificate versions come off the justified block.** `commit_proofs.rs:248-260`. I checked the two call sites resolve to the right block: at `:129` `block` is loaded as `Block::get(tx, &commit_qc.calculate_block_id())`, which is by definition what `commit_qc` justifies; in the loop at `:154` the guard is `block.justifies_parent()`, defined as `justify.calculate_block_id() == *self.parent()` (`block.rs:362`), so `block.justify()` justifies exactly the `parent` now passed in. Hoisting `get_parent` above the push is the same single read, just ordered so the parent is in hand — no extra DB work. Commit proofs are now schedule-free end to end.
- **`zero_block`** is `ProtocolVersion::V0` (`block_header.rs:239`), and the `check_quorum_certificate_signatures` comment at `common.rs:305` states the `check_protocol_version` invariant that makes the schedule lookup equivalent to the header field.
- **`ProposalVote::block_height` doc** (`proposal_vote.rs:21-24`) now says it's part of the signed preimage from V1 and records the bucketing argument for why a disagreeing height can't be folded into an honest quorum.

Once the example is out and CI is green this is good to go from my side.

sdbondi:
Correct, and my fault — `git add -A` swept up an untracked scratch file that was sitting in the working tree, along with the dev-deps it needed. Nothing to do with this PR.

Amended to `cbca2846`: the example, the `ootle-network` / `serde_json` dev-deps in `crates/state_store_rocksdb/Cargo.toml` and the `Cargo.lock` churn are all gone. The commit is back to the seven files it should have touched, and `git diff` against the previous head shows those three as the only difference.

Both CI checks pass locally now — `scripts/file_license_check.sh` and `cargo +nightly-2025-12-05 fmt --all --check` — as does `cargo lints clippy --all-targets`.

The tool itself is kept aside, not lost; if it earns a place it can come back with a licence header, argument handling, and the shard/version filters as arguments, in its own PR.

sdbondi (commented):
`cbca284` → `9fc03a6` cleared the stray example, and `9fc03a67e` ("start unlaunched networks at protocol version 1") is a substantive change of direction rather than a review fixup. Everything from rounds 1 and 2 is still in place, and reverting `zero_block` to `ProtocolVersion::at(network, Epoch::zero())` is now correct — my earlier nit assumed a genesis version that was uniform across networks, which it no longer is. Two blocking items on the new commit.

## Blocking

**The startup guard does not cover a change to the genesis entry.** `protocol_version.rs:42-50`, `:76-77`, `:109`

`check_activation_schedule` compares `scheduled_activations`, which is `activations(network).get(1..)` — everything *after* genesis. Moving index 0 from `V0` to `V1` leaves `scheduled_activation_epochs` empty before and after, so `recorded` and `to_record` are both `[]`, `check_schedule` returns `Ok` on the first branch, and `protocol_activation.rs:43` records nothing new. The guard is structurally blind to exactly the edit this commit makes.

The consequence for a node on MainNet, StageNet, NextNet, Igor or LocalNet that has block history: it starts clean on this binary, then produces V1 block IDs and signs V1 vote preimages while the rest of its committee is still on V0. Each side rejects the other's proposals. The `InvalidProtocolVersion` alarm added in the previous round does fire — but only once the node has already stalled, and it tells the operator to upgrade this node, which is the opposite of what the operator of the node that just upgraded needs to hear.

Blast radius is bounded: `substate_hasher.rs:32` maps V0 and V1 to the same substate schema, so committed state hashes are untouched. This is a block-layer fork, not state corruption.

All of which is fine *if* none of those five networks has a running Ootle deployment. That assumption is doing all the work here and the code neither records nor enforces it. MainNet/StageNet/NextNet not having launched is easy to believe and LocalNet is disposable by definition; Igor is the one I'd want stated explicitly, since it has a base-layer life of its own. Two things worth doing:

- Say it in the `activations` doc comment, network by network — the current wording ("A network with no history to preserve may start at any version") states the rule but not which networks it was applied to or on what basis. The next person to touch a genesis entry needs to know the guard will not catch them.
- Consider extending `check_schedule` to compare the genesis version when `last_known_epoch` is `Some`. It already has the epoch history in hand, and a node that previously ran under a different genesis version is in the same position as one that missed a scheduled activation — which is the case the function exists to catch.

**The PR description now contradicts the code.** Four places:

- "Adds `ProtocolVersion::V1`, **not scheduled on any network**. `ProtocolVersion::at` resolves every epoch to `V0`, so blocks are produced, hashed, signed and stored exactly as they are today."
- "## No reset or coordinated restart required" — that heading now holds for esmeralda only.
- "Every v0 encoding is byte-identical to today, and V1 is unscheduled, so this is a normal rolling upgrade."
- Breaking Changes: "No network data is invalidated and no restart coordination is needed."

Five of six networks now produce V1 blocks from genesis. The description should name which networks move, give the reason each is safe to move, and say that esmeralda stays on V0 because it is the live one.

## Non-blocking

**`every_network_starts_at_its_genesis_schema` no longer asserts anything.** `protocol_version.rs:241-246`

`at(network, Epoch(0))` resolves by construction to `activations(network)[0].1`: the reverse scan takes the last entry with `at <= 0`, and `genesis_is_never_a_scheduled_activation` already forbids a second `Epoch(0)` entry. The test it replaced pinned a real property of the table ("every network starts at V0"); this one restates the implementation. The property actually worth pinning is the one this whole commit turns on — that esmeralda, the live network, stays at `V0`, so nobody moves it without noticing.

CI restarted on the new head; `integration test` is the only job still outstanding as of writing.

PR #2495, opened 2026-09-01
fix(state-sync): start the state stream at the first unpersisted version

sdbondi:
## Problem

A validator node on 0.39.3 (esmeralda) wedged permanently with:

```
JMT Storage error: A node v6384:f9 expected to exist (according to JMT logic) was not found in the storage
```

Once it appears, the node is unrecoverable without a database wipe — every block evaluation and every subsequent state sync fails on the same missing node, so the consensus state machine loops `CheckSync -> Syncing -> Failure -> Sleeping` forever.

## Root cause

`start_state_sync` opened the stream at the shard's **already-persisted** state version:

```rust
let start_state_version = maybe_persisted_state_version.map_or(1, |v| v.max(1));
```

The stream is inclusive of `start_state_version` (the server's `fetch_next_batch` starts at exactly that version, and the client rejects any batch below it), so the peer replays a version the client has already written, and the client writes it again:

```
🛜Syncing from v6384
🛜 Committing 2 state tree changes batch v6384
Batch inserting 5 new nodes and recording 5 stale tree nodes
```

JMT nodes are keyed by `(version, nibble_path)`. Writing a version a second time overwrites the live nodes at those keys, while `batch_put_substate_changes` records those **same keys** as stale at that version. An hour later the stale-node GC (`state_tree_nodes_clear_stale`) reaches them once `max_version - state_history_length` passes that version and deletes them out from under the current tree:

```
Deleted 1126 stale nodes in shard Shard(1) in 46.57ms to version 6466
```

Note the equal new/stale counts in the sync log — that is the signature of the self-overwrite.

Observed sequence on the affected node (times UTC):

| Time | Event |
|---|---|
| 09:54:25 | last shard-1 diff committed: `v6384` |
| 11:01:54 | epoch rollover sync: `🛜Syncing from v6384`, rewrites v6384 |
| 12:01:57 | GC: `Deleted 1126 stale nodes in shard Shard(1) … to version 6466` |
| 13:05:28 | `A node v6384:f9 expected to exist … was not found` — permanent |

## Fix

- Start at `persisted + 1`. This matches what the indexer's sync worker already does (`network_state_sync/worker.rs`: `start_state_version: prev_version + 1`). The server already terminates cleanly with just a completion marker when the start version is past its tip, so no protocol change is needed. The `Some(0)` (freshly bootstrapped, genesis at v0) case still starts at 1, which is also the minimum the peer accepts.
- Reject a non-monotonic version in `calculate_substate_changes` so a same-version write fails loudly instead of silently corrupting the tree. No production caller writes a version twice — consensus (`sharded_state_tree`) always uses `current + 1`, genesis writes once at `GENESIS_STATE_VERSION`. The existing `hash_allows_putting_in_same_version` test covered that now-forbidden capability and is replaced by tests asserting it is rejected.

## Testing

`cargo test -p tari_state_tree` and `cargo test -p tari_state_store_rocksdb --test substates` pass.

## Not addressed here

The same incident had an earlier, non-fatal trigger that is worth a separate issue: after `Syncing::on_enter` clears the transaction pool, the very next epoch's blocks re-proposed a cleared transaction, and `evaluate_local_only_command` no-votes any block whose transaction is not in the pool ("This is likely a previous transaction that has been re-proposed"). The node was then unable to vote for a whole epoch (~65 min) and its state fell behind — which is what put it into the sync path above. It self-healed at the next epoch boundary, so it is a liveness bug rather than a corruption one, and the fix is a consensus behaviour change that deserves its own PR.

### DIFF
--- a/crates/consensus/src/hotstuff/substate_store/sharded_state_tree.rs
+++ b/crates/consensus/src/hotstuff/substate_store/sharded_state_tree.rs
@@ -1,7 +1,7 @@
 //    Copyright 2024 The Tari Project
 //    SPDX-License-Identifier: BSD-3-Clause
 
-use std::collections::HashMap;
+use std::{collections::HashMap, ops::Deref};
 
 use indexmap::IndexMap;
 use log::*;
@@ -176,7 +176,11 @@ impl<TTx: StateStoreReadTransaction> ShardedStateTree<&TTx> {
     }
 }
 
-impl<TTx: StateStoreWriteTransaction> ShardedStateTree<&mut TTx> {
+impl<TTx> ShardedStateTree<&mut TTx>
+where
+    TTx: StateStoreWriteTransaction + Deref,
+    TTx::Target: StateStoreReadTransaction,
+{
     pub fn commit_diffs(
         &mut self,
         diffs: IndexMap<Shard, Vec<PendingShardStateTreeDiff>>,
--- a/crates/rpc_state_sync/src/state_sync.rs
+++ b/crates/rpc_state_sync/src/state_sync.rs
@@ -178,10 +178,14 @@ where TConsensusSpec: ConsensusSpec<Addr = PeerAddress>
             return Ok(None);
         }
 
-        // Bootstrapped genesis state is committed at version 0; consensus/sync state changes begin at
-        // version 1. A freshly bootstrapped node persists version 0, so clamp the sync start to 1 (the
-        // peer rejects start_state_version 0, and genesis is never synced - every node bootstraps it).
-        let start_state_version = maybe_persisted_state_version.map_or(1, |v| v.max(1));
+        // The stream is inclusive of start_state_version, so it must be the first version we have not
+        // yet persisted. A persisted version must never be written a second time: JMT nodes are keyed
+        // by (version, nibble_path), so rewriting a version overwrites live nodes and records those
+        // very keys as stale at that version, and the stale-node GC then deletes them from under the
+        // current tree. Bootstrapped genesis state is committed at version 0 and is never synced -
+        // every node bootstraps it - so a freshly bootstrapped node starts at version 1, which is also
+        // the minimum the peer accepts.
+        let start_state_version = maybe_persisted_state_version.map_or(1, |v| v + 1);
         let mut last_state_version = start_state_version;
         info!(
             target: LOG_TARGET,
--- a/crates/rpc_state_sync/src/state_sync.rs
+++ b/crates/rpc_state_sync/src/state_sync.rs
@@ -238,7 +238,8 @@ where TConsensusSpec: ConsensusSpec<Addr = PeerAddress>
                     if local_state_root != checkpoint_shard_root {
                         error!(
                             target: LOG_TARGET,
-                            "❌ State root mismatch for {shard}. Checkpoint {expected} but got {actual}. Rolling back.",
+                            "❌ State root mismatch for {shard}. Checkpoint {expected} but got {actual}. Everything \
+                             streamed so far is already committed; the next peer resumes after it.",
                             expected = checkpoint_shard_root,
                             actual = local_state_root,
                         );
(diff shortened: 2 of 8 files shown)


sdbondi (commented):
Reviewed at `b7082942`. The diagnosis and the fix both check out against the code.

Verified:

- The stream really is inclusive of `start_state_version`: `state_transitions_get_starting_at` range-scans `(req_shard, state_version)..` (`crates/state_store_rocksdb/src/reader.rs:1732`) and `run_for_shard` seeds `current_state_version = cursor.start_state_version` (`applications/tari_validator_node/src/p2p/rpc/state_sync_task.rs:143`), so the old `v.max(1)` did replay the persisted version.
- Past-the-tip start terminates cleanly: `fetch_next_batch` returns `Ok(None)` → `break` → `send_complete`, as the description claims. Bounded requests report `last_sent_version.unwrap_or(start - 1)` (`state_sync_task.rs:215`), which the client does not trust anyway.
- `Some(0)` → 1 and `None` → 1 both still satisfy the server's `start_state_version == 0` rejection (`state_sync_task.rs:62`).
- No production caller can now trip `NonMonotonicVersion`: consensus uses `next_version = current_version.unwrap_or(0) + 1` (`crates/consensus/src/hotstuff/substate_store/sharded_state_tree.rs:91`), genesis passes `None` once per shard (`applications/tari_validator_node/src/genesis_state.rs:221`), and the sync client's per-version commit advances `maybe_persisted_state_version` before the next batch (`crates/rpc_state_sync/src/state_sync.rs:369-371`).
- `IsNotFoundError` uses `matches!`, so the new variant doesn't silently change not-found classification, and `StateTreeError` isn't matched exhaustively anywhere outside the crate. `tari_state_tree` isn't in `publish_crates.py::CRATES`, so the added variant carries no semver cost.
- The version-goes-backwards case is safe: `state_tree_truncate_to_version` deletes every node and stale record above the target and resets the pointer to the highest surviving version (`utilities/validator_rollback/src/storage/write.rs:86-146`), so `current_version` stays equal to the max version present in the store — which is what makes the `version <= current_version` guard equivalent to "these keys are free".
- Removing `hash_allows_putting_in_same_version` loses no live capability: `put_substate_tree_changes` is called exactly once per block (`crates/consensus/src/hotstuff/common.rs:256`), so nothing batches multiple writes into one version.

### Non-blocking

1. `crates/consensus/src/hotstuff/substate_store/sharded_state_tree.rs:202` — the guard sits on the *compute* path, but consensus computes against a `StagedTreeStore` and writes to the DB later, through `commit_diff`, which calls `record_stale_tree_nodes` / `insert_nodes` / `set_state_version` with no monotonicity check at all. So for the path that produces the vast majority of tree writes, the new invariant is enforced on a throwaway in-memory store and not on the thing that can actually be corrupted. `batch_put_substate_changes` in state sync is the only path where compute and DB write are the same call. If the point is "fail loudly instead of corrupting", the check wants to be at the write funnel — either in `commit_diff` or in `ShardScopedTreeStoreWriter::set_state_version` (`crates/storage/src/state_store/shard_scoped_state_tree.rs:62`), which every writer including genesis and sync already goes through.

2. `crates/rpc_state_sync/src/state_sync.rs:236-246` — the mismatch branch logs "Rolling back." but nothing is rolled back: each version was committed in its own `with_write_tx`, and the root check runs afterwards in a read tx. The peer loop then re-reads `state_tree_versions_get_latest` (line 582) and, with this PR, starts strictly *after* whatever the failed peer wrote. Pre-existing, and the old inclusive start only ever re-fetched the last version so it was no better — but it is the remaining half of the same failure class, and the log line is actively misleading while debugging exactly this kind of incident.

3. No test covers the actual regression — the two new tests pin the `state_tree` guard, which is the backstop, not the off-by-one. The client has no test harness at all, so I don't think that blocks anything, but it does mean the thing that broke esmeralda is still only protected by the guard in (1), which as noted doesn't cover the consensus write path.

### Nit

`StateTreeError::NonMonotonicVersion` names its field `next_version` while `calculate_substate_changes` takes the parameter as `version`; the sibling methods all call it `next_version`. Renaming the parameter would make the guard read consistently with the rest of the file.

PR #2496, opened 2026-09-01
fix(consensus): re-sequence a proposal's transaction when its pool record is gone

sdbondi:
## Problem

A validator node that clears its transaction pool cannot vote on any block carrying a transaction it still holds, for the rest of the epoch.

Seen on esmeralda: after an epoch rollover the node state-synced, `Syncing::on_enter` cleared the pool, and the very next epoch's block 3 re-proposed one of the cleared transactions:

```
🧹 Cleared 2 transaction(s) from the pool after state sync
...
⚠️ Local proposal received ([NodeHeight(3), ... 2 cmd(s)]) for transaction 1e3ee6ff… which is
   not in the pool. This is likely a previous transaction that has been re-proposed. Not voting on block.
❌ State Merkle root disagreement for block [NodeHeight(4) …]
```

From that point its state fell behind the committee and every subsequent block failed the merkle root check. The node was out of consensus for ~65 minutes until the next epoch boundary.

## Why parking does not cover this

The missing-transaction path keys on the transaction **record**: `check_for_missing_transactions` calls `TransactionRecord::get_missing`, which reads the transactions table. `clear_transaction_pool` empties a different table — the **pool record** (stage, evidence, decision).

The body was never gone, so the block logged `✅ has no missing transactions`, was persisted, and only failed later in `evaluate_local_only_command` with `TransactionNotInPool`. There is also nothing to request from a peer: a pool record is derived local state, not something a peer can send.

So the clear is fine; what was missing is any path that re-derives a pool record for a transaction we already hold.

## Fix

Widen the readiness check from "which of this block's transactions do we not have?" to "which are not ready for consensus?":

- missing record → park and request (unchanged)
- record present, pool record absent → re-sequence locally via `OnReceiveNewTransaction::resequence_known_transactions`, which runs the same `validate_new_transaction` → `insert_new_batched` path that `process_requested` already runs for peer-fetched transactions

Safety is enforced where it already was, rather than by wholesale deletion. `validate_new_transaction` refuses to sequence an id whose finalized decision is a commit, or whose `TransactionReceipt` exists in state — explicitly covering freshly-synced nodes. So re-sequencing can never resurrect a transaction the synced state has committed, which is exactly what the clear exists to prevent.

Additionally narrowed: `resequence_known_transactions` skips any id with *any* finalized record. Whether a previously-aborted id may be sequenced again is the mempool's decision; repairing the pool during block validation must not trigger the abort-replay path as a side effect.

All five `TransactionNotInPool` no-vote sites are in `on_ready_to_vote_on_local_block`, reached from both paths this repair covers (live local proposal and catch-up import).

This is also a step toward the `TODO` already sitting above `clear_transaction_pool`: *"the transaction pool only holds pending (derived) state and should not be persisted at all"*. Once the pool is re-derivable on demand, dropping it after a sync is harmless.

## Testing

New `consensus_tests::transaction_missing_from_pool_is_resequenced_from_its_record`: two validators, proposals held back so the drop point is deterministic, validator "2"'s pool record cleared the way a state sync does, then consensus released. Quorum requires "2", so the transaction only commits if "2" re-derives its pool record.

Verified as a real regression test: with the re-sequencing disabled, zero blocks commit and the test fails; with it, the transaction commits on both validators.

## Related

Same incident, separate root cause: #2495 fixes the state-sync off-by-one that corrupted the JMT and permanently wedged the node.

### DIFF
--- a/crates/consensus/src/hotstuff/on_message_validate.rs
+++ b/crates/consensus/src/hotstuff/on_message_validate.rs
@@ -244,9 +244,10 @@ impl<TConsensusSpec: ConsensusSpec> OnMessageValidate<TConsensusSpec> {
             });
         }
 
-        let missing_tx_ids = self
-            .store
-            .with_write_tx(|tx| self.check_for_missing_transactions(tx, epoch_state, &proposal, new_transactions))?;
+        let missing_tx_ids = self.store.with_write_tx(|tx| {
+            self.resequence_unpooled_transactions(tx, epoch_state, &proposal, new_transactions)?;
+            self.check_for_missing_transactions(tx, epoch_state.local_committee_info(), &proposal)
+        })?;
 
         if missing_tx_ids.is_empty() {
             return Ok(MessageValidationResult::Ready {
@@ -341,9 +342,10 @@ impl<TConsensusSpec: ConsensusSpec> OnMessageValidate<TConsensusSpec> {
         proposal: ProposalMessage,
         new_transactions: &OnReceiveNewTransaction<TConsensusSpec>,
     ) -> Result<MessageValidationResult<TConsensusSpec::Addr>, HotStuffError> {
-        let missing_tx_ids = self
-            .store
-            .with_write_tx(|tx| self.check_for_missing_transactions(tx, epoch_state, &proposal, new_transactions))?;
+        let missing_tx_ids = self.store.with_write_tx(|tx| {
+            self.resequence_unpooled_transactions(tx, epoch_state, &proposal, new_transactions)?;
+            self.check_for_missing_transactions(tx, epoch_state.local_committee_info(), &proposal)
+        })?;
 
         if missing_tx_ids.is_empty() {
             return Ok(MessageValidationResult::Ready {
@@ -366,14 +368,40 @@ impl<TConsensusSpec: ConsensusSpec> OnMessageValidate<TConsensusSpec> {
         })
     }
 
-    fn check_for_missing_transactions(
+    /// Rebuilds pool records for this block's transactions that we hold but that are absent from the
+    /// transaction pool, so that a proposal is not rejected over derived state no peer can resend.
+    ///
+    /// Only the commands the rebuilt `New` stage can serve are offered: a later command is not made
+    /// votable by a `New` record, and that record would then be proposed at a stage the committee has
+    /// moved past. See [`OnReceiveNewTransaction::resequence_known_transactions`].
+    fn resequence_unpooled_transactions(
         &self,
         tx: &mut <TConsensusSpec::StateStore as StateStore>::WriteTransaction<'_>,
         epoch_state: &EpochState<TConsensusSpec::Addr>,
         proposal: &ProposalMessage,
         new_transactions: &OnReceiveNewTransaction<TConsensusSpec>,
+    ) -> Result<(), HotStuffError> {
+        let ids = proposal
+            .block
+            .commands()
+            .iter()
+            .filter_map(|cmd| cmd.local_only().or_else(|| cmd.local_prepare()))
+            .map(|atom| atom.id)
+            .collect::<Vec<_>>();
+
+        if ids.is_empty() {
+            return Ok(());
+        }
+
+        new_transactions.resequence_known_transactions(tx, epoch_state.epoch(), ids, epoch_state.local_committee_info())
+    }
+
+    fn check_for_missing_transactions(
+        &self,
+        tx: &mut <TConsensusSpec::StateStore as StateStore>::WriteTransaction<'_>,
+        local_committee_info: &CommitteeInfo,
+        proposal: &ProposalMessage,
     ) -> Result<HashSet<TransactionId>, HotStuffError> {
-        let local_committee_info = epoch_state.local_committee_info();
         if proposal.block.commands().is_empty() {
             debug!(
                 target: LOG_TARGET,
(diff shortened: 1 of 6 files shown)


sdbondi (commented):
Reviewed at `b3215a7`. Read the three consensus files at head plus `finalize_block`, `clear_transaction_pool`/`reconcile_transaction_pool`, `Syncing::on_enter`, the five `TransactionNotInPool` sites, and the rocksdb finalized/execution readers.

What holds up:

- The diagnosis is right. `TransactionRecord::get_missing` reads the transactions table and `clear_transaction_pool` empties the pool table, so the record-keyed park/request path genuinely cannot see or repair this. Confirmed all five `TransactionNotInPool` no-vote sites are in `on_ready_to_vote_on_local_block.rs` (604, 963, 1046, 1307, 1500), reachable from both paths the repair now covers.
- The commit-safety argument checks out. `finalize_block` removes the pool record and calls `TransactionRecord::finalize_all` in the same write tx (`on_ready_to_vote_on_local_block.rs:1750-1766`), so pool-absent ⟺ finalized for the normal path; `reconcile_transaction_pool`'s removals are covered by the `receipt_exists` check in `validate_new_transaction`. There is no window where a finalizing transaction reads as re-sequenceable.
- `get_any` de-dupes through a `HashSet`, so a block carrying two commands for the same id cannot double-insert into `insert_new_batched`.
- `check_local_proposal` (leader strategy + signature) runs before `handle_missing_transactions_local_block`, so the new write is not reachable from an unauthenticated proposal.

## Blocking

**1. A re-sequenced record starts at stage `New` and is `is_ready = true`, so the node will propose it — even when the committee is already past that stage.** `on_receive_new_transaction.rs:207-215`

The repair only makes a proposal votable when the block's command for that id is the initial one. If the committee is mid-flight — pool record at `LocalPrepared`, block carries `LocalAccept` — the fresh `New` record does not fix the vote: `on_ready_to_vote_on_local_block.rs:1049` sends it down `initial_prepare_multishard` and out via `OutputOnlyDisagreement`, and 1310/1503 reject it with `StageDisagreement`. Same no-vote as before, different reason.

What is new is the side effect. The record is inserted ready, so `transaction_pool_get_many_ready` will pick it up (`transaction_pool.rs:136`) and `has_ready_or_pending_transaction_updates` reports work pending. Every leader slot this node takes now produces a block carrying a stale-stage command for that id, which the rest of the committee no-votes — a dead round per rotation, persisting for as long as the id sits unadvanceable in our pool. Previously the node only failed to vote; it never proposed.

This is reachable: `Syncing::on_enter` clears the pool immediately after `state_sync.sync(...)` returns and the node re-enters consensus straight away, so in-flight committee transactions at `LocalPrepared`/`LocalAccepted` are the expected case, not an exotic one. It is also the exact failure `clear_transaction_pool`'s own doc comment exists to prevent ("Re-proposing such a transaction breaks liveness"), and the guards that catch the dangerous version of it (`finalized decision`, `receipt_exists`) do not catch this one, because the transaction is genuinely unfinalized — just ahead of us.

Either bound the repair to the case it can actually fix (only re-sequence when the block's command for that id is the initial stage command), or say why emitting stale proposals is acceptable. If it is the latter, the doc comment on `resequence_known_transactions` should stop claiming re-derivation "is what makes such a proposal votable again" — it is only true for `New`-stage commands.

## Non-blocking

**2. `get_finalized_decision` is not a finalized-marker read.** `on_receive_new_transaction.rs:173`

The new narrowing guard is documented as "skips any id with *any* finalized record", but `TransactionRecord::get_finalized_decision` (`storage/src/consensus_models/transaction.rs:225`) goes through `finalized_transaction_execution_get`, which returns `NotFound` when the `FinalizedTransactionLinkCf` marker exists but no `BlockTransactionExecution` row does (`state_store_rocksdb/src/reader.rs:653-677`). So it means "has a finalized *execution*". In practice output-only groups do record one from the foreign proposal (`foreign_proposal_processor.rs:163/266/436`), so I could not construct a live case — but the guard's stated intent and its implementation are not the same predicate, and only `finalized_transaction_execution_get_finalized_time` reads the marker directly. Worth either a marker-level read or a comment narrowing the claim.

**3. `check_for_missing_transactions` now writes.** `on_message_validate.rs:389`

The name, the doc on `process_catch_up_response` ("missing-transaction parking are identical to the live-proposal path") and the `Ok(HashSet::new())` early return all read as a pure check. The repair could equally live in the worker between `handle()` returning `Ready` and dispatch, keeping this function read-only. If it stays here, the function needs a name that admits the mutation.

**4. Repeated full validation for ids that stay unpooled.** `on_receive_new_transaction.rs:184-196`

An id with `must_sequence == false` — no local inputs and foreign pledges not yet arrived, a normal multishard state — is never inserted, so it fails the `transaction_pool.exists` check on the *next* proposal too. Each proposal carrying it therefore re-runs `validate_full` (structural: signature verification, template existence — `block_transaction_validator.rs:40`) and `rec.save(tx)`, for as long as the pledges are outstanding. The window is usually short, but this is on the proposal hot path and is a new per-proposal cost that scales with commands-per-block.

**5. Duplicated guards.** The loop at `on_receive_new_transaction.rs:167-175` checks `transaction_pool.exists` and `get_finalized_decision`, then `validate_new_transaction` re-checks both at 228 and 239 for every surviving id.

**6. `epoch_state.epoch()` on the catch-up path.** `on_message_validate.rs:391` — a catch-up block from an older epoch validates its transactions against the *current* epoch, so anything whose `max_epoch` has passed silently won't re-sequence. Same as `process_requested` today, so not a regression, just noting the repair is a no-op there.

**7. Return value is unused** at the only call site; `usize` buys nothing over `()` unless a caller logs it.

**8. `Syncing::on_enter` comment is now stale.** `state_machine/syncing.rs:46` still says "Genuinely-pending transactions return via mempool gossip" — this PR exists precisely because they don't. Worth updating in the same change.

## Nits

**9.** `consensus.rs:1875` — "shorten the round and allow more than the default single-round budget" sits on the `modify_consensus_constants` closure, but only the first half is in that closure; the budget half is `with_test_timeout(30s)` on the next line (default is `DEFAULT_PACEMAKER_BLOCK_TIME + 2s`, `harness.rs:659`).

**10.** `consensus.rs:1866`/`1868` — "held back" / `hold_proposals`: the filter drops proposals, it doesn't buffer and release them. `drop_proposals` would describe it.

PR #2498, opened 2026-09-01
feat(indexer): invalidate the substate cache from the transition stream

sdbondi:
## What

Lets the indexer serve substate reads from its local cache indefinitely, invalidated by the state
transition stream rather than expired by a timer. No consensus changes.

Today an unversioned substate read is servable from cache for **two seconds**
(`DEFAULT_LATEST_VERSION_TTL`). Every read past that costs a validator committee round trip — for a
wallet account refresh, one per vault and per resource.

That bound was not arbitrary. Invalidation came from `DiffSummary.upped` on synced transaction
receipts, held in `SubstateVersionTracker`: in memory, capacity bounded, empty after a restart, and
positive-only by construction, so `is_superseded() == false` meant *"not known superseded"*, never
*"current"*. With no way to establish that an entry was still current, a short timer was the only
safe backstop.

## The invariant

A cached entry for substate `S` answers for `S`'s current state if every commit that supersedes or
destroys `S` has already reached the indexer through `shard(S)`'s transition stream. Given that,
presence in the cache *is* validity — no TTL, no version comparison at read time. The staleness of a
served value is exactly the wall-clock lag of the shard's watermark.

Three changes make the argument checkable.

**1. The down feed is completed (commit 1).** `ALL_HASHES` was honoured on the up arm of
`state_transitions_get_starting_at` but not the down arm, so a subscriber asking for hashes of
filtered-out substates never learned one had been *destroyed*. That gap is only visible for a
terminal down — a substate destroyed with no successor version, which is exactly a spent `Utxo`, a
spent `ConfidentialOutput`, or a `ValidatorFeePool` drained to zero. Every other destroy is paired
with a create at the next version, which `ALL_HASHES` already carried. A terminal down appears in no
later up and in no receipt field at any watermark, so without it the stream is not a complete record
of which substates changed, and those entries would be served as live indefinitely.

A destroy carries no value, so admitting it costs an id and a version. `UP_ONLY` still suppresses
every down: the from-scratch sync path wants the set that is live now, not the history that produced
it.

**2. Invalidation commits with the watermark.** The cache moves from a `cacache` directory into the
indexer's own SQLite database. The reason is atomicity, not storage: invalidation has to commit in
the same transaction that advances `Key::SyncProgress`, and two separate stores cannot do that. A
reader that saw one without the other would break the argument the cache rests on.

**3. A shard must be demonstrably kept up with.** `ShardWatermarks` records, per shard and *per
process run*, that a stream's completion marker put this indexer level with the committee. The cache
serves a substate only while its shard was confirmed within `substate_cache_max_serve_lag`.

Two consequences worth spelling out:

- A fresh or restarted indexer has confirmed nothing, whatever its stored progress says, so it serves
  from cache only once its first round lands — which is also when the transitions it missed while
  down arrive.
- Only the *completion marker* confirms, never a mid-stream batch. Part-way through a catch-up the
  indexer genuinely holds every transition up to some version while the chain is arbitrarily far
  ahead of it, and confirming there would open the cache on a stale entry.

This is also the only bound on a validator that stops serving transitions: it freezes the watermark,
and the lag gate closes the cache for its shards.

## One entry per substate: its head

`HeadIndex` holds exactly one `(version, is_up)` per substate, and upping a substate downs its
predecessor (`working_state.rs:1962`, `pending_store.rs:143-151`). So **a version that is up *is* the
head** — there is never a second live version, and versions are contiguous from 0.

The cache therefore keys on `substate_id` alone and holds that substate's head. Only a result that
establishes the head is written: an `Up` at any query, or any result to a query that named no
version. A named version that came back `Down` bounds the head from below without naming it, so it is
not written.

**A cached head settles every version below it, locally.** A substate that ever reached `L` has
versions `0..L` down permanently, so a versioned read below the head is answered `Down` without a
committee round trip *and without the watermark* — a cached head is a lower bound on the real one, so
a stale `L` only makes the conclusion more certainly true. An inferred `Down` is exactly what a
validator would have answered.

| request | answer | watermark gate |
|---|---|---|
| `N < head` | `Down { version: N }`, computed locally | no |
| `N == head` / unversioned | the cached entry | yes |
| `N > head` | miss → fetch | — |

There is no *successful* read that needs a historic row: a versioned read returns either `Up` at `v`
(so `v` is the head) or `Down`, which `SubstateManager` turns into `InputSubstateIsDown` regardless.
Validators also clear downed values at epoch GC (`substates_prune_downed_values`), so historic values
are not reliably refetchable in the first place.

## The write race, and a head that is too high

A fetch that began at watermark `W₀` and lands after a transition for the same substate arrived would
resurrect a stale entry. `substate_cache_invalidations` journals each touched substate at the state
version that touched it, and `substate_cache_put` reads that journal **inside the same write
transaction** as the insert. A result observed before a transition the cache already holds is
discarded; the caller still has its freshly fetched value and the next read fetches again. The
journal only has to span an in-flight fetch, so it is a few hundred rows rather than the tracker's
50k.

The head is also monotonic, so a committee member that is behind cannot walk it backwards. That guard
needs two escapes, because **no transition retires a version above the head** — one recorded too high
is corrected by nothing:

- an arriving result that is verified displaces an **unverified** head regardless of version;
- an **unverified** head older than `head_ttl` is displaced by anything.

Both escapes are confined to an unverified head, and that is the whole rule: a proof attests that a
version existed, never that it is current, so a verified head is a lower bound on the real one and
nothing — a proof, or elapsed time — may walk it back. It is also the only kind that can be recorded
*above* the real version, so it is the only kind either escape was needed for.

That matters because a committee member which is merely behind can prove an older version against an
older signed root, and the trusted-root ring accepts that by design. Without the qualification, an
idle substate's proven head would be walked backwards by the next such answer once it aged out.

The first escape covers verification-on: `fetch_and_cache_substates` serves `POST /substates/fetch`
over the batch RPC, which carries no proofs and so writes `verified: false` whatever the setting,
meaning one validator could otherwise park an unverified head that blocks the proven one forever. The
second covers verification-off, where nothing outranks anything and time is the only correction. The
read path ages the head out on the same bound, so an inference is drawn only from evidence younger
than `head_ttl`.

## Config

| | |
|---|---|
| removed | `latest_substate_cache_ttl` |
| added | `substate_cache_max_serve_lag` (default 300s) |
| added | `substate_cache_max_entries` (default 100k, one entry per substate) |
| changed | `DEFAULT_CACHE_TTL` 300s → 900s, now also the head TTL |

`cache_ttl` stays, demoted from a correctness mechanism to a coarse backstop for the case where the
transitions that would have corrected an entry never arrive.

`max_serve_lag` must comfortably exceed a *full sync round* — `state_scanning_interval` plus however
long it takes to sync every shard group — not just the interval, or the cache closes between rounds.
Set too high, it widens the window in which a withholding validator can hold a stale value open.
Ordinary staleness is bounded by the sync round, not by this.

**API surface:** `GetIndexerInfoResponse.latest_substate_cache_ttl_secs` →
`substate_cache_max_serve_lag_secs`, with the TS binding updated by hand.

## Deleted

- `crates/indexer_lib/src/substate_versions.rs` and its wiring
- `applications/tari_indexer/src/substate_file_cache.rs` and the `cacache` dependency
- `latest_version_ttl` / `DEFAULT_LATEST_VERSION_TTL`

## Tests

Store-level: retire-on-transition, terminal destroy, a higher cached head surviving a lower
transition, a lower version not displacing the head, a verified result displacing an unverified head,
a stale head not blocking a lower version, the write-race veto, and prune/eviction ordering.

Cache-adapter level: a version below the head answered `Down` with no committee, that inference
working on a shard with no watermark while current-state claims on the same shard do not, the
inference stopping once the head ages out, the head answering for itself and for an unversioned read,
and a version above the head being a miss.

Plus `ShardWatermarks` (cold shard, highest version wins, stale confirmation closes the shard) and one
rocksdb test that `ALL_HASHES` streams downs for filtered-out substates.

## Notes for the reviewer

- **Eviction is oldest-written, not true LRU.** Recording `last_read_at` would put a write on every
  cache hit; an evicted entry costs one round trip to restore.
- **Every transition now journals**, which is one indexed upsert per transition per state version
  where there was none before. Necessary — a substate not yet in the cache still has to be able to
  veto an in-flight fetch.
- **Unversioned reads can now return a cached `Down`** (a terminal down is the head). `SubstateManager`
  turns it into `InputSubstateIsDown` exactly as it does a freshly fetched one, so nothing observable
  at the API changes — but it is a new path.
- **`<data_dir>/substate_cache`** (the old cacache directory) is orphaned on upgraded nodes rather
  than deleted.
- **The migration has been edited in place** rather than superseded, since it is unreleased. Anyone
  who ran an earlier revision of this branch must wipe their indexer DB.
- **`IndexerConfig` is `#[serde(deny_unknown_fields)]`**, so an operator `config.toml` still setting
  `latest_substate_cache_ttl` fails to load rather than ignoring the key. Nothing in-repo sets it.
- **Not in this PR:** single-flight coalescing of concurrent committee fetches, the batched read path
  (needs proofs on the batch RPC), metrics for invalidations / evictions / refused-stale, a log line
  when a read is refused for a stale watermark, and regenerating
  `docs/developer-docs/public/indexer/openapi.json` + `indexer-api.html`.
- **Assumption worth a second opinion:** that committed state is never rolled back in-protocol.
  `rollback_history.rs` reads as an offline operator tool (`tari_validator_rollback`), which would
  make "presence is validity" safe. If in-protocol rollback exists, the model needs revisiting.


sdbondi (commented):
Reviewed at `83777ff`. The core argument holds up: `ALL_HASHES` now completes the down feed (the terminal-down gap was real and the rocksdb test pins it), invalidation commits in the same write tx as the state version it came from, the journal veto closes the in-flight-fetch race, and `ShardWatermarks` being per-process-run — confirmed only by the completion marker — is the right conservative choice. I checked the two claims the model rests on: `to_shard` reads only byte 0 of the object key, so `shard_of(id)` is version-independent and matches the shard a transition is recorded under (including `Shard::global()` for global ids); and `last_state_versions` is never removed from `SyncProgress`, so a shard cannot fall back to the `UP_ONLY` from-scratch path with a populated cache.

One blocking correctness bug, in the `is_latest` bookkeeping.

## Blocking

**`is_latest` can be claimed by two rows at once, and the unversioned read returns the older one.** `substate_cache_put` never clears `is_latest` on lower versions of the same substate (writer.rs:582-590, deliberately, per the comment), and `substate_cache_get` selects `is_latest = true` with no ordering and takes `.first()` (reader.rs:1126-1136). The comment in the migration — "`is_latest` marks the row that answers an unversioned read" — assumes at most one claimant; nothing enforces that.

The reachable sequence needs no TTL expiry at all:

1. v5 is cached `is_latest`, fresh.
2. The chain moves to v6; the `Created{v6}` transition has not been synced yet, so v5 is not retired.
3. A client hits `POST /substates/fetch` (not `cached_only`) → `fetch_and_cache_substates` fetches v6 and writes it with `is_latest: true` (cached_substate_manager.rs:307-320). The journal cannot veto it: no transition for the substate has arrived.
4. Both v5 and v6 now claim `is_latest`.
5. The next unversioned `get_substate` returns **v5**.

Confirmed against SQLite with this schema: with rows (c1,5,is_latest) and (c1,6,is_latest), `select version from substate_cache where substate_id='c1' and is_latest=1 limit 1` returns `5` — the planner uses `sqlite_autoindex_substate_cache_1` and walks it in `(substate_id, version)` order, so the *lowest* version wins. `idx_substate_cache_latest` is not chosen for this query at all.

So a superseded version is served as the latest even though the correct one is sitting in the cache — the exact failure the PR sets out to remove, and it survives until the retiring transition lands (a full sync round).

There is a second, quieter path to the same state: any cache miss that fetches a newer version — `age > cache_ttl`, or an unverified entry with verification on — leaves the old claimant behind. There the stale row is usually not *served* (its `cached_at` is old), so the symptom is instead that the substate's cache is dead: every read pays a round trip. But a versioned lookup of that same old version refreshes it — the `on_conflict ... do_update` sets `cached_at = now` while leaving `is_latest = true` (writer.rs:569-580) — and it becomes servable again.

Fix either end: `ORDER BY version DESC` on the unversioned read (it then uses the PK index directly, and `idx_substate_cache_latest` can go), and/or clear `is_latest` on `version < entry.version` for the same substate when writing a latest. I'd do the ordering regardless — relying on an unordered `.first()` to pick a specific row is not something a reader can check. The store tests never put two live versions of one substate in the cache; worth a case for it.

## Non-blocking

**Versioned reads are gated on the shard watermark, and needn't be.** `SqliteSubstateCache::read` returns `None` for any substate whose shard has no fresh watermark (substate_cache.rs:122), including `version: Some(v)` lookups. A `(substate_id, version)` entry is immutable — both `Up` and a terminal `Down` are permanent facts about that version — so the completeness argument the watermark supports is only needed for the `is_latest` claim. As written, a freshly restarted indexer, or one whose shard is behind, pays a committee round trip for reads that were always safe to serve, in exactly the window where round trips are most expensive. Gating on `version.is_none()` would keep the model intact.

**Unversioned reads now serve `SubstateResult::Down`.** Dropping the `is_up` guard is consistent with the new model (a terminal down is the truth, and a successor's `Created` retires the entry), and `get_cached_substates` filters it via `into_up()` so only `get_substate` is affected — but it is a behaviour change that isn't in the description. Worth a line, since a cached `Down` reaching a caller that previously only ever saw a freshly-fetched one is a new path.

**`latest_substate_cache_ttl` removal is a hard startup failure, not an ignored key.** `IndexerConfig` is `#[serde(deny_unknown_fields)]` (config.rs:90), so any deployed `config.toml` that sets the old key now fails to load rather than falling back to the default. Nothing in-repo sets it, so this is only about operator configs — but the failure lands as a config parse error at startup on upgrade. A retained-and-ignored field, or a release note, would take the sharp edge off.

**Generated API docs are stale.** `docs/developer-docs/public/indexer/openapi.json` and `indexer-api.html` still advertise `latest_substate_cache_ttl_secs`. #2461 and #2479 both regenerated these when they changed the response; `scripts/indexer-gen-api-docs.sh`.

**No signal when the cache closes.** If a full sync round ever exceeds `substate_cache_max_serve_lag`, every read silently reverts to a committee round trip with nothing in the logs to say why — and `ALL_HASHES` makes each round strictly heavier than the measurements the 300s default was presumably chosen against (`state_scanning_interval` defaults to 60s, so the headroom is ~5 intervals). Metrics are explicitly out of scope, but a `debug!`/`warn!` on the refused-because-stale branch in `read` would make it diagnosable without them.

## Nits

- `SqliteSubstateCache::shard_of` (substate_cache.rs:100-106) re-implements the canonical mapping including the `is_global` special case; `VersionedSubstateIdRef::new(id, 0).to_shard(NumPreshards::current())` already is that function. Two copies of a rule that must not drift.
- The `UPDATE ... SET is_latest = true` at writer.rs:582-590 is a no-op on the insert path (the insert already sets it) — only the conflict path needs it. Folding it into `do_update` as `is_latest = is_latest OR excluded.is_latest` would say the "never cleared" rule in one place.
- `pruning_evicts_down_to_the_cap_and_expires_the_journal` asserts two entries survive but not *which* two — every row is written with `cached_at: 0`, so the eviction ordering it exists to check isn't actually exercised.

sdbondi (commented):
Re-reviewed at `1f0dc48` (one row per substate, lower versions inferred down).

The inference is sound, and it's a better design than the one it replaces — the `is_latest` bookkeeping and the whole class of bug I raised are gone rather than patched. I checked the invariant it rests on rather than taking it on faith: `pending_store.rs:143-151` inserts a `Down` for the predecessor on every up and sets `next_version = version + 1`, and the engine's in-place update is `Substate::new(existing_state.version() + 1, ...)` (working_state.rs:1962). Versions are contiguous from 0 and an up downs its predecessor, so a head at `L` does imply `0..L` down. The monotonicity that makes it un-staleable is enforced in both directions: `substate_cache_put` refuses a version below the cached head (writer.rs:558-567), and invalidation only ever deletes. `is_head = specific_version.is_none() || result.up().is_some()` (cached_substate_manager.rs:216) is the right admission rule — a named version that came back `Down` bounds the head from below without naming it, and is correctly not written.

The watermark gating is now correct by construction rather than by convention: only claims about current state consult it, which also resolves the "versioned reads are gated needlessly" point from the last round. And an inferred `Down` for a version below the head is exactly what a validator would have answered, so nothing observable at the API changes.

## Non-blocking

**A head recorded too high is permanent, and nothing re-validates it.** The head only moves up, the inferred `Down` is stamped `cached_at: now` so it bypasses `cache_ttl`, and the inference bypasses the watermark by design. If a head is ever recorded above the real one, the entry cannot recover: incoming `Created{v}` transitions retire `<= v-1`, always below the inflated head, and a later fetch carrying the true lower head is rejected by the new monotonicity guard — which also means its `cached_at` stops being refreshed, so only `max_entries` eviction ever clears it.

This is reachable with `verify_substate_proofs` on, because `fetch_and_cache_substates` writes `verified: false` unconditionally (cached_substate_manager.rs:322) — the batch RPC carries no proof, as the description notes. A single validator answering `POST /substates/fetch` with an inflated version parks an unverified head that the serve-time `verified` check then refuses forever, *and* that blocks the verified single-fetch path from ever writing the true head. The result is not a wrong answer while verification is on — it is a substate whose cache is permanently dead. With verification off it is a wrong answer: live versions reported `Down` indefinitely.

The guard that closes it is small: let a verified result displace an unverified head regardless of version, or apply the `cache_ttl` age check to the head row backing an inference. The second also gives the whole scheme the same backstop `cache_ttl` provides everywhere else, for the same reason — the case where the transitions that would have corrected an entry never arrive.

**Stale wording now that rows are per substate:**
- config.rs:145 — "Maximum substate versions held in the cache" is now one row per substate, not per version.
- The PR description still describes the previous design throughout: `(substate_id, version)` keying, the `is_latest` paragraph, the whole "Retiring versions" section, and the test list ("versioned-vs-unversioned claims").

**Migration edited in place, third time.** Correct for an unreleased migration, but anyone running this branch has a `substate_cache` table with `is_latest` and the old primary key, and diesel will not re-run it — expect a column error until the dev DB is wiped.

**Still open from the last round:** unversioned reads now serve `Down` (undocumented behaviour change); `#[serde(deny_unknown_fields)]` turns the removed `latest_substate_cache_ttl` key into a hard startup failure on upgrade; `docs/developer-docs/public/indexer/openapi.json` and `indexer-api.html` still advertise `latest_substate_cache_ttl_secs`; no log line when a read is refused because a shard's watermark went stale.

CI is re-running on this head — clippy, both test jobs and the integration test are still pending.

sdbondi (commented):
Re-reviewed `2532098`, `86ef767`, `1205845`. All three of the last round's points are closed, and the first-creation change is a good one I hadn't asked for.

**The over-recorded head is bounded.** The two escapes are the right pair, and the `!cached_verified` qualification on the time escape (`1205845`) is the part that matters: a proof attests a version existed, so a verified head is a genuine lower bound and time must not walk it back — while an unverified head, which is a lower bound on nothing, yields both to a proof and to time. The dead-cache case I described is closed by `outranked`, and `a_verified_result_displaces_an_unverified_head` pins it. I also checked the direction that worries me more: `outranked` lets a verified *lower* head displace an unverified higher one, and since a verified result can legitimately be behind (an older version proven against an older signed root, which the trusted-root ring accepts), the head can move down — but only to another value the substate provably reached, so it stays a lower bound and the `Down` inference below it stays true. The cost is refetches, not wrong answers.

**Not journalling a first creation is safe.** The argument holds at the write path: nothing can be cached for a substate before its first version, because the only result a lookup could return is `DoesNotExist`, which fails the `result.version()` gate in `get_substate` and is never written. So there is no in-flight fetch to veto. The batch path can only observe an `Up`, and an observation of v0 concurrent with the creation of v0 is not stale. Any later transition still journals normally, so nothing after the first creation loses its veto. Keying the skip on `version == 0` via `checked_sub` is exactly the first-creation case, and dropping the whole struct rather than just the `DELETE` takes the write out too — which for transaction receipts, created once and never updated, means the entire class stops journalling.

The eviction test now asserts *which* entries survive against descending `cached_at`, so it actually exercises oldest-first. Config wording fixed.

## Two small things

**The description's second escape is stated wider than the code.** "a head older than `head_ttl` is displaced by anything" — the code requires `!cached_verified` (writer.rs:583), which is precisely what `1205845` added and what `an_aged_verified_head_is_still_not_walked_backwards` pins. The following paragraph implies the qualification but the bullet contradicts it; a reader taking the bullet at face value gets the rule wrong.

**The read path ages the inference out regardless of `verified` (substate_cache.rs:162).** The write path draws the verified/unverified distinction deliberately; the read path does not. A verified head cannot have been recorded too high, so the conclusion it supports cannot be wrong, and ageing it out only buys committee round trips on substates whose head was proven and then went quiet. Gating it the same way the write path does would make the two consistent. Conservative as it stands, so a preference rather than a defect.

## Still open from earlier rounds

Unversioned reads now serve `Down`, which is a behaviour change not in the description; `#[serde(deny_unknown_fields)]` turns the removed `latest_substate_cache_ttl` key into a hard startup failure on upgrade rather than an ignored one; `docs/developer-docs/public/indexer/openapi.json` and `indexer-api.html` still advertise `latest_substate_cache_ttl_secs` (`scripts/indexer-gen-api-docs.sh`); and nothing logs when a read is refused because a shard's watermark went stale.

Nothing here blocks. CI is still running on this head — clippy, the three test jobs, integration, ledger and web ui are pending.

PR #2503, opened 2026-09-01
feat(engine)!: add caller_component/direct_caller_template rule requirements

### LINKED ISSUE #2500: `component(addr)` method access rule is checked against the callee, not the caller
_Written by a zed-harnessed deepseek v4 pro and reviewed by hooman brian_

**Describe the bug**

`rule!(component(addr))` — which expands to `RuleRequirement::ScopedToComponent(addr)` — does not work as
documented when used as a **component method** access rule (`.method("bar", rule!(component(caller_addr)))`).
The rule is evaluated against the **callee's** own address instead of the **caller's** address, so a
cross-component call made by the intended caller component is rejected with `AccessDenied`.

The same primitive works correctly for **resource** rules (e.g. `mintable(rule!(component(x)))`), which is why
this has gone unnoticed: that is the only path with existing test coverage.

**Expected behavior**

Per the authorization guide (https://ootle.tari.com/guides/authorization-and-access/):

> `AccessRule::Restricted(vec![identity1, identity2, ...])`: Only allows callers whose identity matches one of
> the provided identities. Identities can be `ComponentAddress`, `ResourceAddress`, `TemplateAddress`, etc.

and

> `ComponentAddress`: When a component calls another component.

So `component(caller_addr)` on a method should mean "only callable from the component whose address is
`caller_addr`", and a call made from exactly that component should succeed.

The source-level doc comment agrees: `RuleRequirement::ScopedToComponent` is documented as
"Requires execution within a specific component" (`crates/template_lib_types/src/access_rules.rs`).

This matches Radix's `global_caller(ComponentAddress)` requirement — asserting that a specific component is the caller — which tari-ootle is inspired by.

**Root cause**

1. `rule!(component(x))` maps to `RuleRequirement::ScopedToComponent(x)`
   (`crates/template_lib_types/src/access_rules.rs`, `__rule_requirement` macro).

2. The engine evaluates it in `crates/engine/src/runtime/tracker_auth.rs:261`:

   ```rust
   RuleRequirement::ScopedToComponent(address) => Ok(state.current_component()? == Some(*address)),
   ```

3. `WorkingState::current_component()` returns the **top** call frame's component
   (`crates/engine/src/runtime/working_state.rs:1488`):

   ```rust
   pub fn current_component(&self) -> Result<Option<ComponentAddress>, RuntimeError> {
       let frame = self.call_frames.last().ok_or(RuntimeError::NoActiveCallFrame)?;
       Ok(frame
           .scope()
           .get_current_component_lock()
           .and_then(|lock| lock.substate_id().as_component_address()))
   }
   ```

4. In `TransactionProcessor::invoke_component` the **callee** frame is pushed _before_ the method access rule is
   checked (`crates/engine/src/transaction/processor.rs:920` and `:930`):

   ```rust
   runtime.interface_mut().push_call_frame(PushCallFrame::ForComponent { ... })?;

   // This must come after the call frame as that defines the authorization scope
   runtime.interface_mut().check_component_access_rules(method)?;
   ```

Consequently, at check time `current_component()` is the **callee**, so `component(caller_addr)` compares
`callee == caller_addr` and denies the caller.

Resource rules are unaffected because they are evaluated while the caller's frame is still the current/top frame
(no new component frame is pushed for a resource action).

**Impact**

A template cannot express "only this component may call this method" with the documented `component(addr)`
primitive. The only workarounds are heavier and don't map cleanly onto the documented semantics (e.g. minting and
passing a badge proof, or manually storing the allowed caller and comparing it inside the method).

**To Reproduce**

The quickest path is to run the regression test on the repro branch
(`n0izn0iz/tari-ootle` → `repro/component-rule-on-method`):

```bash
# from an existing tari-ootle checkout
git fetch https://github.com/n0izn0iz/tari-ootle.git repro/component-rule-on-method
git checkout FETCH_HEAD
rustup target add wasm32-unknown-unknown   # only if not already installed
cargo test -p tari_engine --test method_rule_component
```

The test (`component_rule_on_method_is_checked_against_the_callee_not_the_caller`) asserts the documented
behavior — that the caller is allowed — so it currently **fails** against the buggy engine. That failure is the
demonstration of the bug.

The two templates are minimal (the `#[template]` macro supports only one component struct per crate, hence two
crates):

`callee` template:

```rust
#[template]
mod callee {
    use super::*;

    pub struct Callee;

    impl Callee {
        pub fn new(gate: ComponentAddress) -> Component<Self> {
            let access_rules = ComponentAccessRules::new()
                .method("bar", rule!(component(gate)))
                .method("ping", rule!(allow_all))
                .default(rule!(deny_all));

            Component::new(Callee)
                .with_access_rules(access_rules)
                .with_owner_rule(OwnerRule::None)
                .create()
        }

        pub fn bar(&self) {}                 // restricted to `gate`
        pub fn ping(&self) -> u64 { 42 }     // control, allow_all
    }
}
```

`caller` template:

```rust
#[template]
mod caller {
    use super::*;

    pub struct Caller;

    impl Caller {
        pub fn new() -> Component<Self> {
            let access_rules = ComponentAccessRules::new()
                .method("call_bar", rule!(allow_all))
                .method("call_ping", rule!(allow_all))
                .default(rule!(deny_all));

            Component::new(Caller)
                .with_access_rules(access_rules)
                .with_owner_rule(OwnerRule::None)
                .create()
        }

        pub fn call_bar(&self, callee: ComponentAddress) {
            ComponentManager::get(callee).invoke("bar", args![]);
        }

        pub fn call_ping(&self, callee: ComponentAddress) -> u64 {
            ComponentManager::get(callee).call("ping", args![])
        }
    }
}
```

Test flow (using `TemplateTest`):

1. Create `Caller` → obtain `caller_addr`.
2. Create `Callee` with `new(caller_addr)`.
3. `caller.call_ping(callee)` → returns `42` (proves cross-component calls work).
4. `caller.call_bar(callee)` → **fails with `AccessDenied`**; expected to succeed.

**Suggested fix direction**

Evaluate `ScopedToComponent` against the **caller** rather than the top frame for the component-method case.

A likely-safe approach is to capture the caller component address in `invoke_component` _before_
`push_call_frame`, and thread it into the `ScopedToComponent` check — while leaving the current behavior for
resource rules, where the current frame already _is_ the caller. Reordering the `push_call_frame` /
`check_component_access_rules` lines is not obviously safe: the comment at `processor.rs:929` notes the call
frame defines the authorization scope used by the ownership check and other rule types.

**Where a regression test should go**

A regression test already exists and can be lifted into a fix PR:

- test: `crates/engine/tests/method_rule_component.rs`
- templates: `crates/engine/tests/templates/repro_caller` and `crates/engine/tests/templates/repro_callee`

It asserts the documented behavior (`execute_expect_success`, i.e. the caller is allowed) and currently **fails**
against the buggy engine; it should turn green once the fix lands. Alternatively, fold the case into
`crates/engine/tests/access_rules.rs` alongside `it_restricts_resource_actions_to_component` (which only covers the resource path).

**Environment / version**

- Reproduced against `tari_template_lib 0.31` / `tari_engine 0.39.3` (via `tari_template_test_tooling 0.39`), and
  confirmed by running the regression test against current `development` (`tari_engine 0.40.0`). The file/line
  references above are from a fresh `development` checkout.

**Additional context**

- `tari-project/tari-ootle#1145` is the origin of the `component(addr)` primitive, and it scopes the feature to
  **resource** actions ("only the owning component is authorised to perform resource actions") — consistent with the resource-only test coverage.
- No issue/PR currently tracks the method-rule behavior (a search for `ScopedToComponent`, `ScopedTo`,
  `current_component`, `locked_to_component`, etc. returns nothing relevant).
- The online authorization guide's concrete code sample is stale (it uses a non-existent `rule!(require(...))` and an old `create_component(name, state)` API), so the high-level `AccessRule::Restricted` bullet is the
  authoritative intent source.

### LINKED ISSUE #2502: `template(addr)` method access rule is checked against the callee's template, not the caller's
_Written by a zed-harnessed deepseek v4 pro and reviewed by hooman brian_

**Describe the bug**

`rule!(template(addr))` — which expands to `RuleRequirement::ScopedToTemplate(addr)` — does not work as
documented when used as a **component method** access rule (`.method("bar", rule!(template(caller_template)))`).
The rule is evaluated against the **callee's** template instead of the **caller's** template, so a cross-template
call made by the intended caller template is rejected with `AccessDenied`. (If the rule happens to name the
callee's own template, it is instead trivially satisfied by _any_ caller.)

This is the sibling of the `component(addr)` method-rule bug (#2500); both share the same root cause, differing only in which top-frame field is read (`current_component()` vs `current_template()`).

**Expected behavior**

Per the developer skills docs (`docs/skills/*/SKILL.md`):

```rust
rule!(component(component_address))  // Only callable from this component
rule!(template(template_address))    // Only callable from this template
```

**Intended semantics (resolving the ambiguity).** The engine models a caller as both a template and, optionally, a component — see `AuthHookCaller` in `crates/template_lib_types/src/auth_hook.rs`:

- `template_address` — always present; the template the calling code belongs to.
- `component_address` — `Some` only when the caller is a component instance; `None` for a static template function.

Therefore:

- `rule!(component(addr))` = "the caller is the specific component instance `addr`" (a static function never
  matches, since its `component_address` is `None`).
- `rule!(template(addr))` = "the caller belongs to template `addr`" — matching **both** a component instance of that template and a static function of that template.

This matches Radix's `package_of_direct_caller(package)` requirement — "the immediate caller's package" — which tari-ootle is inspired by.

So `template(caller_template)` on a method should mean "only callable from code belonging to template
`caller_template`" (component instance or static function), and a call made from either should succeed. This is
exactly how resource rules already behave (the check runs while the caller is the current frame); the method-rule
path is broken because it evaluates against the callee instead of the caller, so neither caller kind is
recognized.

The source-level doc comment agrees: `RuleRequirement::ScopedToTemplate` is documented as "Requires execution within a specific template" (`crates/template_lib_types/src/access_rules.rs`).

**Root cause**

1. `rule!(template(x))` maps to `RuleRequirement::ScopedToTemplate(x)`
   (`crates/template_lib_types/src/access_rules.rs`, `__rule_requirement` macro).

2. The engine evaluates it in `crates/engine/src/runtime/tracker_auth.rs:262`:

   ```rust
   RuleRequirement::ScopedToTemplate(address) => {
       let current = state.current_template()?;
       Ok(current == address)
   },
   ```

3. `WorkingState::current_template()` returns the **top** call frame's template
   (`crates/engine/src/runtime/working_state.rs:1461`), which for a `ForComponent` frame is the **callee's**
   template.

4. As in the `component(addr)` sibling, `TransactionProcessor::invoke_component` pushes the **callee** frame
   _before_ checking the method access rule (`crates/engine/src/transaction/processor.rs:920` and `:930`):

   ```rust
   runtime.interface_mut().push_call_frame(PushCallFrame::ForComponent { ... })?;

   // This must come after the call frame as that defines the authorization scope
   runtime.interface_mut().check_component_access_rules(method)?;
   ```

Consequently, at check time `current_template()` is the **callee's** template, so `template(caller_template)`
compares `callee_template == caller_template` and denies the caller (or, when the rule names the callee's own
template, tautologically allows everyone).

Resource rules are unaffected because they are evaluated while the caller's frame is still the current/top frame.

**Impact**

A template cannot express "only callers from template X may call this method" with the documented `template(addr)` primitive.

**To Reproduce**

The quickest path is to run the regression test on the repro branch
(`n0izn0iz/tari-ootle` → `repro/template-rule-on-method`):

```bash
# from an existing tari-ootle checkout
git fetch https://github.com/n0izn0iz/tari-ootle.git repro/template-rule-on-method
git checkout FETCH_HEAD
rustup target add wasm32-unknown-unknown   # only if not already installed
cargo test -p tari_engine --test method_rule_template
```

The two tests (`template_rule_on_method_is_checked_against_the_callee_not_the_caller` and
`template_rule_on_method_from_static_function_is_checked_against_the_callee_not_the_caller`) assert the documented behavior — that the caller is allowed — so they currently **fail** against the buggy engine. That failure is the
demonstration of the bug.

The two templates are the same shape as the `component(addr)` sibling, but the callee gates `bar` with
`rule!(template(caller_template))` instead of `rule!(component(caller_addr))`:

- `callee` template: `new(gate_template: TemplateAddress)` gating `bar` via
  `.method("bar", rule!(template(gate_template)))`; plus an `allow_all` `ping` control.
- `caller` template: `call_bar(callee)` (component method) and `call_bar_static(callee)` (static function) both do
  `ComponentManager::get(callee).invoke("bar", args![])`.

Test flow (using `TemplateTest`) — two cases, both currently failing:

1. Obtain the caller's template address: `let caller_template = test.get_template_address("TemplateCaller");`
2. Create `TemplateCallee` with `new(caller_template)`.
3. **Component caller:** `caller.call_bar(callee)` → **fails with `AccessDenied`**; expected to succeed.
4. **Static-function caller:** `call_bar_static(callee)` → **fails with `AccessDenied`**; expected to succeed.

A practical wrinkle versus the `component(addr)` sibling: `TemplateAddress` is less ergonomic to obtain at
runtime. There is no `CallerContext::current_template_address()`; a component can recover its own template via
`ComponentManager::current().get_template_address()`, but a _static_ template function has no direct equivalent.
The repro above sidesteps this by taking the caller's template address from the test harness at construction time.

**Suggested fix direction**

Evaluate `ScopedToTemplate` against the **caller's** template rather than the top frame for the component-method case — most naturally by capturing the caller template address in `invoke_component` _before_ `push_call_frame` (alongside the caller component address needed for the `component(addr)` sibling) and threading it into the `ScopedToTemplate` check, while leaving the resource-rule behavior unchanged.

A shared fix is likely desirable: both `ScopedToComponent` and `ScopedToTemplate` are read from the top frame in `check_requirement` and both are mis-evaluated for method rules for the same reason, so a single change that captures the caller's component address _and_ template address before the callee frame is pushed would address both.

**Where a regression test should go**

A regression test already exists and can be lifted into a fix PR:

- test: `crates/engine/tests/method_rule_template.rs` (two cases: component caller and static-function caller)
- templates: `crates/engine/tests/templates/repro_template_caller` and `crates/engine/tests/templates/repro_template_callee`

Both cases assert the documented behavior (`execute_expect_success`, i.e. the caller is allowed) and currently
**fail** against the buggy engine; they should turn green once the fix lands. Alternatively, fold the cases into
`crates/engine/tests/access_rules.rs` alongside `it_restricts_resource_actions_to_component` (which only covers
the resource path).

**Environment / version**

- Reproduced by running the regression test against current `development` (`tari_engine 0.40.0`) — both cases
  fail with `Access Denied`. The file/line references above are from a fresh `development` checkout. The
  `component(addr)` sibling (#2500) has also been reproduced.

**Additional context**

- `tari-project/tari-ootle#1145` is the origin of the `component(addr)` primitive and scopes it to **resource** actions; `template(addr)` was added alongside it with the same top-frame evaluation.
- The `component(addr)` sibling is tracked in #2500; this issue covers the `template(addr)` variant. No prior issue/PR tracked either behavior (searches for `ScopedToComponent`, `ScopedToTemplate`, `current_component`, `current_template`, `locked_to_component`, etc. return nothing relevant).
- `template(addr)` has weaker runtime ergonomics than `component(addr)` (no `current_template_address()` in `CallerContext`) and no pre-existing upstream test coverage, so this issue is lower-confidence than its sibling and is filed separately so the two can be triaged independently.

n0izn0iz:
_Written by a zed-harnessed deepseek v4 pro and reviewed by hooman brian_

> **Updated after review:** the approach changed from overloading `component(addr)` / `template(addr)`
> to adding distinct `caller_component(addr)` / `caller_template(addr)` requirements. The original
> description is kept at the bottom for review context.

Closes #2500
Closes #2502

Description
---

Adds `caller_component(addr)` and `caller_template(addr)` as new `rule!` requirements for component
method access rules, meaning "the caller of this method is the given component/template".

`component(addr)` / `template(addr)` are unchanged and continue to mean "execution within / current
frame". A new `RuleContext::{CurrentFrame, Caller(Option<MethodCaller>)}` threads the caller into both
the method-rule and ownership checks; `Caller(None)` (a top-level transaction signer) denies
`caller_component` / `caller_template`.

The new requirements are only accepted where they're meaningful: `component(..)`/`template(..)` are
rejected on component method rules, and `caller_component(..)`/`caller_template(..)` are rejected on
resource rules, so a silently-constant rule can't be constructed.

`Closes #2500` and `Closes #2502` are resolved by supplying `caller_component` / `caller_template` as
the correct primitive — `rule!(component(x))` on a method was not changed and still means "execution
within".

Motivation and Context
---

Previously there was no way to express "only component A may call this" as a single access rule:
`component(addr)` / `template(addr)` evaluate against the current frame (the callee on method rules),
not the caller. The badge-based alternative requires minting a badge, vaulting it in the caller, and
passing a proof on every call.

`caller_template(addr)` matches both a component instance of that template and a static function of it.

How Has This Been Tested?
---

- `crates/engine/tests/method_rule_component.rs` — `caller_component` allows the gated caller and denies
  unrelated components, static functions, and top-level signers; an
  `OwnerRule::ByAccessRule(rule!(caller_component(...)))` variant short-circuits only for the gated caller.
- `crates/engine/tests/method_rule_template.rs` — `caller_template` allows a component caller and a
  static-function caller, and denies a component of another template.
- `crates/engine/tests/access_rules.rs` — `caller_component` is observed through the resource auth-hook
  path (the acting account is the caller) and is denied for a non-gated acting account; `update_access_rule`
  rejects a `caller_component` resource rule with `RuntimeError::InvalidArgument` rather than aborting the
  validator node; `set_access_rules` rejects a `component`/`template` method rule (the #2500 shape) with
  `InvalidArgument`.
- `crates/template_lib_types` unit tests — the new requirements are accepted in valid contexts and panic
  in degenerate ones (constant rules rejected at construction).

```bash
cargo test -p tari_engine --test method_rule_component --test method_rule_template --test access_rules -- --test-threads=1
cargo test -p tari_engine --all-features -- --test-threads=1
```

---

1. Check out the branch and run the test commands above.
2. Run `cargo +nightly-2025-12-05 fmt --all -- --check` to confirm formatting matches CI.
3. Run `cargo clippy --all-targets --all-features` to confirm no new lint errors.

Breaking Changes
---

- [ ] None
- [x] Other - Please specify

Template *source* using `rule!(component(..))` / `rule!(template(..))` on a component method rule, or
`rule!(caller_component(..))` / `rule!(caller_template(..))` on a resource rule, now panics at construction —
a source-level break (templates recompiled from such source); already-deployed WASM is unaffected by these
builder asserts.

The engine also rejects these degenerate rules on the dynamic `ResourceAction::UpdateAccessRule` and
`ComponentAction::SetAccessRules` paths with `RuntimeError::InvalidArgument`, replacing the previous behavior of
silently installing a constant rule. This applies to all templates (including deployed WASM) but only ever
rejects rules that were already meaningless, so no data migration is required.

**Open question (for maintainers):** `RuleRequirement` gains CBOR variants `n(4)`/`n(5)`. Existing
encodings are untouched, but an older binary would hit `unknown_variant` decoding a substate that uses
them, so during a rolling upgrade this may need to sit behind a `ProtocolVersion` activation (see
`crates/engine_types/src/protocol_version.rs`). Deferring that decision to the maintainers.

<details>
<summary>Original description (kept for review context — superseded)</summary>

Description
---

Fixes `rule!(component(addr))` and `rule!(template(addr))` when used as **component method** access
rules. They were evaluated against the **callee** (the top call frame) instead of the **caller**, so a
cross-component / cross-template call made by the intended caller was rejected with `AccessDenied`.

The fix adds `WorkingState::method_caller()` (`crates/engine/src/runtime/working_state.rs`), which reads
the frame immediately below the callee and returns the caller's component and template addresses, then
threads that caller through access-rule evaluation in `crates/engine/src/runtime/tracker_auth.rs`.
`check_current_component_access_rules` now passes the caller; resource and ownership paths pass `None`
and keep their existing top-frame semantics.

The `component(addr)` and `template(addr)` bugs are addressed together because they share a single root
cause — `ScopedToComponent` and `ScopedToTemplate` are both read from the callee in the same
`check_requirement` match — and the same caller-threading change fixes both.

Note: this fixes the component/template **caller** path. A top-level transaction signer that invokes a
method directly (no component/template caller) still falls back to the callee, so a self-gated
`component(own_address)` / `template(own_template)` method remains callable directly by a signer — the
pre-existing resource-rule behavior (see the `None` case in `check_requirement`).

Motivation and Context
---

`TransactionProcessor::invoke_component` pushes the **callee** call frame _before_ running
`check_component_access_rules` (`crates/engine/src/transaction/processor.rs`). At check time,
`WorkingState::current_component()` / `current_template()` return the **top** frame — now the callee — so
`ScopedToComponent` / `ScopedToTemplate` compared the callee against itself and never the caller.

Resource rules were unaffected because they are evaluated while the caller is still the top frame, which
is why the bug went unnoticed: resource-rule coverage existed, but no method-rule coverage did.

`rule!(template(addr))` is interpreted as "the caller belongs to template `addr`", matching both a
component instance of that template and a static function of that template (the caller shape already
modeled by `AuthHookCaller` and Radix's `package_of_direct_caller`).

How Has This Been Tested?
---

New regression tests:

- `crates/engine/tests/method_rule_component.rs` — a component caller is allowed to call a method gated
  with `rule!(component(caller))`.
- `crates/engine/tests/method_rule_template.rs` — two cases: a component caller **and** a static-function
  caller are allowed to call a method gated with `rule!(template(caller_template))`.

```bash
cargo test -p tari_engine --test method_rule_component --test method_rule_template -- --test-threads=1
```

No regression in the existing authorization tests:

```bash
cargo test -p tari_engine --test access_rules -- --test-threads=1
```

Also ran the full `tari_engine` suite (`cargo test -p tari_engine --all-features`): all targets pass (a
few timing-sensitive targets required an isolated re-run). `--test-threads=1` avoids a pre-existing race
in the harness's on-demand wasm template compilation (unrelated to this change).

---

1. Check out the branch and run the two regression tests above — each asserts the caller is allowed
   (`execute_expect_success`) and fails on the unfixed engine with `AccessDenied`.
2. Run `cargo test -p tari_engine --test access_rules` to confirm the resource-rule path is unchanged.
3. Run `cargo +nightly-2025-12-05 fmt --all -- --check` to confirm formatting matches CI.

Breaking Changes
---

- [ ] None
- [x] Other - Please specify

`rule!(component(addr))` / `rule!(template(addr))` on a component **method** now evaluate against the
caller instead of the callee. This is the intended/documented semantics, but it changes runtime behavior
for any existing component that relied on the previous callee-based evaluation — for example, a method
gated with `rule!(component(own_address))` previously allowed any caller and now restricts to that
component itself.

</details>


sdbondi (changes requested):
Reviewed against the code rather than just the diff. The root cause and direction are right: `invoke_component` pushes the callee frame at `processor.rs:920` before `check_component_access_rules` at `:930`, so on a method rule `ScopedToComponent`/`ScopedToTemplate` could only ever be a tautology (`addr == callee`) or a contradiction. Reading frame `len-2` is a sound way to get the caller — frames are only pushed at real call sites, and spend-script frames can't be callers since `call_invoke` rejects on `is_cross_template_calls_allowed()` (`impl.rs:3639`).

Two things I'd like addressed before merge, plus some notes on the new attack surface this opens up.

## 1. The `None` fallback preserves the exact bug being fixed, for the most natural idiom

A top-level instruction has zero call frames, so a direct `CallMethod` gives `call_frames.len() == 1`, `method_caller()` returns `None`, and the rule compares the callee against itself — and passes. So a method gated `rule!(component(own_address))`, the "only I may call this" idiom that `Callee::new_self_gated` in this PR's own test template uses, is callable by any signer with a plain `CallMethod` instruction. Same for `template(own_template)`. `component_rule_on_method_denies_an_intruder_when_gated_on_own_address` only covers the cross-component case; there is no test for the direct top-level call, which succeeds.

The PR documents this as intentional and justifies it as "the pre-existing resource-rule behavior", but I don't think that holds. On the resource path the top frame *is* the actor performing the action, which is why current-frame semantics are correct there. On the method path the top frame is the callee, which is never an actor. When there is no caller frame the honest answer is "there is no component/template actor", i.e. deny. Given this is already a breaking change, it seems better to close the hole now than to leave a self-gate that silently doesn't gate.

The underlying reason it can't be closed as written: `Option<MethodCaller>` conflates "no caller frame" with "this isn't a method check, use the current frame". Those need to be distinguishable, e.g.

```rust
enum RuleContext {
    /// Resource, ownership and covenant checks: the current frame is the actor.
    CurrentFrame,
    /// Component method checks: the actor is the frame below the callee.
    Caller(Option<MethodCaller>),
}
```

with `Caller(None)` evaluating to `false`.

## 2. The ownership path at the same call site has the identical bug, and it outranks the method rule

`check_current_component_access_rules` runs `check_ownership(...)` first (`tracker_auth.rs:43`), which reaches `check_access_rule(state, scope, rule, None)` for `SubstateOwnerRule::ByAccessRule` — and it is evaluated at the same post-push point, so `current_component()` is the callee.

That means a component created with `OwnerRule::ByAccessRule(rule!(component(own_address)))` is owned by everyone: the ownership check short-circuits before any method rule is consulted, granting access to every method. It also affects the migration path, where `check_component_ownership` is likewise called after the frame is pushed (`processor.rs:559` then `:561`), so anyone could re-template such a component.

This is the same callee/caller confusion, in a path that grants strictly more authority than the one being fixed. Since the fix is already touching `check_current_component_access_rules`, it would be good to thread the caller into the ownership check on this path too (the `RuleContext` above makes that natural), or at minimum call it out explicitly as known-unfixed.

## Reentrancy / new attack surface

I went looking for reentrancy problems specifically. The good news first:

- Ambient authority genuinely does not propagate. `CallScope::update_from_parent` (`scope.rs:195`) is a no-op and signer virtual proofs are only seeded into the *first* frame (`push_frame`, depth 0), so cross-component calls carry no authority and proofs must be passed explicitly.
- Same-component reentrancy is structurally blocked by the lock manager: `LockedSubstates::try_lock` (`locking.rs:23`) rejects a second write lock (`MultipleWriteLockRequested`) and rejects read-while-write, so an `&mut self` method cannot be re-entered while it is on the stack.

Neither is changed by this PR. But it does make caller identity a usable authority for the first time, and unlike a proof it is conferred without the caller intending it. Two shapes worth being aware of:

- **Proxy methods.** Any `allow_all` method that forwards a caller-supplied `(component, method)` turns that component's identity into a public capability. This is exactly the shape of the test templates here (`call_bar(callee: ComponentAddress)`) — fine for a test, but it is the canonical footgun and deserves a warning in the `rule!` docs.
- **Auth hooks transfer caller identity.** `invoke_resource_access_hook` (`impl.rs:395-441`) calls `invoke_component_method` on the hook component while the *acting* component's frame is still on top, so the hook's callee observes `caller = A` even though A's code never made that call. An attacker who can get component `A` (or any component of template `T`) to act on an attacker-authored resource can therefore reach a method on a third component gated `component(A)` / `template(T)`. It is narrow — `check_resource_auth_hook` (`impl.rs:558`) constrains the callee to `(ResourceAuthAction, AuthHookCaller) -> ()` — but it is new surface created by making these rules functional, and worth a regression test.

Not a problem today, but adjacent: `PushCallFrame::MigrationContext` (`processor.rs:538`/`:548`) sets the frame's template to the caller-supplied `new_template` while the component is the one being migrated. Only that template's own migration function ever executes in that frame, so the identity is honest — but `template(...)` becomes forgeable the moment anything else runs in a migration frame. Might be worth a note at that construction site.

## Smaller points

- **The docs now contradict the behaviour.** `RuleRequirement::ScopedToComponent` is documented as "Requires execution within a specific component" (`access_rules.rs:180`), and the in-tree idiom `mintable(rule!(component(allocation.get_address())))` (`tests/templates/access_rules/src/lib.rs:195`, `builtin_component_state.rs:115`) depends on that reading. After this PR the same token means "was called by" on a method rule and "is executing as" everywhere else. My preference would be distinct `caller_component` / `caller_template` requirements rather than overloading one — the overloading is a good part of why this bug stayed invisible, and Radix separates `package_of_direct_caller` from the executing actor for the same reason. If overloading stays, the `rule!` and `RuleRequirement` doc comments need to state the split.
- **`template(T)` matching static functions** means the rule really reads "anyone who can get *any* public function or method of `T` to make this call", since every public function of `T` is callable by anyone. Reasonable as designed, but far broader than the natural reading of "only template T may call this" — please document it.
- `MethodCaller.template` is `Option<TemplateAddress>` but is always constructed `Some(*caller.current_template())`. Making it a plain `TemplateAddress` removes a template-less-caller case that cannot occur and simplifies the `ScopedToTemplate` arm.
- `method_caller()` returns `Result` but has no failure path.

## On the premise

For what it's worth, I don't think `component(addr)` on a method is a niche feature that could be left broken. Because proofs don't propagate across frames, the badge-based way to say "only component A may call this" is: mint a badge, vault it in A, and have A create and pass a proof on every call — a lot of ceremony for a static structural relationship. These requirements exist to express that without a badge, and the resource-side idiom is already load-bearing in this repo. So fixing it is the right call; I'd just rather it land fully fixed.

sdbondi (commented):
Re-reviewed at `bcadb9cf`. Both blocking items are resolved, and the new approach is better than what I suggested — introducing `caller_component` / `caller_template` as distinct requirements rather than redefining `component` / `template` keeps the existing semantics intact and makes the two concepts impossible to confuse at the call site. `RuleContext` is the right shape, `Caller(None)` correctly denies, and `MethodCaller.template` is no longer a phantom `Option`. Thanks for also adding `caller_component_rule_denies_a_top_level_signer` and `caller_component_owner_rule_short_circuits_only_for_the_caller` — those are exactly the cases I was worried about.

Particularly appreciated `it_transfers_caller_identity_through_auth_hook`: turning the auth-hook caller transfer into an asserted, named behaviour rather than leaving it as folklore is the right call, and the negative twin pins it down.

`require_ownership_from_caller` is correctly scoped — `check_component_ownership` has exactly one caller (`processor.rs:563`, the migration path), which is the one that runs post-push, so no other ownership site changes meaning.

A few loose ends before I approve.

## Generated TS bindings are stale

`bindings/src/types/RuleRequirement.ts` still lists only four variants:

```ts
export type RuleRequirement =
  | { Resource: ResourceAddress }
  | { NonFungibleAddress: NonFungibleAddress }
  | { ScopedToComponent: ComponentAddress }
  | { ScopedToTemplate: Hash32 };
```

`CallerComponent` / `CallerTemplate` need adding (both are `ComponentAddress` / `Hash32` shaped, so it is a two-line hand edit).

## Title and description now describe the abandoned approach

Worth a pass over both. As it stands the code does not "evaluate component/template method access rules against the caller" — `ScopedToComponent` / `ScopedToTemplate` ignore `RuleContext` entirely and still read the top frame. Concretely:

- The `fix(engine)!:` marker and the "Breaking Changes → Other" section no longer apply. Every pre-existing rule evaluates exactly as before; the change is purely additive (two new CBOR variants plus a new `Authorization` method). That is a nice property and it would be good for the description to say so outright.
- The description still explains the `None`-falls-back-to-callee design and the self-gate caveat, neither of which exists any more.
- `Closes #2500` / `Closes #2502`: these are now addressed by supplying a correct primitive rather than by fixing the reported rules, which I think is the right resolution — but please say that explicitly in the description, so that a reader six months from now doesn't come away believing `rule!(component(x))` on a method was fixed.

## The degenerate case that remains

`rule!(component(addr))` on a *method* rule is still a tautology when `addr` is the component's own address and a contradiction otherwise, and the same holds for `OwnerRule::ByAccessRule(rule!(component(own_address)))`, which silently means "owned by everyone" once the callee frame is on the stack. That is now consistent with the documented "execution is within this component" reading, so it is not a correctness bug — but it is precisely the trap that produced #2500 in the first place, and the fix does not remove it.

Since `ComponentAccessRules` and owner rules are constructed through builders, would it be worth rejecting `ScopedToComponent` / `ScopedToTemplate` in a component *method* rule and in an owner rule outright? A rule that can only ever be constant is never what the author meant. The mirror case is the same shape: `caller_component` / `caller_template` in a `CurrentFrame` context (a resource rule, a covenant) silently evaluate to `false`, so `mintable(rule!(caller_component(x)))` yields a permanently unmintable resource with no diagnostic.

Either direction is fine by me — validation at construction, or a prominent warning in the `rule!` docs on both halves — but silently-constant rules in an authorization API seem worth closing off given the history here.

## Wire compatibility

`RuleRequirement` derives minicbor `Decode`, so a node on an older binary hits `unknown_variant` on index 4/5 rather than degrading gracefully. Existing encodings are untouched, so this only bites for substates created *after* the new variants are in use, during a rolling upgrade. Does that need to sit behind a protocol version activation, or is the release cadence expected to cover it? Genuine question — you have the better view of the activation schedule than I do.

## Scope

The docs commits carry a wholesale Prettier reformat of `stealth-resources.mdx` (+300/-156, almost entirely quote style, table padding and SVG reflow) plus reflowed tables in `api-overview.mdx` and `types.mdx`. None of it relates to access rules and it makes the guide's history harder to read. The substantive rewrite of `authorization-and-access.mdx` is very welcome — the old examples were not even syntactically valid (`rule!(require(admin_badge)])`) — but I would split the pure reformatting out or drop it.

sdbondi (changes requested):
Bindings fix and the docs-churn revert both look right, thanks. The constant-rule validation is the right idea and the recursive `contains_requirement` + test coverage are good — but one of the six `assert!` sites is reachable from inside the validator node rather than from inside the WASM sandbox, and it aborts the process.

## `ResourceAccessRules::set_access_rule` asserts inside the engine

Five of the new asserts sit on builder methods (`ComponentAccessRules::add_method_rule` / `default`, `ResourceAccessRules::mintable` / `burnable` / ...). Those only ever run template-side, inside the WASM sandbox, where a panic is a clean transaction reject. No issue there.

`set_access_rule` is the exception: it is not a builder, and the engine calls it.

```
ResourceManager::update_access_rule(action, rule)      // template, attacker-controlled
  -> ResourceAction::UpdateAccessRule
  -> impl.rs:1834  resource_mut.update_access_rule(action, new_rule)
  -> engine_types/src/resource.rs:168
  -> ResourceAccessRules::set_access_rule    <-- assert! runs natively in the VN
```

`assert!` is not compiled out in release, `[profile.release] panic = 'abort'` (`Cargo.toml:331`), and I can't find a `catch_unwind` anywhere in the executor — the only ones in the tree are in `ootle_sdk_ffi_c`. So this looks like it aborts the validator node process, not the transaction.

The authorization check just above it (`impl.rs:1816-1827`) is no barrier: an attacker creates their own resource, is its owner, `UpdateRule::Owner` passes, and then `update_access_rule(Mint, rule!(caller_component(anything)))` panics. Every VN that executes the transaction takes the same path.

Would you mind either:

- returning `RuntimeError::InvalidArgument` (or a dedicated variant) from the `UpdateAccessRule` arm in `impl.rs` and dropping the assert from `set_access_rule`, or
- keeping the validation strictly on the builder methods, which are the template-facing surface this is meant to guard?

The second is closer to the original intent and keeps every assert inside the sandbox. Either way it would be worth a test that submits `update_access_rule(_, rule!(caller_component(..)))` and asserts a reject reason rather than a crash.

## Two smaller notes

**The validation is advisory, not an invariant.** The engine never re-validates decoded rules — `ComponentAction::SetAccessRules` (`impl.rs:1349`) assigns the whole decoded `ComponentAccessRules` without going through the builder — so a hand-written or non-Rust template can still install a constant method rule. That's fine for what this is (a template-author lint that catches the mistake at the point it's made), but the docs shouldn't imply the engine enforces it, and nothing downstream should assume the requirement can't appear.

**This makes the PR breaking again.** Any template source using `rule!(component(x))` on a method rule now panics at runtime where it previously mis-evaluated silently. Deployed wasm is unaffected, so it is a source-level break only — but it belongs in the Breaking Changes section, which currently still describes the abandoned caller-semantics change.

sdbondi (commented):
`10adf518` fixes the crash vector. The assert is gone from `ResourceAccessRules::set_access_rule`, the `UpdateAccessRule` arm returns `RuntimeError::InvalidArgument` before touching the substate, and `update_access_rule_rejects_caller_requirement` pins the reject reason rather than the crash. I checked the remaining five asserts: every one is on a builder method, and the only native (non-WASM) caller of a rule-taking builder in the tree is `template_test_tooling/src/builtin_component_state.rs:115`, which passes `rule!(component(..))` to `mintable` — a scoped requirement on a resource rule, which is permitted. So there's no engine-side panic path left.

Everything I raised across the three rounds is now addressed, and the description carries the wire-compat question for maintainers. Nice work — the shape this landed in is better than what I originally proposed.

I'm leaving the approval to a maintainer rather than clicking it myself, because the one open item is a judgement call I don't have standing to make: whether `RuleRequirement`'s new `n(4)` / `n(5)` variants need to go behind a `ProtocolVersion::V1` activation. Your read of the machinery matches mine — `protocol_version.rs` has only `V0`, every network's schedule is `[(Epoch(0), V0)]`, and `hash_substate` (`substate.rs:122`) takes the creation epoch so the schema does bind into the hash. So the gate exists and has never been used. The question is purely whether any network will have mixed binaries executing transactions that create these rules, which depends on the rollout, not the code.

Two small things for whoever takes it from here, neither blocking:

**The `rule!` doc note is slightly narrower than the truth.** It says the engine "does not re-validate component `SetAccessRules`" — resource *creation* isn't validated either, so a hand-written template can create a resource whose mint rule is `caller_component(..)` and get a permanently unmintable resource. Same harmless-but-constant outcome as the component case, worth folding into the same sentence so the claim stays exact.

**Consider mirroring the engine check on `ComponentAction::SetAccessRules`** (`impl.rs:1349`). Right now `ComponentManager::set_access_rules` is the one path where a template author gets no protection from the builder lint, and it can install `component(own_address)` on a method — the exact shape of #2500, reachable dynamically. It's not a crash and not a privilege escalation beyond what the rule already meant, so it's fine to leave. But rejecting it there would make the two dynamic paths consistent and turn the lint into something closer to an actual invariant.

sdbondi (commented):
`954a8c3c` closes the last dynamic path. Guard is on the whole rule set (default plus every method rule), the detection recurses through `AnyOf`/`AllOf`/`MOfN`, and building the degenerate value through hand-rolled CBOR in the test is the right way to prove the engine check holds independently of the builder lint. The doc note is now accurate about which paths are and aren't validated.

I also checked the traversal can't be turned into a stack overflow: `RestrictedAccessRule`'s hand-written `Decode` bounds nesting at `tari_bor::MAX_DECODE_DEPTH` before any of this runs, so anything reaching `contains_requirement` is already depth-limited. And no builtin template uses a scoped rule or `set_access_rules`, so nothing in-tree breaks on the new rejection.

Nothing further from me. The only thing still outstanding is the `ProtocolVersion::V1` question in the description, which is a maintainer call.

sdbondi (approved):
Thank you 🚀

PR #2505, opened 2026-09-02
feat(indexer): cache substates the committee agrees do not exist

sdbondi:
> **Depends on the streaming work.** This was blocked and drafted until #2517 and #2523 merged. Serving a cached nonexistence is only sound while the transition stream that would retract it is current, which the 60s scan interval could not provide. Rebased onto `bfa9c6e43` and un-drafted on that basis.

Follow-up to #2498. Adds the one result the substate cache refused to hold.

## Why

`DoesNotExist` is the most expensive lookup the indexer makes. `Up` and `Down` return on the first acceptable response; absence has nothing to prove against the state tree, so it is settled by `f + 1` agreement and walks that many committee members — every time, because it was also the one result the cache would not keep.

## Recording it

A cache row with **no version**, not a sentinel. A version is a `u32` and the column is a signed integer, so every value that could stand in for "no version" is one a real head could legitimately take — `-1` collides with `u32::MAX`, and the whole top half of the range lands in the negative space with it. Nullable avoids the question, and `Option`'s own ordering then gives the head-ranking rule for free: nonexistence yields to any version, any version displaces it.

Two gates in `CachedSubstateManager` had to admit it:

- **Serve** — exempt from the proof requirement. A cached nonexistence is never `verified`, so gating it on a proof would mean never serving one. Its evidence is the f+1 agreement that produced it, which is exactly what the uncached path returns.
- **Write** — an unversioned lookup that came back absent establishes the head as much as one that named a version.

## The journal row, and why it is scoped

Retracting a cached negative costs a journal row where none was written before. A first creation retires no version — there is none below 0 — so `created()` returned `None` and **neither** the delete nor the journal insert ran. Both are needed, and they do different jobs:

- the **delete** clears an entry already cached;
- only the **journal** stops a fetch that was already in flight when the creation committed, because that delete runs before there is a row to delete.

Emitting one for every creation would put a row on the sync commit path for each created-once substate in the stream, so it is emitted only where the entry it retracts can exist (`caches_nonexistence`, shared by the invalidation builder and the write gate so the two cannot diverge).

**Transaction receipts are excluded.** Created once and never updated, they are the bulk of the stream by count and would carry nearly all of that cost; a synced receipt is answered from the indexer's own tables without a committee walk at all; and what is left is a caller polling for one that does not exist yet — the single case where a cached negative delays rather than serves.

## Versioned reads still miss

Since #2490 a destroyed substate whose history has been pruned also reports `DoesNotExist`, so a cached negative cannot be read as "never created". It answers an unversioned read only.

## TTL

Correctness rests on the retraction, so the TTL is only a backstop for a stream that does not deliver. Nonexistence gets its own shorter one (60s, `substate_cache_negative_ttl`) because it is the entry whose staleness a caller feels as an absence — a substate being waited on stays missing until it expires.

## Migration

`2026-09-02-000000_substate_cache_nonexistence` drops and recreates `substate_cache` with a nullable version. The cache refills from the committee on demand, so there is nothing to migrate. Verified up and down against sqlite directly, including that `version <= N` does not match a null row (hence the separate `is null` delete).

Not added to `PublishedIndexerConfig`, so no openapi or bindings regeneration. A client using the published `substate_cache_max_serve_lag` as its staleness bound over-estimates for negatives, which errs in the safe direction; happy to publish it if you'd rather it be explicit.

## Serving is gated on stream liveness, not on a staleness budget

A cached head and a cached nonexistence age differently, so they get different bounds.

A head that is behind is still a lower bound on the real one — waiting only makes it more certainly true — so it keeps the loose `substate_cache_max_serve_lag` (300s). Nonexistence is correct at the instant it is taken and false ever after if the substate has since been created, so every version the stream has not yet delivered is a version in which it may already be wrong.

It is therefore served only while the shard's stream is demonstrably alive: `negative_serve_lag`, derived as `state_sync_keepalive_interval * 3`. Derived rather than configured because it is a liveness test, not a staleness budget — a standalone setting could be left behind when the keepalive interval changes. Three intervals tolerates two lost keepalives, enough to ride out an ordinary stream reopen without holding a dead one open.

Reusing `max_serve_lag` here would have meant thirty missed keepalives — a dead stream answering "does not exist" for five minutes, which is the failure that blocked this PR in the first place.

## Testing

The unit tests cannot reach the case that matters, so `Test Results (CI)` is the verdict: the two `transfer.feature` scenarios that create an account and immediately use it are the ones this feature can break, and they were red on the pre-streaming heads.

## Tests

107 pass in `tari_indexer`. The new ones cover the first-creation retraction, the mid-fetch race, destroy leaving a negative alone, the `Option` head ordering, the versioned-read miss, and the watermark requirement. Mutation-checked the two load-bearing guards:

- reverting `created()` to skip first creations fails `a_first_creation_retires_the_nonexistence_it_denies`, `a_creation_landing_mid_fetch_vetoes_the_nonexistence` and the model test;
- dropping the versioned-read guard fails `a_cached_nonexistence_does_not_answer_a_versioned_read`.

### DIFF
--- a/applications/tari_indexer/src/bootstrap.rs
+++ b/applications/tari_indexer/src/bootstrap.rs
@@ -115,6 +115,15 @@ const SUBSTATE_CACHE_JOURNAL_RETENTION: Duration = Duration::from_secs(300);
 
 const SUBSTATE_CACHE_PRUNE_INTERVAL: Duration = Duration::from_secs(60);
 
+/// Keepalives a shard's stream may miss before a record that a substate does not exist stops being
+/// served for it.
+///
+/// Derived from the keepalive interval rather than configured: the bound is not a staleness budget
+/// but a liveness test - a negative is only as true as the stream that would retract it - and a
+/// standalone setting could be left behind when that interval changes. Two may be lost before the
+/// cache closes, which is enough to ride out an ordinary reopen without holding a dead stream open.
+const NEGATIVE_SERVE_KEEPALIVES: u32 = 3;
+
 #[allow(clippy::too_many_lines)]
 pub async fn spawn_services(
     config: &ApplicationConfig,
@@ -294,6 +303,7 @@ pub async fn spawn_services(
         store.clone(),
         shard_watermarks,
         config.indexer.substate_cache_max_serve_lag,
+        config.indexer.state_sync_keepalive_interval * NEGATIVE_SERVE_KEEPALIVES,
         SUBSTATE_CACHE_JOURNAL_RETENTION,
         DEFAULT_CACHE_TTL,
         config.indexer.substate_cache_max_entries,
--- a/applications/tari_indexer/src/bootstrap.rs
+++ b/applications/tari_indexer/src/bootstrap.rs
@@ -333,12 +333,21 @@ pub async fn spawn_services(
     .await
     .context("template manager init thread panicked")??;
 
-    // Dry run - use a shorter cache TTL for more accurate fee estimates. Proof verification is left
-    // off here: dry run only produces a fee estimate, and gating it on proof availability would make
-    // transaction submission fragile when proofs are momentarily unavailable.
+    // Dry run - use a shorter cache TTL for more accurate fee estimates. A nonexistent input is held
+    // for the shorter of the two: an input resolved as absent estimates a transaction that would in
+    // fact have succeeded, which is the same staleness `dry_run_cache_ttl` is set to bound.
+    // Proof verification is left off here: dry run only produces a fee estimate, and gating it on
+    // proof availability would make transaction submission fragile when proofs are momentarily
+    // unavailable.
     let dry_run_substate_manager = substate_manager
         .clone()
         .with_cache_ttl(config.indexer.dry_run_cache_ttl)
+        .with_negative_cache_ttl(
+            config
+                .indexer
+                .dry_run_cache_ttl
+                .min(config.indexer.substate_cache_negative_ttl),
+        )
         .with_substate_proof_verification(false);
     let fee_table = get_fee_table_by_network(config.network);
     let dry_run_transaction_processor = DryRunTransactionProcessor::new(
--- a/applications/tari_indexer/src/bootstrap.rs
+++ b/applications/tari_indexer/src/bootstrap.rs
@@ -306,7 +306,8 @@ pub async fn spawn_services(
         validator_node_client_factory.clone(),
         substate_cache,
     )
-    .with_substate_proof_verification(config.indexer.verify_substate_proofs);
+    .with_substate_proof_verification(config.indexer.verify_substate_proofs)
+    .with_negative_cache_ttl(config.indexer.substate_cache_negative_ttl);
     #[cfg(feature = "metrics")]
     let substate_manager = substate_manager.with_metrics(metrics_registry);
 
--- a/applications/tari_indexer/src/config.rs
+++ b/applications/tari_indexer/src/config.rs
@@ -32,6 +32,7 @@ use tari_common::{
     configuration::{CommonConfig, serializers},
 };
 use tari_crypto::ristretto::RistrettoPublicKey;
+use tari_indexer_lib::cached_substate_manager::DEFAULT_NEGATIVE_CACHE_TTL;
 use tari_ootle_app_utilities::{
     epoch_oracle_config::EpochOracleConfig,
     p2p_config::{P2pConfig, PeerSeedsConfig},
@@ -187,6 +188,15 @@ pub struct IndexerConfig {
     /// are evicted, at the cost of one validator round trip each to fetch again.
     #[serde(default = "default_substate_cache_max_entries")]
     pub substate_cache_max_entries: usize,
+    /// How long the cache serves a substate it has established does not exist. A creation retracts
+    /// that through the transition stream, so this bounds only the case where the stream does not
+    /// deliver.
+    ///
+    /// Held shorter than the other entries because it is the one a caller feels as an absence: a
+    /// substate being waited on stays missing until this expires. Set it to zero to answer every
+    /// nonexistent lookup from the committee, at `f + 1` round trips each.
+    #[serde(default = "default_substate_cache_negative_ttl", with = "serializers::seconds")]
+    pub substate_cache_negative_ttl: Duration,
     /// How many epochs past its terminal epoch a stored transaction is retained before it is pruned.
     /// A transaction's terminal epoch is the epoch it committed in once its receipt has been
     /// indexed, and its `max_epoch` — the last epoch it could still be sequenced in — until then, so
@@ -268,6 +278,10 @@ fn default_substate_cache_max_serve_lag() -> Duration {
 
 fn default_substate_cache_max_entries() -> usize {
     100_000
+}
+
+fn default_substate_cache_negative_ttl() -> Duration {
+    DEFAULT_NEGATIVE_CACHE_TTL
 }
 
 /// The subset of an indexer's configuration that is published over its API, as it affects what
@@ -414,6 +428,7 @@ impl Default for IndexerConfig {
             allow_past_protocol_activation: false,
             substate_cache_max_serve_lag: default_substate_cache_max_serve_lag(),
             substate_cache_max_entries: default_substate_cache_max_entries(),
+            substate_cache_negative_ttl: default_substate_cache_negative_ttl(),
             transaction_retention_epochs: default_transaction_retention_epochs(),
             index_gossiped_transactions: default_index_gossiped_transactions(),
             max_transaction_gossip_queue_bytes: default_max_transaction_gossip_queue_bytes(),
--- a/applications/tari_indexer/src/config.rs
+++ b/applications/tari_indexer/src/config.rs
@@ -189,12 +189,19 @@ pub struct IndexerConfig {
     #[serde(default = "default_substate_cache_max_entries")]
     pub substate_cache_max_entries: usize,
     /// How long the cache serves a substate it has established does not exist. A creation retracts
-    /// that through the transition stream, so this bounds only the case where the stream does not
-    /// deliver.
+    /// that through the transition stream, so ordinary staleness is bounded by the sync round and
+    /// this covers only what that stream cannot correct.
+    ///
+    /// What it cannot correct is an answer that was already wrong when it was cached. The committee
+    /// is chosen from the epoch manager's current epoch, independently of the watermark that gates
+    /// the write, so a lagging view of a shard group split can put the question to a committee that
+    /// honestly no longer holds the substate and agrees it does not exist. Uncached that misleads
+    /// one caller; cached it is served to every caller until this expires.
     ///
     /// Held shorter than the other entries because it is the one a caller feels as an absence: a
-    /// substate being waited on stays missing until this expires. Set it to zero to answer every
-    /// nonexistent lookup from the committee, at `f + 1` round trips each.
+    /// substate being waited on stays missing until this expires. Lowering it towards zero narrows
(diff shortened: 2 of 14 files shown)


sdbondi (commented):
Read at head `d3c4869`. The core of this is sound: nullable version rather than a sentinel is the right call given the `u32`/signed-integer collision, `Option`'s ordering falls out correctly in the write gate (`None > Some(_)` is false, `Some(_) > None` is true), the versioned-read miss is justified by #2490's pruned-destroy behaviour, and the invalidation buffer is filled before the `value_filters` early return in `extend_bufs_from_substate_update`, so the retraction covers substates whose values the stream never carries. The mid-fetch veto test exercises the one race the journal exists for.

One blocking item, which is about where the new TTL does *not* reach.

## Blocking

**`applications/tari_indexer/src/bootstrap.rs:337` — the dry-run manager inherits the 60s negative TTL while overriding the positive one to 10s.**

`substate_manager` gets `.with_negative_cache_ttl(config.indexer.substate_cache_negative_ttl)` at line 308. `dry_run_substate_manager` is cloned from it at 337 and overrides only `cache_ttl` (`dry_run_cache_ttl`, default 10s). `negative_cache_ttl` carries over at 60s.

So on the one path the config explicitly tunes for freshness — "use a shorter cache TTL for more accurate fee estimates" (bootstrap.rs:334), "A shorter TTL reduces the chance of stale fee estimates" (config.rs:126) — negatives are now held **six times longer than positives**, exactly inverting this PR's own stated principle that nonexistence is "held shorter than the other entries because it is the one a caller feels as an absence". Before this PR the dry-run path never cached a negative at all, so this is a regression in that contract, not just an omission.

Concretely: a dry run whose input was created within the last 60s, on a shard whose stream is inside `max_serve_lag` but has not yet delivered the creation, resolves the input as absent and returns a fee estimate for a transaction that would actually have succeeded — for up to 60s, where the 10s TTL was chosen to keep that window at 10.

One line:

```rust
let dry_run_substate_manager = substate_manager
    .clone()
    .with_cache_ttl(config.indexer.dry_run_cache_ttl)
    .with_negative_cache_ttl(config.indexer.dry_run_cache_ttl.min(config.indexer.substate_cache_negative_ttl))
    .with_substate_proof_verification(false);
```

## Non-blocking

**`applications/tari_indexer/src/config.rs:155` — "Set it to zero" does not do what the doc says.** The serve gate is `age <= ttl.as_secs()` (cached_substate_manager.rs:209), so with `ttl = 0` an entry cached in the current second still answers, and the write still happens on every miss. Zero narrows the window to sub-second rather than "answer every nonexistent lookup from the committee". Either reword, or gate on `age < ttl` and accept that the same edge moves for `cache_ttl`.

**`applications/tari_indexer/src/network_state_sync/worker.rs:918` — stale doc comment.** "a substate's first creation retires nothing and is dropped" is precisely what this PR changes: a first creation of anything `caches_nonexistence` admits now reaches `invalidations_buf`. It reads as a live invariant and now contradicts the code two lines below it.

**Negative-cache poisoning across an epoch boundary is worth a sentence somewhere.** `get_specific_substate_from_committee` picks the committee from `current_epoch` independently of the watermark that gates the write. If the epoch manager lags a shard-group split, the old committee can honestly return f+1 `DoesNotExist` for a substate it no longer holds — previously a wrong answer to one caller, now persisted and served to every caller for the TTL. The write-side `watermark.is_some()` gate makes this unlikely (a current shard implies a roughly current epoch view), so I don't think it needs code, but it is the failure mode the TTL is actually insuring against and the config doc frames the TTL purely as a stream-delivery backstop.

## Nits

**`crates/indexer_lib/src/substate_cache.rs:65` — `caches_nonexistence` is not enforced at the storage boundary.** Both current call sites are correct, but `SqliteSubstateCache::write` / `substate_cache_put` will happily persist a null-version row for a receipt, and nothing would ever emit a journal row to retract it — `created()` returns `None` for exactly those. The failure is silent and only bounded by the manager's TTL. A guard in `substate_cache_put` (or a debug assert) makes the invariant local to the thing that can violate it.

**`applications/tari_indexer/src/storage_sqlite/writer.rs:635-651` — two DELETEs per creation.** `version <= N` and `version is null` are disjoint filters on the same row set; `.or_filter()` would issue one statement. Only worth it because this is on the sync commit path for every created substate now.

**`SubstateCacheInvalidation::created` sets `retires_nonexistence: true` unconditionally**, including for a receipt at version >= 1 — where it drives a DELETE that can never match. Harmless (receipts are never updated), but the field then means "this transition would retire a nonexistence" rather than "an entry this can retire may exist", which is the distinction the `created()` early-return is drawing one line above.

sdbondi (commented):
`2f12565` resolves every finding from the previous round — the dry-run TTL clamp, the config doc, the `worker.rs` comment, the epoch-boundary note, and both nits, one of them (`caches_nonexistence` at the storage boundary) better than what I suggested. `rustfmt --check` with the repo config is clean, and clippy, fmt and all three test shards pass.

I was about to approve on that. Then the `Test Results (CI)` check landed red, and it is a real regression, not flake.

## Blocking — a cached negative outlives the creation that should retract it

Two scenarios fail, on **both** heads of this PR (`d3c4869` and `2f12565`):

- `Transfer tokens to account that does not previously exist` — transfer.feature:8
- `Transfer tokens to existing account` — transfer.feature:62

Both fail the same way: the step immediately after a transfer materialises the destination account submits a manifest against it and gets `500 Substate component_… does not exist`.

They are not flaky. #2498's head — the positive cache, everything here minus the negative — is `All 2 021 tests pass`, and I checked the last 25 commits on `development`: zero transfer.feature failures. On this branch it is these two and nothing else, twice.

The indexer log from the run has the whole thing (`cucumber-log-artifacts`, `ootle.log`), second scenario, component `component_280f6d…`:

```
09:00:39  Updating cached substate component_280f6d… with version None   cached_substate_manager.rs:258
09:00:49  Substate cache hit for component_280f6d… with version None     cached_substate_manager.rs:210
09:00:49  New component created: component_280f6d…                        tracker.rs:332
09:00:49  Up: component_280f6d… v0 Shard(41)                              pending_store.rs:256
09:00:50  Substate cache hit for component_280f6d… with version None      <- the 500
```

The first scenario is the same shape inside a single second: cached at 09:00:26, `Up ... v0 Shard(43)` at 09:00:26, served as nonexistent at 09:00:27, and never retracted before the scenario gave up — that hit is the last mention of the substate in the log.

**This is not the mid-fetch race.** That one the journal does close, and `a_creation_landing_mid_fetch_vetoes_the_nonexistence` covers it. The window that is open is the one after: the creation commits on the network, and the transition carrying it has not yet reached this indexer. There is no journal row to veto against and no row present to delete, because the delete and the journal insert both happen on *delivery*, not on commit. Until the stream catches up, the entry is not stale-but-safe — it is wrong.

That window is a full sync round, which `substate_cache_max_serve_lag` sizes deliberately: its own doc says it "must comfortably exceed a full sync round" and that "ordinary staleness is bounded by the sync round". For a head that is fine and the PR argues it well — a cached head is a lower bound on the real one, so age only makes it more certainly true. Absence has no such monotonicity. A negative is correct only at the instant it is taken, and every sync round it survives is a round in which it can be flatly false. The watermark gate asks whether the shard is within a lag budget; serving a negative needs it to be at the tip, which is a different question. The 60s TTL never enters into it — the failure here is at one second.

The two scenarios that break are the two in the suite that create an account and immediately use it, which is also the case the description names as the one where a cached negative "delays rather than serves". It turns out that case is not confined to receipt polling: the wallet's account monitor polls the indexer for a locally-created account on every pass, so the negative is not just cached once but continuously refreshed right up until the account lands on chain.

I do not think there is a one-line version of this. Roughly, the options are to gate serving a negative on the shard being at the tip rather than within `max_serve_lag`; to keep the entry but re-confirm it against the committee before serving (which gives back most of the saving); or to narrow `caches_nonexistence` to substates where "not yet" is harmless — which excludes account components, and they are much of the motivation.

## Also worth knowing: CI will not tell you about this

`integration test` reported **success** on both heads. `integration_tests/tests/cucumber.rs:108` ends in `.filter_run(...)`, not `filter_run_and_exit`, so a failed scenario never fails the process. The only surface is the `Test Results (CI)` check, which is easy to miss under a green job. Worth a separate issue — a red suite that reports green is going to cost someone a merge sooner or later.

Repro without waiting on CI:

```
cargo test --profile ci-build --package integration_tests --test cucumber -- --tags @transfer
```

sdbondi (commented):
Still blocked. The same two scenarios fail on `52c42d7`, third run in a row — and the new logs show the liveness gate cannot reach this case.

## The stream was alive, confirmed, and had delivered

`ootle.log` from run 33866313665, second scenario, `component_9d0ab04c…` on `Shard(158)`, with line numbers so it can be followed:

```
304053  11:55:28  Updating cached substate component_9d0a… with version None   <- account monitor refresh
305945  11:55:30  Substate cache hit … with version None                        (legitimate, not yet created)
306093  11:55:30  Up: component_9d0a… v0 Shard(158)                             <- created
306846  11:55:30  Completed state sync for shard Shard(158) … state version 1   <- confirm
307435  11:55:30  Completed state sync for shard Shard(158) … state version 1   <- confirm
307563  11:55:30  Substate cache hit … with version None                        <- the 404
```

The failing serve is at 307563, **after** two confirmations of that shard's stream. There is no `Updating cached` after 304053, so the entry was not re-written by a later fetch — it simply survived the creation. `NEGATIVE_SERVE_KEEPALIVES` cannot bite here at any value: the shard was confirmed twice in the same second as the read.

## Why tightening the lag cannot fix it

`ShardWatermarks::get` is `confirmed_at.elapsed() <= max_lag` — a recency test on the last confirmation. And `confirm(shard, V)` means "I hold every transition up to V", never "V is the committee's tip". Those are different claims, and only the second is what a negative needs. A substate created at V+1 is invisible to the gate no matter how tight the bound, because the gate never asks whether V is current — which is exactly what 306846 shows: a confirmation to version 1 sitting between the creation and a serve that still answers "does not exist".

So the commit's reasoning is right — a negative is a liveness question, not a staleness budget — but `confirmed_at` is not the liveness signal it needs. It fixes the dead-stream case, which was real and worth fixing, and leaves this one untouched.

## The other half: the wallet refreshes the negative into place

Line 303893, two lines before the fetch: `🏦 Refreshing account component_9d0a…`. The wallet's own account monitor polls the indexer for an account it holds locally, and each poll re-caches the negative with a fresh `cached_at`. So the 60s TTL never expires — the entry is continuously renewed right up to the moment the account is created, and only a delivered retraction can clear it. The wallet is poisoning the cache that its own next transaction then reads.

That also means the receipt exclusion does not bound the exposure the way the description argues. The polling case is not confined to receipts; it is any substate a wallet is waiting on, and account components are the common one.

## A direction, not a demand

The entry already carries a `FetchWatermark`. If it were stored and the serve required the shard's watermark to still equal it — serve only while nothing at all has happened on that shard since the negative was taken — this case is caught: cached at version 0, shard at version 1 by 306846, so 307563 misses. It is a real cost on a busy shard, where the negative would almost never serve, but it is the honest bound: a cached negative is only as good as "nothing has landed here since I asked", and any weaker gate is a guess about a window rather than a proof about it.

Whatever the shape, the check I would want before this goes in is the transfer.feature pair going green on a head that has not otherwise changed.

## Unrelated, still worth doing

`integration test` reported **success** on all three heads while these scenarios failed — `cucumber.rs` ends in `.filter_run(...)`, not `filter_run_and_exit`, so a failed scenario never fails the process. Only the `Test Results (CI)` check surfaces it. Worth its own issue regardless of what happens here.

Repro:

```
cargo test --profile ci-build --package integration_tests --test cucumber -- --tags @transfer
```

PR #2506, opened 2026-09-02
fix(consensus): re-record justified-block evidence on the surviving branch

sdbondi:
## Summary

Fixes the flaky `consensus_tests leader_failure::multi_shard_node_goes_down`, which is a real consensus liveness bug, not a test artifact.

The prepare/accept evidence for a block's `LocalPrepare`/`LocalAccept` commands is recorded by the block that *justifies* it, into that block's `ProposedBlockChangeSet`. A change set only ever applies to the branch its block is on, but the "already justified" guard was a persisted per-block flag: the first block to justify `B` stamped `B.justify_qc_id` straight to the block store (`Block::add_justify_qc` writes to the write tx directly), whether or not that block went on to be certified. Every later block that justified `B` then skipped `process_newly_justified_block`, so when the branch that stamped it was abandoned the evidence went with it.

A multi-shard transaction whose local accept evidence is lost this way sits in the pool at `LocalAccepted` with `is_ready=false` and local evidence stuck at `PREPARED (need accept)`. The foreign shard group has already sent its `LocalAccept` foreign proposal, so nothing will mark it ready again — the shard group proposes empty blocks forever while `has_ready_or_pending_transaction_updates` reports nothing to do and the pool never drains.

## The failure in CI

From [run 33536033189](https://github.com/tari-project/tari-ootle/actions/runs/33536033189/job/99950450348?pr=2501), where the leader that would have certified the justifying block was the offline validator:

| block | what happened |
|---|---|
| h7 `084ddace` | 10 `LocalAccept` commands, evidence `SG(1-128): PREPARED (need accept)` |
| h8 `50d9ad60` | justifies h7 → evidence `SG(1-128): ACCEPTED`, **and stamps `justify_qc_id` on h7** |
| — | h8 never gets a QC (votes go to the offline node 4, leader of height 8) → TC at h9, dummy h8/h9 built from h7, **h8 orphaned** |
| h10 `ed84e8f7` | justifies h7 again → guard says already justified → **skipped**; record loads back as `PREPARED (need accept)` |
| h16/h17 | foreign `LocalAccept` from `SG(129-256)` arrives → `NOT ready for AllAccept … All Justified: false` |
| h18…h51 | 0 cmd(s) every block, 11 transactions stuck in the pool, test times out at the height limit |

`🔍 Updating evidence for LocalAccept command … NodeHeight(7)` appears exactly once in the whole run — inside the orphaned block.

It is flaky because it needs the block immediately after a `LocalAccept`-carrying block to be orphaned, i.e. the offline validator has to land as leader one height later.

## The fix

Always run `process_newly_justified_block` so each branch records the evidence for itself, and drop the two guards that assumed the flag was branch-local:

- `common.rs` — remove the `new_leaf_block.justify_qc_id() == Some(justify_id)` early return. Both the abandoned block and the block that replaces it carry the *same* QC over the justified block, so this fired on exactly the case that needs re-recording.
- `on_ready_to_vote_on_local_block.rs` / `on_propose.rs` — remove the `!justified_block.has_justify_qc()` guard.

Persisting the flag only on `change_set.save()` would not help: validators voted Accept on the orphaned block and saved its change set; the block died afterwards.

The parent walk still terminates on the first ancestor carrying a justify QC, so the extra work is bounded — in an unforked chain each block justifies a different parent and nothing is recomputed. A transaction that has left the pool is now skipped rather than raising an invariant error, since a finalized transaction has no evidence left to update; that path becomes reachable now that a late sibling can re-justify a block whose transactions have since been finalized.

## Test plan

- [x] `cargo test -p consensus_tests` — 67 passed, 0 failed
- [x] `cargo clippy -p tari_consensus --all-targets` — no new warnings

I could not land a deterministic regression test for this. Reproducing it needs a block to be evaluated by a quorum (so it stamps the flag) and then abandoned, and every way to do that from the harness's message filter runs into the same two walls: dropping the block's votes still lets the next leader build the certificate out of the `last_vote` piggybacked on NEWVIEWs, and withholding the block from enough validators to deny the quorum also denies them the parent's certificate, so whichever of them wins the timeout round forks from further back and the scenario evaporates. Both hinge on which validator wins which timeout round, which is exactly the race that makes `multi_shard_node_goes_down` flaky in the first place. Doing it properly wants a harness hook (suppress voting on a chosen block, or drop `last_vote` while keeping the timeout vote) — happy to add that separately if it is wanted here.

### DIFF
--- a/crates/consensus/src/hotstuff/common.rs
+++ b/crates/consensus/src/hotstuff/common.rs
@@ -30,6 +30,7 @@ use tari_ootle_storage::{
         EpochCheckpoint,
         PendingShardStateTreeDiff,
         SubstateChange,
+        TransactionRecord,
         TreeRootSummary,
         ValidatorConsensusStats,
     },
@@ -493,13 +494,24 @@ pub fn process_newly_justified_block<TTx: StateStoreReadTransaction>(
                 .get_transaction_pool_record(tx, new_leaf_block, atom.id())
                 .optional()?
             else {
-                // A transaction that has left the pool has been finalized, so it has no evidence left to update.
-                debug!(
-                    target: LOG_TARGET,
-                    "Transaction {} in newly justified block {} is no longer in the pool",
-                    atom.id(),
-                    new_leaf_block,
-                );
+                // Finalizing a transaction removes its pool record, and a finalized transaction has no evidence
+                // left to update. Any other absence leaves the transaction unable to make progress.
+                if TransactionRecord::is_record_finalized(tx, atom.id())? {
+                    debug!(
+                        target: LOG_TARGET,
+                        "Transaction {} in justified block {} is finalized and no longer in the pool",
+                        atom.id(),
+                        block,
+                    );
+                } else {
+                    warn!(
+                        target: LOG_TARGET,
+                        "❓️ Transaction {} in justified block {} is not in the pool and is not finalized. Its \
+                         evidence cannot be updated.",
+                        atom.id(),
+                        block,
+                    );
+                }
                 continue;
             };
 
@@ -511,7 +523,7 @@ pub fn process_newly_justified_block<TTx: StateStoreReadTransaction>(
                 debug!(
                     target: LOG_TARGET,
                     "🔍 Updating evidence for LocalPrepare command in block {} for transaction {}. {}",
-                    new_leaf_block,
+                    block,
                     atom.id(),
                     pool_tx.evidence()
                 );
@@ -523,7 +535,7 @@ pub fn process_newly_justified_block<TTx: StateStoreReadTransaction>(
                 debug!(
                     target: LOG_TARGET,
                     "🔍 Updating evidence for LocalAccept command in block {} for transaction {}. {}",
-                    new_leaf_block,
+                    block,
                     atom.id(),
                     pool_tx.evidence()
                 );
@@ -537,7 +549,7 @@ pub fn process_newly_justified_block<TTx: StateStoreReadTransaction>(
                     target: LOG_TARGET,
                     "✅ Justified block: Setting READY for transaction {} in block {}",
                     atom.id(),
-                    new_leaf_block,
+                    block,
                 );
                 pool_tx.set_ready(true);
             }
--- a/crates/consensus/src/hotstuff/common.rs
+++ b/crates/consensus/src/hotstuff/common.rs
@@ -438,6 +438,11 @@ pub(crate) fn get_highest_seen_justified_view<TTx: StateStoreReadTransaction>(
     Ok(high_pc.max(high_tc))
 }
 
+/// Records the prepare/accept evidence for the commands in `new_leaf_block` and its not-yet-justified ancestors,
+/// now that `justify_id` justifies them, and marks any transaction that this completes as ready to propose.
+///
+/// The updates are written to `change_set`, which only ever applies to the branch its block is on, so every branch
+/// that justifies `new_leaf_block` must call this for itself.
 #[allow(clippy::too_many_lines)]
 pub fn process_newly_justified_block<TTx: StateStoreReadTransaction>(
     tx: &TTx,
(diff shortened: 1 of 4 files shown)


sdbondi (commented):
Reviewed at `f9ed279`. No blocking findings — the diagnosis holds up against the code and the fix is the right shape.

## What I verified

**The guard really was non-branch-local.** `Block::add_justify_qc` (`crates/storage/src/consensus_models/block.rs:711`) writes straight to the write tx and is called from `on_ready_to_vote_on_local_block.rs` *before* `should_vote`/`change_set.save`, so the flag is stamped whether or not the node votes and whether or not the stamping block survives. Meanwhile the evidence it gates lives only in that block's `ProposedBlockChangeSet`, which is branch-scoped (`transaction_pool_get_for_blocks` overlays updates by walking `get_pending_chain_ordered(to_block_id)`, `crates/state_store_rocksdb/src/reader.rs:1212`). Gating branch-scoped work on globally-persisted state is the bug; removing the gate is the fix.

**The fix closes the class, not just the observed instance.** Evidence for block `A` is recorded into the change set of whichever block carries `QC(A)`. It is lost only when that block is not an ancestor of the surviving leaf — which means the surviving branch forked at or below `A`, and after a TC the new branch's first real block carries the high QC (`build_next_block`, `on_propose.rs:346-348`), i.e. `QC(A)`, so `A` becomes `justified_block` and is now reprocessed. The ancestor walk stopping at the first stamped parent is therefore still sufficient: any block reachable only through the walk (dummies, or a parent whose own QC-carrier survived) either has no commands or has its evidence on the live chain.

**Determinism improves rather than degrades.** Both guards were reads of node-local persisted state, so proposer and voters could disagree about whether to apply the evidence depending on which orphaned blocks each had seen. Now both `on_propose.rs:445` and `on_ready_to_vote_on_local_block.rs:157` apply it unconditionally, and `process_newly_justified_block` loads pool records at the same anchor (`new_leaf_block.as_leaf()`) on both paths.

**No new termination risk from dropping the early return.** The walk from an epoch genesis block reaches the zero block, which is stamped at bootstrap (`worker.rs:1518`, `add_justify_qc(tx, &PcId::zero())`), so it breaks before `get_parent` can hit the `NotFound` on the zero block. And the early return only fired on an *exact* `justify_id` match, so for every other case the walk already ran — the reachable depth is unchanged.

**Evidence updates are monotonic.** `set_prepare_qc`/`set_accept_qc` are plain setters and `set_ready` is only ever raised, so re-running the recording on a live branch can't regress a record. The only remaining consumer of `justify_qc_id` is `has_justify_qc()`; nothing reads the id itself, so the now-unconditional `add_justify_qc` overwriting a dead branch's QC id is inert.

**`can_propose_epoch_end |= justified_block.is_epoch_end()` moving out of the guard is safe.** It only unlocks *voting* on an `EndEpoch` command; the next-epoch hash is still ratified against the local oracle (`on_ready_to_vote_on_local_block.rs:441-460`), so a node still never lends quorum to a hash it hasn't seen.

**The `continue` is justified.** Pool records are removed at `on_commit` (`on_ready_to_vote_on_local_block.rs:1753`) and at the two startup/state-sync reconciliation paths (`worker.rs:1089`, `worker.rs:1142`) — all irreversible — so "not in the pool" does mean finalized, and `on_commit` runs before `process_newly_justified_block` in the same `handle` call, which is exactly how a late re-justification reaches it.

## Non-blocking

1. `common.rs:496` — the skip is now silent at `debug`, and it swallows more than the finalized case: any future bug that loses a pool record for a still-pending transaction (the #2496 class) now degrades to a missing-evidence stall instead of a loud `InvariantError`. Since the justified case is cheap to distinguish, consider confirming the transaction is actually finalized before skipping quietly (`TransactionRecord`'s final decision, or the receipt-exists check `reconcile_transaction_pool` already uses) and `warn!`ing otherwise. At minimum `warn!` rather than `debug!` — this path should be rare enough that the noise is worth it.

2. `common.rs:499` and the two sibling `debug!`s at `:507`/`:518` say "in newly justified block {new_leaf_block}" but the command comes from `block`, which during the ancestor walk is not the leaf. Pre-existing in the other two, but this fix makes the walk fire more often, so the log gets more misleading. Logging `block` (or both) would have made the CI trace in the description easier to read.

3. `justify_qc_id` now means "some branch, somewhere, once justified this block" rather than "the QC that justified this block on our chain", and its only remaining use is as the walk terminator. The new doc comment on `process_newly_justified_block` explains the change-set side well; a one-liner on `Block::justify_qc_id`/`has_justify_qc` saying the field is not branch-local would stop the next reader from reaching for it as a branch-local guard again.

## On the missing regression test

The two walls described (votes reconstructed from `last_vote` on NEWVIEW; withholding the block also withholding the parent's certificate) match what the harness's message filter can express, so I agree a deterministic repro needs a harness hook. Worth doing separately — the failure mode is a permanent liveness stall, and without a test the guard can be reintroduced as an "optimisation" by anyone reading `has_justify_qc()` at face value (see 3).

CI is green apart from `test (1/3)`, still pending at the time of writing.

PR #2510, opened 2026-09-02
fix(walletd): answer a missing substate as not found, not a general error

sdbondi:
Spun out of the discussion on #2505.

## The problem

Resolving a transaction's inputs against a substate that does not exist returns `ApplicationErrorCode::GeneralError` (500). This is what the failing `transfer.feature` scenarios on #2505 surfaced as `500 Substate component_… does not exist`.

A caller polling for something that does not exist **yet** — an account it has just created, an input whose creating transaction has not yet reached the indexer — cannot distinguish that from the wallet being broken. Only one of the two is worth retrying, and a general error tells the caller the wrong one.

## The fix

The distinction was already carried and simply never reached the JSON-RPC layer. `SubstateApiError` implements `IsNotFoundError` and classifies `SubstateDoesNotExist` correctly; the handler returned it as an opaque `anyhow::Error`, so `resolve_any_error` fell through to the general arm.

`OrNotFound::or_not_found` consults that classification and routes a not-found through the **existing** `helpers::not_found`, which is what ~10 other handlers already use. Applied to the input-resolution paths a caller can reach with a substate that is not there yet (`locate_dependent_substates`, four call sites in `transaction.rs`).

## Net-subtractive on the error plumbing

`HandlerError::NotFound` was a second path to the same error code, and a lossy one — it flattened every message to the literal string `"Not found"`. Its three call sites are all "get transaction by id"; they now say which transaction was not found, via the same helper as everything else. That leaves the variant unused, so it and its arm in `resolve_handler_error` are removed.

## Scope

Deliberately limited to the substate/transaction paths this issue is about. Around twenty error types in the wallet implement `IsNotFoundError`; `or_not_found` generalises to any of them, but converting the rest is a separate sweep with its own review surface, and each one is a behaviour change for whatever client depends on the current code.

## Tests

Two new unit tests pin the classification in both directions — a missing resource gets the `NotFound` code and keeps its own message, and anything else is passed through untouched (so nothing unrelated starts looking retryable). 40 pass in `tari_ootle_walletd`.

### DIFF
--- a/applications/tari_walletd/src/handlers/error.rs
+++ b/applications/tari_walletd/src/handlers/error.rs
@@ -5,6 +5,4 @@
 pub enum HandlerError {
     #[error("Error: {0}")]
     Anyhow(#[from] anyhow::Error),
-    #[error("Not found")]
-    NotFound,
 }
--- a/applications/tari_walletd/src/handlers/helpers.rs
+++ b/applications/tari_walletd/src/handlers/helpers.rs
@@ -9,7 +9,10 @@ use tari_engine_types::{
     component::derive_component_address_from_public_key,
     confidential::{AbridgedTransactionKernel, EncodedMerkleProof, MinotariBurnClaimProof},
 };
-use tari_ootle_common_types::{SubstateRequirement, optional::Optional};
+use tari_ootle_common_types::{
+    SubstateRequirement,
+    optional::{IsNotFoundError, Optional},
+};
 use tari_ootle_transaction::TransactionId;
 use tari_ootle_wallet_sdk::{
     WalletSdk,
@@ -175,6 +178,30 @@ pub(super) fn not_found<T: Display>(details: T) -> anyhow::Error {
     .into()
 }
 
+/// Answers a missing resource as [`not_found`] rather than as a general error.
+///
+/// A caller asking for something that does not exist yet - an account it has just created, an input
+/// whose creating transaction has not reached the indexer - has to be able to tell that apart from
+/// the wallet failing. Only the first is worth retrying, and a general error tells it the opposite.
+///
+/// The distinction is already carried by [`IsNotFoundError`]; this puts it where the JSON-RPC layer
+/// reads it, keeping the error's own message.
+pub(super) trait OrNotFound<T> {
+    fn or_not_found(self) -> Result<T, anyhow::Error>;
+}
+
+impl<T, E: IsNotFoundError + Into<anyhow::Error>> OrNotFound<T> for Result<T, E> {
+    fn or_not_found(self) -> Result<T, anyhow::Error> {
+        self.map_err(|e| {
+            if e.is_not_found_error() {
+                not_found(e.into())
+            } else {
+                e.into()
+            }
+        })
+    }
+}
+
 pub(super) fn invalid_request<T: Display>(details: T) -> anyhow::Error {
     application_error(
         ApplicationErrorCode::InvalidRequest,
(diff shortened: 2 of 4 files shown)


sdbondi (commented):
Reviewed at `0844ea9`. Nothing blocking.

## Verified

- The classification source is right and the routing matches the existing convention. `SubstateApiError::is_not_found_error` (`crates/wallet/sdk/src/apis/substate.rs:398`) is `SubstateDoesNotExist` plus a store `NotFound`, and `or_not_found` hands it to the same `helpers::not_found` that ~10 sibling handlers use.
- `or_not_found` cannot over-classify on the paths it is applied to. Every store read inside `locate_dependent_substates` is already `.optional()`-guarded — `get_dependent_substates_inner` returns `Ok(())` on a missing child (`crates/wallet/sdk/src/apis/substate.rs:441`) — so the only not-found reachable through those four call sites is a genuinely absent substate, not a wallet-local inconsistency dressed up as one.
- The `HandlerError::NotFound` removal is complete: no references survive at head, and `server.rs` still needs both `ApplicationErrorCode` and `json!` for the remaining arms.
- The tests type-check against the real API rather than an assumed one — `JsonRpcError::error_reason()` returns an owned `JsonRpcErrorReason` (axum-jrpc 0.8.0 `src/error.rs:113`), so `code` binds as `i32` and the `as i32` compare is well-typed. `JsonRpcError`'s `Display` is `"{reason}: {message}"`, so the `contains("missing")` assertion holds.
- No in-repo consumer matches on the literal `"Not found"` message the three transaction getters used to return, and the web UI already keys off `code === 404` (`web_ui/src/routes/AssetVault/Components/MyAssets.tsx:58`), so widening 404 does not break it.

## Non-blocking

**404 is now overloaded across two different meanings for Rust clients.** `WalletDaemonClientError::is_not_found_error` treats *any* 404 as not-found (`clients/wallet_daemon_client/src/error.rs:64`), which is what makes `.optional()` work on the getters. After this change a `transactions.submit` that fails because an input is missing also returns 404, so a caller doing `client.submit(..).optional()?` gets `Ok(None)` — a failed submit silently indistinguishable from an absent thing. Nothing in-repo does that today (the only `.optional()` calls on a client are the VN and indexer clients), so it is latent rather than broken. Worth an upgrade note whenever 0.41 changelog entries are written, since out-of-repo callers can't see this from the diff.

**The three transaction getters pick up a warn log they did not have.** `HandlerError::NotFound` was built directly in `resolve_handler_error` with no logging; `not_found` produces an `anyhow::Error` that goes through `resolve_any_error`, which `warn!`s unconditionally (`applications/tari_walletd/src/server.rs:340`). A client polling `transactions.wait_result` for an id the wallet does not know now emits a warning per poll. Consistent with the other `not_found` sites, so this is only a heads-up, not a request to change it.

**Message gains a double prefix.** Same path: `resolve_any_error` re-serialises with `error.to_string()`, and `JsonRpcError`'s `Display` prepends the reason, so callers of `transactions.get` now see `ApplicationError(404): Not found: Transaction <id> not found` where they saw `Not found`. Pre-existing for every other `not_found` site — noting it only because these three handlers are newly joining it.

## Nit

`a_missing_resource_is_answered_as_not_found` (`handlers/helpers.rs`) — "resource" reads as `ResourceAddress` in this codebase, and `fetch_resource` is one of the paths this PR deliberately does *not* convert (`accounts.rs:1600`, `accounts.rs:1770`, `confidential.rs:121`). `a_not_found_error_keeps_its_message` would say what the test actually pins.

PR #2511, opened 2026-09-02
fix(vn): answer sync_state only for stored shards, at the epoch of sending

sdbondi:
## Description

A `sync_state` responder read whatever cursors it was given straight out of its local store, and stamped every completion marker with the epoch captured when the stream opened. Both are wrong in ways a caller cannot detect.

**A node that has left a committee answers with silence.** An unbounded (`until_epoch: None`, sync-to-tip) request for shards this node no longer serves streams no transitions — because none arrive for those shards any more — and then closes each shard off with a completion marker. The caller cannot tell "there is nothing more" apart from "I no longer serve this". The indexer's substate cache takes a completion marker as evidence that a shard is level with its committee (`ShardWatermarks::confirm`), so that silence reads as freshness and the cache keeps serving values it believes are current.

`get_substate` already guards against this with `includes_substate_id`; `sync_state` had no equivalent.

**A long-lived stream reports a stale epoch.** `StateSyncTask` took `consensus.current_epoch()` once at request time and repeated it in every `SyncComplete`. The indexer persists that epoch via `add_state_sync_progress(shard, synced_to, msg_epoch)`. `check_consensus_state()` was likewise evaluated once, so a stream that opened while consensus was running could keep asserting "you are level with me" after the node dropped into sync.

## Changes

* `ShardCursor::ensure_all_stored` rejects an unbounded request unless this node's committee for the current epoch covers every requested shard, with an error naming the shard group — the same shape as `get_substate`'s rejection. Every committee holds the global shard, so a cursor for it is always in range (`contains_or_global`).
* `StateSyncTask::authorise_tip` establishes, at send time, both the epoch a completion marker names and this node's standing to make the claim — as one fact. The epoch a marker names is the epoch whose committee was checked to store that shard, and membership is re-established whenever the epoch moves on beneath a stream. Without that, a from-scratch stream crossing a boundary would keep asserting the caller is level with a committee this node has left, stamped with the new epoch, which is the very defect being fixed.
* The standing check applies only to an unbounded stream. A bounded answer is history verified against a quorum-signed checkpoint, and the completion marker is the only path on which `rpc_state_sync` verifies the shard root and returns `Ok` (`state_sync.rs:250`) — replacing it with an error would make a validator discard a round whose contents were entirely valid.
* `ConsensusHandle::can_serve_committed_state()` carries the Running|Idle rule once, shared by `check_consensus_state` and the task.

Membership resolves at the epoch **consensus** is in, not the epoch reached by scanning the base layer. The two differ for a few seconds around a boundary; the consensus epoch is what governs which transitions this node receives, and it is the epoch the marker names, so anchoring both to it keeps the claim and its warrant the same fact.

**Bounded requests are deliberately unaffected.** The validator node's own epoch-change state sync asks *previous*-epoch committees for shards those nodes may no longer serve (`get_committees_overlapping_shard_group(prev_epoch, ..)`, `until_epoch: Some(checkpoint.epoch())`) and verifies what comes back against a quorum-signed checkpoint. Any node still holding that history may serve it; only the "this is my tip" claim requires current membership. The sole unbounded caller in the tree is the indexer.

## Indexer-side isolation

A rejection is expected traffic whenever the two nodes' views of committee membership differ, which is exactly what happens around an epoch boundary — the indexer resolves committees from its own base-layer scan and the responder from its. Previously any stream error propagated out of `sync_shard_group_state` → `sync_state` → `start()`, where `spawn` logs, sleeps `work_interval` and restarts the worker: one non-member responder in one shard group cost every other group its state sync *and* its checkpoint sync.

A peer-attributable failure (`NetworkStateSyncError::is_peer_fault`) now skips that shard group, as session-creation and probe failures already do, leaving the global shard unclaimed so the next committee streams it. Local failures — storage, epoch manager, invariant — still abort the round.

## Motivation and Context

First of three stacked changes taking indexer cache invalidation from one sync round (~60s) down to about one block, by holding the `sync_state` stream open instead of polling it. Follow mode makes both of these defects load-bearing: a held-open stream spans epoch boundaries, and "peer went quiet" must read as *unknown*, never as *caught up*.

Builds on #2498 (cache invalidation on sync transitions) and #2499 (RPC keepalive frames).

## How Has This Been Tested?

`cargo test -p tari_validator_node --lib state_sync_task` — 9 tests pass, two of them new: the global shard is in range for every committee, and a cursor for shard 17 is rejected by a node in `ShardGroup(9-16)`.

`cargo test -p tari_indexer --lib network_state_sync::error` — 2 new tests pinning that a storage or invariant failure is never attributed to the peer, so it can never be skipped past.

`cargo lints clippy -p tari_validator_node -p tari_indexer --all-targets` is clean.

Point an indexer at a committee, then move a validator out of that shard group at an epoch boundary. Before: the indexer's next round completes quietly and the cache keeps serving. After: the responder rejects the request naming its shard group, the indexer's watermark for those shards ages out past `substate_cache_max_serve_lag`, and reads fall back to the committee.

## Breaking Changes

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


sdbondi (commented):
Reviewed at `3f4ed08`. Verified against the tree: `contains_or_global` does admit `Shard(0)` for every group, `ensure_all_stored` mirrors `get_substate`'s rejection shape *and* its epoch source (`epoch_manager.current_epoch()` → `get_local_committee_info`), the `check_consensus_state` refactor is behaviour-preserving, and the claim that the only unbounded caller is the indexer holds — `rpc_state_sync` is the sole `until_epoch: Some(..)` caller. The indexer asks each peer for exactly `global + shard_group.shard_iter()` (`worker.rs:420-428`), so the new guard cannot reject a steady-state round. The core idea is right.

Two things block, both about behaviour the diff introduces rather than the idea.

## Blocking

**1. The send-time standing check is applied to bounded requests, which the PR's own reasoning exempts.** `state_sync_task.rs:230`

`send_complete` gates unconditionally on `can_serve_committed_state()`, but the PR argues — correctly — that a bounded answer is history checked against a quorum-signed checkpoint and needs nothing of the responder's current standing. The effect is a new mid-stream failure for validator-node epoch-change sync: a responder that was `Running` at request time (already enforced by `check_consensus_state` at `rpc_impl.rs:125`) and drops into `CheckSync`/`Syncing` during a long historical stream now sends `Consensus is not running on this node` in place of the completion marker. That marker is the *only* point at which `rpc_state_sync` verifies the shard root against the checkpoint and returns `Ok` (`crates/rpc_state_sync/src/state_sync.rs:213-253`), so the syncing node discards the round and has to go find another peer — for a stream whose contents were entirely valid.

Gate it: `if self.end_epoch.is_none() && !self.consensus.can_serve_committed_state()`.

**2. A rejection now takes down the whole indexer round, and the worker with it.** `applications/tari_indexer/src/network_state_sync/worker.rs:426`

The description frames this as "a quiet round becomes an aborted one", which understates it. The error propagates `stream_shard_state` → `sync_shard_group_state` → `sync_state` → `start_sync_round` → out of `start()`, and `spawn` catches it at `worker.rs:121-143`: log, `sleep(work_interval)`, restart from `wait_for_initial_scanning_to_complete`. So one non-member responder in one shard group costs every *other* shard group its state sync **and** its checkpoint sync, plus a full worker restart cycle.

What makes that matter here is that the rejection is now expected traffic. The indexer's committee view comes from its own base-layer scan (`initialize_sync_plan` → `get_network_description`) and the responder's from its; the two drift around an epoch boundary, which is exactly when membership changes. `get_local_committee_info` also hard-errors with `ValidatorNodeNotRegistered` for a peer that deregistered, which lands the same way.

Session-creation and probe failures already `continue` at `worker.rs:396-418`. A stream error deserves the same three lines, and it belongs with the change that creates the trigger rather than with follow mode — otherwise the intermediate state is worse than either end.

## Non-blocking

**3. Membership is established once, at request time; the epoch is stamped at send time.** `state_sync_task.rs:246` vs `rpc_impl.rs:490-501`

These two can disagree, and the disagreement reintroduces the defect being fixed. A from-scratch 64-shard stream runs for minutes and can cross an epoch boundary: membership was checked for epoch N, the node leaves the committee at N+1, and every remaining marker asserts "you are level with me" stamped `N+1`. The indexer persists that (`add_state_sync_progress`, `worker.rs:594`) and calls `shard_watermarks.confirm`, which refreshes `confirmed_at` and so re-arms the cache's freshness window (`shard_watermarks.rs:44-53`). Same silence-reads-as-freshness failure, displaced to the boundary — and under follow mode the boundary is the normal case, not the tail case.

Either re-check membership in `send_complete` alongside the standing check, or hold the epoch that membership was verified at and stamp *that*, so the marker's epoch and its warrant are the same fact. The second is cheaper and makes the doc comment at `state_sync_task.rs:219-221` literally true — right now it promises "both the epoch and that standing are established at send time" and only the standing half is.

Related: the guard reads `epoch_manager.current_epoch()` while the marker reads `consensus.current_epoch()`. Matching `get_substate` is the right call for the guard, but note `get_checkpoints` (`rpc_impl.rs:428-437`) treats a divergence between those two as an error condition worth rejecting on — which is what makes the window above observable rather than theoretical.

## Nits

- `"Consensus is not running on this node"` is now a literal in two places (`rpc_impl.rs:129`, `state_sync_task.rs:231`). Since `can_serve_committed_state` centralises the rule, the message could sit next to it.
- The test fixture `committee_info(8, 15)` passes `P64` with `num_committees: 8`. Shards number from 1, so 8 committees over 64 preshards give groups `1-8`, `9-16`, …; `ShardGroup(8-15)` straddles two of them. Only `shard_group()` is read so the tests are valid, but a realistic group (`9-16`, rejecting shard 17) would not invite the reader to re-derive whether the fixture is meaningful.

sdbondi (commented):
Re-reviewed at `4d61a6e`.

**Both blocking items are resolved.** `authorise_tip` now returns early for a bounded stream (`tip_authorised_at: None`), so a responder dropping into `Syncing` mid-stream no longer costs a validator node its checkpoint verification — verified against `rpc_state_sync/src/state_sync.rs:213-253`, which only reaches its `Ok` through the completion marker. And `is_peer_fault` + `continue` at `worker.rs:429-441` confines a rejection to its own shard group; leaving `has_synced_global_shard` unset on that path is right, and the classification splits correctly — `EpochManagerError`/`StorageError`/`InvariantError` still abort, which is what you want for broken local state.

Re-anchoring membership at the epoch consensus is actually in, rather than the base-layer scan epoch, is the better call and makes the marker's epoch and its warrant the same fact. One gap remains in how it is re-established.

## Blocking

**`authorise_tip` re-checks one shard but marks the whole stream authorised.** `state_sync_task.rs:291-299`

On an epoch change it verifies only `shard` against the new committee and then writes `tip_authorised_at = Some(epoch)`. Every later shard in the stream sees `epoch == authorised_at` and skips the check, so it is covered by an authorisation that was never taken for it.

Concretely, with cursors `[0, 9..16]` accepted at `E0` in `ShardGroup(9-16)`, and the committee count changing at `E1` so the node lands in `ShardGroup(9-12)`:

- shard 9's marker: `E1 != E0`, refetch, `ensure_shard_is_stored(9, 9-12)` passes, `tip_authorised_at = E1`
- shards 13-16: `E1 == E1`, no check, markers claim tip for shards this node no longer stores

That is the original defect, reached through the new re-authorisation path. It needs a group that shrinks or reshuffles rather than moves wholesale, which is exactly what a change in `num_committees` produces.

The cause is that the cached value records *that* an epoch was authorised rather than *what* it authorised. Cache the committee info instead and check every shard against it:

```rust
// field: tip_authority: Option<(Epoch, CommitteeInfo)>
let Some((authorised_at, committee_info)) = self.tip_authority.as_ref() else {
    return Ok(epoch);
};
if !self.consensus.can_serve_committed_state() { ... }
if epoch != *authorised_at {
    let info = self.epoch_manager.get_local_committee_info(epoch).await...;
    self.tip_authority = Some((epoch, info));
}
ensure_shard_is_stored(shard, &self.tip_authority.as_ref().unwrap().1)?;
```

`rpc_impl.rs:490-506` already has the `CommitteeInfo` in hand, so it can pass `Some((epoch, local_committee_info))` down and the per-shard check costs nothing extra in the common case.

## Non-blocking

**`Epoch(0)` reaches `get_local_committee_info` on a node that has not entered a view.** `state_sync_task.rs:282`, `rpc_impl.rs:491`

`ConsensusCurrentState::Idle` is `#[default]` and `can_serve_committed_state()` admits it, so `check_consensus_state` lets a not-yet-started node through; `CurrentView`'s epoch is an `AtomicU64` that defaults to 0. An unbounded `sync_state` against that node calls `get_local_committee_info(Epoch(0))`, which fails `ValidatorNodeNotRegistered` and comes back as `log_internal_error` — the right outcome (rejection) reported as an internal error, and logged as one on every such request.

Rejecting on `epoch.is_zero()` with a plain status would say what is actually true. This predates the PR by way of `check_consensus_state`, but the switch to the consensus epoch is what routes it into `get_local_committee_info`.

**`ensure_all_stored` is now only reachable from `rpc_impl`,** with the mid-stream path calling `ensure_shard_is_stored` directly. If the suggested change above lands, the two tests at `state_sync_task.rs:430-443` cover the request-time path only; a test that a shard dropped by an epoch change is rejected mid-stream would pin the part that actually regressed here.

sdbondi (commented):
Re-reviewed at `22d6a7b`.

The blocking finding from `1a170d5` is closed: `TipAuthority` carries the `CommitteeInfo`, `authorise_tip` checks every shard against it at `state_sync_task.rs:322` rather than only the one the boundary was noticed on, and `rpc_impl` hands down the info it already resolved so the steady-state path costs no extra epoch-manager call. `a_warrant_does_not_cover_a_shard_its_epoch_dropped_from_the_group` pins exactly the case that was open.

Splitting `Initialising` out of `Idle` is the right root fix rather than the `epoch.is_zero()` patch it replaces — `Idle` now means one thing (registered for no committee) instead of two. I checked the transition table: `Idle` is reachable only from `Initialising(Initialised)` and `Running(NotRegisteredForEpoch)`, both after the scan, so dropping `wait_for_initial_scanning_to_complete` from `Idle::on_enter` is safe, and `Sleeping → Initialising → Idle` is net-identical to the old `Sleeping → Idle`. Every exhaustive match on `ConsensusCurrentState` is updated, and `harness.rs:413` goes through `is_running()` so the consensus tests are unaffected by the `#[default]` move.

## Non-blocking

**The `#[default]` move silently changes `sync_blocks`, and by the PR's own argument it shouldn't.** `rpc_impl.rs:346`, `handle.rs:70-74`

`ConsensusCurrentState::default()` is now `Initialising`, and `consensus/mod.rs:106` seeds the watch channel with it, so `check_consensus_state` rejects until the worker reaches `Idle`. For unbounded `sync_state` that is exactly right. For `sync_blocks` it is a tightening that nothing here argues for: a restarting node with full history now refuses to serve committed blocks for the length of its base-layer catch-up, where it previously served (default was `Idle`). Block sync answers out of history and the caller verifies against a QC — the same reasoning that got the standing check scoped off bounded `sync_state`.

The scan resumes from a persisted height so the window is short, but the effect is to remove a healthy block-sync source from the network during every validator restart, which is when peers are most likely to be looking for one. If that is deliberate, worth saying so; if not, `sync_blocks` wants the pre-split gate.

**`ConsensusState::Initialising = 6` is a new value in a shipped wire enum.** `crates/p2p/proto/rpc.proto:355`

Appending is compatible and the sole consumer degrades correctly — `validator_status.rs:99-108` warns on an unknown discriminant and returns the peer's verified tip rather than disqualifying it. So an older indexer against a new validator loses the diagnostic snapshot (and logs a warning) for any node in `Initialising`, and nothing else. Fine, but it belongs in the description given how this repo handles protocol-visible changes.

**The title and description no longer cover the change.** Three of the four commits are now outside "answer sync_state only for stored shards": the indexer's per-shard-group error isolation, the consensus state machine split, the proto enum, and the TS binding. The `sync_blocks` change above in particular is the sort of thing that wants to be findable from the PR rather than from the state machine diff.

sdbondi (commented):
Re-reviewed at `de551d4`.

`Lagged` is a real stall and the fix is right. Only `RegisteredForEpoch` promotes out of `Idle`, and the sole producer of it in the loop was the event stream, so a dropped `EpochChanged` left the node idle until the *next* boundary — with `wait_for_initial_scanning_to_complete` now moved to `Initialising`, `Idle` blocks purely on that channel and has nothing else to fall back on. Re-reading is the correct recovery.

I checked that the recovery path and the event path agree on what "registered" means: `is_this_validator_registered_for_epoch` resolves through `get_local_committee_info` (`traits.rs:195-213`) and `registered_shard_group` through `get_our_validator_node` (`epoch_manager_service.rs:331-345`) — both are "this node has a validator record for the epoch", so `Lagged` cannot promote where the event would not have.

## Nit

**One description bullet is now stale.** "`ConsensusHandle::can_serve_committed_state()` carries the Running|Idle rule once, shared by `check_consensus_state` and the task" — the task moved to `is_running()` at `state_sync_task.rs:316`, and deliberately so: `Idle` may serve history but must not claim a tip. The bullet describes the shape before that split.

The description otherwise now covers the state-sync and indexer-isolation work well. Still unmentioned: the `Initialising`/`Idle` split, `ConsensusState::Initialising = 6` on the wire, this `Lagged` fix, and the `sync_blocks` gate change that the `#[default]` move brings with it — that last one is the one an operator would want to find here rather than in the state machine diff.

Nothing blocking outstanding on my side. The two original findings and the follow-up on `authorise_tip` are all resolved; what is left is description accuracy and the open question about whether tightening `sync_blocks` is intended.

PR #2512, opened 2026-09-02
feat(rpc): per-request deadline and keepalive, and repair the framework's test suite

sdbondi:
## Description

`deadline` and `keepalive_interval` are already per-request fields on the wire (`rpc.proto`), but the only way to set either was `RpcClientConfig`, fixed when the session is created. `RpcMultiPool` caches one client per peer and every caller shares it — state sync, the cached substate manager and the dry-run manager all reuse the same session — so buying a long deadline for one long-lived stream meant applying it to every ordinary read on that session.

This adds `RpcRequestOptions`, an override of the session defaults that travels with a single call:

```rust
let stream = client
    .sync_state_with_options(
        request,
        RpcRequestOptions::new()
            .with_deadline(Duration::from_secs(600))
            .with_keepalive_interval(Duration::from_secs(30)),
    )
    .await?;
```

Anything left unset takes the value the client was built with, so `server_streaming` is unchanged and every existing call site sends exactly what it sent before.

**The PR grew past that.** Restoring the framework's dead test suite turned up several pre-existing bugs on the paths this change touches, and they are fixed here rather than left under new tests. The full list is below.

### Call surface

The generated clients gain a `<method>_with_options` variant for each **server-streaming** method. Unary methods keep the session values: only a caller holding a response open needs a deadline of its own.

Internally the request and its options travel to the client worker together as a `ClientCall`, and the worker derives the effective `RpcClientConfig` for that request, so the wire values and the client's own read timeout come from one place.

## What else is in here

**Read timeouts actually bound the read.** The client timed each frame at `deadline + grace` and restarted that on *every* frame, keepalives included — so asking for keepalives made the client wait longer, not shorter. With keepalives arriving more often than the deadline, an idle stream never timed out at all. The client now tracks the same budget the peer does: an absolute instant renewed only by a frame carrying a response, never by a keepalive, and no earlier than the peer starts its own. Where keepalives were asked for and the peer has begun streaming, the read is additionally bounded by several intervals — several because a server raises a request below its own minimum and never reports which it chose, which `RpcClientConfig::peer_minimum_keepalive_interval` now states.

**A refused session is reported.** The handshake is one-way and stays that way — an accept reply would cost a round trip on every session opened — so a refusal arrives where the first *response* was expected. It was decoded as a response with no request id and discarded until the session died; `connect` returned `Ok` against a server that had already refused. It is now recognised both on the first frame of a session and, where the peer closed before the client's write landed, from what it had already sent.

**Discarding stale frames no longer kills sessions.** A caller that abandons a stream leaves however many frames were in flight. Past a fixed budget of twenty, the next request tore down the whole session — for every caller on a pooled client. The budget is gone; the deadline the read carries is the bound, and it does not care how many frames arrive or which request they were for.

**An unsolicited keepalive is no longer a protocol error.** That rejection was load-bearing only while every frame renewed the deadline. Now that only a response moves it, an unasked-for keepalive gains the sender nothing, while rejecting one meant an operator enabling keepalives on a server could end every session its peers held.

**The server accepts any substream.** `PeerRpcServer`, `RpcServer::serve` and `Router::serve` named `libp2p::Stream` while `ActivePeerRpcService` was already generic, so no server could be driven without a transport — the reason the end-to-end tests died with the tari_comms port. Inference covers every existing caller; `pub type Substream` stays, still used by `networking/core`.

**The test suite is back.** ~1,900 lines were sitting behind two commented-out `mod` lines. `greeting_service.rs` had its `#[tari_rpc]` attribute commented out with ~250 lines reproducing the macro's output by hand; the macro writes all of it now. `smoke.rs` is rebuilt on `tokio::io::duplex` in place of tari_comms' muxer helpers, with the buffer sized as a flow-control window rather than the frame limit, so tests about abandoned streams test something.

**1,115 lines of pool code deleted** — `client/pool.rs`, `client/tests.rs`, `test/client_pool.rs`, `test/mock.rs`, `server/mock.rs`. All `LazyPool` over tari_comms' `PeerConnection` and `RpcCommsBackend`, commented out of the tree since the port, and superseded by `RpcMultiPool`.

## Motivation and Context

Prerequisite for indexer follow mode, which holds `sync_state` open per shard group instead of re-polling every `state_scanning_interval`. Follow mode needs a deadline much longer than an ordinary read and keepalives to tell "idle" from "dead peer", neither of which may leak onto the reads sharing the same pooled session.

Independent of #2511; different crates, no shared code.

## Known limitations, for PR 3

- **A client cannot cancel an in-flight stream.** There is no protocol message for it, so the session stays blocked until the server finishes or its deadline expires — with follow mode's long deadline, minutes of a pooled session being unusable.
- **The worker serves requests serially.** `run()` awaits each request to completion before the next, and `RpcMultiPool` hands every caller a clone sharing one worker, so a held-open stream blocks every other caller on that peer — not merely lengthens them. Follow mode needs its own session or a multiplexing worker before it can share the pool.
- **`RpcClientConfig { deadline: None }` is not expressible on the wire.** It encodes as `0`, which a server reads as `Duration::ZERO` and rejects as below its minimum. Pre-existing, nothing in-repo sets it, and changing it means changing what `0` means on the wire.

## How Has This Been Tested?

`cargo test -p tari_rpc_framework -p tari_rpc_macros -p tari_validator_node_rpc` — 42 pass, up from 25.

New coverage: per-request options on the wire and their inheritance; deadline and keepalive liveness bounds, including a peer yet to send its first frame and a stalled consumer; stale-frame discarding from more than one request back; request/response and streaming errors; concurrent requests; oversize responses; ping latency; deadlines; session limits, global and per-client; unknown protocols; refused sessions; abandoned-stream recovery; and the handshake's one-way contract.

Every fix in this PR has a test confirmed to fail without it.

Confirm nothing outside a `*_with_options` call site changes what goes on the wire: `RpcRequestOptions::default()` is all-`None` and `apply_to` is `self.x.or(config.x)`. Then read `RpcResponseReader::frame_timeout` against `PeerRpcServer::process_body` and check the two budgets agree on what renews them.

## Breaking Changes

- [x] None
- [ ] Other

`RpcClientConfig` gains a public field, so a full struct-literal construction would break; nothing in-workspace does that. `RpcError::UnexpectedAckResponse` is removed; nothing outside the framework matched on it.


sdbondi (commented):
Reviewed at `bcf56fa`, reading the four changed files at head against `development`, plus the server side (`networking/rpc_framework/src/server/mod.rs`) and `RpcMultiPool`.

What is sound: `apply_to` is `self.x.or(config.x)` over a `Default`-all-`None` struct, so every existing call site puts the same bytes on the wire; threading the derived `RpcClientConfig` into `read_response` keeps the wire deadline and the per-frame read timeout from drifting; and the session-scope argument for keepalive tolerance holds — `next_request_id()` never returns 0, so a leftover keepalive can never be mistaken for a pong, and `a_session_that_asked_for_keepalives_once_tolerates_a_later_stray_frame` does fail if the latch is scoped per request. I also confirmed the description's claim about the server: `process_body` only resets `deadline_at` on a real message, so keepalives do not extend the budget.

## Blocking

**The client's read timeout ignores the keepalive interval, so asking for keepalives makes the client wait longer, not shorter.**
`RpcResponseReader::next` (`networking/rpc_framework/src/client/mod.rs:1114`) bounds each frame by `deadline + grace`, and every frame restarts it — keepalives included. Nothing derives a liveness bound from `keepalive_interval`.

Take the follow-mode shape this PR exists for: `deadline = 600s`, `keepalive = 30s`, session grace `60s`.
- Dead peer: the client notices after `660s`, exactly as it would with no keepalives. `with_keepalive_interval`'s doc ("an idle stream stays distinguishable from a dead peer") buys the client nothing.
- Idle stream: the server abandons it at `t=600` and `break`s out of `process_body` (`server/mod.rs:817-830`) — no FIN, no error frame. The client's timer was last reset by the keepalive at `t≈595`, so it sits until `t≈1255` before returning `ReplyTimeout`, roughly double the deadline it asked for.

That second case is worse than not asking for keepalives at all, and it holds the worker (see the design note below) for the whole extra window. The fix belongs on the client: when the effective config asks for keepalives, bound the inter-frame read by something like `k * keepalive_interval` while the overall deadline still bounds the stream. Fix here or say why it lands with follow mode, but the API shouldn't ship documented as solving dead-peer detection when it doesn't.

## Non-blocking

**`with_deadline`'s doc is wrong about sub-second deadlines** — `client/mod.rs:359-360` says a sub-second deadline "reads as 'no deadline' to the peer". It doesn't: `as_secs()` gives 0, and `handle_request` rejects `deadline < minimum_client_deadline` (default 1s) with a `bad_request` FIN (`server/mod.rs:612-639`). The request fails outright. `RpcClientBuilder::with_deadline` already words this correctly ("the responding peer MAY immediately reject the request"). Given `keepalive_interval` right below it floors at `.max(1)`, the more consistent fix is to floor the deadline the same way rather than only correcting the prose.

**The rationale for restricting the variant to streaming methods doesn't hold** — `generator.rs:189` says "a per-request deadline only says something about a response that arrives over time". The server applies the client's deadline to unary handlers too: `time::timeout(deadline, service_call)` at `server/mod.rs:685`, then `process_body` with the same budget. A slow unary handler is precisely a case where a caller might want a per-request deadline. Restricting the surface for now is a reasonable call; the stated reason for it isn't.

**Design note on the shared session.** The description's premise is right — `RpcMultiPool` hands every caller a clone of one `ValidatorNodeRpcClient` per peer (`crates/validator_node_rpc/src/client.rs:374-417`). But the worker serves requests strictly serially: `run()` awaits `handle_request` → `do_request_response` to completion before the next `request_rx.recv()` (`client/mod.rs:592-607`). So a follow-mode `sync_state` held open on a pooled session doesn't merely leak its deadline onto the neighbours — it blocks them entirely for the life of the stream. Per-request options are the right knob, but follow mode will also need its own session (or a multiplexing worker) before it can share the pool. Worth saying so in the PR that consumes this.

## Nits

- `client/mod.rs:672` — `do_ping_pong` passes `.tolerating_keepalives(self.tolerates_keepalives)`, but `read_ack` never reads the flag; it only matches request ids. Dead configuration. Either drop it, or have `read_ack` skip keepalive frames explicitly so the two readers agree on what a stray ACK means.
- `crates/validator_node_rpc/src/rpc_service.rs:70-88` — the compile-pin test is fine, but it pins generator output from a consumer crate. A `#[tari_rpc]` fixture in `rpc_macros` would fail closer to the change that breaks it; as written, the pin disappears the moment this service stops having a streaming method.

sdbondi (commented):
Re-reviewed `5d855f2` and `5afaf2f` against `bcf56fa`.

All five earlier points are resolved, and I verified each rather than taking the commit message for it:

- **Read timeout now bounded by both budgets.** `frame_timeout()` (`client/mod.rs:1147-1157`) takes `min(deadline_at + grace - now, 3 * keepalive_interval + grace)`, and `deadline_at` is only moved by a frame that carries a response — keepalives are consumed inside `read_response`'s loop and never reach it. Worked through the two cases I raised: with `deadline=600s, keepalive=30s, grace=60s`, an idle stream now ends at ~660s instead of ~1255s, and a silent peer at 150s instead of 660s. The two new tests exercise exactly those directions and both would fail without the change.
- **Sub-second deadline.** `.max(1)` on the wire in `do_request_response` *and* `do_ping_pong`, so the server's `minimum_client_deadline` rejection is out of reach. The ping fix is a real one on its own — the deadline check at `server/mod.rs:617` runs before the ACK branch, so a sub-second session deadline was getting pings rejected too.
- **Generator comment** is now accurate — it states the scope rather than justifying it with a claim about unary deadlines that the server contradicts.
- **Dead `.tolerating_keepalives()`** on the ping reader removed.
- **Compile pin** moved to `rpc_macros/tests/macro.rs` with its own fixture, and it now also covers the `is_unit` branch via `unit_streaming_with_options(options)` — better than what I asked for. `prost` and `tari_rpc_framework` as dev-deps introduce no cycle (`rpc_framework` does not depend on `rpc_macros`), and `rpc_macros` isn't in `publish_crates.py::CRATES`, so the path-only dev-dep has no publishing consequence.

## One thing to fix

**`deadline_at` is renewed before the frame is handed to the consumer, so a slow consumer eats the next frame's budget.** `client/mod.rs:891` sets `deadline_at = now + deadline` immediately on receipt, then `response_tx.send(Ok(resp)).await` at `:901` can block — the channel is 5 deep.

The peer's budget does not run during that time: `process_body` resets `deadline_at` *after* `framed.send(msg).await` returns (`server/mod.rs:796-797`), and the send is outside the `time::timeout`. So a consumer that stalls `S` seconds leaves the client giving up `S - grace` earlier than the peer does. With `S > deadline_grace_period` and an idle peer afterwards — precisely the follow-mode shape, a batch write followed by a quiet period — the client aborts a healthy stream before its deadline.

Moving the assignment below the `response_tx.send` in the `Ok(Ok(resp))` arm lines the two clocks back up. As a bonus it stops a stale frame on the `ResponseIdDidNotMatchRequest` arm from renewing the budget, which it currently does.

## Nit

`MISSED_KEEPALIVES_BEFORE_TIMEOUT` multiplies the interval the client *asked for*, but `keepalive_timeout`'s doc justifies the multiplier by the peer serving a longer one. Those don't connect: a client asking for 1s against the default `minimum_keepalive_interval` of 5s tolerates 3s of silence from a peer that legitimately emits every 5s. The default 60s grace hides it, but a caller who tightens `deadline_grace_period` alongside a short interval gets a stream that cannot survive. Either floor the tolerance at `3 * minimum_keepalive_interval`-ish, or have the server echo the interval it chose so the client can stop guessing — worth a `TODO` at least.

CI: `clippy`, `test (1-3/3)` and `integration test` still pending; I'll wait on those before approving.

sdbondi (commented):
Reviewed `998a90f8` on its own. Both points from the last round are fixed, and the second one is fixed better than I suggested.

**Deadline now starts at the hand-off.** The assignment moved inside the `Ok(Ok(resp))` arm below `response_tx.send(...).await` (`client/mod.rs:909-913`). I walked every arm: the `is_closed` path breaks before it, `Ok(Err(_))` and the id-mismatch `continue` deliberately don't renew, and the first budget is still set before `send_request`. `a_slow_consumer_does_not_spend_the_next_message_s_deadline` is a real regression test, not a tautology — the 6th `send` blocks until t=1.0s, so pre-fix the budget from receipt expired at t=1.2s and the FIN at t=1.5s arrived as a `Timeout`; post-fix the budget restarts at 1.0s and runs to 2.2s. It fails without the change.

**Keepalive tolerance covers the clamp.** `max(asked, minimum_keepalive_interval) * 3 + grace` closes the case I raised, and hoisting `DEFAULT_MINIMUM_KEEPALIVE_INTERVAL` into `lib.rs` so the client's assumption and `RpcServerBuilder`'s policy read from one constant is better than the `TODO` I asked for — the two can't drift now. Checked the unit test's arithmetic: 17s, 92s, `None`. `a_peer_that_stops_sending_the_keepalives_it_promised_is_given_up_on` needed the 100ms minimum to stay under its 3s bound, which is correct rather than a fudge, and `keepalives_do_not_extend_the_deadline...` still ends on the deadline path (1.2s) since its tolerance rose to 15.2s. No in-repo struct literal builds `RpcClientConfig`, so the new public field breaks nothing.

## Worth adjusting

**The two clocks only coincide under backpressure.** The peer restarts its budget when `framed.send` *returns*, which for a small message is immediately — it doesn't wait for the client to read. The equivalence the comment claims holds only once socket buffers are full. The test's own server shows it: it writes all six messages at t≈0 into a 6 MiB duplex with no backpressure at all, so a real peer in that scenario starts counting at t≈0 and abandons at t≈1.0s while this client now waits to 2.2s.

That is the right trade — a late timeout beats aborting a live stream, and the excess is bounded by how long the consumer stalled, which the caller owns. But the worst case is now `stall + deadline + grace`, and a consumer that stalls for minutes leaves the worker (and every other caller queued behind it on the shared session) waiting that much past the peer's abandonment. Worth saying "no earlier than the peer" in the comment at `:909` rather than implying the budgets line up, so the next reader doesn't take the symmetry literally.

## Nits

- `RpcClientConfig::minimum_keepalive_interval` has no `RpcClientBuilder` setter, unlike every other field on that struct. Its own doc says to "set it to the peer's own minimum where that is known to differ" — through the builder there's no way to. One `with_minimum_keepalive_interval` alongside `with_keepalive_interval` would do it.
- The name matches `RpcServerBuilder::minimum_keepalive_interval` but means the opposite side of the exchange — what this client *assumes a peer serves*, versus what that server *will serve*. The doc distinguishes them; the field names don't.
- The tolerance now bounds every request on a session that configures `keepalive_interval`, unary included — and the server emits no keepalives while a unary handler runs, since `service_call` completes before `process_body` starts them (`server/mod.rs:685`, `:770`). A session-level 5s interval therefore caps a slow unary handler at ~75s rather than the 180s the deadline implies. Nothing in-repo sets session keepalives, so it's latent, but the field is public and `with_keepalive_interval`'s doc doesn't mention it.

CI restarted on this head; still waiting on clippy, `test (1-3/3)` and integration before approving. `dead_code` is denied via `lints.toml` and clippy runs `--all-targets`, so the new `rpc_macros` fixture generating an unused `TestRpcServer` is the one thing I can't check without a build — that job will say.

sdbondi (commented):
Reviewed `745c53ec`. This closes the last three open points, and the first of them turned out to be the substantive one.

**Keepalive tolerance now waits for the peer to actually be streaming.** `ReadBounds.keepalives_due` starts false, so the first frame is bounded only by the deadline — which is right, because the server runs `service_call` to completion before `process_body` starts emitting anything (`server/mod.rs:685`, `:770`). I checked both places the flag flips and both are needed rather than redundant: `next()` sets it on the reader so a `read_response` that skips keepalives tightens from the second frame on, and `do_request_response:922` carries it across the per-frame reader rebuilds. `a_peer_yet_to_send_a_first_frame_is_bounded_only_by_its_deadline` fails without the change — its 500ms tolerance would fire well before the peer's 1s first frame. The two older keepalive tests still land where they should: `a_peer_that_stops_sending...` now primes with two frames before going silent, so it times out at ~700ms rather than on the first-frame path, and `keepalives_do_not_extend_the_deadline...` still ends on the deadline (1.2s vs a 15.2s tolerance).

**Builder setter and naming.** `with_peer_minimum_keepalive_interval` makes the field reachable through the normal API, and `peer_minimum_keepalive_interval` with a doc pointing at `RpcServerBuilder::with_minimum_keepalive_interval` settles the which-side-am-I question. The comment at `:934` now states the asymmetry ("no earlier than the peer") and names the cost, which is what I was after.

## Two observations, neither blocking

**A first frame is now unbounded when `deadline` is `None`.** `frame_timeout`'s fallback arm is `deadline.or(keepalive)`, and with `keepalives_due` false the keepalive side is `None` too — so a config with `deadline: None` and keepalives set has no bound at all on the first read, where before `5d855f2` it had `3 × interval + grace`. Only reachable through a struct literal (`with_deadline` always sets `Some`), and a conforming peer rejects a zero deadline outright at `server/mod.rs:617`, so it takes a hostile peer to hang on. Still, an unbounded read in a worker that serves the session serially is worth not having: a floor on that arm would cost nothing.

**The flag flips on any frame, including one the id check then rejects.** Both assignment sites run before `check_response` / `convert_to_result`, so a delayed frame from an abandoned earlier request — the case the `MAX_ALLOWED_IGNORED` leniency exists for — tells this request that its peer is streaming, tightening the bound while its handler may still be silent. Narrow (needs session keepalives, a leftover frame, and a slow handler), and keepalives are deliberately not id-policed so the flag can only ever be a heuristic. Worth knowing it is one.

CI is running on this head — clippy, `test (1-3/3)`, integration and the two build jobs still pending. Once they're green I have nothing left blocking.

sdbondi (commented):
Five new commits — this has grown well past "carry a deadline and keepalive interval per request". Reviewed each on its own. The direction is right and the restored suite is worth having, but the PR body still describes only the original change: a reader arriving at this diff has no warning that it deletes 1,115 lines of pool code, makes the server generic over its substream, changes what an unsolicited keepalive means, and removes the discarded-frame budget. Please fold those into the description before merge.

**`4e39e5303`** — clean. `TSubstream` threads through `PeerRpcServer`, `RpcServer::serve` and `Router::serve` with inference covering existing callers; `pub type Substream` stays used by `networking/core/src/worker.rs`, so removing the import without removing the alias is right.

**`4862cc188`** — the deleted modules were all `PeerConnection`/`RpcCommsBackend`-based and commented out of the tree; `RpcMultiPool` owns pooling now. Nothing to argue with.

**`fcc688fa4`** — `extern crate self as` is correctly `#[cfg(test)]`-gated. The dev-dep cycle (`rpc_framework` dev-depends on `rpc_macros`, which dev-depends back) is the case Cargo explicitly permits, since neither is a normal dependency. `refusal_reason` handling the refusal at either arrival point, and `wait_for_session` polling rather than sleeping, are what keep this suite from being flaky — good calls. `a_session_recovers_from_an_abandoned_stream` genuinely pins the budget removal: 49 leftover frames against a limit of 20.

**`978ada859`** — I checked the safety argument for dropping `MAX_ALLOWED_IGNORED` rather than taking it: a discarded frame does not move `bounds.deadline_at`, and `frame_timeout` recomputes from that absolute instant, so the skip window really is bounded by the deadline no matter how many arrive. Same for tolerating an unsolicited keepalive: `keepalive_timeout()` is `None` when this client never asked, so setting `keepalives_due` on one changes nothing. Both hold.

## Worth fixing

**The budget removal only covers one request back.** The guard is still `actual.wrapping_add(1) == request_id`, so a frame from two requests ago falls to `Err(err) => return Err(err)`, which propagates through `handle_request` to `run()` and terminates the worker — the whole session, for every caller on a pooled client. That is the failure this commit removes for the N-1 case, left in place for N-2. It is reachable exactly where the commit says these frames come from: abandon a stream, abandon the next one before the first's frames have drained, and the third request kills the session. The argument that makes the fix safe — the deadline bounds the skipping, so the count does not matter — applies just as well to the id, so discarding any frame that is not this request's would make the rule uniform instead of leaving a narrower version of the same bug.

**`read_rejection` can wait where the comment says it does not.** `client/mod.rs`'s sibling path is careful; this one goes through `recv_next_frame`, which applies the handshake timeout (5s in the suite, 90s by default). "reads what is there rather than waiting on the peer" holds only because a failed write usually means a closed substream, where `framed.next()` returns `None` at once. A write that fails for any other reason against a live, silent peer now blocks `connect` for the whole handshake timeout instead of returning the IO error. A zero or very short timeout for this one read would match what the comment promises.

**The two rejection decoders disagree.** `decode_session_rejection` requires `Some(SessionResult::Rejected(true))`; `read_rejection` uses `RpcSessionReply::result()`, which maps `session_result: None` to `Err(Unknown(...))` (`message.rs:316-319`). So any buffered frame that happens to decode as a reply with no session result — an `RpcResponse` carrying only flags and payload does — is reported as a refusal with an unrecognised reason instead of the real IO error. The strict form is the correct one; worth using it in both.

## Nit

`self.framed.send(payload.into()).await.and(self.framed.flush().await)` evaluates the flush even when the send failed, since `Result::and` is eager. Harmless — the result is discarded and the error kept is the send's — but it awaits a flush on a sink that just failed. `if let Err(err) = send { ... } else { flush }` says what is meant.

CI is running again on this head; clippy, `test (1-3/3)`, integration and ledger still pending.

sdbondi (commented):
Reviewed `804d7882`. All three items from the last round are closed, each verified rather than taken from the commit message.

**Stale frames from any request.** The `actual + 1 == request_id` guard is gone, so the N-2 case no longer falls through to terminating the worker. The safety argument still holds for the wider rule: a discarded frame doesn't move `bounds.deadline_at`, and `frame_timeout` recomputes from that absolute instant, so skipping stays bounded however many arrive and whichever request they were for. Discarding an id from *ahead* of the current request is new behaviour too, and it's the better failure: the worker is serial so that can't happen legitimately, and a peer that answers with wrong ids now costs one request its deadline instead of costing every caller the session. `a_frame_from_more_than_one_request_ago_is_discarded` fails under the old guard — the stray frame is id 1 against request 3.

**One refusal decoder.** `decode_session_rejection` is `pub(crate)` in `handshake.rs` with the strict `Rejected(true)` check, and both call sites go through it. `recv_next_frame` is still used by the server handshake, so nothing is left dangling. `a_frame_that_is_not_a_refusal_is_not_read_as_one` is deterministic despite looking racy — the server half is written, closed and dropped before the client is built, all in one task, so the client's write always fails, and tokio's duplex still hands back the buffered frame after the close.

**Short bound on the refusal read.** `REJECTION_READ_TIMEOUT` replaces the handshake timeout, and the send/flush is now a `match` so a failed send no longer flushes eagerly.

## Three small things

**The `decode_session_rejection` doc argues against the alternative.** "Deliberately strict: a reply carrying no session result is not a refusal, and treating one as such would report an arbitrary frame ... as a refusal with an unrecognised reason" is the shape `CLAUDE.md` rules out — a present-tense invariant, not "without this, X would happen", and not a reply to review feedback. The behaviour it describes isn't in the file for a reader to contrast against. Something like "A refusal is `Rejected(true)`; a reply with no session result is not one" says the same thing without the counterfactual. Same shape in the new test's comment, though that one reads more as scenario setup.

**100ms is tight for a loaded CI runner.** The frame is already buffered so the read resolves immediately in the normal case, but a scheduling stall past 100ms on a busy runner turns `a_rejection_is_reported_when_the_peer_has_already_closed` into a flake. The path is only reached after a write has already failed, so a second here costs nothing anyone will notice and still avoids the ninety.

**`do_request_response:952` is now doubly unreachable.** `convert_to_result` never returns `ResponseIdDidNotMatchRequest` — it only produces `Ok(Err(..))` — so that arm was already dead, and its comment ("Ignore the response, this can happen when there is excessive latency") now describes handling that genuinely lives at `:1011`. A reader looking for where stale frames are dropped finds the dead copy first. Pre-existing, but this commit is what makes it misleading; worth deleting while you're here.

Nothing blocking. CI is running on this head — clippy, `test (1-3/3)`, integration and the build jobs pending; I'll approve once they're green.

PR #2513, opened 2026-09-02
fix(ootle-rs): resume a truncated UTXO pass from the last delivered version

sdbondi:
## Description

A wallet scanning stealth UTXOs could silently skip outputs — no error, no gap in the frame sequence, and nothing to notice after the fact. Two independent causes, both in the `/utxos/stream` resume protocol.

### 1. A read could stop in the middle of a state version

`state_version` is stamped per *sync batch* (`network_state_sync/worker.rs:638`) and applied to every update in it, so every UTXO a block touches on a shard shares one version. It is not a unique key, and a client's cursor can only say "everything after version V" — it cannot address a row within a version.

`utxos_get_updates` ordered by `state_version` and cut at `LIMIT n` wherever that fell. When the cut landed inside a version group, the reported `max_state_version` was that group's version `V`, and the next request's `> V` filter dropped the rest of the group. One busy block on a shard was enough.

### 2. `EndOfShard` reported a watermark above what was delivered

`EndOfShard.max_state_version` came from `utxos_get_max_state_version` — `MAX(state_version)` over the whole shard, ignoring `from_epoch`, `unspent_only` and the `per_shard_limit` truncation. `ootle-rs`'s `ShardCursor::observe` advanced from it unconditionally, exactly as the frame's doc instructed, so a truncated pass jumped its cursor past its own undelivered remainder.

## Fix

**Reads end on a version boundary** (`storage_sqlite/reader.rs`). The query takes one row beyond the limit to find where the boundary falls and holds back a trailing partial group. Note the `limit + 1` read is only a *probe*: when the whole probe sits at one version that group may be far larger, so a second unbounded read fetches it in full. A version group wider than the limit has no complete predecessor to stop at, so it is served whole and the limit gives way — returning nothing would leave the cursor unable to advance past the shard.

**Truncation is the server's answer, not each client's inference.** `UtxoStateUpdateSet` and protobuf `StartOfShard` carry `has_more` (new field tag 4, additive and wire-compatible), and `EndOfShard.max_state_version` is the resume point the indexer chose: the last delivered version while the shard has more to give, the shard's high watermark once drained. The high watermark still does its job on a drained pass — it carries the cursor past versions the request's own filters excluded.

Both consumers drop their `num_updates >= per_shard_limit` heuristic, which group-trimming makes actively wrong: a short batch would read as a drained shard. `ootle-rs` resumes from `EndOfShard`; `sdk_services`' scanner resumes from `StartOfShard.max_state_version` (now always a complete version) and loops on the authoritative `has_more`.

`per_shard_limit == 0` is now rejected: updates are served in whole state versions, so a zero limit would return one version rather than the empty response it asks for.

## Also in this PR

- **`X-Accel-Buffering: no`** on the three streaming endpoints (`/events`, `/transactions/events/stream`, `/utxos/stream`) via a shared helper. nginx and the proxies following its conventions buffer upstream responses by default, holding a stream's frames until the buffer fills or the response ends — on a long-lived SSE stream, indefinitely.
- **`/utxos/stream` declares its negotiated content type** instead of always answering `application/octet-stream`: `application/x-protobuf` for the length-delimited protobuf framing, `application/x-ndjson` for the newline-delimited JSON one. `Accept` gains `application/x-ndjson` alongside the existing `application/json`. Neither client validated the old header, so this is safe.

Both came out of checking whether the indexer can be horizontally scaled behind a load balancer. Worth recording from that review: the UTXO stream cursor is a network-global `(shard, state_version)` pair and each indexer's `utxos` table is contiguous per shard, so UTXO scanning is safe to round-robin across instances. The bugs above are orthogonal to that and bite a single indexer just as hard.

## Versions

Per `scripts/crate_versioning.py impact tari_indexer_client --breaking` — adding a public field is a minor bump:

| crate | version |
| --- | --- |
| `tari_indexer_client` | 0.41.0 → 0.42.0 |
| `ootle-rs` | 0.22.0 → 0.23.0 |
| `tari_ootle_wallet_sdk` | 0.42.0 → 0.43.0 |
| `tari_ootle_wallet_storage_sqlite` | 0.42.0 → 0.43.0 |
| `tari_ootle_walletd_client` | 0.42.0 → 0.43.0 |

Plus the workspace pins, and `bindings/src/types/UtxoStateUpdateSet.ts` hand-updated for the new field.

## Test plan

- [x] Four store tests in `storage_sqlite/store_factory.rs`: mid-group cut, lossless paged resume (drains a shard and asserts every row is seen), oversized group served whole, drained read reports no more
- [x] `ootle-rs` cursor tests in `utxo_watcher.rs`: monotonic advance, genesis covers every shard at zero
- [x] 36 `tari_indexer` storage tests and 34 `ootle-rs` lib tests pass
- [x] `cargo clippy --all-targets` clean on the touched crates
- [x] `cargo +nightly-2025-12-05 fmt --all`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7901,7 +7901,7 @@ dependencies = [
 
 [[package]]
 name = "ootle-rs"
-version = "0.22.1"
+version = "0.23.0"
 dependencies = [
  "async-stream",
  "async-trait",
@@ -11756,7 +11756,7 @@ dependencies = [
 
 [[package]]
 name = "tari_indexer_client"
-version = "0.41.0"
+version = "0.42.0"
 dependencies = [
  "anyhow",
  "bounded-vec",
@@ -12238,7 +12238,7 @@ dependencies = [
 
 [[package]]
 name = "tari_ootle_wallet_sdk"
-version = "0.42.0"
+version = "0.43.0"
 dependencies = [
  "anyhow",
  "futures 0.3.31",
@@ -12326,7 +12326,7 @@ dependencies = [
 
 [[package]]
 name = "tari_ootle_wallet_storage_sqlite"
-version = "0.42.0"
+version = "0.43.0"
 dependencies = [
  "diesel",
  "diesel_migrations",
@@ -12422,7 +12422,7 @@ dependencies = [
 
 [[package]]
 name = "tari_ootle_walletd_client"
-version = "0.42.0"
+version = "0.43.0"
 dependencies = [
  "hex",
  "ootle_serde",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7901,7 +7901,7 @@ dependencies = [
 
 [[package]]
 name = "ootle-rs"
-version = "0.22.0"
+version = "0.22.1"
 dependencies = [
  "async-stream",
  "async-trait",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,9 +104,9 @@ tari_ootle_p2p = { path = "crates/p2p" }
 tari_ootle_storage = { path = "crates/storage" }
 tari_ootle_storage_sqlite = { path = "crates/storage_sqlite" }
 tari_ootle_wallet_crypto = { path = "crates/wallet/crypto", version = "0.42" }
-tari_ootle_wallet_sdk = { path = "crates/wallet/sdk", version = "0.42" }
+tari_ootle_wallet_sdk = { path = "crates/wallet/sdk", version = "0.43" }
 tari_ootle_wallet_sdk_services = { path = "crates/wallet/sdk_services" }
-tari_ootle_wallet_storage_sqlite = { path = "crates/wallet/storage_sqlite", version = "0.42" }
+tari_ootle_wallet_storage_sqlite = { path = "crates/wallet/storage_sqlite", version = "0.43" }
 tari_epoch_manager = { path = "crates/epoch_manager" }
 tari_epoch_oracles = { path = "crates/epoch_oracles" }
 tari_indexer_lib = { path = "crates/indexer_lib" }
@@ -118,14 +118,14 @@ tari_transaction_manifest = { path = "crates/transaction_manifest", version = "0
 tari_validator_node_rpc = { path = "crates/validator_node_rpc" }
 sqlite_message_logger = { path = "networking/sqlite_message_logger" }
 tari_base_node_client = { path = "clients/base_node_client" }
-tari_indexer_client = { path = "clients/tari_indexer_client", default-features = false, version = "0.41" }
+tari_indexer_client = { path = "clients/tari_indexer_client", default-features = false, version = "0.42" }
 tari_networking = { path = "networking/core" }
 tari_rpc_framework = { path = "networking/rpc_framework" }
 tari_rpc_macros = { path = "networking/rpc_macros" }
 tari_swarm = { path = "networking/swarm" }
 tari_validator_node_client = { path = "clients/validator_node_client" }
 tari_validator_rollback = { path = "utilities/validator_rollback", version = "0.40" }
-tari_ootle_walletd_client = { path = "clients/wallet_daemon_client", version = "0.42" }
+tari_ootle_walletd_client = { path = "clients/wallet_daemon_client", version = "0.43" }
 transaction_generator = { path = "utilities/transaction_generator" }
 
 # Dependency crates on crates.io (update versions here as well when updating the workspace version)
--- a/applications/tari_indexer/src/rest_api/handlers/indexer_events.rs
+++ b/applications/tari_indexer/src/rest_api/handlers/indexer_events.rs
@@ -3,28 +3,27 @@
 
 use axum::{
     Extension,
-    response::{Sse, sse},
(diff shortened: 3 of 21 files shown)


sdbondi (commented):
Verified the core claim against the indexer: `utxos_get_max_state_version` is `MAX(state_version)` over the shard with no `from_epoch`/`unspent_only`/limit filters (`storage_sqlite/reader.rs:782-791`), and `EndOfShard` is emitted from it unconditionally (`streaming/utxo_stream.rs`), so the truncation skip described is real and the drained/truncated split in `resume_watermark` closes it. The `X-Accel-Buffering` change and the `ootle-rs` patch bump are sound.

### Blocking

**A truncated pass can still skip rows that share the last delivered state version** — `crates/wallet/ootle-rs/src/provider/utxo_watcher.rs:218`

`state_version` on a UTXO row is the block-level version stamped by the sync worker (`network_state_sync/worker.rs:969,1028`), so every UTXO created or spent in one block on a shard shares it. The server query is `state_version > cursor ORDER BY state_version LIMIT n` with no secondary ordering (`storage_sqlite/reader.rs:810-814`). When the limit falls inside a version group, `StartOfShard.max_state_version` is that group's version `V`, the truncated branch resumes from `V`, and the next request's `> V` filter drops the rest of the group. Same failure class the PR is fixing, just narrower — one busy block on a shard is enough to hit it.

The new doc on `StealthUtxoFrame` ("every update at or below it has been delivered", line 54) is therefore not true at a group boundary, and the `per_shard_limit` doc on the request (line 33) leans on the same assumption.

This can't be fixed client-side: `Unspent` frames don't carry a state version, so the client cannot tell where the group boundary is. It needs the server to stop splitting a version group — e.g. in `utxos_get_updates`, load `limit + 1`, and if the overflow row shares the last version, drop the trailing partial group so the reported `max_state_version` is the last *complete* version (with the degenerate single-group-larger-than-limit case surfaced as an error rather than an infinite loop). Note that dropping rows makes `num_updates < per_shard_limit`, which both this client and the sdk_services scanner (`has_more: sos.num_updates >= 1000`) read as "drained" — so the truncation signal needs to move to the server too. Cleanest is for the server to put the resume point directly in `EndOfShard.max_state_version` (it already holds both values in `PendingUpdates`), which makes both consumers correct without the `>= per_shard_limit` heuristic. Fine to do in a follow-up if you'd rather keep this PR scoped, but then the doc claims here should be softened until it lands.

### Non-blocking

- `utxo_watcher.rs:53-55` — even for a drained pass, "every update at or below it has been delivered" is not quite right: the high watermark also covers rows the request's own `unspent_only`/`from_epoch` filters excluded. "delivered or excluded by the request's filters" is the accurate statement.

### Nits

- `utxo_watcher.rs:329` — `a_truncated_pass_leaves_every_undelivered_update_above_the_cursor` asserts the same thing as `truncated_pass_resumes_from_the_last_delivered_version` (watermark == 100 implies 101/175/250 are above it). One of the two is enough.

sdbondi (commented):
Re-reviewed at `0cd8571`.

**Blocking item resolved.** `utxos_get_updates` now reads `limit + 1`, cuts back to the last complete version group via `partition_point`, and falls back to serving one oversized group whole rather than returning nothing (`storage_sqlite/reader.rs:812-885`). `has_more` rides on `StartOfShard` (proto field 4, additive) and `EndOfShard.max_state_version` is chosen server-side: last delivered version when `has_more`, high watermark otherwise. Both consumers are correct against this: `ootle-rs` resumes from `EndOfShard` with no local heuristic, and `scanner_round.rs` resumes from `StartOfShard.max_state_version` (now the last *complete* version) and loops on the authoritative `has_more`. The four store tests cover mid-group cut, lossless resume, oversized group, and drained read. Content-Type change is safe — neither client validates it on the UTXO stream.

### Non-blocking

- **PR description is stale.** It still describes `ShardBatch`/`resume_watermark` in the client, `ootle-rs` 0.22.1, and four `utxo_watcher.rs` unit tests. None of that is in the diff any more; the fix moved to the indexer and this is now a wire-protocol change with minor bumps on `tari_indexer_client`, `ootle-rs`, `wallet_sdk`, `storage_sqlite` and `walletd_client`.
- **`ootle-rs` 0.23 against a pre-`0.42` indexer reintroduces the original skip** — `utxo_watcher.rs:177`. An old server sends no `has_more` (prost default `false`) and `EndOfShard` still carries the raw high watermark, so a truncated pass jumps over the remainder exactly as before. Cheap guard that is correct against both servers: when `num_updates >= per_shard_limit`, cap the `EndOfShard` value at `StartOfShard.max_state_version`. Against the new server this is a no-op (both values already agree when truncated; an overrun group is served whole so the SOS version is still complete). Up to you whether mixed-version deployments matter here.

### Nits

- `reader.rs:857` — in the whole-version fallback `has_more` stays `true` even when that group was the last one in the shard, so the caller makes one extra empty pass. Harmless: the empty pass emits no frames and the cursor stays put.
- `reader.rs:812` — `limit == 0` now serves one whole version (overshoot of 1 always trips the fallback). Previously it returned nothing. Probably fine, but worth a `limit.max(1)` or a bad-request check at the handler so the behaviour is deliberate.

PR #2515, opened 2026-09-03
test(integration): cover an epoch change that splits the committee

sdbondi:
## Description

Every integration scenario runs on a single committee, so nothing exercises the case #2511 was written for: a validator that stops storing shards it used to store. The rejection, the indexer's per-shard-group recovery, and the re-authorisation of a tip claim at an epoch boundary all need a network whose committee membership actually changes.

**Why that was out of reach.** `calculate_num_committees` is `max(1, num_vns / committee_size)` and `integration_tests` ran at the devnet default of 7, so a second shard group needed 14 validators. This is also why the repo's own `Leader failure with multiple committees` scenario is `@ignore`d — at ten validators it stopped producing more than one committee and became a misnomer.

### A consensus constants file for LocalNet

Committee size is a consensus constant — a network's nodes have to agree on it — and the only way to change it was `TARI_DEVNET_COMMITTEE_SIZE`, read at startup. That is a per-process setting for a network-wide fact, and no use of it lets one scenario differ from the rest of a test run.

A LocalNet now reads its constants from a file named by `consensus_constants_file` on the validator node and indexer configs, defaulting to `consensus_constants.toml` in the data directory:

```toml
committee_size_per_shard_group = 3
pacemaker_block_time_secs = 5
```

* Every field is optional and applied over the network's own values, so an absent file — or an absent field — changes nothing. Point several nodes at one path and a devnet is configured from a single authored source.
* Read **once at startup**, where `ConsensusConstants::from(network)` was called before. Changing constants means restarting, which is right for something the network has to agree on.
* Read **only on LocalNet**. Every other network's constants are part of what its nodes agree on, so nothing a local file says about them is safe to act on, and a file left behind by a devnet cannot reach a node that is not on one.
* `num_preshards` is not settable. Shard identity is derived from it everywhere, so nodes that disagree on it do not form a misconfigured network, they form no network at all. `deny_unknown_fields` makes an attempt to set it — or a typo — a startup error rather than a silent no-op.
* It sits behind a feature on `tari_ootle_app_utilities`, honouring the existing intent that nothing which merely prices a transaction has to carry the consensus engine. `TARI_DEVNET_COMMITTEE_SIZE` gives way to it.

**What this does not do:** it does not stop nodes being handed *different* files. Nothing today commits to the constants — `epoch_hash` is base-layer-derived or a static function of the epoch — so a mismatch still forks quietly. Attesting the constants in something peers already exchange is the fix for that, and is worth doing independently of where the values are read from.

### The test the file enables

* The harness authors one file per scenario from a `consensus_constants` block on the network spec, hands it to every node it spawns, and takes its own constants from the same file so that assertions and nodes cannot disagree.
* A new step, `the network has {int} shard group(s) according to indexer {word}`, polls the indexer's `network/stats`. It is how a scenario states the shape of the network it built, and how it waits for a registration to take effect at an epoch boundary.
* `committee_split.feature`: a five-validator network at committee size 3, a sixth validator registered mid-scenario, an assertion that the following epoch has two shard groups, and reads that must keep working across the split.

## Motivation and Context

Follow-up to #2511, which made a `sync_state` responder refuse an unbounded request for shards it no longer stores, and made the indexer confine that refusal to the shard group it came from. That merged on unit tests and review because no integration scenario could produce a committee change: `calculate_num_committees` is `num_vns / committee_size`, so at the devnet default of 7 a second shard group needed fourteen validators. It is also why the repo's own `Leader failure with multiple committees` scenario is `@ignore`d — at ten validators it stopped producing more than one committee.

## How Has This Been Tested?

Six unit tests on the constants file: an empty file keeps every network value, a set field replaces only itself, a burn rate above the maximum is refused, a network that is not local does not read the file, a local one does, and `num_preshards` is not accepted.

`cargo check --workspace --all-targets` and clippy are clean across the touched crates.

The scenario itself passes, run locally on top of #2516:

```
CUC_DEBUG=1 cargo test -p integration_tests --test cucumber --release -- --tags "@committee_split"
```

**It requires #2516.** On `development` as it stands the scenario fails at `all validator nodes have started epoch 4`: the committee split leaves every validator looping between `Running` and `CheckSync`, because the epoch checkpoint is rolled back by the very error that asks for a sync and `check_sync` cannot see that a shard group moved. This scenario is what found that, and it is the test for it — merge #2516 first.

`cargo test -p integration_tests --test cucumber -- --tags '@committee_split'`.

To see it fail without #2511's responder change, revert `ensure_all_stored` in `sync_state`: the indexer goes on treating a non-member's silence as a completed round, and its sync progress for the shards that moved stops advancing. That comparison only holds because `I wait for the indexer to sync with the network` now takes its target as the union of state versions across every running validator — a single validator reports its own shard group only, which after the split is the half it kept rather than the half that moved.

## Breaking Changes

- [x] None
- [ ] Other - Please specify

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5396,6 +5396,7 @@ dependencies = [
  "tari_validator_rollback",
  "time",
  "tokio",
+ "toml 0.9.11+spec-1.1.0",
  "tonic 0.13.1",
  "yaml_serde",
 ]
@@ -11972,6 +11973,7 @@ dependencies = [
  "tari_ootle_transaction",
  "tari_template_lib",
  "tari_transaction_components",
+ "tempfile",
  "thiserror 2.0.18",
  "tokio",
  "toml 0.9.11+spec-1.1.0",
--- a/applications/tari_indexer/Cargo.toml
+++ b/applications/tari_indexer/Cargo.toml
@@ -32,7 +32,7 @@ tari_indexer_client = { workspace = true, default-features = false, features = [
 tari_indexer_lib = { workspace = true }
 tari_ootle_template_metadata = { workspace = true }
 tari_networking = { workspace = true }
-tari_ootle_app_utilities = { workspace = true, features = ["epoch_oracle", "p2p"] }
+tari_ootle_app_utilities = { workspace = true, features = ["consensus_constants", "epoch_oracle", "p2p"] }
 tari_ootle_common_types = { workspace = true }
 tari_ootle_p2p = { workspace = true }
 tari_ootle_storage = { workspace = true }
--- a/applications/tari_indexer/src/bootstrap.rs
+++ b/applications/tari_indexer/src/bootstrap.rs
@@ -415,6 +415,7 @@ pub async fn spawn_services(
     save_identities(config, &keypair)?;
     Ok(Services {
         network: config.network,
+        consensus_constants: consensus_constants.clone(),
         published_config: PublishedIndexerConfig::from(&config.indexer),
         keypair,
         networking,
@@ -435,6 +436,7 @@ pub async fn spawn_services(
 
 pub struct Services {
     pub network: Network,
+    pub consensus_constants: ConsensusConstants,
     pub published_config: PublishedIndexerConfig,
     pub keypair: RistrettoKeypair,
     pub networking: NetworkingHandle<TariMessagingSpec>,
--- a/applications/tari_indexer/src/config.rs
+++ b/applications/tari_indexer/src/config.rs
@@ -82,12 +82,19 @@ impl ApplicationConfig {
         self.to_data_dir().join("state.db")
     }
 
-    pub fn consensus_constants_path(&self) -> PathBuf {
-        if self.indexer.consensus_constants_file.is_absolute() {
-            return self.indexer.consensus_constants_file.clone();
+    /// The configured consensus constants file, resolved against the data directory.
+    pub fn localnet_consensus_constants_path(&self) -> Option<PathBuf> {
+        let path = self.indexer.localnet_consensus_constants_file.as_ref()?;
+        if path.is_absolute() {
+            return Some(path.clone());
         }
 
-        self.to_data_dir().join(&self.indexer.consensus_constants_file)
+        Some(self.to_data_dir().join(path))
+    }
+
+    /// Where a consensus constants file is picked up from when none is configured.
+    pub fn default_localnet_consensus_constants_path(&self) -> PathBuf {
+        self.to_data_dir().join("consensus_constants.toml")
     }
 
     pub fn global_db_path(&self) -> PathBuf {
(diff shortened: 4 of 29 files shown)


sdbondi (commented):
### Blocking

**The final assertion does not cover the shards that moved.** `I wait for the indexer INDEXER to sync with the network` (`integration_tests/tests/steps/indexer.rs:435-460`) takes `state_versions` from the first running validator, and the VN reports those for its *local* shard group only (`applications/tari_validator_node/src/json_rpc/handlers.rs:770-775`). After the split that is half the shard space — the half VN1 kept. The shards VN1 stopped storing, which are exactly the ones #2511's responder refusal and the indexer's per-shard-group recovery are about, are never compared. So with `ensure_all_stored` reverted the indexer's progress on the moved shards stalls and this step still passes, contrary to the "to see it fail" claim in the description. Fix: union `state_versions` across every running validator (or every shard group at the current epoch) before comparing, either in this step or a dedicated one used by this scenario.

### Non-blocking

- **Swarm template advertises `base_layer_confirmations` as tunable, but the daemon does not read it.** `mine_in_validator_registrations` (`applications/tari_swarm_daemon/src/process_manager/manager.rs:741`) uses `ConsensusConstants::from(self.network)` with a comment saying it must match the validator's oracle lag. Uncommenting that line in the generated file changes the validators but not the daemon's mining target, which is the overshoot that comment warns about. Either have the daemon load the file too or drop that line from the template.
- **Indexer economics endpoint bypasses the loaded constants.** `get_economics` (`applications/tari_indexer/src/rest_api/handlers/network.rs:57`) calls `ConsensusConstants::from(context.network())`, so a devnet that sets `exhaust_burn_rate_bps` reports the built-in target. The constants are already loaded in `lib.rs`; thread them into the handler context.
- **`committee_size_per_shard_group = 0` is accepted by the loader** and only fails at bootstrap via the `NonZeroU32` conversion. `apply_to` already validates the burn rate; a zero committee size deserves the same treatment so the error names the file.

### Nits

- Template values in `CONSENSUS_CONSTANTS_TEMPLATE` (`instances/manager.rs:32-45`) mostly mirror devnet defaults, but `max_transaction_validity_epochs = 10` (default 2160) and `epoch_end_spread_blocks = 0` (default 1) do not. A reader will assume the commented values are the current ones. Either use the defaults throughout or say they are suggestions.
- `swarm_path()` recovers the swarm dir via `processes_path.parent().expect(...)`. `ProcessContext::new` is built from `self.base_path` a few lines up; passing it directly avoids the invariant.

### Verified sound

- Loader semantics: LocalNet-only gating, configured-vs-default path handling, `deny_unknown_fields` rejecting `num_preshards`, feature gate on `tari_ootle_app_utilities`, and no remaining `TARI_DEVNET_COMMITTEE_SIZE` references anywhere in the repo.
- Path resolution: VN resolves relative paths in `set_base_path` (called at config load), indexer via `to_data_dir()`; the harness passes absolute paths.
- Feature epoch numbering: `create_network` ends with all validators at epoch 3, so `epoch 4` after VN6's registration is correct, and `all_validators_have_started_epoch` includes VN6.
- Swarm template is written in `fork_all` at startup, before any `fork_new`, so a configured file always exists when a node starts.
- Committee size flows from the loaded constants into both epoch managers (`indexer/bootstrap.rs:233`, `validator_node/bootstrap.rs:268`), so the file actually drives `calculate_num_committees`.

PR #2517, opened 2026-09-03
feat(indexer): follow each shard group's tip on a held-open sync_state stream

sdbondi:
## Summary

The indexer's network state sync no longer polls. Each shard group holds a `sync_state` stream open past the validator's tip and receives transitions as they commit, with a per-request deadline and keepalive interval. This is the follow-mode half of phase 2 of the substate cache work (after #2498, #2499, #2511, #2512).

### Wire (`crates/p2p/proto/rpc.proto`)

`SyncStateRequest.follow`. A following responder keeps an unbounded stream open once every cursor has reached its tip, streams each shard's new versions as they commit and closes off each such burst with a further `SyncComplete`. No marker on a followed stream is final. Refused on a bounded request. An old responder ignores the flag and closes at the tip with a final marker, so a following client degrades to polling.

### Responder (`state_sync_task.rs`)

- Subscribes to hotstuff events before the catch-up, then passes over every cursor on each `BlockCommitted`, streaming a shard only if its tip moved past the cursor.
- `EpochChanged` forces a marker for every shard. The marker re-runs `authorise_tip` at the new epoch, so a shard this node no longer stores ends the stream promptly instead of leaving the caller to wait out the deadline on silence.
- The commit event fires inside the committing transaction, so a pass can run before its transitions are visible. The next commit's pass catches them up; the pacemaker commits on a timer.
- The cursor advances monotonically past each marker.

### Indexer (`worker.rs`)

- Each shard group runs its own loop (session, probe, stream, reopen) and the groups run concurrently under a `FuturesUnordered`. The global shard is claimed by the lowest group.
- `sync_state_with_options` with `state_sync_stream_deadline` (120s) and `state_sync_keepalive_interval` (10s) from config.
- Stream ends: timeout on a followed stream is the ordinary quiet end and reopens at once (the reopen refreshes each shard's watermark with a marker); a final marker means the responder does not follow, so the reopen waits the work interval; anything else is a peer fault and also waits.
- An epoch change tears every stream down and re-plans against the new committees. Cursors survive in the persisted progress. Per-group teardown (item 4 of the handoff) and the `num_committees` re-plan (item 5) are the follow-up PR.
- Progress is shared between group loops behind a tokio mutex held from snapshot to write, since it is persisted as one value.
- `StreamOrder` no longer treats a marker as ending a shard. Version monotonicity and chunk integrity are unchanged.
- `SyncStats` is shared counters, reported every work interval.

### Config

`state_sync_stream_deadline` must stay comfortably under `substate_cache_max_serve_lag`: a quiet shard is confirmed only when its stream is reopened, and the client notices a quiet stream ended up to a minute after the deadline (the RPC grace period). With defaults that is ~180s against a 300s lag. Threading keepalive liveness into `ShardWatermarks` (item 3) lifts that constraint later.

## Test plan

- [x] `cargo test -p tari_indexer --lib network_state_sync`
- [x] `cargo test -p tari_validator_node --lib state_sync_task`
- [x] clippy clean on `tari_indexer`, `tari_validator_node`, `tari_rpc_state_sync`
- [x] CI integration tests, including `committee_split.feature` from #2515

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11745,6 +11745,7 @@ dependencies = [
  "time",
  "tokio",
  "tokio-stream",
+ "tokio-util 0.7.18",
  "toml 0.9.11+spec-1.1.0",
  "tower-http",
  "tracing",
--- a/applications/tari_indexer/Cargo.toml
+++ b/applications/tari_indexer/Cargo.toml
@@ -91,6 +91,7 @@ tokio = { workspace = true, features = [
     "rt-multi-thread",
 ] }
 tokio-stream = { workspace = true }
+tokio-util = { workspace = true }
 tower-http = { workspace = true, features = ["default", "cors", "limit", "trace"] }
 url = { workspace = true, features = ["serde"] }
 prometheus-client = { workspace = true, optional = true }
--- a/applications/tari_indexer/docs/configuration.md
+++ b/applications/tari_indexer/docs/configuration.md
@@ -62,8 +62,20 @@ Main indexer application settings.
 # Block scanning interval in seconds (default: 10)
 #block_scanning_interval = 10
 
-# State scanning interval in seconds (default: 60)
+# How long, in seconds, a shard group waits before reopening its state sync stream after it fails,
+# or after a validator that does not follow its tip closes it (default: 60)
 #state_scanning_interval = 60
+
+# The longest, in seconds, a validator holds a state sync stream open with no transition to send. The
+# stream follows the validator's tip and is reopened once this passes, one completion marker per
+# shard. The substate cache serves a quiet shard only while its last marker arrived within
+# substate_cache_max_serve_lag, so keep this comfortably shorter than that (default: 120)
+#state_sync_stream_deadline = 120
+
+# How often, in seconds, a validator is asked to show it is still there while the state sync stream
+# has nothing to send. Must be well under state_sync_stream_deadline; validators serve no shorter
+# than their own minimum (5s by default) (default: 10)
+#state_sync_keepalive_interval = 10
 
 # How many epochs past its terminal epoch a transaction submitted through this indexer is retained.
 # A transaction's terminal epoch is the epoch it committed in once its receipt has been indexed, and
--- a/applications/tari_indexer/src/bootstrap.rs
+++ b/applications/tari_indexer/src/bootstrap.rs
@@ -274,6 +274,8 @@ pub async fn spawn_services(
         NetworkWideStateSyncConfig {
             event_filters: Arc::from(config.indexer.event_filters.clone()),
             work_interval: config.indexer.state_scanning_interval,
+            stream_deadline: config.indexer.state_sync_stream_deadline,
+            keepalive_interval: config.indexer.state_sync_keepalive_interval,
             watched_templates: watched_templates.clone(),
         },
         event_notifier.clone(),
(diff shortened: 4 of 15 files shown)


sdbondi (commented):
Read both commits in full against `ce3baecec`, and traced the deadline/keepalive path through `networking/rpc_framework` (server `process_body` idle budget, client `ReplyTimeout` → `RpcStatus::timed_out`), the `BlockCommitted`/`EpochChanged` publish sites in consensus, and `pacemaker.set_epoch` ordering. The responder design holds up: the cursor advance is monotone, the forced marker on `EpochChanged` re-runs `authorise_tip` after `current_view` has moved to the new epoch, and the "pacemaker commits on a timer" claim checks out (`block_timer` → force beat), so the pass-before-visible race is bounded by one block time. On the client side, `StreamEnd::TimedOut` correctly matches what the framework delivers after deadline + grace, and holding the progress lock from snapshot to write is the right shape.

### Blocking

**1. Epoch teardown can roll a shard's cursor back and double-count economic totals.** `applications/tari_indexer/src/network_state_sync/worker.rs:166-182` drops the `sync_plan` future the moment an epoch event arrives. A group loop parked inside `with_write_tx` (`worker.rs:778-819`) is not cancelled by that drop — `SqliteIndexerStore::with_write_tx` runs the closure via deadpool `interact` on a blocking thread and only awaits the result, so the transaction keeps going and commits after the drop. `initialize_sync_plan` (`worker.rs:224-231`) then reads `Key::SyncProgress` and can observe the pre-commit map (readers are not blocked until the writer's commit lock). The new plan's `SharedSyncProgress` is now stale for that shard, and the next whole-map persist from any group — including `sync_checkpoints` at `worker.rs:324-329`/`386-403`, which runs before any stream is opened — writes the stale cursor over the committed one. On reopen the responder re-streams the already-committed version; `StreamOrder` is seeded from the same stale cursor so it accepts it, and the read-modify-write on `TariAccumulatedClaimed`/`Fees`/`ReceiptExhaustBurn` (`worker.rs:808-816`) counts it twice. UTXO/receipt inserts are idempotent; the totals are not.

This is new in this PR: the base awaited `start_sync_round` inside the select arm, so an epoch event never interrupted an in-flight write. Two ways to close it, either is fine:
- Tear down cooperatively: signal the group loops (a `Notify`/`CancellationToken` checked between messages in `stream_shard_state`) and await `sync` to completion before `initialize_sync_plan`, rather than dropping it.
- Make the persist monotone: have the write closure read the persisted map and merge `max` per shard/group before `key_value_set`, so a stale snapshot can never move a cursor backwards. The doc on `SharedSyncProgress` (`sync_progress.rs:41-45`) already states the invariant this would enforce.

### Non-blocking

**2. The follow path on the responder has no tests.** `state_sync_task.rs` tests cover cursor validation and `TipAuthority` only. `follow_shard`'s moved check (`tip >= start_state_version`), the forced marker, and the `max(start, synced_to + 1)` advance at `state_sync_task.rs:329-330` are the load-bearing pieces and are cheap to unit test if the advance is lifted into a pure fn on `ShardCursor`. The client-side `StreamOrder` change is well covered.

**3. A responder task outlives its client until the next commit.** After the server abandons an idle stream at the deadline (`server/mod.rs:812-827`), `follow_tip` (`state_sync_task.rs:218-238`) sits in `events.recv()` and only learns the client is gone when the next pass calls `send`. With force beats that is one block time, so it's bounded; a `select!` on `self.sender.closed()` would end it at once and is a two-line change.

### Nits

- `worker.rs:490-493`: the `StreamEnd::Final` log says the validator "closed the stream at its tip", but `sync_shard_group_state` also returns `Final` at `worker.rs:544-546` when there was no incremental cursor to follow (every shard still at version 0). Worth a distinct variant or message so a quiet log isn't misread as an old responder.
- `worker.rs:36`: importing `Either` via `tari_rpc_framework::__macro_reexports` is pre-existing but this PR touches the line; `futures::future::Either` is already a direct dependency.

PR #2519, opened 2026-09-03
feat(engine)!: add native intrinsics and move signature verification onto them

sdbondi:
## Why

Templates had no hashing at all — `EngineOp` carried no hash op and `template_lib` exposed nothing beyond hashmap helpers — no access to group or scalar arithmetic, and one bespoke engine op for signature verification. Anything cryptographic beyond that op had to be compiled to WASM, where it costs roughly ten times what the same work costs natively: a Groth16 verification measured at ~112M metering points against ~9.9M running natively.

This adds a native intrinsic set and moves signature verification onto it.

## The API

```rust
use tari_template_lib::prelude::*;   // re-exports the `intrinsics` module

intrinsics::ristretto_add(&a, &b)              intrinsics::scalar_add(&a, &b)
intrinsics::ristretto_sub(&a, &b)              intrinsics::scalar_sub(&a, &b)
intrinsics::ristretto_negate(&a)               intrinsics::scalar_mul(&a, &b)
intrinsics::ristretto_mul(&point, &scalar)     intrinsics::scalar_negate(&a)
intrinsics::ristretto_mul_base(&scalar)        intrinsics::scalar_invert(&a)      -> Option
intrinsics::ristretto_msm(&points, &scalars)   intrinsics::scalar_from_uniform_bytes(&[u8; 64])
intrinsics::ristretto_is_identity(&a) -> bool  intrinsics::scalar_is_canonical(&a) -> bool
intrinsics::ristretto_is_canonical(&a) -> bool

intrinsics::hash_blake2b(data) / _sha256 / _keccak256 / _sha512
intrinsics::hash_blake2b_parts(&[a, b])        // one engine crossing per Merkle level
intrinsics::schnorr_verify_with_domain(..)
```

Every intrinsic is a pure function of its arguments — no substate, no workspace, no caller context. That restriction is what lets the engine price one from its declared arguments *before* running it and charge it through `charge_native_execution`, and what makes its use reproducible on every validator.

### Failure convention

Malformed input panics and aborts the transaction, which keeps the common path free of `Result` threading. A template handling untrusted bytes checks them once at its boundary with `ristretto_is_canonical` / `scalar_is_canonical` and then operates freely — cheaper and clearer than threading a `Result` through every step.

An `Option` or `bool` return means the absence of an answer *is* the result — a zero scalar has no inverse, a signature is or is not valid. It never signals a malformed argument. So no operation doubles as a validity oracle; validity testing has its own intrinsics.

## Extending the set later

Intrinsics are addressed by a permanent numeric `IntrinsicId` behind a single `EngineOp::IntrinsicInvoke`, rather than an enum behind an op each. Adding one — Groth16, say — changes no wire type and leaves every published template working; it needs a coordinated validator and indexer upgrade, and nothing more.

The id is a plain integer rather than an enum on purpose. A validator that does not know an id answers `IntrinsicNotSupported` naming it, which is the correct diagnosis when a template built against a newer `template_lib` meets a validator that has not been upgraded. An enum would instead fail inside the CBOR decoder as an unknown variant, which reads like a malformed transaction.

`every_id_is_priced_and_dispatched` fails the build if an id reaches `price` without `dispatch` or the reverse: the first charges for work never done, the second runs unmetered native code, which is the hole the pricing discipline exists to close.

## Signature verification consolidated

`EngineOp::SignatureInvoke` is retired and discriminant `0x10` reserved. The template-facing interface is unchanged — `Signature::verify` and `Verifiable` behave exactly as before and their four tests pass untouched — they simply route through `IntrinsicId::SCHNORR_VERIFY`.

**This changes what a verification costs.** It was a flat `per_signature_verification_cost` fee charge (10 µT) and zero metering points. It is now `NativeExecutionPoints::PER_SCHNORR_VERIFY`, which prices the real ~100µs of validator CPU and counts toward the block execution budget as every other native verification does. The flat rate, its `RuntimeEvent::SignatureVerified` and `FeeSource::SignatureVerification` are removed rather than left to double-charge. `FeeSource` discriminant 4 is left reserved for future use.

Intrinsics get **no fee source of their own**. They are charged under the existing `FeeSource::NativeExecution`, alongside stealth, confidential and burn-claim verification. A source per kind of native work would widen `FeeReceipt::widest`, and so the receipt size bound on every transaction, to itemise a breakdown the shared per-point rate already makes comparable — the same reasoning `RuntimeCall` already documents for merging two unlike prices.

## Tests

`crates/engine/tests/intrinsics.rs`, 11 tests, plus a test template exercising the surface.

Per-primitive equality against natively computed values catches a wrong implementation. It does **not** catch a primitive wired to the wrong operation, because both sides move together — so the tests that earn their place are the algebraic identities:

- `(a + b)*G == a*G + b*G`, which only holds if scalar addition, fixed-base multiplication and point addition all do what they claim
- `commit(v1,r1) + commit(v2,r2) == commit(v1+v2, r1+r2)`, the property any confidential scheme built on these would depend on
- a multi-scalar multiplication against the same terms multiplied and summed one at a time
- hashing parts against hashing the concatenation, without which a template walking a Merkle path would compute a different root than the one that committed it

Plus `intrinsics_are_charged_as_native_execution`, which checks the charge lands under `NativeExecution` and that an 8-term MSM costs more than a 1-term one — that last part is what proves the price is a function of the declared arguments rather than a constant.

## Versioning

Workspace 0.40.0 -> 0.41.0 plus the crates re-exposing the removed items, per `scripts/crate_versioning.py impact --breaking`, with one deviation: a crate whose current version is not yet on crates.io absorbs this change in the version it already carries, so `ootle-rs`, `tari_indexer_client`, `tari_ootle_wallet_sdk`, `tari_ootle_wallet_storage_sqlite` and `tari_ootle_walletd_client` stay put rather than being bumped past an unreleased number. Every target version was checked against crates.io.

Two pins live outside `[workspace.dependencies]` and so fall outside the script's tier-3 sweep — `ootle-wasm-core` in `ootle-wasm`, `ootle_ledger_client` in `ootle-rs`. Both updated; a repo-wide scan reports no remaining stale pin.

## Notes for review

- **The prices are provisional.** `PER_RISTRETTO_MUL`, `PER_SCHNORR_VERIFY` and the rest are derived from the ratio implied by the existing `PER_INPUT` calibration, not measured. They want a run of the kind `native_points_calibrate` does for the stealth constants before this carries real value.
- Retiring `EngineOp::SignatureInvoke` breaks any already-published template that calls it. Accepted as testnet-only.
- `Hash32` is now re-exported from the `template_lib` prelude, since hashing is a first-class template capability and its return type should be reachable without a separate import.
- Signing is deliberately absent. A template has no secrets — its code is public, its state is public substate, and every validator executes it — so a `schnorr_sign` taking a secret key could not produce a signature only that component could have produced.

Verified locally: `cargo lints clippy --all-targets` clean, `cargo machete` clean, fmt clean, and the intrinsics / signature / fees / engine-lib suites pass.


crates/template_lib/src/intrinsics.rs
```diff
@@ -0,0 +1,243 @@
...
+
+        #[doc = concat!("The ", $alg, " digest of every part concatenated, in order.")]
+        /// One engine call rather than one per part, which is what makes walking a Merkle path
+        /// affordable: the parts are concatenated engine-side and hashed once.
+        pub fn $parts(parts: &[&[u8]]) -> $out {
+            let args = parts
+                .iter()
+                .map(|part| Bytes::from_vec(part.to_vec()))
+                .collect::<Vec<_>>();
+            invoke($id, invoke_args![args])
```
sdbondi: To avoid copy and allocations, we need a BytesRef which holds a slice and encodes as CBOR bytes

sdbondi: Done — added `BytesRef<'a>` in `template_lib_types::bytes`: a borrowed slice encoding as CBOR major type 2, the same wire shape `Bytes` produces. Encode-only, since the engine decodes into the owning `Bytes`.

The hash `_parts` builder and `scalar_from_uniform_bytes` use it now, so a blob being hashed is no longer copied into an owned `Bytes` before the encoder copies it again. Follows the `SignatureVerifyArgRef` convention already in `engine_args`.

sdbondi (commented):
**Review verdict: changes requested** (posted as a comment — GitHub does not allow request-changes on one's own PR).

Reviewed at ac7a7c3. The shape is sound: pure-function intrinsics behind one op with a numeric id, price-before-dispatch through `charge_native_execution` (so both the per-tx native ceiling and the compute allowance gate the call before any work), and the `every_id_is_priced_and_dispatched` guard. The `SignatureInvoke` retirement is clean — no stale references left in Rust, bindings, or the three fee-label maps. `BytesRef` is the right fix for the hash/uniform-bytes copies. The identity/homomorphism tests do earn their place.

Two things block, both in the metering/dependency story the description leans on.

### Blocking

**1. MSM is priced as a batch but executed as a loop — under-metered.**
`crates/engine/src/intrinsics/mod.rs:159` does `acc + (decode_point(p)? * decode_scalar(s)?)` per term: one decompression, one full variable-base multiplication and one addition each. That is exactly the work of `RISTRETTO_MUL` (`PER_RISTRETTO_MUL = 550_000`) plus an `OP`, yet each term is charged `PER_RISTRETTO_MSM_TERM = 250_000` (`crates/engine_types/src/limits.rs:139`) on the stated rationale that "the batch shares the accumulation". Nothing is shared. A template can therefore buy N variable-base muls at under half price by routing them through `ristretto_msm`, and the template-lib doc at `crates/template_lib/src/intrinsics.rs:86` ("the engine batches the terms") is describing behaviour that does not exist.

`tari_crypto` already has the real thing: `RistrettoPublicKey::batch_mul(&scalars, &points)` (from the `PublicKey` trait) calls `RistrettoPoint::multiscalar_mul` (Straus/Pippenger). Use it, and the per-term price becomes defensible. If you would rather keep the loop, the per-term price must be at least `PER_RISTRETTO_MUL + PER_RISTRETTO_OP` and the "cheaper than a loop" docs go.

**2. The `curve25519-dalek` dependency is unnecessary and on the wrong major.**
`crates/engine/Cargo.toml:26` adds it "solely for `scalar_invert` — `tari_crypto` exposes no inversion". It does: `RistrettoSecretKey::invert(&self) -> Option<Self>` in tari_crypto 0.23, returning `None` for zero — precisely the semantics the intrinsic wants. That replaces `dalek_scalar` and the whole `SCALAR_INVERT` arm at `crates/engine/src/intrinsics/mod.rs:196` with `a.invert()` and drops the direct dep.

Beyond redundancy: the workspace pins `curve25519-dalek = "4.1"`, but tari_crypto 0.23.2 resolves against `curve25519-dalek 5.0.0` (see `Cargo.lock`). So the engine would carry two majors of the curve library and round-trip scalars between them through bytes. The description's "already resolved at 4.1.3 transitively" is true only because of `ed25519-dalek`/`snow`/`monero` — not the crypto path this code is on.

### Non-blocking

- **`docs/developer-docs/src/content/docs/reference/fees.mdx:179,225,347`** still document `SignatureVerification` / `per_signature_verification_cost` at 10 µT. Since this PR removes the source and the fee-table field, the reference doc should move with it.
- **`SCHNORR_VERIFY` is flat-priced (`mod.rs:80`) but hashes `domain || message` of unbounded length.** Every other variable-size intrinsic is priced off its byte length; a `PER_HASH_BYTE * (domain.len() + message.len())` term would keep the "price is a function of the declared arguments" invariant honest here too. The args are already decoded in `dispatch`, so decoding once in `price` and threading through is the cheap way (see next point).
- **`price` and `dispatch` each decode the arguments** (`hash_input_len` at `mod.rs:254`, and the `Vec<RistrettoPublicKeyBytes>` at `:61` vs `:145`). For the hash intrinsics that is a second full copy of the blob being hashed, done before any charge. Decoding once and passing the parsed value to `dispatch` avoids both the copy and the unmetered work.

### Nits

- `IntrinsicId::ALL` is the only thing tying the guard test to the id set; a new `pub const` that forgets `ALL` escapes it. A `const _: () = assert!(ALL.len() == N)` or listing the consts from one macro would close that.
- `hash_sha512` returns `[u8; 64]`, which minicbor encodes as a 64-element array (~130 bytes) rather than a byte string. Not wrong, but a `Hash64`/`ByteArray` wrapper would match how `Hash32` crosses the boundary.
- `// _Reserved = 0x10` and `// 0x10 => Some(EngineOp::_Reserved)` in `ops.rs` are commented-out code; a plain `// 0x10 reserved` reads cleaner.

crates/template_lib/src/intrinsics.rs
sdbondi: Reconsidered the `hash_sha512` nit — you were right, so `Hash64` moved to `template_lib_types` rather than being duplicated.

`engine_types::Hash64` already used `#[cbor(with = "minicbor::bytes")]`, so this is a pure move with no wire-format change; `engine_types/src/hash.rs` is now a re-export and keeps its `serialize_deserialize` test, which asserts `Value::Bytes` and still passes. `hash_sha512` returns `Hash64`, and `Hash64` joins `Hash32` in the template-lib prelude.

Two things that made the move safe rather than disruptive: every `Hash64` use outside its own file is a return type in `crypto/messages.rs` plus one wallet-crypto test, so there was no wide API surface to preserve (I mirrored the full method set anyway — `leading_bytes`, `trailing_bytes`, `as_slice`, `zero`, `DerefMut`, `Default`). And there are two distinct `HashParseError` types: `template_lib_types`' unit struct and `engine_types`' thiserror enum. Only the former has consumers (`storage_sqlite`), so `Hash64` switching to it breaks nobody.

Also added sha512 to `hashes_match_native` — it had no coverage before, which is partly why the encoding slipped past me.

sdbondi (commented):
**Review verdict: approve** at 7700a41 (posted as a comment — GitHub does not allow approving one's own PR; another maintainer must approve).

Re-verified each item from the earlier review against the diff, not the commit messages:

- **MSM** now goes through `RistrettoPublicKey::batch_mul` (real `multiscalar_mul`), so the per-term price matches work actually done; the "engine batches" docs are now true.
- **`curve25519-dalek`** removed from the workspace and engine; `SCALAR_INVERT` uses `RistrettoSecretKey::invert()`, which has exactly the `None`-for-zero semantics the intrinsic exposes.
- **Schnorr and hash pricing** now scale with `EngineArgs::encoded_len()`, charged before decode — closes both the flat-price gap and the double-decode copy.
- **`fees.mdx`** no longer documents `SignatureVerification`.
- **`IntrinsicId::ALL`** guarded by a `const` assertion on `ALL_LEN`; `ops.rs` reserved slots are plain comments.
- **`Hash64`** moved to `template_lib_types` for the SHA-512 return: same derives, same CBOR byte-string encoding, same serde hex behaviour (`fixed_hex` matches the old `ootle_serde::hex` branch for branch), `engine_types` re-export keeps existing callers compiling. Wire-compatible.
- PR description no longer claims the dalek dependency.

CI: clippy, fmt, machete, docs, web UI, ledger, integration, fuzz and test shards 2/3 and 3/3 are green. Shard 1/3's `cargo test` step has succeeded; the job shows pending only because the runner's "Stop containers" teardown is still finishing.

PR #2520, opened 2026-09-03
feat(engine)!: burn a share of collected fees instead of charging it on top

sdbondi:
## Summary

Switches the exhaust burn from a surcharge to a share of what was collected.

| | user pays | leader gets | burn | burned fraction |
|---|---|---|---|---|
| before | `F·(1+r)` | `F` | `F·r` | `r/(1+r) ≤ 1/2` |
| after | `F` | `F − B` | `B = ⌊F·s/10_000⌋` | `s ∈ [0, 1]` |

The fee table alone is the user's price; the rate only splits what was collected. `s = 10_000` burns everything, which consensus already tolerates (zero leader fee is handled throughout). Every network stays at 500 bps; the policy value is a one-line follow-up.

**Breaking.** Every validator computes `transaction_fee`, `exhaust_burn` and the header burn total differently from a node running the surcharge model, even at the same configured rate, so this needs a coordinated upgrade or a testnet reset. Receipts persisted under the surcharge model still decode: `exhaust_burn` defaults to 0, and the old breakdown slot is kept as `FeeSource::Reserved` (same CBOR/borsh index, JSON alias for the old name) so wallet and indexer transaction listings keep rendering.

## Engine

- The burn is no longer a charge. `FeeModule` stops assigning `FeeSource::ExhaustBurn`; `finalize_fees_and_refunds` computes `⌊total_fees_paid × s / 10_000⌋` once at settlement (overcharge included, so `split_overcharge` goes) and records it on a new additive `FeeReceipt::exhaust_burn` field (`#[n(4)] #[cbor(default)]`, so receipts already on testnets decode).
- `FeeSource::ExhaustBurn` becomes `FeeSource::Reserved`: never written, same index, `#[serde(alias = "ExhaustBurn")]` for JSON-persisted receipts. Can be dropped at any later reset.
- `pre_burn_fees_paid()` keeps its formula, so nothing above the engine changes. `total_fees_charged()` and `required_fees()` no longer include a burn.
- Tracker: `spendable_on_charges` / `payment_covering_charges` became the identity and are removed. The compute allowance is `payments − charges`, `required_fee_payment` is `total_charges`.
- The rate reaches settlement as `ExhaustBurnRate` (processor → tracker → fee state), so the `≤ 10_000` bound the split relies on is enforced by the type.

## Estimation

- `FeeRates` loses `exhaust_burn_rate`; `FeeTable::to_rates()` and `fee_estimate_allowance()` lose their rate argument.
- `FEE_ESTIMATE_ALLOWANCE` drops from 25 to 12 (drift alone). `required_fees_covers_a_real_run_at_any_max_fee` still passes at every rate.
- `MAX_EXHAUST_BURN_RATE_BPS` / `ExhaustBurnRate` docs: the ceiling now means "burn all of it".
- `tari_ootle_app_utilities`: `exhaust_burn_rate_by_network` and the consensus-mirroring test are deleted, along with the test-only `tari_consensus` dev-dependency.

## Docs

`reference/fees.mdx` rewritten for the share model (checkpoint test, compute allowance, charge tables, settlement, the exhaust burn section, estimation allowance).

## Web UIs

Walletd and indexer receipt views read `exhaust_burn` from the receipt and show it as a share of the total paid; the validator node block view gains a per-block "Fee Burn" row (this block's step in the accumulated burn, as a share of the leader fee plus burn).

## Consensus

No logic change. `TransactionExecution::exhaust_burn()` reads the receipt field instead of the breakdown. `ConsensusConstants::exhaust_burn_rate` doc updated.

## Indexer

`fee_volume` is now total fees paid by users, so `receipt_exhaust_burned / fee_volume` is the realised share. Existing accumulators are in the old units; a resync recomputes them.

## Tests

- `working_state.rs`: settlement split property test (`burn + leader == paid`, floor share, full rate leaves leaders zero) replaces `burn_is_a_floor_divided_percentage_of_base_fees`.
- `fees.rs`: `the_exhaust_burn_compounds_the_drift` replaced by `the_exhaust_burn_does_not_move_the_charges`; `spend_the_whole_allowance_and_still_pay` asserts the burn is taken out of the payment.
- `stealth_fee_estimate.rs`: estimates no longer vary with the rate.
- `consensus_tests` test receipt builds a share-model receipt.

Ran locally: `cargo lints clippy --all-targets` on the touched crates, `tari_engine_types`, `tari_engine` (lib + `fees` + `stealth_fee_estimate`), `tari_ootle_app_utilities`, `tari_indexer` economics, and `consensus_tests`.


sdbondi (commented):
Read the full diff against `development` at 7a5e202 and traced the settlement path, the receipt helpers, the consensus consumers and the indexer accumulators.

**Sound**

- Settlement: `exhaust_burn_share(total_fees_paid, rate)` in `working_state.rs:2154` is computed once over the collected total, overcharge included; `pre_burn_fees_paid()` stays `paid − burn`, so `calculate_leader_fee` and the header burn totals need no change. The u128 cast comment holds for every `ExhaustBurnRate`.
- Tracker: with the burn out of the charges, `spendable_on_charges`/`payment_covering_charges` really are the identity, and the fee-checkpoint test in `impl.rs:3572` is the direct `payments < charges` comparison again. Removing them is correct, not just simplifying.
- Receipt: `#[n(4)] #[cbor(default)] #[serde(default)]` is the right additive shape; `widest()` and `widest_bounds_every_other_receipt` include the new field. Old receipts decode with `exhaust_burn = 0`, and since `TransactionPoolRecord` already carries its own `exhaust_burn`, nothing re-derives a burn from a stored pre-share receipt.
- Estimation: dropping the rate from `FeeRates`/`to_rates()`/`fee_estimate_allowance()` is consistent everywhere I looked (walletd, sdk fixture, stealth estimate). `FEE_ESTIMATE_ALLOWANCE` 25 → 12 matches the derivation test and the fixture delta (`dry_run.json` estimated_fee moves by 13, i.e. the old `+1` burn boundary and the burn share of the drift).
- Integration test: the second claim can collect at most the first claim's fee (`MAX_FEE = 50_000` in `wallet_daemon_client.rs:78`), so "increased by at most 50000 since the last check" is the right bound and no longer depends on the absolute pool size.
- CI: the only red job is `test (1/3)` failing on `leader_failure::single_shard_node_goes_down_and_catches_up`, which is the known unrelated flake with a fix in a separate PR.

**Blocking**

1. `docs/developer-docs/src/content/docs/reference/fees.mdx` still documents the surcharge model this PR removes, and every recent fee PR that changed the model updated this file in the same change (#2482, #2447, #2441). As it stands the reference contradicts the code in at least these places:
   - `:209` — "the check is `spendable(payments) ≥ total charges` … the exhaust burn is taken *on top of* the charges".
   - `:283` — the `ExhaustBurn` row in the charge table, and `:98` in the pipeline SVG listing it as a charge.
   - `:308`–`:314` — "the overcharge is split … burn share moves into `ExhaustBurn`".
   - `:319`–`:330` — "The exhaust burn is charged **on top of** the accrued fee" and the ceiling rationale about the allowance.
   - `:353` — "`ExhaustBurn` | finalisation | 5% of the total".
   - `:365`, `:373`–`:376` — `FEE_ESTIMATE_ALLOWANCE` (25 µT), "the exhaust burn re-multiplies both", "at any exhaust burn rate up to `MAX_EXHAUST_BURN_RATE_BPS`".

**Non-blocking**

2. This changes what every validator computes for `transaction_fee`, `exhaust_burn` and the header burn total, so it is consensus-breaking even though the configured rate is unchanged. Neither the commit titles nor the PR title carry the `!` the repo uses for this (`feat(engine)!:` in #2482/#2503), and the description says "pure model change". Worth marking as breaking and stating that it needs a coordinated upgrade or testnet reset, so the release notes don't understate it.
3. `exhaust_burn_share` (`working_state.rs:2154`) takes a raw `u16` and relies on the `≤ 10_000` invariant for the lossless cast, but the value reaches it via `FeeState::burn_rate_bps: u16` seeded from `TransactionProcessor`/`TemplateTest::set_burn_rate_bps(u16)`, which nothing bounds. The old `calculate_burn_amount` returned an `Overflow` error above the ceiling; the new helper silently truncates. Either take `ExhaustBurnRate` through the processor/fee-state or `debug_assert!(rate_bps <= MAX_EXHAUST_BURN_RATE_BPS)` in the helper.

**Nits**

4. `bindings/src/types/TransactionPoolRecord.ts:40` still references `FeeReceipt::exhaust_burn_charged`, which no longer exists; the Rust doc it mirrors was updated in `transaction_pool.rs:437`.
5. `docs/developer-docs/public/indexer/openapi.json` carries the old `fee_volume`/`target_burn_rate_bps` descriptions that `clients/tari_indexer_client/src/types.rs` changed; regenerate if that is part of the usual flow.

sdbondi (commented):
Re-verified at 04921f6: docs rewritten to the share model everywhere flagged, breaking marker added, the burn rate travels as `ExhaustBurnRate` end to end, TS comment fixed. All CI checks green, including test (1/3) this time. Would approve; GitHub refuses approvals on one's own PR, so another maintainer needs to.

crates/engine_types/src/fees.rs
```diff
@@ -182,6 +181,13 @@ pub struct FeeReceipt {
     /// Breakdown of fee costs
     #[n(3)]
     cost_breakdown: FeeBreakdown,
+    /// The share of `total_fees_paid` that is burned rather than paid to leaders: `⌊paid × rate / 10_000⌋` at
+    /// the exhaust burn rate in force for the execution epoch. Settled over what was collected, so it is never
+    /// charged to the payer and never appears in `cost_breakdown`.
+    #[n(4)]
+    #[cbor(default)]
+    #[serde(default)]
+    exhaust_burn: u64,
```
sdbondi, line 190: For future: this will decode, but it will break resyncing because it changes the JMT hash (borsh)

PR #2523, opened 2026-09-04
feat(indexer): keep quiet shards confirmed and reopen per shard group on an epoch change

sdbondi:
## Summary

Second half of the follow-mode work, after #2517. Three things:

### 1. Keepalives refresh `ShardWatermarks`

`ClientStreaming::keepalives()` (rpc_framework) is a `watch` counter moved by each keepalive addressed to that response. `stream_shard_state` waits on it alongside the stream and, on each one, re-stamps the watermark of every shard the stream has closed off with a marker. A shard still being caught up is not level and is not touched. A keepalive left over from an abandoned request is still discarded for the read but does not count for the new signal.

This removes the coupling between `state_sync_stream_deadline` and `substate_cache_max_serve_lag`: a quiet shard stays served for as long as its validator keeps answering. The deadline default moves to 600s and is now only how often a quiet stream is reopened. `substate_cache_max_serve_lag` now needs to cover a validator going away: several missed keepalives plus the reopen, a minute or two against the 300s default.

### 2. Per-group teardown and reopen on epoch advance

An epoch event that keeps the shard-group partition is sent to the group loops on a `watch<Epoch>`. Each group winds its stream down between messages (`StreamEnd::EpochAdvanced`), syncs the new epoch's checkpoints, resolves its committee at the new epoch through the pool and reopens from the cursor it holds. Groups do this independently; nothing is re-planned or re-streamed. Checkpoint sync moves into the group loop and runs before every stream it opens (a no-op once recorded), so a group whose checkpoint source fails retries alone.

### 3. Global re-plan on a partition change

Retained from #2517 for the one case that cannot be per-group: when the set of shard groups changes, every stream is wound down and a new plan drawn. Every exit from the epoch-event arm, including the error paths, winds the plan down first, for the reason established in #2517's review (a write commits on a blocking thread whether or not the future survives).

## Test plan

- [x] `cargo test -p tari_rpc_framework --lib` (new: keepalives for this request are reported, others are not)
- [x] `cargo test -p tari_indexer --lib network_state_sync` (new: refresh keeps a confirmed shard open at its version; does not confirm one that never was)
- [x] clippy clean on `tari_rpc_framework`, `tari_indexer`, `tari_validator_node`
- [x] CI, including `committee_split.feature`, which exercises the per-group epoch path end to end


sdbondi (commented):
Read at 093909c, every changed file at head, traced into `rpc_framework` client/server and the epoch manager.

### Blocking

**1. An indexer started at epoch zero never begins syncing once the same-partition path is taken** (`applications/tari_indexer/src/network_state_sync/worker.rs:190-204`, `:295-299`)

`sync_plan` still short-circuits at epoch zero and parks on `cancel.cancelled()`. Before this PR any `EpochChanged` cancelled and re-planned, so that was fine. Now the event arm compares partitions first: `get_network_description` at epoch 0 reports `max(1, 0 / committee_size) = 1` shard group (`crates/epoch_manager/src/service/epoch_manager.rs:539-546`), and epoch 1 on any network with fewer than `2 * committee_size` validators reports the same one group. The partitions match, `epoch_tx` is bumped, and `continue` is taken. Nothing is listening to that watch: the parked future is not a group loop. The indexer stays at "Current epoch is zero, nothing to sync" until either the partition changes or the broadcast lags.

Any fix works as long as the zero-epoch plan is not treated as a plan that can absorb an epoch change per group. Smallest: in the same-partition guard, also require the plan's epoch to be non-zero, and re-plan otherwise. Alternatively drop the zero-epoch short-circuit entirely: `sync_group_checkpoints` already returns `Ok` when `epoch.checked_sub(1)` is `None`, `new_session` fails at epoch 0 with no committee, and `pause` wakes on `epoch.changed()`, so the group loops handle it on their own.

### Non-blocking

- **A transient epoch manager failure now restarts the whole sync** (`worker.rs:210`). The `Err` from `get_network_description` in the event arm winds the plan down and propagates out of `start`, and `run` sleeps a work interval and re-plans. That is a full teardown of every stream for what was previously a no-op event handler. It is safe, and re-planning is the conservative answer, but a `Lagged`-style "re-plan" (`Ok(())`) would recover a work interval sooner and log at the same level. Your call.

- **Docs claim on `substate_cache_max_serve_lag`** (`config.rs:176-183`) checks out: `keepalive_timeout` is `max(interval, peer_min) * MISSED_KEEPALIVES_BEFORE_TIMEOUT + grace`, plus a `work_interval` pause on "the peer failed", plus a reopen. "A minute or two" against 300s is accurate at the defaults.

### Verified sound

- Keepalive counting in `RpcResponseReader::read_response` only bumps the watch when `check_response` accepts the request id, and the read path still discards mismatched keepalives without failing, matching the existing `keepalives_left_over_from_an_abandoned_request_do_not_end_the_session` test. `keepalive_tx` lives for the whole `do_request_response`, so the receiver's `changed()` errors exactly when the response is over, which is what the `keepalives_open` flag in `stream_shard_state` relies on.
- A server keepalive is sent only when the body stream yielded nothing for the interval (`server/mod.rs:772-810`), and the responder follows the tip on commit events rather than polling (`state_sync_task.rs:229-255`), so "closed off and idle" really is level. Refreshing only `level_shards` is the right restriction: a shard still catching up on the same stream stays unconfirmed.
- `ShardWatermarks::refresh` on a never-confirmed shard is a no-op, covered by the new test.
- Winding down before every exit of the event arm, including the error paths, preserves the #2517 invariant on persisted cursors.
- `borrow_and_update` at the top of each group iteration plus `Ok(()) = epoch.changed()` in both `pause` and the stream select means a change between the read and the reopen is not lost.
- Nothing unrelated swept into the diff.

sdbondi (commented):
Re-checked b1d68ac against 093909c.

- **Epoch-zero stall (blocking): resolved.** `plan_absorbs_epoch` refuses to absorb any epoch into a plan drawn at epoch zero, so the parked future is always wound down and re-planned at the first real epoch. The three unit tests cover the kept, split, and zero-plan cases.
- **Network description error (non-blocking): resolved.** A failed read now logs and re-plans instead of propagating out of `start`.

One thing swept in: `crates/wallet/ootle-rs/examples/stealth_transfer.rs` changes the example recipient address, which has nothing to do with this PR. Harmless, but it belongs elsewhere.

Approving once CI is green.

PR #2528, opened 2026-09-04
refactor(indexer)!: serve only the head from the substate cache

sdbondi:
Stacked on #2527 (both edit `SqliteSubstateCache::read`); rebase onto `development` once that merges.

## Summary

The indexer is a gateway to the network's current state, not to its history, so a lookup that names a version is only ever asking whether that version is still current. This drops the cache's version-aware read path in favour of a head-only cache:

- `SubstateCache::read(id)` returns the head; there is no `version` parameter.
- What the head says about a named version is decided once, in `SubstateCacheEntry::answer_at`: it is the head → hit; below it → `Down` (versions are contiguous and upping a substate downs its predecessor); above it → miss; nonexistence answers an unversioned lookup only. `CachedSubstateManager::get_substate` applies it to the entry it already has.
- The read-side `head_ttl` check goes, along with the synthesised `Down` entries stamped `cached_at: now`. A `Down` derived from the head carries the head's own age, so the manager's `cache_ttl` bounds it like every other entry.
- `head_ttl` stays on the write path (`substate_cache_put`), where it lets a proven or fresher result displace an unverified head recorded too high.

This settles phase 3 item 8 (whether to gate the down-inference age-out on `verified`) by removing the read-side TTL it was about.

## Behaviour change

A version below the head previously answered without a watermark; it now needs the shard's watermark like every other cached answer. Conservative, and one fewer special case.

## Tests

`answer_at` is covered in `tari_indexer_lib::substate_cache::tests`; the `SqliteSubstateCache` tests are reduced to the watermark gates.

### DIFF
--- a/applications/tari_indexer/src/store.rs
+++ b/applications/tari_indexer/src/store.rs
@@ -260,7 +260,10 @@ pub trait IndexerStoreReadTransaction {
     // -------------------------------- Substate Cache -------------------------------- //
 
     /// The cached head version of `substate_id`. Says nothing about whether it is fresh enough to
-    /// serve, nor what it implies for lower versions - see [`crate::substate_cache::SqliteSubstateCache`].
+    /// serve - see [`crate::substate_cache::SqliteSubstateCache`] - nor what it implies for a
+    /// lookup at a particular version, which is
+    /// [`SubstateCacheEntry::answer_at`](tari_indexer_lib::substate_cache::SubstateCacheEntry::answer_at)'s
+    /// to decide.
     fn substate_cache_get(&mut self, substate_id: &SubstateId) -> Result<Option<SubstateCacheEntry>, StorageError>;
 }
 
--- a/applications/tari_indexer/src/substate_cache/mod.rs
+++ b/applications/tari_indexer/src/substate_cache/mod.rs
@@ -1,10 +1,7 @@
 //   Copyright 2026 The Tari Project
 //   SPDX-License-Identifier: BSD-3-Clause
 
-use std::{
-    sync::Arc,
-    time::{Duration, SystemTime, UNIX_EPOCH},
-};
+use std::{sync::Arc, time::Duration};
 
 use log::*;
 use tari_engine_types::substate::{SubstateDiff, SubstateId};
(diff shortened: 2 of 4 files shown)


sdbondi (commented):
Reviewed the commit unique to this PR (`ed7513b`, on top of #2527's head). No blocking findings.

**Sound**

- `SubstateCacheEntry::answer_at` reproduces the old read-path semantics exactly: head/unversioned → hit, below head → `Down` with the head's `cached_at` and `verified`, above head → miss, nonexistence → unversioned only. The manager's `verified || !verify_substate_proofs` gate therefore still refetches a `Down` inferred from an unverified head, as before.
- The read-side `head_ttl` removal is a no-op on the effective bound: `bootstrap.rs:306-313` passes `DEFAULT_CACHE_TTL` as `head_ttl`, and the production manager never calls `with_cache_ttl`, so the inferred `Down` now ages out under the manager's `cache_ttl` at the same value it did under `head_ttl`. The description's "bounds it like every other entry" claim checks out.
- `head_ttl` is still consumed on the write path (`writer.rs:599`, unverified heads only), so nothing became dead.
- Only one `SubstateCache` implementor exists (`SqliteSubstateCache`); both `read` call sites in `cached_substate_manager.rs` are updated.

**Non-blocking**

- `applications/tari_indexer/src/store.rs:263`: the `substate_cache_get` doc still says "nor what it implies for lower versions - see `SqliteSubstateCache`", but that type's docs no longer discuss lower versions. Point it at `SubstateCacheEntry::answer_at` instead.
- The behaviour change is slightly wider than "needs the shard's watermark": a versioned read below a head that has aged past `cache_ttl` now goes to the committee every time, since a `Down` for a named version is never written back as head (`is_head` is false for `Down` with `specific_version`). Previously the same reads were served from the head for `head_ttl` regardless of watermark, then likewise fell through. Same steady state, so fine, just worth knowing that a hot substate whose head is only ever queried by version never refreshes its cache entry. Pre-existing, not introduced here.

CI: clippy and unit tests still pending at time of review.

sdbondi (commented):
Re-reviewed `b6362e3` against `ed7513b`. The only change is the `substate_cache_get` doc pointer in `store.rs`, which now names `SubstateCacheEntry::answer_at`; the code is byte-identical. Nothing outstanding from my side.

CI is red on "Test Results", but none of the three failed scenarios trace to this PR:

- `transfer.feature:8` and `transfer.feature:62` (`Substate component_… does not exist` on the second manifest submit) fail identically on `development` at `68a9eca` (#2505's merge, run 33869968740) and on #2527's head (run 33874136149). They appear to be a regression from #2505's cached nonexistence, not from this change.
- `indexer.feature:8` ("did not sync with the network in time", epoch 2) exercises the state-sync watermark path, which this PR does not touch; it passed on the two runs above and looks like a timing flake.

Approval needs another maintainer. I would not merge this on top of the transfer failures until #2505's regression is understood, even though this PR is not the cause.

PR #2529, opened 2026-09-04
fix(indexer): refuse a cache write below the version the stream has already shown

sdbondi:
## Description

`substate_cache_invalidations` recorded the *shard* state version a transition committed at. That answers one question: did this fetch start before the transition, and get overtaken by it? `substate_cache_put` discards a write whose captured watermark is below the journalled shard version, which closes that race.

It says nothing about a fetch that starts **after** the transition and lands on a committee member that is behind. Such a member answers with a version this indexer has already watched the substate pass — legitimately, from its own point of view — and the transition has already deleted the cached row that would have ranked the answer down (`substate_cache_invalidate` deletes on `version <= retires_up_to`). With no row to rank against and the shard-version guard satisfied, the stale answer is installed as the head and served until it ages out of `DEFAULT_CACHE_TTL` (900s), for as long as the shard stays level.

The trusted-root ring accepts this by design — `substate_cache_put` already says "a committee member that is behind can prove an older version against an older signed root" — and `get_specific_substate_from_committee` keeps `unproven_result` precisely because "a member that is still syncing may respond with a stale copy". Both anticipate the peer; nothing refused its answer.

## Changes

* The journal carries `substate_version`: the version the transition showed, which is the version created or the version destroyed. A write below it is refused.
* It is a floor, not a target. A fetch that catches up with the substate, or overtakes it, is recorded exactly as before — including the destroyed version itself.
* The floor only rises: the journal upsert keeps `max(substate_version, excluded.substate_version)`, so the transitions of one batch (a destroy of v6 and a creation of v7, in either order, from the stream or from `retire_committed`) leave the floor at 7.
* A record that the substate does not exist is left to the existing version ranking. `DoesNotExist` is settled by f + 1 members rather than one, so a single member being behind cannot produce it, and refusing it here would suppress the legitimate one that follows a destroy.
* `SUBSTATE_CACHE_JOURNAL_RETENTION`'s comment now states the quantity it has to outlive: the committee lag, not the length of an in-flight fetch. Both are covered by the same 300s today; tuning it down to a fetch would silently reopen this.

The journal is derived from the stream and spans a fetch, so the migration recreates the table rather than migrating rows into it.

## What this does not close

The guard holds only while the journal row is retained. A member lagging by more than 300s can still answer with a version whose journal row has been pruned, and nothing then remembers the substate reached it.

A destroy with no successor leaves a second, narrower gap: the floor admits the destroyed version itself, so a member that is behind can answer `Up` at that version with the live value and have it installed as the head for the read path's TTL, where a level member answers `Down`. Refusing that needs the journal to record that the row came from a destroy, which is not done here; a destroy without a following creation is the rarer case.

The retention window is the residual hole, and it is bounded: one occurrence serves a stale head for at most the read path's TTL before refetching. Closing it permanently means the head row carrying the version the stream last showed — writing the new head on an up transition rather than deleting the row — which trades a row per substate the indexer has ever seen for the guarantee. Deliberately not done here: the delete is what keeps the cache's storage proportional to what is being read rather than to what has ever existed.

Retention covers the realistic case rather than an arbitrary slice of it. A member answers staley because it is behind, so it answers staley about *recent* changes: a member lagging five minutes serves stale copies of substates changed in the last five minutes and correct copies of everything older.

## Motivation and Context

Raised in review of the phase 2 work, against the substate cache added in #2498 and extended in #2505.

## How Has This Been Tested?

Five tests in `store_factory.rs`: a version the stream has already seen past is refused, a version at or above it is recorded, a destroy records the version it saw (and still admits that version as a head), the floor is the highest version a batch showed regardless of order, and a nonexistence after a destroy is still recorded. The first, third and fourth fail without the guard.

`cargo test -p tari_indexer --lib storage_sqlite` — 48 pass. Clippy clean.

## Breaking Changes

- [x] None
- [ ] Other - Please specify

The migration recreates `substate_cache_invalidations`. The journal is rebuilt from the stream within one retention window, and the substate cache itself is rebuilt from the committee on demand, so no data is lost.

### DIFF
--- a/applications/tari_indexer/src/bootstrap.rs
+++ b/applications/tari_indexer/src/bootstrap.rs
@@ -108,9 +108,13 @@ use crate::{
 
 const LOG_TARGET: &str = "tari::indexer::bootstrap";
 
-/// How long a substate stays journalled as recently changed. The journal exists only to stop a
-/// committee fetch that started before a transition arrived from writing its result back as current,
-/// so it needs to outlive an in-flight fetch and nothing more.
+/// How long a substate stays journalled as recently changed.
+///
+/// The journal stops a fetch from writing back a version the stream has moved past, which covers two
+/// things with different horizons. A fetch the transition overtook is bounded by how long a fetch
+/// takes. A fetch answered by a committee member that is behind is bounded by how far behind that
+/// member is, which is the longer of the two and the one this must outlive: a member lagging by less
+/// than this serves stale copies only of substates whose journal rows are still here to refuse them.
 const SUBSTATE_CACHE_JOURNAL_RETENTION: Duration = Duration::from_secs(300);
 
 const SUBSTATE_CACHE_PRUNE_INTERVAL: Duration = Duration::from_secs(60);
--- a/applications/tari_indexer/src/storage_sqlite/migrations/2026-09-04-000000_substate_cache_invalidation_version/down.sql
+++ b/applications/tari_indexer/src/storage_sqlite/migrations/2026-09-04-000000_substate_cache_invalidation_version/down.sql
@@ -1,0 +1,10 @@
+drop table substate_cache_invalidations;
+
+create table substate_cache_invalidations
+(
+    substate_id    text   not null primary key,
+    state_version  bigint not null,
+    invalidated_at bigint not null
+);
+
+create index idx_substate_cache_invalidations_expiry on substate_cache_invalidations (invalidated_at);
--- a/applications/tari_indexer/src/storage_sqlite/migrations/2026-09-04-000000_substate_cache_invalidation_version/up.sql
+++ b/applications/tari_indexer/src/storage_sqlite/migrations/2026-09-04-000000_substate_cache_invalidation_version/up.sql
@@ -1,0 +1,22 @@
+-- Record the substate version a transition showed, alongside the shard version it committed at.
+--
+-- The shard version answers "did this fetch start before the transition", which catches a fetch the
+-- transition overtook. It says nothing about a fetch that starts afterwards and lands on a committee
+-- member that is behind: that member answers with a version this indexer has already watched the
+-- substate pass, and with no cached row left to rank against - the transition deleted it - the
+-- answer is installed as the head. Holding the version the stream showed gives that write something
+-- to be refused by.
+--
+-- The journal is derived from the stream and spans a fetch, so it is recreated rather than migrated.
+drop table substate_cache_invalidations;
+
+create table substate_cache_invalidations
+(
+    substate_id      text    not null primary key,
+    state_version    bigint  not null,
+    -- The substate version the transition showed: the version created, or the version destroyed.
+    substate_version integer not null,
+    invalidated_at   bigint  not null
+);
+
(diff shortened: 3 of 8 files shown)


sdbondi (commented):
The guard itself is sound: reading the journal inside the write transaction, refusing `version < observed_version`, and leaving `DoesNotExist` to the ranking below are all right, and the `store_factory.rs` tests cover the stream path. One ordering problem undermines it on the retire-ahead path.

## Blocking

**The journal upsert overwrites `substate_version` unconditionally, so the floor regresses when a substate's destroy is journalled after its create** (`applications/tari_indexer/src/storage_sqlite/writer.rs:829-838`).

`retire_committed` (`applications/tari_indexer/src/substate_cache/mod.rs:145-151`) chains `up_iter` before `down_iter`. For the common case of one transaction that downs v6 and ups v7, that produces `created(7)` then `destroyed(6)`. The second row hits `on_conflict … do_update` and sets `substate_version = 6`. The floor is now 6, so a lagging member's `Up { v6 }` is admitted as the head. That is exactly the write this PR sets out to refuse, and it happens on the path whose caller is about to read what it just created. The window closes when the stream's own row arrives, but the description claims the guard holds for the full retention, and here it does not hold at all.

The stream path is only safe by the accident of transition order within a state version (Destroy before Create in `block.rs:852-861`; the rocksdb iterator order is not stated as a guarantee anywhere).

Fix either by making the upsert order-independent, which is the version I'd take:

```rust
.set((
    substate_cache_invalidations::state_version.eq(state_version.as_u64() as i64),
    substate_cache_invalidations::substate_version.eq(diesel::dsl::sql::<diesel::sql_types::Integer>(
        "max(substate_version, excluded.substate_version)",
    )),
    substate_cache_invalidations::invalidated_at.eq(now),
))
```

or, at minimum, by chaining `destroyed` before `created` in `retire_committed` with a comment stating that the journal takes the last write. A test that journals `created(7)` then `destroyed(6)` in one batch and asserts `put(6)` is refused would pin it either way.

## Non-blocking

**A destroy floor admits `Up` at the destroyed version.** The description's justification, "it is down, which is what a lookup for it answers", holds only when the member returns `Down { 7 }`. A member that is behind returns `Up { 7 }` with the live value, and with the floor at 7 and no cached row to outrank it, that is installed and served as the live head via `answer_at` for the TTL. Every test `put_entry` (`store_factory.rs:1097-1123`) builds a `SubstateResult::Down`, so `a_destroy_records_the_version_it_saw` cannot see this. The writer has `entry.substate_result` in hand and could refuse `Up` at exactly the destroyed version if the journal recorded that the row came from a destroy. Not a regression, and a destroy with no successor is the rarer case, but the PR text should at least not claim it is closed.

## Nits

- The "Requires data directory to be deleted" box is ticked, but the migration drops and recreates the table itself and the description says as much. It reads as "migration must run", which is every migration. Untick it or reword.
- `SUBSTATE_CACHE_JOURNAL_RETENTION` doc (`bootstrap.rs:111-117`) is clear; no change needed.

CI: integration test still pending, everything else green.

PR #2531, opened 2026-09-05
fix(indexer): retire cached substates a finalized result supersedes

sdbondi:
## Summary

Fixes the `transfer.feature` failures ("`Substate component_… does not exist`" on the second manifest submit) that started with #2505 — they fail on `development` at `68a9eca20` and passed on the commit before it; cucumber's retry keeps the job green so only "Test Results" shows them.

The race: `TransactionManager::get_transaction_result` answers from the **committee**, ahead of the indexer's own transition stream. The wallet had earlier asked whether the destination account existed (to decide on `create_account!`), which #2505 cached as `DoesNotExist`. It then learns the transfer committed, reads the new account, and gets the stale cached nonexistence. The same ordering bites a cached *old head* when a transaction replaces it.

## Change

- On a finalized accept, the transaction manager hands the result's `SubstateDiff` to `SqliteSubstateCache::retire_committed`, which applies the same `created`/`destroyed` invalidations the stream would.
- Each is journalled at `shard watermark + 1` (new `substate_cache_retire_ahead` store op) so a fetch captured before the commit cannot re-cache the stale value; the stream's own journal row replaces it once the transition arrives. A shard with no watermark can't be served from anyway, so its entries are only deleted.
- Failure to retire is logged, not surfaced: the result itself is still correct.

## Tests

`a_finalized_result_retires_what_it_created_ahead_of_the_stream`: cached nonexistence is gone after the result, a write captured at the old watermark is vetoed, one captured past it is accepted.

### DIFF
--- a/applications/tari_indexer/src/bootstrap.rs
+++ b/applications/tari_indexer/src/bootstrap.rs
@@ -314,7 +314,7 @@ pub async fn spawn_services(
         store.clone(),
         epoch_manager.clone(),
         validator_node_client_factory.clone(),
-        substate_cache,
+        substate_cache.clone(),
     )
     .with_substate_proof_verification(config.indexer.verify_substate_proofs)
     .with_negative_cache_ttl(config.indexer.substate_cache_negative_ttl);
@@ -430,6 +430,7 @@ pub async fn spawn_services(
         config.network,
         consensus_constants.max_transaction_weight,
         consensus_constants.max_transaction_validity_epochs,
+        substate_cache,
     );
 
     // Save final node identity after comms has initialized. This is required because the public_address can be
--- a/applications/tari_indexer/src/network_state_sync/worker.rs
+++ b/applications/tari_indexer/src/network_state_sync/worker.rs
@@ -1194,7 +1194,7 @@ fn extend_bufs_from_substate_update(
 ) -> Result<(), NetworkStateSyncError> {
     invalidations_buf.extend(match &update {
         SubstateUpdateProof::Create(create) => {
-            SubstateCacheInvalidation::created(create.substate.substate_id().clone(), create.substate.version)
+            SubstateCacheInvalidation::created(create.substate.substate_id(), create.substate.version)
         },
         SubstateUpdateProof::Destroy(destroy) => Some(SubstateCacheInvalidation::destroyed(
             destroy.substate_id.clone(),
--- a/applications/tari_indexer/src/storage_sqlite/models/substate_cache.rs
+++ b/applications/tari_indexer/src/storage_sqlite/models/substate_cache.rs
@@ -35,14 +35,14 @@ impl SubstateCacheInvalidation {
     /// retraction of a cached nonexistence. Where nothing caches that, it carries nothing, and is
     /// not one of these at all: emitting it would put a journal row on the sync path for every
     /// created-once substate in the stream and buy nothing with it.
-    pub fn created(substate_id: SubstateId, version: u32) -> Option<Self> {
+    pub fn created(substate_id: &SubstateId, version: u32) -> Option<Self> {
         let retires_up_to = version.checked_sub(1);
-        let retires_nonexistence = caches_nonexistence(&substate_id);
+        let retires_nonexistence = caches_nonexistence(substate_id);
         if retires_up_to.is_none() && !retires_nonexistence {
             return None;
         }
         Some(Self {
-            substate_id,
+            substate_id: substate_id.clone(),
             retires_up_to,
             retires_nonexistence,
         })
@@ -91,7 +91,7 @@ mod tests {
 
     #[test]
     fn a_first_creation_retires_nothing_but_the_nonexistence() {
-        let invalidation = SubstateCacheInvalidation::created(substate(), 0).unwrap();
+        let invalidation = SubstateCacheInvalidation::created(&substate(), 0).unwrap();
         assert_eq!(invalidation.retires_up_to(), None);
         assert!(invalidation.retires_nonexistence());
     }
@@ -101,17 +101,17 @@ mod tests {
     #[test]
     fn a_first_creation_is_dropped_where_nonexistence_is_not_cached() {
         assert!(!caches_nonexistence(&receipt()));
-        assert!(SubstateCacheInvalidation::created(receipt(), 0).is_none());
+        assert!(SubstateCacheInvalidation::created(&receipt(), 0).is_none());
         // Above the first version it carries a retirement that stands on its own, and does not
         // attempt one that could never match.
-        let above = SubstateCacheInvalidation::created(receipt(), 1).unwrap();
+        let above = SubstateCacheInvalidation::created(&receipt(), 1).unwrap();
         assert_eq!(above.retires_up_to(), Some(0));
         assert!(!above.retires_nonexistence());
     }
 
     #[test]
     fn a_creation_retires_every_version_below_it() {
-        let invalidation = SubstateCacheInvalidation::created(substate(), 6).unwrap();
+        let invalidation = SubstateCacheInvalidation::created(&substate(), 6).unwrap();
         assert_eq!(invalidation.retires_up_to(), Some(5));
         assert!(invalidation.retires_nonexistence());
     }
--- a/applications/tari_indexer/src/storage_sqlite/store_factory.rs
+++ b/applications/tari_indexer/src/storage_sqlite/store_factory.rs
@@ -1178,7 +1178,7 @@ mod tests {
         assert!(put_nonexistent(&store, &id, 100).await);
         assert!(is_nonexistent(&read_entry(&store, &id).await.unwrap()));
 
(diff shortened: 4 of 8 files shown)


sdbondi (commented):
The mechanism is sound: journalling at `watermark + 1` gives exactly the veto the fetch path checks (`invalidated_at_version > watermark` in `substate_cache_put`), the shard is the substate's own shard so the comparison is against the same watermark the fetch captured, and `confirm()` only ever raises a shard's version, so the stream's later `do_update` at its real state version can't walk the row backwards. `created(id, 0)` still carries `retires_nonexistence`, so the account case in the description really is covered. `Duration::MAX` in `ShardWatermarks::get` is safe (`elapsed() <= MAX`), and journalling an unconfirmed shard at `0` is consistent with `cached_substate_manager` skipping the write when `watermark()` is `None`.

Two things block.

## Blocking

**1. `accept()` skips fee-only accepts, which do commit a diff — `transaction_manager/mod.rs:181`**

```rust
finalized.execute_result.as_ref().and_then(|r| r.finalize.accept())
```

`TransactionResult::accept()` (`crates/engine_types/src/commit_result.rs:358`) matches `Accept(_)` only. `AcceptFeeRejectRest(diff, _)` is `Decision::Commit` (`crates/consensus_types/src/decision.rs:70`) and its `SubstateDiff` *is* committed — the fee withdrawal downs and ups the fee-paying account, and the receipt is created. So the exact staleness this PR fixes survives for every fee-rejected transaction: the caller is handed a commit, the account's cached head is superseded, and nothing retires it until the stream arrives.

Use `any_accept()`. Gating on `final_decision` is not needed on top of it — `TransactionExecution::decision()` is derived from `finalize.result`, so a diff being present already implies a commit.

**2. A write transaction on every read of any finalized result — `transaction_manager/mod.rs:176-189`**

`retire_committed` runs unconditionally on every `get_transaction_result`, which is a public REST endpoint (`rest_api/handlers/transactions.rs:231`, `/transactions/{id}/result`) that callers poll. Each call takes SQLite's single write lock and upserts one journal row per substate in the diff — receipts, vaults, fee pools, everything — for a commit that in the steady state was retired long ago.

That is the write pattern the guard 20 lines above deliberately avoids, with the reason spelled out in its own comment: *"Re-recording it on every read would needlessly take SQLite's single write lock and contend with other writers under load."* The new path does what that comment rules out, on every finalized result rather than only aborts.

It also re-vetoes: each poll re-journals at the *current* watermark + 1, so a substate that was evicted (or never re-cached) cannot be written back while polling continues, because every fetch captured at the current watermark is refused. The cached head itself survives — `retires_up_to = v - 1` doesn't match a row at `v` — so this is contention and cache-miss cost, not incorrectness, but it is unbounded in the number of callers.

Retiring once per finalized result is enough; the in-flight-fetch protection only matters at the moment this indexer first learns of the commit. A small in-memory set/LRU of already-retired transaction ids gives that. Note that the tempting cheap guard — "skip if the receipt is already in the local store" — is not equivalent: the receipt's shard and the created substate's shard are streamed independently, so the receipt can land while the account's shard is still behind.

## Non-blocking

**`substate_cache: Option<SqliteSubstateCache>` — `transaction_manager/mod.rs:49`, `bootstrap.rs:434`**

There is one construction site and it always calls `with_substate_cache`. An `Option` that is never `None` makes a correctness invariant look optional, and a future construction site silently reintroduces the bug this PR fixes. Make it a `new()` parameter. Separately, `TransactionManager` is generic over `TStore` but now names the concrete `SqliteSubstateCache`; if that genericity is meant to hold, `impl SubstateCache` is the matching bound.

**Test covers only the nonexistence case — `substate_cache.rs:483`**

The description claims "the same ordering bites a cached *old head* when a transaction replaces it", and that path goes through a different branch (`retires_up_to` and the `destroyed` invalidation, not `retires_nonexistence`). A second case — cache a head at `v`, `down(id, v)` + `up(id, v+1)`, assert the old head is gone and a write captured at the old watermark is refused — would pin the half the current test doesn't reach.

## Nits

- `substate_cache.rs:141` — `id.clone()` happens before `created()` decides to return `None`, so every transaction receipt in the diff pays for a discarded clone. `filter(|(id, _)| ...)` first, or clone inside.
- `store.rs:368` — the doc says the state version "is not yet known", but the caller does pass one; "is not the transition's own" reads closer to what it does.

sdbondi (commented):
Re-reviewed `4efe4ea..7f0d74d`. Both blockers are genuinely addressed.

**1 — fee-only accepts.** `any_accept()` at `transaction_manager/mod.rs:203` now covers `AcceptFeeRejectRest`, and gating on the diff's presence is still sufficient: `TransactionExecution::decision()` derives from `finalize.result`, so a diff implies `Decision::Commit`. Resolved.

**2 — write on every poll.** `first_sight_of_result` (`transaction_manager/mod.rs:93`) takes the write path only once per transaction. Checked: the guard drops before the `.await`, so no lock is held across it; `Arc` is shared across `Clone`s of the manager; poisoning is recovered the same way `ShardWatermarks` does; and the set is only touched when there is a diff, so rejects don't consume slots. Resolved.

**Also confirmed.** `substate_cache` is a `new()` parameter with the one call site updated (`bootstrap.rs:433`). The new `filter` is exactly the negation of `created`'s `None` condition (`version == 0 && !caches_nonexistence`), so nothing is dropped. `a_finalized_result_retires_the_head_it_replaced_ahead_of_the_stream` exercises the branch the first test didn't: `created(id, 7)` retires up to 6, which is the cached head, and the 100/101 pair pins the veto boundary. `SubstateCacheEntryRef` is `Copy`, so reusing `entry` across the two writes is fine.

## One new item (non-blocking)

`transaction_manager/mod.rs:199-208` — the id is recorded *before* the retire runs, so a failed retire is never retried:

```rust
self.first_sight_of_result(transaction_id) &&
    let Err(e) = self.substate_cache.retire_committed(diff).await
```

On `Err` the warning is logged and the id stays in the set, so every later poll short-circuits. A `SQLITE_BUSY` on that write — which is likeliest under exactly the load this guard exists to reduce — leaves the stale entry in place with no second attempt, which is the bug this PR fixes.

The same ordering also narrows the guarantee for concurrent pollers: a second caller arriving while the first's retire is in flight sees `false`, returns the result immediately, and can read the substate before the retirement commits.

Both go away by making it check-then-act: test membership before, insert after a successful retire. Concurrent callers then each run their own (idempotent) retire and await it, and a failure leaves the id absent so the next poll tries again.

## Nit

`substate_cache.rs:151` — the "version 0 and nothing caches its nonexistence" rule now lives both in the `filter` and inside `SubstateCacheInvalidation::created`, and a future change to `created` would silently drop invalidations here. My clone nit wasn't worth that; `created` taking `&SubstateId` and cloning internally would keep the rule in one place. Fine to leave.

Holding off on sign-off until `clippy` / `test (1-3)` / `integration test` report — they're still pending.

sdbondi (commented):
Re-reviewed `7f0d74d..3b074dc`. Both follow-up items are resolved and CI is fully green — signing off. (GitHub won't take a formal approval from me on my own PR, so this is a comment; it still needs another maintainer.)

**Retry / concurrent-poll window.** `is_result_retired` before, `mark_result_retired` only on `Ok` (`transaction_manager/mod.rs:93-108`, `:210-219`). A failed retire now leaves the id absent and is retried on the next poll, and a second poller arriving mid-flight sees `false`, runs its own idempotent retire and awaits it. Both lock guards still drop before the `.await`.

**Duplicated rule.** `created` takes `&SubstateId` and clones internally, so the "version 0 and nothing caches its nonexistence" decision lives in one place again and the parallel `filter` is gone. Checked all 13 call sites; `create.substate.substate_id()` at `network_state_sync/worker.rs:1197` already returned `&SubstateId`, so dropping the `.clone()` is type-correct and the borrow outlives the `extend`.

**The fix is confirmed by CI, not just by the unit tests.** "Test Results (CI)" on `development` at `68a9eca2` reports exactly the two scenarios this PR targets:

```
Scenario: Transfer tokens to account that does not previously exist ... failed
Scenario: Transfer tokens to existing account ....................... failed
```

Neither appears on this head. The one failure that showed up mid-run, `state_tree::check_state_transitions (consensus_tests)`, is a pre-existing flake — it fails on `development` at `5faf1b5f` and `a98d679e`, in code this PR does not touch — and it passed on the rerun, so every check is now green including "Test Results (CI)".

**Verified across the three heads:** journalling at `watermark + 1` matches the veto `substate_cache_put` actually applies; the shard is the substate's own, so it is compared against the watermark the fetch captured; `confirm()` only raises a version, so the stream's later `do_update` cannot walk the row back; `any_accept()` covers fee-only commits and a diff's presence already implies `Decision::Commit`; unconfirmed shards journal at `0`, which is consistent with `cached_substate_manager` skipping the write when `watermark()` is `None`; and the two tests pin both branches — the retracted nonexistence and the replaced head — at the 100/101 veto boundary.

PR #2532, opened 2026-09-05
docs(skills): add OpenCode agent skill with corrected fee and size guidance

m4r1m0:
## Summary

Adds `docs/skills/opencode/SKILL.md` — an OpenCode adaptation of the existing `claude-code` skill (frontmatter + tool-name mentions), with stale guidance corrected based on community review of a production template (confidential-rcv-template):

- **Fees:** replaces the hardcoded publish fee guidance ("150k–250k units") with dry-run fee estimation — `build -> dry-run -> read estimated fee -> submit with margin`. Publishing a naive ~300 KB template actually cost ~7.8M fee units; fee amounts are sats-like integer units.
- **WASM size:** documents `wasm-opt -Oz --enable-bulk-memory` minification before publishing and a ~250 KB size target; notes that dual `cdylib` + `rlib` crate types silently drop `-C lto` (~76 KB bloat) and that pure logic should live in a standalone crate instead.
- **Versions:** bumps `tari_template_lib` to 0.31 and `tari_template_test_tooling` to 0.40, and notes the tooling crate re-exports `tari_ootle_transaction` (no separate dev-dependency needed).
- **Publish workflow:** Web UI remains the easiest path for one-off publishing; client apps may publish programmatically when using dry-run fee estimation, never hardcoded fees.

Also lists `opencode/` in the `docs/skills/README.md` table and adds a manual-install section for OpenCode.

## Notes

- The `claude-code/` skill is left untouched; this adds a parallel skill for OpenCode (same content lineage, corrected guidance).
- Local harness copies of the skill can be updated from this file.

### DIFF
--- a/docs/skills/README.md
+++ b/docs/skills/README.md
@@ -19,6 +19,7 @@ AI agents can automatically discover and load skills from this endpoint. No manu
 | Directory | Agent | Description |
 |-----------|-------|-------------|
 | [`claude-code/`](claude-code/) | [Claude Code](https://claude.ai) | Claude Code / Anthropic Claude |
+| [`opencode/`](opencode/) | [OpenCode](https://opencode.ai) | OpenCode CLI agent |
 | [`cursor/`](cursor/) | [Cursor](https://cursor.com) | Cursor AI editor |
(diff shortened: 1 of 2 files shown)


sdbondi (approved):
Thanks :rocket:

PR #2538, opened 2026-09-07
feat(indexer): export a gauge of active SSE connections

sdbondi:
## Description

The indexer has no metric for how many streaming connections it is serving, and nothing existing stands in for one:

- `api_http_requests_pending` decrements as soon as `next.run(req)` returns, which for a stream is before a single event has been sent. An open stream counts as zero pending requests.
- `api_http_response_time_seconds` observes the same interval, so a stream's observation is handler setup time, not connection lifetime.

This matters because moving the wallet daemon off its poll loop and onto the `/events` SSE stream would make concurrent stream count the indexer's principal load signal from wallets.

## Motivation and Context

Adds `api_sse_connections_active`, a gauge labelled by endpoint. `/events`, `/transactions/events/stream` and `/utxos/stream` all route through the same `SseConnectionLimiter` but serve quite different consumers, so a single number would blur them. Label values are the fixed route paths chosen at router construction and are never taken from the request, so cardinality is bounded by the number of streaming routes.

The count is maintained by `sse_limit_middleware` next to the per-IP connection slot, carried in the response body's stream state rather than the response extensions, so it falls when the stream is fully drained or the client disconnects.

It is kept independently of `SseLimitConfig::enabled`. The middleware previously early-returned before `try_acquire` when limiting was off; a gauge hung off the slot guard would then read zero on exactly the node an operator is most likely to be debugging. The middleware now takes no slot in that case but still wraps the body and counts the stream.

Only a successful response counts. The streaming handlers validate their arguments before they open a stream, and `stream_utxo_updates` has eight early rejections, so counting before the handler ran would let a client sending bad arguments in a loop show up as load. A rejection passes through untouched, keeping its `Content-Length` and its place in the response-size histogram.

Registration moved to the top of `Server::spawn` so one `api` sub-registry produces both the request metrics and the connection gauge, and each streaming route gets its own `SseLimitConfig` carrying that route's gauge handle while sharing one limiter.

Also registers `response_body_size_histogram`, which was constructed with `Histogram::new` and observed on every response but never registered, so it was exported nowhere. It is now `api_http_response_body_size_bytes`.

## New metrics

Both are exported on the metrics listener (`metrics_listen_address`), not the public REST API, and both are behind the default-on `metrics` feature.

| Metric | Type | Labels | Meaning |
| --- | --- | --- | --- |
| `api_sse_connections_active` | Gauge | `endpoint` | Streaming connections currently being served |
| `api_http_response_body_size_bytes` | Histogram | none | Response body size, for responses of known length |

`endpoint` takes one of three values, one per streaming route:

- `/events` — indexer event stream
- `/transactions/events/stream` — transaction event stream
- `/utxos/stream` — UTXO update stream

`api_http_response_body_size_bytes` uses exponential buckets from 100 B, doubling, 15 buckets. Streaming responses have no known length and so are not observed.

## How Has This Been Tested?

`cargo test -p tari_indexer --lib` — 120 passed.

Four new middleware tests drive a real server over TCP: counting a stream until the client disconnects, counting when the limiter is disabled, not counting a rejected 429, and per-endpoint isolation. One new metrics test asserts every registered name appears in the exposition.

Both fixes were confirmed to be covered:

- the disabled-limiter test fails against the naive implementation that counts inside the enabled branch (`gauge settled at 0, expected 1`);
- the exposition test fails when the body-size histogram is left unregistered (`api_http_response_body_size_bytes missing`).

`cargo lints clippy -p tari_indexer --all-targets` is clean, and `cargo check -p tari_indexer --all-targets --no-default-features` confirms the metrics-free build still compiles.

Run an indexer with metrics enabled, open a stream against `/events`, and scrape `/_metrics` on the metrics listener. `api_sse_connections_active{endpoint="/events"}` should read 1 while the stream is open and return to 0 once the client goes away. Repeat with `rate_limits.enabled = false` to check the count is still reported.

## Breaking Changes

- [x] None
- [ ] Other

### DIFF
--- a/applications/tari_indexer/src/rest_api/metrics.rs
+++ b/applications/tari_indexer/src/rest_api/metrics.rs
@@ -25,7 +25,7 @@ use prometheus_client::{
 use tari_ootle_app_utilities::tcp::try_bind_with_fallback;
 use tari_shutdown::ShutdownSignal;
 
-use crate::metrics::CollectorRegister;
+use crate::{metrics::CollectorRegister, rest_api::rate_limit::SseConnectionMetrics};
 
 const LOG_TARGET: &str = "tari::ootle::indexer::rest_api::metrics";
 
@@ -71,27 +71,44 @@ pub struct RequestMetrics {
     response_body_size_histogram: Histogram,
 }
 
-pub fn register(registry: &mut Registry) -> RequestMetrics {
+/// Everything the REST API exports, all of it under the `api` prefix.
+#[derive(Clone)]
+pub struct ApiMetrics {
+    /// State for the [`layer`] middleware wrapping the whole router.
+    pub requests: RequestMetrics,
+    /// Shared with the SSE limit middleware, which maintains the count.
+    pub sse: SseConnectionMetrics,
+}
+
+pub fn register(registry: &mut Registry) -> ApiMetrics {
     let registry = registry.sub_registry_with_prefix("api");
 
-    RequestMetrics {
-        request_counter: Counter::default().register_at(
-            "http_requests_total",
-            "Total number of HTTP requests received",
-            registry,
-        ),
-        response_time_histogram: Histogram::new(
-            exponential_buckets(0.001, 2.0, 15), // buckets from 1ms, doubling, 15 buckets
-        )
-        .register_at("http_response_time_seconds", "HTTP response times in seconds", registry),
-        requests_pending: Gauge::default().register_at(
-            "http_requests_pending",
-            "Number of HTTP requests currently being processed",
-            registry,
-        ),
-        response_body_size_histogram: Histogram::new(
-            exponential_buckets(100.0, 2.0, 15), // buckets from 100B, doubling, 15 buckets
-        ),
+    ApiMetrics {
+        requests: RequestMetrics {
+            request_counter: Counter::default().register_at(
+                "http_requests_total",
+                "Total number of HTTP requests received",
+                registry,
+            ),
+            response_time_histogram: Histogram::new(
+                exponential_buckets(0.001, 2.0, 15), // buckets from 1ms, doubling, 15 buckets
+            )
+            .register_at("http_response_time_seconds", "HTTP response times in seconds", registry),
+            requests_pending: Gauge::default().register_at(
+                "http_requests_pending",
+                "Number of HTTP requests currently being processed",
+                registry,
+            ),
+            response_body_size_histogram: Histogram::new(
+                exponential_buckets(100.0, 2.0, 15), // buckets from 100B, doubling, 15 buckets
+            )
+            .register_at(
+                "http_response_body_size_bytes",
+                "HTTP response body sizes in bytes, for responses of known length",
+                registry,
+            ),
+        },
+        sse: SseConnectionMetrics::register(registry),
     }
 }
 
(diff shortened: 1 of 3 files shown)


sdbondi (commented):
Reviewed at `368d93b`. Nothing blocking.

## Verified sound

- The gauge token lives in the unfold state next to `SseConnectionGuard`, so both drop together when the stream drains or the peer vanishes. `futures::stream::unfold` moves the state into the per-poll `async move`, so dropping the body mid-poll drops it too — same lifetime the slot guard already relied on.
- The 429 path returns before `active_connections.open()` (`rate_limit.rs:459-474`), and `it_does_not_count_a_rejected_connection` pins it. Note it only passes because `extract_ip`'s no-`ConnectInfo` fallback is `127.0.0.1` and the test serves with `into_make_service()` — worth a line of comment, since a change to that fallback silently turns the test into a no-op assertion.
- Hoisting `SseConnectionLimiter` out and cloning it per route preserves the previous semantics exactly: `Clone` shares the `Arc<DashMap>` and copies `max_per_ip`, which is what the old `sse_limiter.clone()` did too. Per-IP limits still span all three routes.
- Label values are `&'static str` fixed at router construction, cardinality bounded at 3, and the three strings match the actual nested paths (`/transactions` + `/events/stream`, `/utxos` + `/stream`, top-level `/events`).
- `Family::get_or_create_owned` returns a clone of the child `Gauge`, and `Gauge` is `Arc<AtomicI64>` — so the handle in the middleware and the child the registry encodes are the same counter. `Family` clone shares its `Arc<RwLock<HashMap>>`, so `ApiMetrics::sse.clone()` is safe.
- Registering `response_body_size_histogram` is a real fix: it was constructed, observed on every response, and exported nowhere.
- The reasoning for counting independently of `SseLimitConfig::enabled` is right, and the description's claims about `api_http_requests_pending` and `api_http_response_time_seconds` check out against `metrics::layer` — both stop at the moment the handler returns headers.

## Non-blocking

**1. Error responses on the streaming routes get counted as connections.** `rate_limit.rs:477-478` opens the gauge before `next.run(req)`, and everything the handler returns is then wrapped. `stream_utxo_updates` has eight `ErrorResponse::bad_request` early returns (`handlers/utxos.rs:49,57,61,72,88,101,135,157`), so a client sending `per_shard_limit: 0` in a loop shows up in `api_sse_connections_active{endpoint="/utxos/stream"}` as streams. It self-corrects once the small JSON body is drained, so this is inaccuracy rather than a leak, but the gauge is meant to be the load signal and a 400 is not load.

Dropping `slot`/`open` and returning the response untouched when `!response.status().is_success()` fixes it and also addresses (2).

**2. The unconditional re-wrap keeps those same responses out of the histogram this PR just registered.** `metrics::layer` only observes `response.size_hint().exact()`, and `Body::from_stream` has no exact hint. Previously, on a node with `rate_limits.enabled = false`, an error response from these three routes passed through untouched and *was* observed; now it is not. It also means those small error bodies go out chunk-encoded rather than with `Content-Length`. Harmless to clients, but it is a behaviour change on the disabled-limiter path that the description doesn't mention (it only claims "takes no slot but still wraps the body and counts the stream").

**3. `IntervalStream` compiles by accident of feature unification.** `rate_limit.rs:733` uses `tokio_stream::wrappers::IntervalStream`, which is behind tokio-stream's `time` feature; the workspace dep at `Cargo.toml:277` declares only `features = ["sync"]`. `cargo tree -p tari_indexer -e features` shows `tokio-stream feature "time"` arriving via tonic's defaults. It builds today, but the new test breaks the moment that transitive path changes. Add `"time"` to the workspace features, or build the ticker from `tokio::time::interval` directly.

## Nits

**4.** `metrics.rs:202` — `metrics.sse.endpoint("/events")` isn't doing what it looks like it's doing. `Registry::encode` writes the `# HELP`/`# TYPE` descriptor for every registered collector before calling `encode` on it, so a `Family` with no children still puts `api_sse_connections_active` in the exposition and the assertion passes without this line. Fine to keep, but it reads as a precondition when it isn't — and it means the test can't distinguish "registered" from "registered and populated".

**5.** `rate_limit.rs:271,304` — `Default` is derived on both metric types regardless of feature, so with `metrics` on it hands out an unregistered gauge that increments happily and exports nowhere. That is exactly the failure mode the "kept independently of `enabled`" reasoning is guarding against. The derive is only needed for the `#[cfg(not(feature = "metrics"))]` branch in `server.rs:103`; gating it that way would push the tests onto `SseConnectionMetrics::register(&mut Registry::default())`, which is no harder to write.

PR #2539, opened 2026-09-07
feat(walletd): drive transaction finality from the indexer event stream

sdbondi:
## Summary

The wallet's transaction service learned of finality only by polling the indexer for every pending transaction every 5 seconds, so a commit was noticed up to 5 seconds after the indexer knew about it. It now subscribes to the indexer's `/events` SSE stream and queries a transaction's result the moment a `TransactionFinalized` notification names it. Same strategy `ootle-rs` already uses in `PendingTransaction::watch()`.

## Aborts

The stream only carries transactions that committed a receipt: an aborted transaction writes no substate and is silent. The 5s poll stays as a backstop, but while the stream is connected it only queries transactions that have been pending longer than `silent_transaction_timeout` (10s). So commits are reported as soon as the indexer's stream delivers the receipt, and an abort is noticed within ~10s instead of ~5s, at a third of the previous query load. The first check 750ms after submission is kept and always queries directly, so an abort decided that quickly is still caught immediately.

While the stream is down every poll queries every pending transaction (previous behaviour), and each (re)connection requests a full check for anything finalized while disconnected. `/events` has no replay, so this is what covers a dropped connection. A full check that cannot run because a poll is already in flight stays requested until one does.

## Substate cache

The `TransactionFinalized` event is emitted while the indexer buffers the receipt, before the batch is persisted, so a wallet acting on it can query before the indexer's own write lands. That is safe because the result query is the same `get_transaction_result` path as before: it asks the committee first and calls `retire_committed` on the accepted diff itself, so the answer does not depend on the local receipt row and the cache reconciliation from #2531 applies unchanged. No new path hands a client a commit.

## Changes

- `WalletNetworkInterface::subscribe_transaction_finalized` and `TransactionFinalizedStream`; the indexer implementation filters `/events` by event name and skips an undecodable payload rather than ending the subscription.
- `IndexerEvent::{NEW_EPOCH,TRANSACTION_FINALIZED}_EVENT_NAME` constants so the client and the SSE handler share the names.
- `FinalizedWatch`: subscription task with fixed-backoff reconnect, spawned from `run()` with the shutdown signal, feeding the service `Connected` / `Finalized` / `Disconnected` events.
- The service keeps the set of transactions it is waiting on in memory (seeded from the store, maintained from wallet events) and filters the network-wide stream against it before touching the store.
- `TransactionServiceConfig` exposes the poll interval, post-submit delay, silent timeout and reconnect backoff. Defaults match the values above; `TransactionService::new` is unchanged.
- Tests in `sdk_tests/tests/transaction_service.rs` cover: a notification resolves a transaction without waiting for the poll (and foreign notifications are ignored), an abort is reported with no notification, and every pending transaction is polled while the stream is down. The first fails without the notification handling.

`tari_ootle_wallet_sdk` gains a required trait method, so the next release needs the breaking bump `crate_versioning.py impact tari_ootle_wallet_sdk --breaking` reports; that is left for the release chore commit as per convention.


sdbondi (commented):
Reviewed at `16f5977`. The state machine is the right shape: `FinalizedWatch` covers all three ways the subscription dies (stream `None`, stream `Err`, `subscribe` `Err`) and each reconnect triggers a catch-up, which is exactly what the indexer's `.take_while(|res| res.is_ok())` on the broadcast stream (`indexer_events.rs:22`) needs — a lagging client's SSE stream is simply ended, with no replay. `check_and_store_finalized_transaction` returns `Ok(None)` for a still-pending result without touching the row (`apis/transaction.rs:244`), so `updated_at` is not refreshed by a failed check and `CheckScope::PendingFor` measures real age from submission rather than from the last query — that's the load-bearing detail and it holds. The double `is_pending_in_wallet` check around the permit correctly prevents a duplicate `TransactionFinalizedEvent`, since `check_and_store_finalized_transaction` re-returns an already-stored result and would re-notify.

## Blocking

**A dropped `on_poll` silently costs the transaction 30s** — `service.rs:207-213` + `service.rs:311-317`

`check_all_on_next_tick` is cleared unconditionally on the tick, but `on_poll` acquires the poll permit with `try_acquire_owned` and returns `Ok(())` without doing anything if a poll is already in flight. So:

- `T`: transaction submitted, `TransactionSubmitted` → `trigger_poll` → `check_all_on_next_tick = true`, interval reset to 750ms.
- `T+750ms`: tick, scope `All`, flag cleared, `on_poll` finds the permit held by the previous 5s poll (which queries the committee once per pending transaction, serially — >750ms is unremarkable with a few pending) and drops the check.
- `T+5.75s` and every tick after: scope is `PendingFor(30s)`, so the transaction is skipped.
- The commit notification does arrive on the stream, so in practice this is only fatal for a transaction whose notification is also lost — but on the abort path there is no notification at all, and this is now the *only* early check. An abort that would previously have been reported at ~5s is reported at ~30s.

`WatchEvent::Connected` at `service.rs:186-190` has the same hole and it matters more there: that `on_poll(CheckScope::All)` is the sole mechanism covering the reconnect gap (`/events` has no replay), and if it is swallowed the next tick is back to `PendingFor(30s)`.

`on_poll` should report whether it actually ran, and the flag should only be cleared when it did:

```rust
// on_poll -> Result<bool, _>, false when the permit was unavailable
if !self.on_poll(scope)? {
    self.check_all_on_next_tick = true;
}
```

and the `Connected` arm should set `check_all_on_next_tick = true` rather than relying on its own `on_poll` landing.

## Non-blocking

**The description's ordering claim is false** — `network_state_sync/worker.rs:1257` vs `:950`

> the SSE event is emitted after the indexer's own stream has written the receipt

It isn't. `extend_bufs_from_substate_update` notifies while *buffering* the update (`:1257`); the receipt is not persisted until `batch_insert_transaction_receipts` inside the write tx at `:950`, which only runs when the batch flushes (`has_more == false`). A wallet that queries the instant it sees the event can beat the indexer's own write.

The conclusion still holds, but for a different reason than the one given: `TransactionManager::get_transaction_result` asks the committee first (`transaction_manager/mod.rs:169-178`) and does its own `substate_cache.retire_committed` on the accepted diff (`:207-216`), so the answer doesn't depend on the local receipt row and the cache is reconciled on that path regardless. Worth correcting in the description, since the stated premise is what the "#2531 is unchanged" argument rests on.

**30s abort latency, and `TransactionServiceConfig` isn't reachable** — `service.rs:56,67`

`TransactionService::new` is the only constructor any production caller uses (`services/mod.rs:50`, plus two in `handlers/accounts.rs`), so `silent_transaction_timeout` is fixed at 30s and the new config struct is effectively test-only. Input-lock aborts are a common and already-frustrating path, and `transactions.wait_result` callers (CLI, cucumber) now sit on them for 30s instead of 5s. Either wire the config through `WalletDaemonConfig`, or pick a smaller default — the cost being traded away is one committee query per pending transaction per 5s, which at 10s instead of 30s is still a 3x reduction with a third of the regression.

**Construction has a network side effect** — `service.rs:135`

`FinalizedWatch::spawn` in `with_config` means building the service opens an SSE connection and requires a Tokio runtime. The two `TransactionService::new` calls in `handlers/accounts.rs` (1973, 2258) exist only to get a handle and point at `http://127.0.0.1:18300` with nothing listening, so those unit tests now run a connection-refused retry loop for their whole duration. Spawning from `run()` keeps `new` pure and ties the task's lifetime to the thing that has the shutdown signal.

**One spawn + one blocking SQLite read per network-wide finalization** — `service.rs:343-352`

The stream carries every transaction the network commits, and `on_transaction_finalized` spawns a task and runs a synchronous `transactions_get` for each one, on a runtime worker thread, to discover that ~all of them are irrelevant. This is the one part of the design that scales with network throughput rather than wallet activity. The service already sees every submission and every finalization, so it could keep the pending ids in memory and filter in the `select!` arm before spawning anything.

**A single undecodable event tears down the subscription** — `indexer_rest_api.rs:215`

`try_filter_map` returning `Err` makes `FinalizedWatch` treat one malformed `TransactionFinalized` payload as a disconnect and reconnect. Recovery is correct, but a `warn!` and `Ok(None)` would skip the bad event without dropping the connection — the outer `map_err` on `:207` still handles genuine transport errors.

## Nits

- `finalized_watch.rs:51` — the task has no shutdown signal; it exits only when it next tries to send and finds the receiver gone, so after `run()` returns it holds the SSE connection open until either an event arrives or the reconnect sleep expires.
- `tari_ootle_wallet_sdk` is published (0.43.0) and a required method on the public `WalletNetworkInterface` trait is breaking for any external implementor. `scripts/crate_versioning.py impact tari_ootle_wallet_sdk --breaking` wants 0.44.0 for it plus `tari_ootle_wallet_storage_sqlite` and `tari_ootle_walletd_client`. Repo convention does this in a separate chore commit, so just flagging it for the next release.
- `crates/wallet/sdk_tests/Cargo.toml` — the last line is rewritten and still has no trailing newline.

sdbondi (commented):
Re-reviewed `a2878b0c3` and `c86bac6e7`. Everything from the first pass is addressed; two notes on the vault commit, neither blocking.

## `a2878b0c3` — verified

- `on_poll` now returns whether it started (`service.rs:334`), the tick clears `check_all_on_next_tick` only when it did (`service.rs:234`), and the `Connected` arm re-arms it when the permit was busy (`service.rs:203`). That closes both holes.
- The `pending` set is a better answer than the in-loop filter I suggested. I checked that it can't miss a transaction: `handle_submit_transaction` and `resubmit_new_transactions` are the only two paths that move a transaction to `Pending` (the only other `submit_transaction` caller is `tariswap_test_bench`, which has no service), and both emit `TransactionSubmitted`; `TransactionFinalized`/`TransactionInvalid` are emitted on every terminal path including aborts, which go out as `TransactionInvalid` because `execution_result` is `None`. A `RecvError::Lagged` on the notify subscription would skip an insert, but that already skipped the `trigger_poll` before this PR and the `PendingFor` backstop covers it.
- `FinalizedWatch` moved into `run()` and given the shutdown signal. The `select!` cancels `next_event()` only on the break path, so the half-assigned `Reconnecting` state can't be observed.
- Undecodable payload skipped; 10s silent timeout; description corrected.

## `c86bac6e7` — correct, but two things

The change itself is right. `tari_template_lib::types::Amount` is the unsigned 128-bit `Amount`, so `Amount::MIN` is zero and `saturating_sub` clamps there — the doc comment on it referring to `saturating_sub_positive` is a leftover from the signed `PrecisionAmount` and doesn't apply. Both callers get zero rather than anything negative, and the new test holds.

**It's unrelated to this PR.** Nothing else in the diff touches vault balances, and the PR title and body are entirely about the finalization stream. If the panic was hit while testing this change it would still land more cleanly as its own commit on `development` — it's a user-visible crash fix that shouldn't have to wait for this to merge.

**`confidential_transfer.rs:147` has the same class of bug, unfixed.** `PreferRevealed` there computes `cmp::min(src_vault.revealed_balance, spend_amount)` against the *raw* balance while the stealth equivalent (`stealth_transfer/api.rs:193`) uses `available_revealed_balance()`. The confidential path will therefore happily plan to spend revealed funds another transaction already has locked, and then `lock_funds_in_vault` stacks a second lock on top. Pre-existing and out of scope, but it's the sibling of the line you just touched.

## Nit

Masking `locked > revealed` means a genuinely leaked lock now reports zero spendable with no signal at all, where before it was a loud (if unacceptable) panic. `clear_stale_locks` is the recovery path, but a `warn!` when the subtraction saturates would make the difference between "transient, resolves in a second" and "this account is stuck" visible in the logs.

sdbondi (commented):
CI is green on `de32d94d4` — clippy, all three test shards, the 1h16m integration suite, ledger app, fmt, machete. Every finding from both reviews is resolved.

Verified on this head:

- `on_poll` reports whether it started (`service.rs:334`); the tick clears `check_all_on_next_tick` only when a poll ran (`service.rs:234`) and the `Connected` arm re-arms it otherwise (`service.rs:203`). A submission's full check and the reconnect catch-up can no longer be swallowed by a poll already holding the permit.
- The `pending` set cannot miss a transaction: `handle_submit_transaction` and `resubmit_new_transactions` are the only paths that move a transaction to `Pending` and both emit `TransactionSubmitted`; every terminal path emits `TransactionFinalized` or `TransactionInvalid`, aborts included. A notify lag degrades to the `PendingFor` backstop, which is the designed fallback.
- `FinalizedWatch` spawns from `run()` with the shutdown signal; the `select!` only cancels `next_event()` on the break path, so the half-assigned `Reconnecting` state is unobservable. An undecodable payload is skipped instead of ending the subscription.
- `available_revealed_balance` uses `checked_sub` and warns with the vault id and both amounts before returning zero. `Amount` here is the unsigned 128-bit type, so the zero is a real floor and neither caller can see a negative.
- `confidential_transfer.rs:148` now nets out locks like the stealth path. `available <= revealed_balance` keeps `confidential_to_spend` from underflowing.

Not approving only because GitHub won't take an approval from the PR author's own account — this needs a second maintainer. Nothing blocking from my side.

One thing left for the release, not this PR: `tari_ootle_wallet_sdk` is published at 0.43.0 and the new required method on `WalletNetworkInterface` is breaking. `scripts/crate_versioning.py impact tari_ootle_wallet_sdk --breaking` wants 0.44.0 for it plus `tari_ootle_wallet_storage_sqlite` and `tari_ootle_walletd_client`, which by repo convention lands as its own chore commit.

PR #2540, opened 2026-09-07
feat(vn): give the validator node a chosen memory ceiling

sdbondi:
## Why

The node's derived memory ceiling was 3.2 GiB, of which ~1.4 GiB was RocksDB and libp2p defaults — a number that fell out of the libraries rather than one anybody chose.

RocksDB was opened with neither a write buffer size nor a shared block cache, so each of the 9 column families got its own 64 MiB × 2 memtable budget and its own 32 MiB internal cache (1152 + 288 MiB). Gossipsub's per-connection send queue was libp2p's default 5000 messages which, against a 2 MiB max message size, is not a bound in any useful sense.

This makes the ceiling a number someone picked.

## What changed

**One RocksDB budget** (`crates/state_store_rocksdb`). Every column family shares a single `Cache`, and a `WriteBufferManager` charges memtable memory to that same cache, so one capacity bounds reads and writes together. Defaults: 768 MiB total, half of it memtables, 32 MiB per-family write buffers, `max_write_buffer_number` set explicitly at 2. Settable via `state_store_memory_budget_bytes`.

Two decisions worth reviewing:

- `allow_stall: false`. Over budget triggers earlier flushes rather than blocking writers. **The bound is therefore soft**: memtable memory can overshoot by as much as the buffers already in flight — `write_buffer_bytes × max_write_buffer_number × cf_count` = 576 MiB — and an LRU cache admits before it evicts. The budget line prints that figure. `allow_stall: true` would make it hard, but a writer blocked inside a block commit can miss a proposal, and `missed_proposal_suspend_threshold` is 5; a memory bound must not be able to cost liveness. Choosing `true` needs commit-latency measurements first.
- `db_write_buffer_size` is deliberately not set. RocksDB's `options.h` says a write buffer manager supersedes it, so setting it as a backstop would be dead configuration.

**Gossipsub bounds set explicitly** (`networking/swarm`). The per-connection send queue drops from 5000 to 256 — `RpcOut::Forward` holds a full `RawMessage` per peer, so this is the real memory term. `history_length` (5), `history_gossip` (3) and `duplicate_cache_time` (60s) keep libp2p's values but are now set in our config with doc comments saying what each trades away; `history_length` in particular is the deadline for reporting a validation verdict under `validate_messages`, so it is a liveness setting and not only a memory one.

**A byte cap at ingress, and the gossip limit derived from it** (`crates/transaction_validation`, `crates/p2p`). `max_transaction_weight` bounds work, not bytes: blob payloads are charged at a divisor and `Blobs` is a transaction-level list, so a transaction could carry ~2.8 MiB and still be admitted. That left the gossip limit nothing to be derived from, and 2 MiB has in fact been *below* what ingress accepts since the template binary limit dropped to 1.5 MiB — a transaction every validator accepts could fail to gossip, as a codec frame error rather than a per-message drop.

`ConsensusConstants::max_transaction_size_bytes` (1.75 MiB = template binary limit + 256 KiB envelope) bounds bytes directly, enforced by `TransactionSizeValidator` at ingress. The measure is `Transaction::encoded_size` — the canonical CBOR length, which is exactly what the p2p message carries in `bor_encoded` and what the state store persists, so every node computes the same number. `max_gossip_message_size` is that cap plus protobuf framing, and both the VN and indexer set their gossip limit from it: a transaction ingress admits is one the mesh will carry, by construction.

Gossip messages drop from 2 MiB to 1.77 MiB. Every gossipsub bound is a message count multiplied by this, so it tightens the memory model and the queue-flood worst case together.

The envelope allowance is held honest by building a real max-size template publish and measuring it (measured overhead: 132 bytes against 256 KiB), and a second test asserts the byte cap binds before the weight cap so it cannot become dead configuration.

**This narrows a wire parameter.** 1.77 MiB is below the 2 MiB it replaces, so during a mixed rollout a node on the new limit cannot decode a frame an unupgraded peer will relay. That is safe here only because every validator and indexer on these networks is upgraded together; the condition and what a deployment without lockstep upgrades must do instead are recorded on `max_gossip_message_size` rather than left to this description.

**Metrics** (`state_store_metrics.rs`). `StateStoreMemoryCollector` reports the budget's capacity and live usage at scrape time, mirroring `InboundQueueCollector`. This is what turns the ceiling from a one-time calculation into a continuously checked invariant.

Reading the budget handles directly rather than RocksDB's `cur-size-all-mem-tables` / `block-cache-usage` properties, because `TransactionDB` in the `rocksdb` crate exposes only default-CF properties and the store uses none — the properties would have read zero. The `Cache` and `WriteBufferManager` report the global figures directly.

**Startup check** (`memory_budget.rs`). Sums the caps the node enforces, logs the table, warns when `MemAvailable` is below them plus headroom:

```
Memory budget:
     256 MiB  Consensus gossip queue
     128 MiB  Transaction gossip queue
     128 MiB  Consensus messaging queue
     768 MiB  State store block cache and memtables
     200 MiB  Compiled template module cache
      64 MiB  Mempool dedup cache
      20 MiB  WASM instance memory at maximum call depth
     1.5 GiB  enforced caps
     2.3 GiB  required, including headroom
```

It warns rather than refuses. The caps are burst-and-attack ceilings, not steady state, so refusing would stop an undersized node that has been serving fine from coming back after a restart — a worse failure than the one it prevents. The comparison needs `/proc/meminfo`; on the Windows and macOS builds the table is still logged and a line says the comparison was skipped.

**Bench** (`tari-vn-bench`) reads the RocksDB, swarm and template defaults from the crates that own them. The three queue caps remain literals — they live behind private defaults in the node binary's own config module. The node keeps a second, shorter table over the same model (enforced caps only, larger headroom factor, because its startup check must not fail on terms outside the node's control); each file points at the other.

## Result

```
             before          after
capped      796.0 MiB      1.5 GiB
estimated     1.8 GiB    384.0 MiB
ceiling       3.2 GiB      2.4 GiB
```

"Capped" means the level enforcement acts at, not a level allocation cannot reach — see `allow_stall` above. What is left estimated — gossipsub's caches and send queues, the block execution working set — is bounded by message counts rather than byte budgets. The counts turn into bytes via the gossip message limit, which is now derived from `max_transaction_size_bytes` and therefore capped; peer count is not. Nothing bounds how many peers a node holds connections to, and that is the remaining structural gap.

## Docs

Adds a validator node requirements page: CPU, memory and storage speed from measurements; **disk capacity and bandwidth marked as not established** rather than guessed, with what it would take to establish each. The page says explicitly that block execution is serial, so single-thread speed sets the floor and core count is a throughput lever. `tari-vn-bench`'s README now records that it is Linux-only.

## Verification

```
cargo lints clippy -p tari_state_store_rocksdb -p tari_swarm -p tari_networking \
  -p tari_validator_node -p tari_indexer -p tari_ootle_walletd -p tari_vn_bench \
  -p db-inspector -p tari_validator_rollback --all-targets     # clean
cargo test -p tari_state_store_rocksdb          # 17 pass
cargo test -p tari_validator_node --lib         # 28 pass
./target/release/tari-vn-bench --skip-execution --skip-native --skip-storage
```

Not verified: behaviour of a real node under sustained write load with the new memtable budget. The write path is exercised by the state store tests, but whether 384 MiB against 32 MiB buffers is the right flush-frequency trade at real block commit rates wants a node run.


sdbondi (commented):
The RocksDB half of this is well built and I checked the wiring rather than taking the description's word for it. One gossipsub change I think is a net loss and want justified or reverted.

## Blocking

**`gossip_sub_history_length: 3` — `networking/swarm/src/config.rs:71`**

`history_length` is not only the IWANT recovery window. Because this node runs gossipsub with `validate_messages()`, it is also the deadline for the application's validation verdict, and two things break when the verdict misses it (libp2p-gossipsub `behaviour.rs::report_message_validation_result`):

- `Accept` finds the message gone from the mcache and returns `false` without forwarding — the node silently stops relaying.
- `Reject` also finds nothing and returns `false` before reaching `peer_score.reject_message` — P4 invalid-message scoring is the *only* scoring signal `peer_score_params()` (`networking/swarm/src/behaviour.rs:280-301`) leaves enabled, so the flooder stops being penalised.

That verdict is reported only after the message has been drained from the bounded inbound queue, which `applications/tari_validator_node/src/config.rs:125-129` says the mempool drains serially and which is sized at 128–256 MiB *precisely to absorb multi-second bursts*. So the failure mode lands exactly in the scenario this PR exists to bound, and 5 → 3 tightens the deadline from ~5s to ~3s at the 1s default heartbeat.

Second effect: the window in which we can still serve an IWANT for something we advertised is `history_length - history_gossip` (`mcache.rs`: `history[..gossip]` is advertised, all of `history` is servable). That was 5-3 = 2 heartbeats; it is now 3-2 = 1. An IHAVE we cannot serve becomes a broken promise at the requester after `iwant_followup_time` (3s), scored as P7 = `excess² × behaviour_penalty_weight`. `peer_score_params()` zeroes `ip_colocation_factor_weight` and the topic weights but leaves `behaviour_penalty_weight` at libp2p's default −10, so **three** unserved IWANTs inside one decay interval puts us at −90, past `graylist_threshold` (−80), at that peer.

Against that, the memory saved is small. The worker reports `Ignore` for a queue-rejected message, which removes it from the mcache immediately, so the mcache's term *over and above* the inbound queues you already count is `history_length` heartbeats of validated, forwarded traffic — a couple of heartbeats of throughput, not a fraction of 512 MiB.

Suggest keeping `history_length` at 5 and lowering `history_gossip` instead if IHAVE volume is the concern, or stating the measured mcache size that makes 3 worth it. The rest of the gossipsub block I have no objection to.

## Non-blocking

**The warning names a knob that does not exist — `applications/tari_validator_node/src/memory_budget.rs:127-131`.** It tells the operator to "lower the queue and cache budgets in the configuration", but the largest single line, the 768 MiB state store budget, has no config path: `DatabaseOptions` carries no serde and `bootstrap.rs:258-261` builds it from `::default()` with only `with_debugging_data` / `with_prune_transaction_history`. Only the three queue caps are settable. Either wire `memory_budget_bytes` through `ValidatorNodeConfig` or narrow the advice to the queues.

**"A node cannot exceed their sum by any amount of traffic" — `memory_budget.rs:8`.** Not true of the state-store line as configured. `allow_stall: false` makes the memtable budget advisory, so memtable memory can overshoot to `write_buffer_bytes × max_write_buffer_number × cf_count` = 32 MiB × 2 × 9 = 576 MiB while triggered flushes are in flight, and `Cache::new_lru_cache` is non-strict, so the cache admits past capacity as well. The PR body says "effective but soft", which is right; the in-tree doc claims hard, and so does `Bound::Capped` on that line in `utilities/tari_vn_bench/src/memory.rs:136`. Given the whole point is that the ceiling is a checkable claim, the doc should say soft.

**`every_configured_cap_is_counted` cannot fail — `memory_budget.rs:163-175`.** It asserts `capped_bytes == lines.iter().map(|l| l.bytes).sum()`, which is verbatim how `capped_bytes` is computed thirty lines above. The name promises a coverage check — that no enforced cap is missing from `lines` — and nothing in the test does that. `clones_share_one_budget` (`crates/state_store_rocksdb/src/memory_budget.rs:110-116`) has the same shape: comparing `capacity_bytes()` and two zero `memtable_bytes()` would pass for a deep copy. If sharing is the invariant, write into one and read usage through the other.

**`gossip_sub_history_gossip`'s "must not exceed" is unenforced — `config.rs:33-35`.** gossipsub's builder returns `Err` for that pair and `behaviour.rs:96` `.unwrap()`s it, so a mis-set pair is a startup panic rather than a config error. Not reachable today (code-only field, all callers use `..Default::default()`), but it is a knob now.

**`gossip_sub_duplicate_cache_time: 60s` is exactly libp2p's default.** Fine as "state it explicitly", but the PR body reads as though it were tightened; it changes nothing.

**`connection_handler_queue_len` 5000 → 256 is right on the memory** — `RpcOut::Forward` holds a full `RawMessage` per peer, so this is the real term. One side effect worth knowing: a failed push calls `peer_score.failed_message_slow_peer` (`behaviour.rs:3181-3183`), so *our* local queue pressure penalises the *remote* peer. At `slow_peer_weight` −0.2 with decay 0.2 it takes ~40 sustained drops/s to one peer to cross `gossip_threshold`, so it is unlikely to bite — but it is ~20× closer than it was.

**Two budget tables now drift independently.** `applications/tari_validator_node/src/memory_budget.rs` and `utilities/tari_vn_bench/src/memory.rs` have different line sets and different overhead factors (1.5 vs `ALLOCATOR_OVERHEAD`). The bench's file comment is accurate that the queue caps can't be imported (`mod config` is private), but the PR body's "reads the real defaults from the crates that own them instead of hardcoding them" holds only for the RocksDB, swarm and template lines — the three largest capped terms are still literals.

## Nits

- `applications/tari_validator_node/src/p2p/services/mempool/mod.rs:33` drops the blank line before `pub use error::*;` — unrelated churn.
- `crates/state_store_rocksdb/src/options.rs` gains `with_memory_budget_bytes` and `with_memtable_budget_bytes` but no `with_write_buffer_bytes`, though the field is public alongside them.

## Verified sound

- The shared-budget wiring genuinely shares. `Options::clone` carries `outlive.write_buffer_manager` and `outlive.block_based.block_cache` (`rocksdb-0.24.0/src/db_options.rs:120-140`, `2819-2824`, `563-568`), so the per-CF descriptors cloned from one `build_default_store_opts` really do hold one `Cache` and one `WriteBufferManager`, and the manager reaches RocksDB through the db-level options it is also set on. `new_write_buffer_manager_with_cache(memtable, false, cache)` matches the crate signature.
- The memtable clamp, and the reasoning for `allow_stall: false` against `missed_proposal_suspend_threshold`.
- `StateStoreMemoryCollector` matches `InboundQueueCollector` exactly, including registering on the root registry rather than the `tari` sub-registry — consistent with its sibling at `bootstrap.rs:176-181`, so not a naming inconsistency despite the comment at `bootstrap.rs:288`.
- The table arithmetic: 256+128+128+768+200+64+20 = 1564 MiB → "1.5 GiB", ×1.5 = "2.3 GiB", and the docs page reproduces it. Queue defaults match `config.rs:147-157`; WASM is 32 pages × 64 KiB × depth 10 = 20 MiB; `MEM_MAX_TRANSACTIONS_DEDUP × 67 B` lands within ~3% of what two hashbrown generations of 500k `[u8;32]` actually allocate at power-of-two bucket counts.
- `Cargo.lock` contains only tari_vn_bench's three new deps — nothing swept in.
- The docs page lands in the autogenerated `reference` sidebar, so no `astro.config.mjs` change is needed.

sdbondi (commented):
Re-checked bb33ab1 against each finding rather than the commit message. All resolved.

**Blocking — `history_length`.** Back to 5/3 (`networking/swarm/src/config.rs:82-83`). The doc comment now names it as the validation-verdict deadline and states the serve margin is the difference between the two, which is the constraint that was missing. Cleared.

**Config knob.** `state_store_memory_budget_bytes` is serde-backed with `DEFAULT_MEMORY_BUDGET_BYTES` as its default (`config.rs:145-147`, `165-167`), `bootstrap.rs:258-265` feeds both budgets from it, and the warning names the real setting. Cleared.

**Soft vs hard.** Module doc separates the two enforcement modes and bounds the overshoot to a multiple of `write_buffer_bytes`; `MAX_WRITE_BUFFER_NUMBER = 2` is now set explicitly (`store.rs:90`) rather than inherited, and the bench's `source` string computes 576 MiB from it instead of asserting a bare cap. Cleared.

**Tests.** `every_configured_cap_reaches_the_total` moves each input by a delta and requires the total to move by it — that one can fail. `a_clone_observes_writes_made_through_the_original` opens a real DB against the original's `WriteBufferManager`, writes 4 MiB and reads non-zero usage through the clone; a deep copy fails it. Both are real now.

**`.build()?`.** `ConfigBuilderError` has `impl std::error::Error` (gossipsub `error.rs:151`) and the closure already boxes errors for `try_from_owned` and `mdns::tokio::Behaviour::new`, so the conversion holds. A `history_gossip > history_length` pair is now a startup error.

Nits, slow-peer note, cross-references between the two tables, and the PR body — all done.

## Two small things, neither blocking

**The config→`DatabaseOptions` seam is the one thing the new test does not cover.** `every_configured_cap_reaches_the_total` mutates `DatabaseOptions::memory_budget_bytes` directly, so if someone dropped `.with_memory_budget_bytes(...)` from `bootstrap.rs:261`, the store would open at the crate default while the config said otherwise and every test would still pass. That is exactly the failure the test was rewritten to catch, one layer up. Threading `config.state_store_memory_budget_bytes` through a small helper that both `bootstrap` and the test call would close it.

**`COLUMN_FAMILIES: u64 = 9` — `utilities/tari_vn_bench/src/memory.rs:35`.** `all_column_families_iter()` is public and the bench already depends on the crate, so this one can be imported (`.count()`) rather than hardcoded — and the file's own doc says the queue caps are the only exception.

Not verified here and unchanged from your list: the memtable budget under sustained write load. Waiting on `test (1-3/3)` and `integration test`; the rest of CI is green.

sdbondi (commented):
Reviewed c08e9c2. The `state_store_options()` extraction closes the seam I flagged (`config.rs:171-180`, and `the_constant_caps_are_present` now goes through it), and the `COLUMN_FAMILIES` nit is unaddressed but immaterial. The gossip-limit commit has a problem.

## Blocking

**`MAX_GOSSIP_MESSAGE_SIZE` is below what the network's own ingress validator accepts — `crates/p2p/src/gossip.rs:26-27`.**

The derivation assumes the largest legitimate message is one template binary plus an envelope. It isn't. `MAX_PUBLISH_TEMPLATES_PER_TRANSACTION` caps `PublishTemplate` *instructions*; it says nothing about total payload. `Blobs` is a transaction-level list and `calc_blobs_weight` (`crates/transaction/src/v1/transaction.rs:435-444`) charges every blob uniformly — the comment there says so explicitly: "uniform across all blob references — `PublishTemplate` binaries and `Blob` instruction args alike". `BlobReferenceValidator`'s own doc puts it plainly: "a transaction can carry arbitrary payload up to the weight cap".

The weight cap is the only byte bound at ingress, and every byte-carrying term is divided by 3 (`BLOB_BYTE_DIVISOR`, `LITERAL_BYTE_DIVISOR`). At `max_transaction_weight = 1_000_000` (`consensus_constants.rs:152` and every other network) that admits roughly **2.86 MiB** of payload. The new limit is 1.5 MiB + 256 KiB = **1.75 MiB**.

Concretely — a transaction every validator accepts and this node will not gossip:

```
PublishTemplate, 1.5 MiB binary   →   1_572_864/3 + 32 = 524_320
one Blob arg, 512 KiB             →     524_288/3 + 32 = 174_795
                                       ───────────────────────
                                       699_115  ≪ 1_000_000
serialized ≈ 2.0 MiB              >   MAX_GOSSIP_MESSAGE_SIZE (1.75 MiB)
```

Both failure modes are silent-ish and bad: `publish_gossip` fails on the originating node, and a peer whose limit is smaller than the frame it receives gets a `prost_codec` length error rather than a per-message drop (the per-message `max_transmit_size_for_topic` check at `protocol.rs:349-352` reads the *per-topic* map, which nothing here populates, so only the codec frame limit applies).

`consensus_constants.rs:402-421` already has `max_transaction_weight_admits_max_template_publish` asserting the weight cap sits above a max-size publish — the same relationship the gossip limit needs, and the same place to derive it from. `max_transaction_weight × BLOB_BYTE_DIVISOR` plus framing would be a derivation that actually bounds what ingress admits. Failing that, keeping 2 MiB and adding a byte-size ingress validator would make the two agree from the other direction.

**Worth a decision either way: this narrows a network-visible constant.** 2 MiB → 1.75 MiB against peers that have not upgraded. The doc comment says agreement across the mesh is required, and I agree — so is this something the protocol activation schedule needs to carry, rather than shipping as a plain constant change?

## Non-blocking

**A comment this commit's research contradicts — `applications/tari_validator_node/src/config.rs:132-136`.** It says the consensus gossip topic "carries `HotStuffMessage`s between shard groups, including block-sized foreign proposals", and that is the stated reason the consensus gossip queue is budgeted at 256 MiB, the largest queue line in the memory table. The new comment in `gossip.rs` says the opposite, and the new comment is right: the only thing broadcast on that topic is `HotstuffMessage::ForeignProposalNotification` (`crates/consensus/src/hotstuff/on_receive_local_proposal.rs:1144`, the sole `.broadcast(` in the crate), and everything else goes over direct messaging. Pre-existing, but it means the largest queue in a memory budget rests on a premise this PR has just disproved.

**`SwarmConfig::default()`'s 2 MiB is now a value no node uses.** Both bootstraps override it, and the field doc says callers must. Leaving a plausible-looking default there invites the next caller to skip the override and silently sit on a different mesh limit — consider dropping the field from `Default` or making the type carry it as a required constructor argument.

**`COLUMN_FAMILIES: u64 = 9` — `utilities/tari_vn_bench/src/memory.rs:35`** still hardcoded where `all_column_families_iter().count()` is importable. Unchanged from last round; take it or leave it.

## Sound in this commit

`state_store_options()` puts the config→`DatabaseOptions` mapping in one place that both `bootstrap.rs:257` and the test go through, which is what I asked for. `TRANSACTION_ENVELOPE_ALLOWANCE` being generous rather than tight is the right instinct — the problem is the term it is added to, not the allowance.

sdbondi (commented):
Reviewed 28b149b. This is the right shape — the byte cap is the thing that was missing, `encoded_size` is the right quantity to measure, and the two new consensus-constants tests are the ones that keep it honest. I verified the encoding claim rather than trusting the comment: `Transaction` derives `minicbor::CborLen`, `tari_bor::encode_to_vec` is `minicbor::to_vec_with` (`crates/tari_bor/src/lib.rs:42`), and `bor_encoded` is produced by exactly that (`crates/p2p/src/conversions/transaction.rs:69`) — so `minicbor::len` really is the wire length, and the check is deterministic across nodes.

One thing to settle before merge.

## Blocking-ish: the gossip limit is now tighter than the old one, and it does not need to be

`max_gossip_message_size(1.75 MiB) = 1_851_392` bytes. The limit it replaces was `2 * 1024 * 1024 = 2_097_152`. That is the same narrowing against unupgraded peers that c08e9c2 was reverted for — a 246 KiB band where an unupgraded node admits and relays a transaction whose frame an upgraded node cannot decode. During rollout that surfaces as a `prost_codec` length error and a torn-down substream, not a per-message drop, and it costs the honest relaying peer.

Your own note one commit ago was that making valid transactions invalid "belongs with a protocol activation rather than in a constant". The ingress cap has a decent case for not needing one — it is mempool admission, like `max_transaction_weight`, and `ProtocolVersion` is about substate hashing schema, not admission. The *gossip frame limit* is a different matter, because it is a wire parameter both ends must agree on.

The good news is that the invariant you need is one-directional. Gossip must carry anything ingress admits; nothing requires it to be tight. So:

```rust
pub const fn max_gossip_message_size(max_transaction_size_bytes: usize) -> usize {
    // never narrow below what the mesh already accepts
    let derived = max_transaction_size_bytes + GOSSIP_FRAMING_ALLOWANCE;
    if derived > LEGACY_GOSSIP_MESSAGE_SIZE { derived } else { LEGACY_GOSSIP_MESSAGE_SIZE }
}
```

keeps the derivation, keeps the guarantee, and removes the rollout hazard for nothing — the memory cost is the 246 KiB you were already paying yesterday. If the answer is instead "these networks have no traffic that could be in that band, so ordering does not matter", say so in the commit and I am satisfied; I just do not want it decided implicitly.

## Non-blocking

**`Transaction::encoded_size` bypasses the crate's own wrapper — `crates/transaction/src/transaction.rs:99`.** `tari_bor::encoded_len` is this function (`tari_bor/src/lib.rs:95-97`, `minicbor::len_with` under a `Result` kept for API compatibility), and `calc_stealth_statement_weight` in this very crate already reaches for it. Going through `minicbor::len` directly means the transaction crate now has two conventions for the same measurement.

**The size validator sits in a different place in each builder.** `create_structural_transaction_validator` puts it before `BlobReferenceValidator` (`builder.rs:38-41`, with a comment justifying bytes-before-weight), while `create_node_transaction_validator` puts it after (`bootstrap.rs:551-553`) — and there it lands under the pre-existing comment "Cheap structural check — reject over-weight transactions before verifying signatures", which now reads as describing the size check. Same verdict either way, but the deliberate ordering argument only holds in one of them.

**`max_transaction_size_bytes` is a fifth copy of the same literal.** Every network's `ConsensusConstants` block now carries an identical four-line comment above `max_transaction_size_bytes: MAX_TRANSACTION_SIZE_BYTES`. The constant already carries its derivation at `consensus_constants.rs:31-35`; the per-network repetition adds nothing that a reader of the constant does not have, and is four more places to forget when it changes.

## Verified sound

- `encoded_size` matches the wire bytes, as above, and `#[derive(CborLen)]` makes it allocation-free.
- `max_transaction_size_admits_max_template_publish` builds and measures a real max-size publish rather than reasoning about the envelope — that is the assertion that keeps `TRANSACTION_ENVELOPE_ALLOWANCE` honest, and it fails loudly if the encoding or the template limit moves.
- `the_byte_cap_binds_before_the_weight_cap` pins the relationship in the other direction, so the new cap cannot silently become dead configuration. 1.75 MiB against the 2.86 MiB the weight cap admits.
- `max_transaction_size_bytes` is threaded through `ConsensusConstantsFile`'s `set!` (`consensus_constants_file.rs:64`), and that path is LocalNet-only, so a stray file cannot desynchronise a real network's gossip limit.
- Both entry points that build validators — the node's and the indexer's, plus the indexer's `TransactionManager` structural pass — pass the new constant; nothing constructs a validator set without it.
- The `TransactionSizeValidator` tests exercise the case that motivates it: a blob-heavy transaction the weight cap admits and the byte cap catches.

sdbondi (commented):
c939a051 clears everything I raised. The narrowing decision is now recorded with the condition it depends on and what a deployment that cannot meet it must do instead — that was the "say so explicitly" option, and it is the right one to take as the maintainer of these networks. Ordering is unified and the weight comment sits back above the weight check; the per-network comment duplication is gone; `encoded_size` goes through `tari_bor::encoded_len`.

Nothing blocking left. One nit on the last of those.

**`unwrap_or(0)` fails open — `crates/transaction/src/transaction.rs:100`.** You are right that it is unreachable: `encoded_len` is `Ok(minicbor::len_with(..))` for anything deriving `CborLen`, so there is no error path today. But this value's only consumer is a cap check, and zero is the one answer that admits every transaction. If the fallback is unreachable, its direction is free — so pick the one that fails closed (`unwrap_or(usize::MAX)`), or `expect` and let it be loud. Same cost, no scenario where it is the wrong choice.

Watching CI; I will confirm when the full set is green. As before, GitHub will not take an approval from me on your own PR, so it needs another maintainer for the formal one.

PR #2541, opened 2026-09-07
feat(rpc)!: prove substates on the batched read path

sdbondi:
## Description

`get_substate_batch` streamed bare `proto::consensus::Substate` values with no proof of any kind, so `CachedSubstateManager::fetch_and_cache_substates` wrote every batch result with `verified: false`. With proof verification on, the read path refuses those entries and refetches singly — the batch populated a cache it could not then use.

This makes the batch carry proofs, and hoists the parts that do not vary per substate out of the per-substate loop.

### The hoist

A `SubstateValueProof` has three parts, with three different natural scopes:

| Part | Scope | Cost |
|---|---|---|
| Commit proof (the anchor) | once per response | ~10–30 KB, ~C signature verifications *on the client* |
| Level 2: shard root in the shard-group root | one tree per response, one leaf per shard | ~250 B |
| Level 1: JMT leaf proof | genuinely per substate | ~1 KB, an O(log N) traversal |

`SubstateProofGenerator` holds the per-shard-group state — the ordered roots, the single `RootProofTree` built over them, each shard's root and version, and level-2 proofs extracted on first use of a shard. `generate_substate_proof` is a one-shot wrapper over it, so the single-substate path behaves exactly as before.

Measured on the worst case (256-shard group, 50 substates in 50 distinct shards; `proof_cost`, release):

```
one-shot x50:        10.68 ms     <- what the batch would have cost unhoisted
generator construct:  228 µs
generator x50:        221 µs      <- ~4.4 µs per substate
hoisted total:        449 µs
one substate alone:   219 µs      <- essentially all fixed cost
```

### On the 50-substate cap and DoS

Those numbers say the cap is not the DoS lever. Cost per request is ~220 µs fixed (reading every shard root in the group and building one tree over them) plus ~4.4 µs per substate. So one request answering 50 costs ~450 µs, while **50 single-substate requests cost ~11 ms for the same data** — 25× more validator CPU. Lowering the cap would make a flood *more* expensive to serve, not less. What 50 actually bounds is the response, which the responder materialises before streaming. The rationale is recorded at the constant so it isn't re-litigated blind. Concurrency is separately bounded by the RPC framework's `BoundedExecutor`.

### The anchor

Anchor, substates and value proofs are all read in one transaction, so every proof in a stream verifies against the anchor's shard-group root — the responder's existing `TODO` about wanting a snapshot is the same concern. The shard group a proof is built against is taken **from the anchor's own block header**, not from an epoch lookup, so the two cannot disagree across an epoch boundary or before consensus has started.

On the client, `trusted_root_from_commit_proof` establishes that root once per batch: a hit in the trusted-root store (recorded by `validator_status.probe`) skips QC-chain validation and the committee lookup entirely; a miss validates the commit proof and warms the store. Each result then costs only its Merkle path. The single-substate path is rewritten in terms of the same helper, so there is one place where a root becomes trusted. The root is trusted because a quorum signed the header committing it, independently of any value proof that cites it — that is what makes it safe to establish once for a batch and to record.

A member that answers without an anchor has nothing committed to prove against. `race_substate_batch` holds that answer and keeps asking the rest of the committee — the shape `CommitteeReadTally` uses for single substates — and serves the held batch only once every member has failed, with a warning, never cached.

### Two responder defects fixed alongside

- **It never checked it stores the requested ids.** `get_substate` guards with `includes_substate_id`; the batch responder read whatever ids it was handed out of its local store. Same class as the `sync_state` hole fixed in #2511.
- **`missing` was logged and dropped.** It is now on the wire and reaches `SubstateBatch::missing`, where the indexer logs it per chunk. It is *not* wired into any caching decision, and must not be: it names ids a response does not answer for, for any reason — including a substate the responder holds but cannot prove — so it is never evidence of nonexistence.

Down heads are also now decoded as `Down` instead of failing the whole batch on `SubstateValue::from_bytes(&[])`, and are cached as such — which is what #2528's head-only cache wants. A substate whose shard has no committed state is reported as missing rather than failing the batch: `SubstateProofGenerator::generate` returns `Ok(None)` for it, while a read failure stays an `Err`.

## Correcting the handoff on one point

The workstream-D notes suggested a batch could answer "provably not here" for ids it holds no record of, using an exclusion proof. It cannot. `SpreadPrefixKeyMapper::map_to_leaf_key` hashes the whole `VersionedSubstateId`, so a leaf key is version-scoped and an exclusion proof states only that *one version* is not up — which is equally true of a live substate at a higher version. Absence of an id is not provable under this scheme, which is why the existing single-substate path already leaves `DoesNotExist` to f+1 agreement.

## Known gaps

- A batch from a validator with **no anchor at all** is served unproven and uncached, after every other member has been asked. Making that a hard failure would turn a legitimate post-epoch-change state into an error for the REST batch endpoint.
- `verify_substate_batch` and `race_substate_batch` have no unit tests — exercising them needs a mock `ValidatorNodeRpcClient` and cache, which `indexer_lib` has no scaffolding for. The proof generation half is covered.

## Test plan

`crates/state_store_rocksdb/tests/substate_proofs.rs`:

- proofs for a batch spanning several shards all verify against one shard-group root
- a reused (memoised) shard-root proof still verifies for its own substate
- a batched proof is byte-identical to the single-substate `generate_substate_proof`
- a version that is not up gets a verifying exclusion proof, and fails inclusion
- a substate outside the generator's shard group is refused
- a shard with no committed state yields `Ok(None)` rather than an error

Mutation-checked: keying the level-2 memo by a constant shard fails 4 of the 6.

`crates/state_store_rocksdb/tests/proof_cost.rs` is the measurement above, `#[ignore]`d because wall-clock thresholds would only make CI flaky.

```
cargo test -p tari_state_store_rocksdb -p tari_indexer_lib -p tari_ootle_storage -p tari_state_tree   # 31 binaries, all pass
cargo lints clippy -p tari_ootle_storage -p tari_state_tree -p tari_validator_node_rpc \
  -p tari_indexer_lib -p tari_state_store_rocksdb -p tari_validator_node -p tari_indexer --all-targets   # clean
```

Integration tests are left to CI.

## Breaking change

`GetSubstatesBatchResponse` becomes a `oneof` carrying the anchor, proven substates, and the missing ids; `GetSubstatesBatchRequest` gains `include_proofs`. Both sides ship together. The client rejects a response that accounts for none of the requested ids, so a pre-oneof responder fails loudly rather than looking like "none of these exist". `verify_substate_value_proof` is deleted — the block this PR rewrote was its last caller.

### DIFF
--- a/applications/tari_validator_node/src/p2p/rpc/rpc_impl.rs
+++ b/applications/tari_validator_node/src/p2p/rpc/rpc_impl.rs
@@ -263,9 +263,9 @@ fn latest_commit_proof<TTx: StateStoreReadTransaction>(
     tx: &TTx,
     epoch: Epoch,
 ) -> Result<Option<CommittedBlockProof>, StorageError> {
-    // `last_executed_get` reports both "nothing committed here" cases as `NotFound`: no row at all,
-    // and a row belonging to another epoch - the latter being exactly the window after an epoch
-    // change, and the former a node that has not started consensus since restarting. Neither is a
+    // `last_executed_get` reports both "nothing committed here" cases as `NotFound`: a node that has
+    // never committed anything has no row, and one whose row belongs to another epoch is either in
+    // the window after an epoch change or has not started consensus since restarting. Neither is a
     // failure to read; both mean there is nothing to anchor to.
     let Some(last_executed) = tx.last_executed_get(epoch).optional()? else {
         return Ok(None);
--- a/applications/tari_validator_node/src/p2p/rpc/rpc_impl.rs
+++ b/applications/tari_validator_node/src/p2p/rpc/rpc_impl.rs
@@ -82,7 +82,6 @@ use tari_ootle_storage::{
         SubstateValueFilterFlags,
         TransactionRecord,
     },
-    generate_substate_proof,
 };
 use tari_ootle_transaction::{Transaction, TransactionId};
 use tari_rpc_framework::{Request, Response, RpcStatus, Streaming};
@@ -152,8 +151,14 @@ impl<TStateStore: StateStore> ValidatorNodeRpcServiceImpl<TStateStore> {
         };
 
         let shard_group = proof_shard_group(&commit_proof).map_err(RpcStatus::log_internal_error(LOG_TARGET))?;
-        let value_proof = generate_substate_proof(tx, shard_group, &substate.to_versioned_substate_id(), num_preshards)
-            .map_err(RpcStatus::log_internal_error(LOG_TARGET))?;
+        // The anchor cannot speak for this substate - its shard is outside the group the anchor
+        // commits, or has nothing committed. Answer unproven rather than not at all.
+        let Some(value_proof) = SubstateProofGenerator::new(tx, shard_group, num_preshards)
+            .and_then(|mut generator| generator.generate(&substate.to_versioned_substate_id()))
+            .map_err(RpcStatus::log_internal_error(LOG_TARGET))?
+        else {
+            return Ok(());
+        };
 
         resp.commit_proof = commit_proof.to_bytes();
         resp.substate_value_proof =
@@ -216,7 +221,7 @@ fn read_substate_batch<TTx: StateStoreReadTransaction>(
             let Some(proof) = generator.generate(&substate.to_versioned_substate_id())? else {
                 warn!(
                     target: LOG_TARGET,
-                    "{} is stored but its shard has no committed state; reporting it as missing",
+                    "{} is stored but cannot be proved against the anchor; reporting it as missing",
                     substate.substate_id()
                 );
                 missing.push(substate.substate_id().to_bytes());
@@ -258,7 +263,13 @@ fn latest_commit_proof<TTx: StateStoreReadTransaction>(
     tx: &TTx,
     epoch: Epoch,
 ) -> Result<Option<CommittedBlockProof>, StorageError> {
-    let last_executed = tx.last_executed_get(epoch)?;
+    // `last_executed_get` reports both "nothing committed here" cases as `NotFound`: no row at all,
+    // and a row belonging to another epoch - the latter being exactly the window after an epoch
+    // change, and the former a node that has not started consensus since restarting. Neither is a
+    // failure to read; both mean there is nothing to anchor to.
+    let Some(last_executed) = tx.last_executed_get(epoch).optional()? else {
+        return Ok(None);
+    };
     if last_executed.height.is_zero() {
         return Ok(None);
     }
--- a/applications/tari_validator_node/src/p2p/rpc/rpc_impl.rs
+++ b/applications/tari_validator_node/src/p2p/rpc/rpc_impl.rs
@@ -141,7 +141,6 @@ impl<TStateStore: StateStore> ValidatorNodeRpcServiceImpl<TStateStore> {
     fn attach_substate_proof<TTx: StateStoreReadTransaction>(
         &self,
         tx: &TTx,
-        shard_group: ShardGroup,
         num_preshards: NumPreshards,
         substate: &SubstateRecord,
         resp: &mut GetSubstateResponse,
@@ -152,6 +151,7 @@ impl<TStateStore: StateStore> ValidatorNodeRpcServiceImpl<TStateStore> {
             return Ok(());
         };
 
+        let shard_group = proof_shard_group(&commit_proof).map_err(RpcStatus::log_internal_error(LOG_TARGET))?;
         let value_proof = generate_substate_proof(tx, shard_group, &substate.to_versioned_substate_id(), num_preshards)
             .map_err(RpcStatus::log_internal_error(LOG_TARGET))?;
 
@@ -163,10 +163,21 @@ impl<TStateStore: StateStore> ValidatorNodeRpcServiceImpl<TStateStore> {
     }
 }
 
+/// The shard group a commit proof's block header commits the state merkle root of.
+///
+/// Value proofs must be built against this group rather than one read from an epoch lookup: the two
+/// disagree across an epoch boundary at which the committee count changed, and level-2 proofs shaped
+/// by the wrong group verify against a root the anchor does not commit. Taking it from the anchor
+/// makes the pair consistent by construction.
+fn proof_shard_group(commit_proof: &CommittedBlockProof) -> Result<ShardGroup, StorageError> {
+    commit_proof.shard_group().map_err(|e| StorageError::QueryError {
+        reason: format!("commit proof has an invalid shard group: {e}"),
+    })
+}
+
 /// What the responder needs to prove a batch of substates against its own committed state.
 struct BatchProofContext {
     epoch: Epoch,
-    shard_group: ShardGroup,
     num_preshards: NumPreshards,
     include_proofs: bool,
 }
@@ -186,16 +197,17 @@ fn read_substate_batch<TTx: StateStoreReadTransaction>(
     ids: &[SubstateId],
     ctx: BatchProofContext,
 ) -> Result<Vec<batch_response::Response>, StorageError> {
-    let (substates, mut missing) = SubstateRecord::get_any_max_version(tx, ids)?;
-    let mut missing = missing.drain().map(|id| id.to_bytes()).collect::<Vec<_>>();
+    let (substates, missing) = SubstateRecord::get_any_max_version(tx, ids)?;
+    let mut missing = missing.into_iter().map(|id| id.to_bytes()).collect::<Vec<_>>();
 
     let mut generator = None;
     let mut messages = Vec::with_capacity(substates.len() + 2);
     if ctx.include_proofs &&
         let Some(commit_proof) = latest_commit_proof(tx, ctx.epoch)?
     {
+        let shard_group = proof_shard_group(&commit_proof)?;
         messages.push(batch_response::Response::CommitProof(commit_proof.to_bytes()));
-        generator = Some(SubstateProofGenerator::new(tx, ctx.shard_group, ctx.num_preshards)?);
+        generator = Some(SubstateProofGenerator::new(tx, shard_group, ctx.num_preshards)?);
     }
 
     for substate in substates {
@@ -300,13 +312,17 @@ impl<TStateStore: StateStore + Clone + Send + Sync + 'static> ValidatorNodeRpcSe
             .ok_or_else(|| RpcStatus::bad_request("Missing substate requirement"))?;
 
         // We need our local committee info to (a) confirm we store a non-global substate and (b) know
-        // our shard group + preshard count when generating a proof. It is read at the epoch consensus
-        // is committing at - the epoch the proved state and its anchor belong to - so that a shard
-        // group from a newer epoch cannot shape a proof the anchor's root does not commit.
+        // the preshard count when generating a proof. The shard group a proof is built against comes
+        // from the anchor rather than from here, so this is only ever a question about the node.
         let local_committee_info = if !substate_requirement.substate_id().is_global() || req.include_proof {
+            let current_epoch = self
+                .epoch_manager
+                .current_epoch()
+                .await
+                .map_err(RpcStatus::log_internal_error(LOG_TARGET))?;
             let info = self
                 .epoch_manager
-                .get_local_committee_info(self.consensus.current_epoch())
+                .get_local_committee_info(current_epoch)
                 .await
                 .map_err(RpcStatus::log_internal_error(LOG_TARGET))?;
             if !substate_requirement.substate_id().is_global() &&
@@ -375,7 +391,7 @@ impl<TStateStore: StateStore + Clone + Send + Sync + 'static> ValidatorNodeRpcSe
             let info = local_committee_info
                 .as_ref()
                 .expect("committee info is fetched whenever include_proof is set");
-            self.attach_substate_proof(&tx, info.shard_group(), info.num_preshards(), &substate, &mut resp)?;
+            self.attach_substate_proof(&tx, info.num_preshards(), &substate, &mut resp)?;
         }
 
         Ok(Response::new(resp))
@@ -748,15 +764,18 @@ impl<TStateStore: StateStore + Clone + Send + Sync + 'static> ValidatorNodeRpcSe
             .collect::<Result<Vec<_>, _>>()
             .map_err(|e| RpcStatus::bad_request(format!("Invalid substate ID: {e}")))?;
 
-        // The state being read and the anchor proving it both belong to the epoch consensus is
-        // committing at, so the committee info that shapes the proof must come from that epoch too:
-        // a shard group read at a newer epoch would produce level-2 proofs against a root the anchor
-        // does not commit. It tells us (a) which of the requested ids we are responsible for and
-        // (b) our shard group + preshard count.
-        let consensus_epoch = self.consensus.current_epoch();
+        // Which of the requested ids we are responsible for. This is a membership question about the
+        // node, not about the state being proved - the shard group a proof is built against comes
+        // from the anchor - so it is asked at the epoch manager's epoch, which is set before
+        // consensus starts.
+        let current_epoch = self
+            .epoch_manager
+            .current_epoch()
+            .await
+            .map_err(RpcStatus::log_internal_error(LOG_TARGET))?;
         let local_committee_info = self
             .epoch_manager
-            .get_local_committee_info(consensus_epoch)
+            .get_local_committee_info(current_epoch)
             .await
             .map_err(RpcStatus::log_internal_error(LOG_TARGET))?;
         if let Some(id) = ids.iter().find(|id| !local_committee_info.includes_substate_id(id)) {
@@ -770,7 +789,9 @@ impl<TStateStore: StateStore + Clone + Send + Sync + 'static> ValidatorNodeRpcSe
         let (sender, receiver) = mpsc::channel(req.substate_ids.len() + 2);
 
         let store = self.state_store.clone();
-        let shard_group = local_committee_info.shard_group();
+        // The epoch to look for a committed block at. Before consensus starts this is zero, which
+        // finds nothing and yields no anchor, so results go out unproven rather than failing.
+        let consensus_epoch = self.consensus.current_epoch();
         let num_preshards = local_committee_info.num_preshards();
         let include_proofs = req.include_proofs;
         let responses = task::spawn_blocking(move || {
@@ -779,7 +800,6 @@ impl<TStateStore: StateStore + Clone + Send + Sync + 'static> ValidatorNodeRpcSe
             store.with_read_tx(|tx| {
                 read_substate_batch(tx, &ids, BatchProofContext {
                     epoch: consensus_epoch,
-                    shard_group,
                     num_preshards,
                     include_proofs,
                 })
--- a/applications/tari_validator_node/src/p2p/rpc/rpc_impl.rs
+++ b/applications/tari_validator_node/src/p2p/rpc/rpc_impl.rs
@@ -65,11 +65,14 @@ use tari_ootle_p2p::{
         SyncBlocksResponse,
         SyncStateRequest,
         SyncStateResponse,
+        get_substates_batch_response as batch_response,
     },
 };
 use tari_ootle_storage::{
     StateStore,
     StateStoreReadTransaction,
+    StorageError,
+    SubstateProofGenerator,
     consensus_models::{
         Block,
         BookkeepingEpochAgnosticRead,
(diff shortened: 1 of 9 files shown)


sdbondi (commented):
Reviewed at `dec03f03b`, files read at head against `origin/development`.

## Blocking

**1. An unanchored batch is settled by the first member asked, with no attempt at a proven copy.** `crates/indexer_lib/src/cached_substate_manager.rs:410-413`

`verify_substate_batch` returns `Ok(false)` when `batch.commit_proof` is `None`, `fetch_and_cache_substates` then increments `batch_count` and moves on. The batch is not cached, but its values *are* inserted into `results` and returned to the REST caller, and no other committee member is ever asked for that chunk.

The single-substate path deliberately does not do this. `CommitteeReadTally::observe` (`crates/indexer_lib/src/committee_read.rs:57-77`) holds an unproven `Up`/`Down` as `unproven_result` and keeps racing the rest of the committee; it is only served from `conclude`, after every member has failed to prove it, and with a warning. The batch path skips both the retry and the warning.

Failure mode: one member that is behind (or is in the window right after an epoch change, where `last_executed_get(epoch).height.is_zero()` and `latest_commit_proof` returns `None` — `rpc_impl.rs:243`) answers the whole 50-id chunk with no anchor. The indexer serves those 50 values to the caller unproven while four other members could have proven them. That is strictly weaker than what `verify_substate_proofs` buys on the single path, and the asymmetry isn't visible from either call site.

Suggest: while `self.verify_substate_proofs`, treat a missing anchor the same as a failed proof — `break` to the next member — and keep the best unproven batch as a fallback used only once all five members are exhausted, mirroring the tally.

## Non-blocking

**2. The responder builds the proof at one epoch and the anchor at another.** `applications/tari_validator_node/src/p2p/rpc/rpc_impl.rs:740-770`

`local_committee_info` (and hence `shard_group` + `num_preshards`, which decide `SubstateProofGenerator`'s `ordered_roots`) comes from `epoch_manager.current_epoch()`, while the anchor comes from `self.consensus.current_epoch()`. In the window where the epoch manager has advanced and consensus has not, and the committee count changed at that boundary, the generator's `ordered_roots` are over the *new* shard group while the anchor block header commits the *old* group's root. Every level-2 proof in the batch then fails `verify_shard_root`.

This mismatch already exists in `attach_substate_proof` (line 149 vs 369), so it isn't introduced here — but the blast radius is new. On the single path a bad proof just disqualifies one member and the tally moves on; here it disqualifies all five and `fetch_and_cache_substates` returns a hard `Failed to get all substate batches for shard group ...` (`cached_substate_manager.rs:392`) where previously the batch simply came back unproven. Deriving both from `consensus.current_epoch()` — the epoch the state being proved actually belongs to — would close it in both places.

**3. One unprovable substate fails the whole batch.** `rpc_impl.rs:186-196`, `substate_value_proof.rs:88-92`

`generator.generate` returns `Err` when the substate's shard has no committed state; `read_substate_batch` propagates it, so the RPC 500s and the other 49 results are lost. A `SubstateRecord` existing for a substate whose shard tree is empty is genuinely inconsistent, so erroring on *that substate* is right — but dropping it from the stream (or reporting it in `missing`) keeps the rest of the batch useful, and the client already refetches anything it didn't get.

**4. `verify_substate_value_proof` now has no callers.** `crates/storage/src/state_store/substate_value_proof.rs:151`

The last one was the block this PR rewrote. It's `pub` so nothing warns. Either delete it or say in the doc comment that it's the external-consumer entry point; right now it reads as live code.

**5. Reusing proto tag 1 for a differently-typed field.** `crates/p2p/proto/rpc.proto:341-349`

`bytes commit_proof = 1` takes the tag that was `tari.ootle.consensus.Substate substate = 1`, and `ProvenSubstate` takes tag 2. A pre-upgrade client against an upgraded server skips unknown tag 2 and gets an **empty** batch — silently, as "these substates don't exist" — rather than a decode error. Both sides ship together and the title is marked `!`, so this is defence-in-depth only, but fresh tags (4/5/6) plus `reserved 1;` would make a version skew fail loudly instead of looking like absence.

**6. Nothing bounds or binds what the responder streams back.** `crates/validator_node_rpc/src/client.rs:355-378`

`batch.substates` is pushed to without limit and without checking the returned ids are ones that were asked for. The old `try_collect` had the same property, but the loop is explicit now and a `substate_ids.len()` cap plus a membership check are two lines. (Unrequested ids currently fall out harmlessly — `watermarks.get` misses so nothing is cached — but they do land in `results`.)

**7. `batch.missing` is decoded and then dropped.** The PR says missing ids are "now reported to the client"; they reach `SubstateBatch::missing` and no caller reads it. Fine if it's staged for #2528's negative cache, but as it stands the client-visible behaviour for a missing id is unchanged.

## Nits

**8. `a_reused_shard_root_proof_belongs_to_its_own_shard` doesn't test what its docstring says.** `crates/state_store_rocksdb/tests/substate_proofs.rs:135-160`

For every substate after the first in each shard, both `first` and `again` are memo reads. Key the memo by a constant shard and both sides are wrong in the same way, so the assertion still holds — the mutation is caught by tests 1 and 3, not this one. Verifying each reused proof against `group_root` (or against a fresh `SubstateProofGenerator`) would make it test the stated property; as written it asserts determinism.

**9. Redundant global guard.** `rpc_impl.rs:751-753`: `!id.is_global() && !local_committee_info.includes_substate_id(id)` — `includes_substate_id` already returns `true` for globals (`crates/common_types/src/committee.rs:300`). Mirrors `get_substate`'s shape, so consistent, but the conjunct is dead in both.

**10.** `mpsc::channel(req.substate_ids.len() + 2)` incidentally removes a panic: an empty `substate_ids` used to reach `mpsc::channel(0)`, which tokio panics on. Worth an early return on an empty request so that isn't load-bearing on the `+ 2`.

## Verified sound

- The hoist is faithful. `shard_iter_with_global()` is `once(global).chain(shard_iter())` (`shard_group.rs:127`), matching the old explicit `[global, shard_0, ...]`, and `generate` reuses `state.root`, which is the same value the old code re-read via `get_root_hash(version)`. `a_batched_proof_matches_the_single_substate_proof` pins that byte-for-byte, and it compares a memo hit against a fresh single proof, so it does catch a mis-keyed memo.
- Memoising level 2 per shard is safe even when two shards share a root (both empty → placeholder): the group tree keys leaves by `HashIdentityKeyMapper` on the root itself, so identical roots collapse to one leaf in the header's computation too.
- `trusted_root_from_commit_proof` is a clean extraction, and keying trust on the root value rather than on the substate is the right axis — the fast path returns the same root the store vouched for.
- Down heads: exclusion is the correct shape, since a down leaf is deleted from the shard JMT at that version and `proof_epoch` is unused for exclusion. Deciding `Down` from `substate.destroyed.is_some()` matches what `SubstateStatus::Down` means on the single path.
- Rejecting `DoesNotExist` inside an anchored batch is right: the responder never emits it, so accepting one would cache an unproven answer as verified.
- Fail-closed ordering is correct — `batch_count` is incremented only after verification, so a rejected chunk is retried against the next member rather than skipped.
- The PR's correction on exclusion proofs checks out: `SpreadPrefixKeyMapper::map_to_leaf_key` hashes the whole `VersionedSubstateId` (`key_mapper.rs:17-20`), so a leaf key is version-scoped and absence of an id is not provable. `missing` being explicitly unproven is the right call.
- `verify_substate_proofs == false` is unchanged end to end: `include_proofs` is false, the responder builds no generator, and the cache write condition collapses to the old behaviour.

CI is green except `integration test`, still pending.

sdbondi (commented):
Re-reviewed at `9c9442184` (`fix(indexer,rpc): prefer a member that can prove a batch, and degrade per substate`), diffed against `dec03f03b` and re-read at head.

## Resolved

- **(1, blocking) Unanchored batch settled by the first member.** `race_substate_batch` now tries up to `BATCH_READ_ATTEMPTS` members per chunk, holds an `Unanchored` batch in `unproven`, and serves it only after every member has failed to prove the chunk — with a warning, and never cached. That is the `CommitteeReadTally` shape. Making the trichotomy explicit as `BatchTrust` is better than the `bool` I was reading against: `NotRequired` short-circuits on the first answer, `Unanchored` keeps racing, `Err` disqualifies and continues. Chunking moving outside the member loop also drops the `batch_count`/`skip` bookkeeping, so a member that fails chunk 2 no longer costs chunk 3 an attempt.
  - One deliberate divergence from the tally worth noting: it keeps the *highest-version* unproven result, `get_or_insert` keeps the *first* unproven batch. There's no single version to rank a batch by, so first-wins is the reasonable call — just not the same rule.
- **(3) One unprovable substate no longer fails the batch.** `generate` returning `Ok(None)` for a shard with no committed state, reported as missing and skipped, is the right shape — `Ok(None)` for "nothing to prove against" vs `Err` for "the read failed" is a real distinction and the doc comment draws it. `a_shard_with_no_committed_state_proves_nothing` covers it, and `generate_substate_proof` re-raises `None` as an error so the single path is unchanged.
- **(4)** `verify_substate_value_proof` deleted along with the now-unreachable `SubstateProofVerifyError::CommitProof` variant and its imports. Checked: no remaining references to either.
- **(5)** Addressed at the client instead of the proto, which is the better trade — the "accounted for none of the N requested" guard turns the tag-skew case (a pre-oneof responder's tag-1 `Substate` parsing as opaque `commit_proof` bytes) into a loud error rather than silent absence. Verified it can't fire on a legitimate response: `read_substate_batch` puts every unanswered id in `missing`, and empty requests are now rejected outright, so `substates` and `missing` cannot both be empty.
- **(6)** Bounded by `substate_ids.len()` and gated on a `requested` set. Both checks read correctly (`HashSet<&SubstateId>::contains(&SubstateId)` resolves via `&T: Borrow<T>`).
- **(8)** The test now fills the cache and then verifies each cached proof against `group_root`, so a memo keyed by a constant shard genuinely fails it. It tests the stated property now.
- **(9, 10)** Redundant `is_global()` conjunct dropped in the batch path; empty requests rejected before `mpsc::channel`.

## New: the epoch fix hard-fails before consensus starts

`rpc_impl.rs:303-308` and `rpc_impl.rs:745-755`

Using the consensus epoch for `get_local_committee_info` is right — the shard group that shapes a proof has to be the one the anchor's root commits, and that closes the mismatch in both handlers. But `ConsensusHandle::current_epoch()` reads `CurrentView`, which is `AtomicU64::default()` until `PacemakerHandle::set_epoch` runs (`current_view.rs:15-27`, `pacemaker_handle.rs:157-160`). Until consensus starts, that is `Epoch(0)`, and `get_local_committee_info(Epoch(0))` goes to `get_validator_node_by_public_key(Epoch(0), ..)` → `ValidatorNodeNotRegistered` (`epoch_manager.rs:413-420`).

So on a VN whose consensus has not started, both handlers now return an internal error where they previously worked. For `get_substate` the old code only reached the consensus epoch inside `attach_substate_proof`, where `Epoch(0)` degraded to "no anchor" and the result went out unproven; the committee lookup used the epoch manager and succeeded. The batch path is worse: `get_local_committee_info` is unconditional, so this also breaks `include_proofs = false` — a deployment with `verify_substate_proofs` off now gets 500s from any starting VN.

Neither handler calls `check_consensus_state()`, though `sync_blocks` and `sync_state` do (`rpc_impl.rs:434`, `:565`) and `CONSENSUS_NOT_RUNNING` exists for exactly this. Gating both on it would turn a 500 into the typed refusal clients already expect; falling back to the epoch manager's epoch when the view is still zero would preserve the old behaviour. Either is fine — silently 500ing is the one option I'd avoid, since the client can only read it as "this member is broken".

## New: `missing` now means two different things

`rpc.proto:350` and `:366`, `client.rs:86-88`

A substate whose shard has no committed state is pushed into `missing` (`rpc_impl.rs:203-211`) even though the responder *does* hold a record of it. Both docs still say the opposite:

- proto: "naming the requested ids the responder holds no record of" / "Encoded SubstateIds the responder holds no record of at any version"
- `SubstateBatch::missing`: "Ids the responder holds no record of at any version"

Harmless today because nothing reads `missing` beyond a debug log. It stops being harmless the moment #2528's negative cache consumes it: "no record at any version" is exactly the premise for caching nonexistence, and a live substate that merely could not be proved would get cached as not existing. Either widen the docs to "ids this responder did not answer for, for any reason — never evidence of nonexistence", or keep the two cases distinct on the wire so a consumer can tell them apart. The former seems proportionate given how rare the unprovable case is.

## Nit

`rpc_impl.rs:189-190`: `let (substates, mut missing) = ...;` immediately followed by `let mut missing = missing.drain()...` — `into_iter()` drops both the `mut` on the first binding and the shadow.

## Still open from the first pass

**(7)** `batch.missing` is still decoded and only logged. Fine as staging, noting it so it isn't mistaken for wired up.

## Status

Blocking finding (1) is genuinely resolved — I traced the new control flow rather than taking the commit message for it. The epoch change is correct in substance but needs the `Epoch(0)` window handled before it lands; that is the one thing I'd want addressed. CI on this head is still running (`clippy`, `test 1-3/3`, `integration test` all pending).

sdbondi (commented):
Re-reviewed at `52b484648` (`perf(storage): build the shard-group root tree once per batch, not once per shard`).

## The new commit is sound

`RootProofTree` is a faithful extraction: `compute_proof_for_hashes` becomes `build(..).get_proof(..)`, so the one-shot path is the same code by construction, and `get_proof` only wraps `&self.store` in a fresh `JellyfishMerkleTree` rather than rebuilding anything. The generator now pays one O(n log n) build in `new` instead of one per distinct shard, which is the last piece of per-shard-group work that was still inside the loop — the level-2 memo now saves a traversal rather than a tree build. `RootProofTree` is exported via `pub use tree::*` (`lib.rs:23`).

The `MAX_REQUESTS` justification checks out against the test it cites: `helpers::num_preshards()` is `P256`, so `ShardGroup::all_shards` really is the 256-shard worst case, and `seed << 24` spreads 50 substates across ~50 distinct shards — nearly every substate a level-2 memo miss. The per-substate figure is therefore a conservative bound, not a best case, which is the right way round for an argument about what a flood costs to serve.

One consequence worth noting rather than fixing: `a_batched_proof_matches_the_single_substate_proof` now compares a generator against `generate_substate_proof`, which is itself a generator. It still catches a mis-keyed memo, but it no longer pins the output against an independently-built tree. Given `compute_proof_for_hashes` is now literally `RootProofTree::build(..).get_proof(..)`, there is nothing left for it to disagree with — noting it so the test isn't later read as stronger coverage than it is.

Nit: `proof_cost.rs` re-implements the commit-substates setup that `substate_proofs.rs` already has; `helpers.rs` is imported by both and would be the natural home.

## Still open

Neither finding from the previous round is addressed by this commit:

1. **`consensus.current_epoch()` is `Epoch(0)` until consensus starts**, so `get_local_committee_info(Epoch(0))` fails and both `get_substate` and `get_substate_batch` return an internal error in that window — the batch path including when `include_proofs` is false, which previously worked without consensus. `check_consensus_state()` / `CONSENSUS_NOT_RUNNING` already exists for this (`rpc_impl.rs:434`, `:565`).
2. **`missing` now carries stored-but-unprovable ids** while the proto (`rpc.proto:350`, `:366`) and `SubstateBatch::missing` still document it as "ids the responder holds no record of at any version" — the exact premise a negative cache would act on.

(1) is the one I'd want resolved before this merges. CI on this head is still running.

sdbondi (commented):
Re-reviewed at `80506f9f7` (`fix(rpc): take a proof's shard group from its anchor`).

## Resolved

- **(2) `missing` semantics.** Both docs rewritten, and the new wording says the thing that matters — "never evidence that a substate does not exist… use it to decide what to ask for again, never to conclude absence". That closes the trap for #2528's negative cache.
- **Nit.** `into_iter()` in place of `drain()`.
- **The shard-group fix is better than what I suggested.** Taking the group from `commit_proof.shard_group()` — the header field whose `state_merkle_root` commits exactly those per-shard roots (`committed_block_proof.rs:72-78`) — makes the proof and its anchor consistent *by construction* rather than by picking a matching epoch. It also removes the reason to read committee info at the consensus epoch, so the membership check goes back to the epoch manager's epoch and `include_proofs = false` no longer touches the consensus epoch at all. That half of my finding (1) is genuinely fixed.

## Blocking: `latest_commit_proof` errors where both callers expect `None`

`rpc_impl.rs:261`

`tx.last_executed_get(epoch)?` returns `Err(StorageError::NotFound)` in *both* absent cases, not a zero-height default:

- no `LastExecuted` row at all → `get_raw_key` → `NotFound` (`state_store_rocksdb/src/cf_api.rs:80-83`)
- a row exists but at a different epoch → explicit `NotFound` (`state_store_rocksdb/src/reader.rs:430-437`)

So the `height.is_zero()` → `Ok(None)` branch only fires when a row is stored *at the queried epoch* with height zero. Every other "no anchor" case propagates an error.

Two things follow:

1. **The comment added at `rpc_impl.rs:792-793` is wrong.** "Before consensus starts this is zero, which finds nothing and yields no anchor, so results go out unproven rather than failing" — `last_executed_get(Epoch(0))` finds an error, not nothing: `NotFound` on a node with a row at any other epoch, and `NotFound` on a fresh node with no row. With `include_proofs`, `read_substate_batch` propagates it and the batch 500s. So the `Epoch(0)` failure I raised is fixed for `include_proofs = false` only — the proving path, which is the one this PR is about, still hard-fails on a VN that has restarted and not yet started consensus.
2. **The PR description's "Known gaps" describes behaviour the code does not have.** "A batch from a validator with **no anchor at all** (nothing committed since an epoch change) is still returned to the caller, just not cached" — right after an epoch change the stored `last_executed.epoch` is the *previous* epoch, so `last_executed_get(new_epoch)` is `NotFound` and the request errors instead of returning unproven results. That is the single scenario the gap is written about.

This shape predates the PR (the original `get_substate` code had the same `?`), so it isn't introduced here — but this PR documents it as working in two places and the batch path now depends on it. `tx.last_executed_get(epoch).optional()?` gives both callers the `None` they already assume; `.optional()` is used three times in this file already (`rpc_impl.rs:356`, `:412`, `:418`).

## Non-blocking: an id in the node's group but outside the anchor's fails the whole batch

The membership check now runs against `local_committee_info` at the epoch manager's epoch, while the generator runs against the anchor's group at the consensus epoch. When the committee count *decreases* at a boundary, the node's new group is a superset of the anchor's, so an id can pass `includes_substate_id` and still be outside the generator's group. `SubstateProofGenerator::generate` then returns `Err("… outside this shard group")` (`substate_value_proof.rs:92-95`), which `read_substate_batch` propagates via `?` — all 50 ids fail on account of one.

This is the same class as "shard has no committed state", which this series already decided to degrade to `missing`. Same treatment fits: an id the anchor's group cannot cover goes to `missing`, and the up-front membership check keeps catching the real wrong-node case as a `bad_request`. `attach_substate_proof` has the same exposure through `generate_substate_proof`, at one substate rather than fifty.

Worth noting these two are related: with `.optional()` in `latest_commit_proof`, a node whose anchor is stale simply proves nothing and answers unproven, which also takes most of the pressure off this one.

## Status

Not approving. The `latest_commit_proof` error path is the blocker — it is small to fix, but it means the no-anchor degradation this PR relies on and documents does not actually happen. CI on this head has only just started.

sdbondi (commented):
Re-reviewed at `b0fb384a6` (`fix(rpc): degrade to unproven when there is nothing to anchor to`).

## Both blocking findings resolved

**`latest_commit_proof` no longer errors on a missing anchor.** `tx.last_executed_get(epoch).optional()?` (`rpc_impl.rs:271`) maps both `NotFound` cases to `Ok(None)`, which is what both call sites already assumed. Checked that `.optional()` only swallows `is_not_found_error()` — a real rocksdb failure is `RocksDbError` and still propagates, and the `Block::get` / `get_commit_qc` reads that follow are untouched, so a dangling `last_executed.block_id` is still the hard error it should be. The "Known gaps" scenario in the description now actually behaves the way it is written.

**Outside-the-anchor's-group degrades instead of failing the batch.** `generate` returning `Ok(None)` (`substate_value_proof.rs:92-94`) folds it into the same path as "shard has no committed state": the batch reports it in `missing` and answers for the rest, the single path answers unproven. Verified the single path returns *before* `resp.commit_proof` is set (`rpc_impl.rs:157-162`), so a client can never see an anchor with an empty value proof — it takes the `commit_proof.is_empty()` → `proof = None` branch and the tally holds the result as a fallback. Fail-open stays bounded: nothing on either path is cached as verified.

Dropping `generate_substate_proof` and inlining the generator in `attach_substate_proof` is the right consequence — the wrapper existed only to turn `None` back into an error, which is precisely the behaviour that was wrong. Confirmed no callers remain and the `pub use substate_value_proof::*` glob means nothing else needed updating.

The test moves are sound: `commit_substates` and `PROOF_TEST_TREE_VERSION` in `helpers.rs` are shared by both files, `a_substate_outside_the_shard_group_proves_nothing` now asserts the new contract, and deleting `a_batched_proof_matches_the_single_substate_proof` alongside the function it compared against is consistent — as noted last round, it had become generator-vs-generator and could no longer disagree. The properties that matter are still covered by `proofs_for_a_batch_verify_against_one_shard_group_root` and `a_reused_shard_root_proof_belongs_to_its_own_shard`, both of which verify against an independently computed `group_root`.

Also checked the file list: nine files, all on-topic, no lockfile or generated artifact swept in.

## Nits

- `substate_value_proof.rs:78`, `:106`, `:117`, `:190` still prefix their error messages with `generate_substate_proof:`, which is now a name that exists nowhere in the tree. Grepping a production log for it will find nothing. `SubstateProofGenerator` would be the name to use.
- `rpc_impl.rs:266-269`: the comment pairs the two cases the wrong way round. "A node that has not started consensus since restarting" has a row belonging to *another* epoch (the latter case), not no row at all (the former) — a node with no row has never committed anything. The mapping to `NotFound` is right for both; only the attribution is swapped.

Neither blocks.

## Status

No blocking findings remain from my side. Holding off on a final sign-off until CI is green — `clippy`, `test (1-3/3)` and `integration test` are all still pending on this head, and none of the six heads I have reviewed has yet completed a full run, so I would rather not sign off into pending checks on a change that touches the proof path.

sdbondi (commented):
LGTM at `cd6a4beb0`. All 18 checks green, including `integration test`.

Every finding I raised across six heads is resolved, and I re-verified each against the code rather than the commit messages:

**Blocking, round 1** — an unanchored batch was settled by the first member asked. `race_substate_batch` now tries up to `BATCH_READ_ATTEMPTS` members per chunk, holds an `Unanchored` batch, and serves it only after all of them fail to prove — with a warning, never cached. That is the `CommitteeReadTally` contract, and the `BatchTrust` trichotomy makes the three cases legible in a way the original `bool` did not.

**Blocking, round 4** — `latest_commit_proof` propagated `NotFound` where both callers expected `None`, so the no-anchor degradation the PR relies on and documents did not actually happen. `.optional()?` fixes it; I checked it only swallows `is_not_found_error()`, so a genuine rocksdb failure and a dangling `last_executed.block_id` both still error.

**Epoch mismatch** — resolved better than I suggested. Taking the shard group from `commit_proof.shard_group()` makes proof and anchor consistent by construction rather than by choosing a matching epoch, and removes the reason to read committee info at the consensus epoch at all.

**Degradation** — outside-the-anchor's-group and shard-with-no-committed-state both return `Ok(None)` and fold into one path: reported in `missing` for a batch, answered unproven for a single read. I confirmed the single path returns before `resp.commit_proof` is set, so a client can never see an anchor paired with an empty value proof.

Also verified: the level-2 hoist is faithful (`shard_iter_with_global` ordering, root identity), per-shard memoisation is safe when two shards share a root, `RootProofTree` is a faithful extraction, exclusion is the right shape for a down head, rejecting `DoesNotExist` inside an anchored batch is correct, `missing` is now documented as never being evidence of absence, the response is bounded and id-bound, version skew fails loudly instead of reading as absence, and the `verify_substate_proofs = false` path is unchanged end to end. The `MAX_REQUESTS` argument checks out against the test it cites — `P256`, ~50 substates across ~50 shards, so the per-substate figure is a conservative bound.

Two caveats for the record, neither blocking and both already noted in the description or the code:

- A batch from a validator with no anchor is still returned to the caller unproven (uncached) once every member has failed to prove it. That is the documented gap, and it now behaves as documented.
- `SubstateBatch::missing` is decoded but not yet consumed. The docs now say plainly that it is never evidence of absence, which is what matters for #2528.

Note: GitHub will not let me record a formal approval on your own PR, so this is a comment. Another maintainer still needs to approve before merge.

PR #2542, opened 2026-09-07
feat(swarm): page log files instead of loading them whole

sdbondi:
## What

The swarm web UI's log view fetched every byte of a node log on each poll and rendered it in one pass. On a swarm that had been up for a while that locked the tab up or killed it outright. This reworks how swarm reads and shows logs.

### Paged, mmap-backed reads

New `logfile` module in the swarm daemon returns a byte window of a log file, trimmed to whole lines, read through `memmap2`. The `get_file` RPC now takes an optional `end` and `max_bytes` and returns `{contents, start, end, file_size}`; passing a chunk's `start` back as the next `end` walks backwards through the file with no gaps or repeats. A bare path string is still accepted and means "the tail".

The web UI keeps at most six 256 KiB chunks. It opens on the tail with the newest lines at the top; scrolling down pages further back and pauses follow mode; re-enabling follow clears the buffer and returns to the head. Trimming the buffer holds the scroll position against the oldest rendered line, so paging back does not jump.

### Crash reports on stdout

A process that exits non-zero now prints its panic message — or, failing that, the last 50 lines of its stderr and stdout — to the swarm daemon's own stdout, instead of leaving the reason buried in a file the operator has to go find.

The daemon's log formatter reparses forwarded child output to recover the child's target and level from the message text. That now applies only to records actually forwarded from a child (target `swarm`), so a multi-line report is printed as written rather than being mangled by the naive parser.

### Only live log links

The log list showed `ootle.log` alongside `ootle.1.log` … `ootle.5.log`, so every entry appeared six times over. Rotated siblings are now filtered out.

### Indexer log format

The indexer logs through `tracing`, not log4rs, and its file layer was writing ANSI colour escapes into `indexer.log` — a file that is never a terminal. The view rendered those as literal text, which made indexer logs unreadable.

Two halves to the fix: the indexer's file layer no longer emits ANSI (and its stdout layer only colours when stdout is really a terminal), and the UI's line parser understands `tracing`'s `<timestamp> <LEVEL> [<spans>: ]<target>: <message>` shape and strips escapes, so logs already on disk render correctly too.

## Testing

- 16 unit tests in `tari_swarm_daemon`, covering line-boundary trimming, backwards paging over a multi-megabyte file (reconstructs the file byte for byte), UTF-8 boundaries at every window size, rotated-file detection, both `get_file` request shapes, and panic extraction vs. tail fallback.
- The UI's `parse`/`normalise` was run over real `indexer.log`, validator-node `ootle.log`, wallet-daemon `ootle.log` and both stdout captures from a local swarm run: every line comes out with a level and, where the format carries one, a target, with no escapes left behind.
- `cargo clippy -p tari_swarm_daemon --all-targets` clean; webui `tsc --noEmit` and `eslint --max-warnings 0` clean; `vite build` succeeds.

Not verified in a browser — no extension connected in this environment — so the scroll and follow interactions are worth a quick manual pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12693,6 +12693,7 @@ dependencies = [
  "indexmap 2.13.0",
  "lockfile",
  "log",
+ "memmap2 0.9.9",
  "mime_guess",
  "minotari_node_grpc_client",
  "minotari_wallet_grpc_client",
--- a/applications/tari_indexer/src/bin/main.rs
+++ b/applications/tari_indexer/src/bin/main.rs
@@ -20,7 +20,7 @@
 // WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-use std::{fs, panic, path::PathBuf, process};
+use std::{fs, io::IsTerminal, panic, path::PathBuf, process};
 
 use anyhow::Context;
 use futures::stream;
@@ -85,21 +85,20 @@ fn init_tracing_subscriber(cli: &Cli) -> anyhow::Result<WorkerGuard> {
     let (ootle_log, guard) = tracing_appender::non_blocking(appender);
 
     tracing_subscriber::registry()
-        .with(
-            fmt::Layer::new()
-                .with_writer(ootle_log)
-                .with_filter(filter::Targets::new().with_targets([
-                    ("tari::application", tracing::Level::DEBUG),
-                    ("tari::indexer", tracing::Level::DEBUG),
-                    ("tari::ootle", tracing::Level::DEBUG),
-                    ("tower_http", tracing::Level::DEBUG),
-                ])),
-        )
+        .with(fmt::Layer::new().with_writer(ootle_log).with_ansi(false).with_filter(
+            filter::Targets::new().with_targets([
+                ("tari::application", tracing::Level::DEBUG),
+                ("tari::indexer", tracing::Level::DEBUG),
+                ("tari::ootle", tracing::Level::DEBUG),
+                ("tower_http", tracing::Level::DEBUG),
+            ]),
+        ))
         .with(
             fmt::Layer::new()
                 .without_time()
                 .with_target(false)
                 .with_writer(std::io::stdout)
+                .with_ansi(std::io::stdout().is_terminal())
                 .with_filter(filter::Targets::new().with_targets([
                     ("tari::application", tracing::Level::INFO),
                     ("tari::indexer", tracing::Level::INFO),
--- a/applications/tari_swarm_daemon/Cargo.toml
+++ b/applications/tari_swarm_daemon/Cargo.toml
@@ -39,6 +39,7 @@ lockfile = "0.4.0"
 indexmap = { workspace = true }
 slug = "0.1.6"
 log = { workspace = true }
+memmap2 = { workspace = true }
 mime_guess = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
 serde_json = { workspace = true }
--- a/applications/tari_swarm_daemon/src/logfile.rs
+++ b/applications/tari_swarm_daemon/src/logfile.rs
@@ -1,0 +1,211 @@
+//   Copyright 2024 The Tari Project
+//   SPDX-License-Identifier: BSD-3-Clause
+
+//! Bounded, mmap-backed reads over process log files.
+//!
+//! Swarm log files grow without bound, so nothing here ever reads a whole file: callers ask for a byte window and
+//! get back the lines fully contained in it.
+
+use std::{fs::File, io, path::Path};
+
+use memmap2::{Mmap, MmapOptions};
+
+/// Window size used when a caller does not ask for one.
+pub const DEFAULT_CHUNK_BYTES: u64 = 512 * 1024;
+/// Ceiling on a single window, so one request cannot pull a multi-gigabyte log into memory.
+pub const MAX_CHUNK_BYTES: u64 = 4 * 1024 * 1024;
+/// How far back [`tail_lines`] is willing to look for its lines.
+const TAIL_SCAN_BYTES: u64 = 1024 * 1024;
+
+/// A contiguous byte window of a log file, trimmed to whole lines.
+#[derive(Debug)]
+pub struct LogChunk {
+    pub contents: String,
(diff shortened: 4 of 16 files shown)


sdbondi (commented):
Read the whole change at `10f8a2c8c` (both commits), traced the new module into its callers and checked the `process_dir` / `base_path` split against `instances/manager.rs`. The paging design is sound: `end` returned unchanged plus "drop the leading partial line" gives a gapless, repeat-free backwards walk, and `paging_a_multi_megabyte_file_reconstructs_it_exactly` is the right test for it. The `record.target() != FORWARDED_TARGET` gate in `logger.rs` is a strict improvement — the naive `split_once(']')` parser was mangling any daemon-side record whose message contained a `]`. `is_rotated` matches log4rs' actual `{name}.{}.log` roller pattern (`log4rs_sample.yml:37,54,71,…`) and correctly leaves `json_rpc.log` / `some.name.log` alone.

Two correctness problems below, then some non-blocking items.

## Blocking

### 1. A window with no newline never advances, and the UI spins on it

`applications/tari_swarm_daemon/src/logfile.rs:63-67` — when the window contains no `\n` at all, `start` is set to `end`, so the chunk is empty *and* addresses nothing new.

Client side (`webui/src/routes/LogView.tsx:172-179`) that chunk is appended with `start === oldest.start`. Consequences, in order:

- `atStartOfFile` (`LogView.tsx:120`) never becomes true, so the "Start of file" terminator is unreachable.
- Every subsequent `loadOlder` computes the same `end` and refetches the same empty window.
- After `MAX_CHUNKS` iterations `slice(-MAX_CHUNKS)` (`LogView.tsx:178`) has evicted every chunk that actually had lines. The view is now empty, `el.scrollHeight <= el.clientHeight` holds, and the effect at `LogView.tsx:242-247` re-fires `loadOlder` on each render — `chunks` changes identity every round, so `loadOlder`'s identity changes, so the effect re-runs. That is an unbounded request loop against the daemon with a blank log view.

Trigger is any single log line ≥ `CHUNK_BYTES` (256 KiB) — a dumped substate or a large payload in a validator-node log will do it.

The window is already trimmed to whole lines only as a courtesy; when there is no line boundary to find, returning the raw window with `start` left at the window start keeps paging monotonic and costs one rendered partial line. Whatever the fix, `read_chunk` should guarantee `start < end` whenever `end > 0`, and it's worth a test that pages a file containing one oversized line to completion.

### 2. `tail_lines` drops a real line on any file over 1 MiB

`applications/tari_swarm_daemon/src/logfile.rs:96-99` — the fragment guard runs *after* `.rev().take(n)`:

```rust
let mut lines = text.lines().rev().take(n).map(...).collect::<Vec<_>>();
lines.reverse();
if start > 0 && lines.len() > 1 { lines.remove(0); }
```

The mid-line fragment is the first line of the *whole mapped window*, but once the window holds more than `n` lines `take(n)` has already excluded it. So for any log larger than `TAIL_SCAN_BYTES` this silently deletes the 50th-from-last line of the crash report — and if that line was the `thread '…' panicked at` header, `find_panic` (`crash_report.rs:59`) no longer sees it and the report degrades to the plain tail, which is exactly the case the feature exists for.

Drop the fragment from `text` before taking the last `n` (e.g. skip to the first `\n` in the window when `start > 0`). `tail_lines_returns_the_last_lines_in_order` uses a 14-byte file, so `start > 0` is never exercised — the new test needs a file over `TAIL_SCAN_BYTES`, or make the scan size injectable.

## Non-blocking

### 3. Reversing lines *within* a chunk breaks multi-line entries

`LogView.tsx:207` — `chunks.flatMap((chunk) => chunk.lines.slice().reverse())` reverses every line, not just chunk order. Panic backtraces in `stderr.log`, wrapped log4rs messages and the new crash-report block all render bottom-up. Given that making crash output readable is one of this PR's stated goals, that is an odd place to land. Keeping chronological order inside the view and pinning the tail with `flex-direction: column-reverse` on `.logview` (`theme.css:929`) gets newest-first-on-open and scroll-down-for-older without inverting the lines themselves.

### 4. No way back to newer lines except Follow

`LogView.tsx:178` trims from the front, so the newest chunks are the ones dropped. After paging back six chunks, scrolling *up* just stops — the only route to newer lines is Follow, which discards the buffer and jumps to the tail. The bottom of the view has `.logend` to explain itself; the top has nothing. A matching marker, or a "Load newer" counterpart, would make the asymmetry legible.

### 5. `list_stdout_files` has never found a file, and this PR adds the accessor that fixes it

`webserver/rpc/logs.rs:118-142` builds `instance.base_path.join("stdout.log")`, where `InstanceInfo.base_path` is `Instance::base_path()` (`handle.rs:192`) = `<process dir>/<network>` (`instances/manager.rs:283`). But `forward_logs` writes to `<process dir>/stdout.log` (`instances/manager.rs:257-258`). Every `.exists()` is false, so the stdout list is always empty. Pre-existing, but this PR introduces exactly the right accessor (`Instance::stdout_log_path`, `instance.rs:83`) — plumbing `process_dir` onto `InstanceInfo` closes it in a couple of lines. While there: `stderr.log` is pushed with the display name `"stdout"` (`logs.rs:127` and `:141`).

### 6. `max_chunk_bytes` is written but never read

`logs.rs:176` documents the field as "so a client can size its own buffer", but `ChunkResponse` (`LogView.tsx:35-40`) does not declare it and `CHUNK_BYTES` is a hardcoded 256 KiB. Either clamp the client request against it or drop the field — a documented contract with no consumer will drift.

### 7. Path check is component-wise with no canonicalisation

`logs.rs:186` — `path.starts_with(&context.config().base_dir)` compares path *components*, so `<base_dir>/../../etc/anything.log` satisfies it. Pre-existing and the daemon is a local dev tool, but `get_file` is the one handler that turns a client-supplied path into a file read, and a `canonicalize()` before the prefix check is cheap.

### 8. Smaller things

- `crash_report.rs:41` — `TAIL_LINES.max(PANIC_CONTEXT_LINES)` reads as though it should be a sum. As written the panic header has to fall inside the last 50 lines to be found at all, and then at most 30 lines from it are shown; the `.max()` suggests more context is fetched than is.
- `LogView.tsx:171-186` — if the `setChunks` guard at `:175` bails (follow raced the fetch), state identity is unchanged, so the layout effect at `:223` never runs and `anchor.current` stays set. The next unrelated `visible` change — a level toggle, a search keystroke — will then scroll to a stale anchor. Clearing `anchor.current` in the `finally` closes it.
- `logfile.rs:105` — the SAFETY note covers truncation but not the other mmap hazard: if a log file is replaced by a shorter one mid-read, touching a page past the new EOF raises SIGBUS rather than returning an error. Log rotation is by rename so the mapped inode survives, which is worth saying explicitly since that is the property the safety argument actually rests on.

PR #2543, opened 2026-09-07
fix(indexer): refuse a live head at a version the stream saw spent

sdbondi:
Two follow-ups to #2529 and #2531, in the same three files.

## A destroy with no successor admitted a live head at the spent version

`SubstateCacheInvalidation::destroyed(id, v)` journalled `observed_version = v`, and the guard in `substate_cache_put` refused only `version < observed_version`. A write of `Up { v }` at exactly the destroyed version was therefore admitted, so a lagging committee member could install a spent substate as the live head where a level member would answer `Down`.

Only the no-successor case was open: a destroy with a successor is accompanied by `created(id, v + 1)` in the same batch, and the `max(...)` upsert lifts the floor to `v + 1`, which already refuses `Up { v }`.

One integer cannot express both "nothing below v" and "v itself is spent, unless you are claiming it is down", so the journal now carries the provenance alongside the version. The guard refuses an `Up` result at the floor when the floor came from a destroy, and still admits a `Down` at it — that is a legitimate head, and what a lookup for that version answers.

The flag follows the same comparison the floor does rather than being overwritten: a `created(v + 1)` arriving after a `destroyed(v)` both raises the floor and clears the flag, while a `destroyed(v)` arriving after a `created(v + 1)` does neither. At the one version both a creation and a destroy can show — a substate created and destroyed within a single batch — the spend stands whichever order they arrive in.

The journal is ephemeral, rebuilt from the stream within one retention window, so the migration recreates the table as #2529's did.

## Result-driven retirements were not counted

`substate_cache_retire_ahead` returned `()`, so the retirements #2531 performs from a finalized transaction result never reached `substate_cache_invalidations` — the counter quietly meant "retirements from the stream". It now returns the count, and `retire_committed` feeds it to the metric the way the eviction and refused-stale paths already do.

## Tests

In `store_factory.rs`, alongside #2529's journal tests: a destroy with no successor refuses `Up` at the destroyed version and still admits `Down` at it; a destroy with a successor admits the created version live in both arrival orders; a version both created and destroyed in one batch is spent in both arrival orders; and the count returned by `substate_cache_retire_ahead` matches the rows removed. In `substate_cache/mod.rs`, the end-to-end path through `retire_committed`.

Each was checked to fail without its half of the change, including against a fix written as a plain floor bump (which the "still admits `Down`" case catches) and against an unconditional flag overwrite in the upsert.

### DIFF
--- a/applications/tari_indexer/src/storage_sqlite/migrations/2026-09-07-000000_substate_cache_invalidation_spent/down.sql
+++ b/applications/tari_indexer/src/storage_sqlite/migrations/2026-09-07-000000_substate_cache_invalidation_spent/down.sql
@@ -1,0 +1,11 @@
+drop table substate_cache_invalidations;
+
+create table substate_cache_invalidations
+(
+    substate_id      text    not null primary key,
+    state_version    bigint  not null,
+    substate_version integer not null,
+    invalidated_at   bigint  not null
+);
+
+create index idx_substate_cache_invalidations_expiry on substate_cache_invalidations (invalidated_at);
--- a/applications/tari_indexer/src/storage_sqlite/migrations/2026-09-07-000000_substate_cache_invalidation_spent/up.sql
+++ b/applications/tari_indexer/src/storage_sqlite/migrations/2026-09-07-000000_substate_cache_invalidation_spent/up.sql
@@ -1,0 +1,23 @@
+-- Record whether the version a transition showed is itself spent, alongside the version.
+--
+-- The version alone is a floor: a result below it is refused. It cannot also express that the floor
+-- version is down, which a destroy with no successor leaves true - the successor is what would
+-- otherwise raise the floor past it. Without the provenance a lagging committee member's `Up` at the
+-- destroyed version is admitted as the live head and a spent substate is served as spendable, while
+-- a `Down` at that same version is a legitimate head and must still be admitted.
+--
+-- The journal is derived from the stream and spans a fetch, so it is recreated rather than migrated.
+drop table substate_cache_invalidations;
+
+create table substate_cache_invalidations
+(
+    substate_id      text    not null primary key,
+    state_version    bigint  not null,
+    -- The substate version the transition showed: the version created, or the version destroyed.
+    substate_version integer not null,
+    -- Whether `substate_version` was destroyed rather than created.
+    spent            boolean not null,
+    invalidated_at   bigint  not null
+);
+
+create index idx_substate_cache_invalidations_expiry on substate_cache_invalidations (invalidated_at);
--- a/applications/tari_indexer/src/storage_sqlite/models/substate_cache.rs
+++ b/applications/tari_indexer/src/storage_sqlite/models/substate_cache.rs
@@ -26,6 +26,7 @@ pub struct SubstateCacheInvalidation {
     retires_up_to: Option<u32>,
     retires_nonexistence: bool,
     observed_version: u32,
+    spent: bool,
 }
 
 impl SubstateCacheInvalidation {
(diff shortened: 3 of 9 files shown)


sdbondi (commented):
No blocking findings. Both halves check out against the surrounding code.

## Verified

**The spend provenance closes the hole it claims to.** The old floor admitted `Up { v }` at exactly `observed_version` because `substate_cache_put` refused only `version < observed_version` (`writer.rs:589`), and `destroyed(id, v)` journals `observed_version = v` (`models/substate_cache.rs:62`). The new guard at `writer.rs:602-611` refuses precisely that pair and nothing else — `version == Some(observed_version) && Up`, so a `Down` at the same version still lands, which is what a lookup for a destroyed version legitimately answers.

**The upsert's `case` is genuinely order-independent.** I ran the SET clause standalone against sqlite3 for all four arrival orders — `created(7)`/`destroyed(6)` both ways, and `created(7)`/`destroyed(7)` both ways — and got `(7, spent=0)`, `(7, spent=0)`, `(7, spent=1)`, `(7, spent=1)`. SQLite evaluates every SET expression against the pre-update row, so the `spent` case and the `max(...)` on `substate_version` see the same old value and cannot disagree about which version the floor ended up holding (`writer.rs:855-863`).

**The batching assumption behind "a destroy with a successor is accompanied by `created(v + 1)` in the same batch" holds.** `extend_bufs_from_substate_update` pushes both a `Create` and a `Destroy` into `invalidations_buf` (`network_state_sync/worker.rs:1205-1212`), and the buffer is drained once per state version at `has_more == false` into a single `substate_cache_invalidate` (`worker.rs:909, 941`). A cross-batch split is only possible in stream order (destroy at N, create at N+1), where the plain `max` already lifts the floor and clears the flag.

**`version` and the `Up`'s own version agree at both write callers**, so keying the guard off `entry.version` is safe today: `cached_substate_manager.rs:264` takes `lookup_result.result.version()`, and `cached_substate_manager.rs:366` takes `substate.version()`.

**The metric is not double-counted.** `retire_committed` deletes the rows first, so when the stream later journals the same transition its `retires_up_to` delete matches nothing and adds zero. Anything it does count in between is a genuinely new entry written after the result landed.

**Migration** follows the same recreate-not-migrate precedent as `2026-09-04-000000_substate_cache_invalidation_version`, column order in `schema.rs:158` matches `up.sql`, and `down.sql` restores the previous shape and index exactly. Nothing unrelated is swept into the diff.

## Non-blocking

- `substate_cache/mod.rs:170` — the debug line now fires on every `retire_committed`, including the common case of zero retirements, whereas the two sibling counters gate on it (`prune` at `mod.rs:181`, the stream at `worker.rs:977`). The `{transitions}` half makes it more informative than those, so this is a deliberate difference if you want it; just noting it is the one of the three that will log on a no-op.

- `writer.rs:603` — the guard reads the version from `entry.version` rather than from the `Up`'s own `substate.version()`. Both callers derive one from the other so the pair cannot disagree today, but the guard would silently not fire for a caller that passed a mismatched pair. `matches!(entry.substate_result, SubstateResult::Up { substate } if substate.version() == observed_version)` would make it independent of that. The pre-existing `version < observed_version` guard above has the same dependence, so this is consistency, not a defect.

## Nit

- The fix's protection lasts exactly as long as the journal row does — after `journal_retention` prunes it (`writer.rs:695`), a lagging member's `Up` at the spent version is admissible again. That is the same bound the floor itself has had since #2529, so it is not a regression, but the PR description's "refuse a live head at a version the stream saw spent" is true within the retention window rather than absolutely.

Not built or tested locally; leaving that to CI.

PR #2544, opened 2026-09-08
perf(abi): carry engine responses as their encoding, not a Value tree

sdbondi:
## What

An engine call's response crossed the WASM ABI as `InvokeResult(Result<tari_bor::Value, String>)`. The host built that tree with `to_value`, which is an encode followed by a decode; the guest tore it down with `from_value`, which is an encode followed by another decode. Neither side wanted the tree, and on the guest it is metered.

`tari_bor::RawCbor` holds one CBOR item as its own encoding — decoding captures the item's bytes without interpreting them, encoding writes them back verbatim — so `InvokeResult` now carries the response's encoding. The host encodes once, the guest decodes once. `ComponentManager::set_state` stopped routing state through a `Value` first, for the same reason.

## Results

Measured on this head against the branch point, by `cargo run -p tari_engine --example abi_measure --release --features abi-metrics`. Guest metering points are deterministic, so the first two tables are exact; host times vary between runs and are indicative.

**Per transaction, guest metering points.** These run the real account, faucet and tariswap templates, unchanged between the two runs, so it is a clean comparison.

| Workload | Before | After | Change |
|---|---|---|---|
| Account create + fund | 91,155 | 70,049 | −21,106 (−23%) |
| Tariswap swap | 218,319 | 102,563 | −115,756 (−53%) |
| Account balance read | 34,496 | 17,544 | −16,952 (−49%) |

**Per operation, guest metering points.** Good to about 1%: the probe template gained functions between the runs, which shifts code layout slightly.

| Operation | Before | After | Change |
|---|---|---|---|
| Engine call round trip, 32 B response | 4,140 | 3,110 | −25% |
| `GetState`, state ≥ 1 KiB | 6,600 | 4,323 | −34% |
| `SetState`, state ≥ 1 KiB | 8,093 | 4,942 | −39% |
| `&mut self` method, both halves | 14,693 | 9,265 | −37% |
| Empty invocation, no engine call | 786 | 798 | unchanged |

The last row is the control. An invocation that makes no engine call is untouched, which is what the change predicts, since it only affects how a response crosses back.

**Host serialisation time per transaction.** Two after-runs are shown because the account workload's wall time swings by around 40% between runs; the other two are stable.

| Workload | Before | After, run 1 | After, run 2 |
|---|---|---|---|
| Account create + fund | 15.3 µs | 13.0 µs | 18.6 µs |
| Tariswap swap | 19.0 µs | 14.7 µs | 14.6 µs |
| Account balance read | 4.5 µs | 3.1 µs | 3.2 µs |

Points are user fees, so the first two tables are a direct cost reduction for anyone transacting.

## The wire does not move

`to_value(&T)` already produces the bytes `encode(&T)` would, so a `RawCbor` field and a `Value` field encode identically. **A published template decodes responses with its own compiled-in `Result<Value, String>` and keeps working unchanged**; only a template recompiled against this gets the saving.

Two cases in `tari_bor::Value` are lossy and would move the wire if they arose: `f16`/`f32` widen to `f64`, and indefinite-length items collapse to definite. Neither can occur — there is no float field anywhere in `tari_template_lib` or `tari_template_lib_types`, and minicbor's derive never emits indefinite-length items.

Tests in `args/result.rs` pin byte identity against the exact shape templates were published with, across null, both integer extremes, bytes, text, arrays, maps, tags and the error variant. Note what that does **not** cover: every `Value` variant survives the round trip by construction, so adding an `f32` to a response type would fail no test. `raw.rs` documents that boundary directly instead — `a_float_is_the_one_shape_a_value_does_not_reproduce` asserts the widening, so the constraint is written down and pinned, but keeping response types float-free is a convention, not something the build enforces. The practical exposure is nil either way: a published template decodes an `f32` item into `Value::Float(f64)` with the same numeric value it got before.

## How this was decided

The first commit is measurement only: an `abi-metrics` feature on `tari_engine` recording per-op host timing and wire bytes, a `metrics` feature on `tari_template_lib`, an `abi_probe` template that pairs each ABI hop with a twin doing identical work minus the boundary crossing, and `abi_measure` to drive them. Both features compile to empty inline functions when off, so a validator build carries none of it.

That measurement is what rejected the original plan. A zero-copy codec (rkyv) targets per-byte marshalling, and per-byte turned out to be near free in metering points: `memory.copy` is charged a flat 2 points whatever its length. The cost was per call and structural, which is what this PR removes instead.

Two things it also says no to:

- **Replacing the guest's `encoded_len` pre-pass with a fixed reserve.** Wins ~70 points/call in the probe, wins nothing across real transactions: 191,048 points with the pre-pass against 191,058 with a 1 KiB reserve, and a 4 KiB reserve gives byte-identical counts to the 1 KiB one. A guest heap busy with the template's own allocations does not behave like the probe's quiet one. The pre-pass stays; the probe stays with it so the trade can be re-tested.
- **Removing the `GetState`/`SetState` round trips entirely** by carrying state in the `CallInfo`. That is where the remaining ~9,400 points per `&mut self` method live, but it changes hop 1's framing and needs a template ABI version the engine switches on — a migration, not a refactor. Not attempted here.

## Separate finding, not addressed here

Metering charges `memory.copy` a flat 2 points regardless of length, which prices copying about 10,000x cheaper per microsecond of validator time than arithmetic. The 250M point per-transaction cap buys ~30 ms of arithmetic but ~336 s of copying at the 4.2 GB/s the copies run at. Reproducible with the `bulk_copy` and `spin` probes. This wants its own issue.

## Versions

`tari_bor` 0.15.0 → 0.15.1: the one crate here whose released and working versions matched. `RawCbor` is additive, which `scripts/crate_versioning.py impact` prices as a patch, and the workspace pin moves with it so a published `tari_template_lib` cannot resolve a `tari_bor` predating the type it compiles against. `tari_template_abi`, `tari_template_lib`, `tari_engine` and the tier-3 rollup already sit ahead of their released versions, so the breaking `InvokeResult::from_value` signature rides the bump they carry.

## Testing

`tari_bor` and `tari_template_lib` unit tests, and the `tari_engine` suites: `test`, `tariswap`, `account`, `fees`, `resource`, `fungible`, `events`, `cross_template`, `access_rules`, `caller_context`, `recall`, `freeze`, `intrinsics`, `limits`, `metering_budget`, `compute_fee_budget`, `validator_fees`, `complex_fee_payment`. Clippy and fmt clean.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11205,7 +11205,7 @@ dependencies = [
 
 [[package]]
 name = "tari_bor"
-version = "0.15.0"
+version = "0.15.1"
 dependencies = [
  "borsh",
  "indexmap 2.13.0",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -145,7 +145,7 @@ tari_template_lib = { version = "0.32", path = "crates/template_lib" }
 tari_template_lib_types = { version = "0.32", path = "crates/template_lib_types", default-features = false }
 tari_template_abi = { version = "0.20", path = "crates/template_abi", default-features = false }
 tari_template_macros = { version = "0.23", path = "crates/template_macros" }
-tari_bor = { version = "0.15.0", path = "crates/tari_bor", default-features = false }
+tari_bor = { version = "0.15.1", path = "crates/tari_bor", default-features = false }
 
 ootle_byte_type = { path = "crates/ootle_byte_type", version = "0.13" }
 ootle_serde = { path = "crates/ootle_serde", version = "0.5", default-features = false }
--- a/crates/engine/Cargo.toml
+++ b/crates/engine/Cargo.toml
@@ -37,6 +37,11 @@ memmap2 = { workspace = true, optional = true }
 
 [features]
 default = []
+# Instruments the WASM ABI boundary (`tari_engine::abi_metrics`) so
+# `examples/abi_measure.rs` can attribute host time and wire bytes per engine
+# op. Measurement only: it adds two clock reads per ABI phase and is never
+# enabled by a validator build.
+abi-metrics = ["tari_template_lib/metrics"]
 # Enables the on-disk wasmer module cache (`tari_engine::wasm::cache`). Pulls
 # in `tari_template_builtin` to bypass caching for builtin addresses, plus the
 # filesystem I/O surface (memmap + bytes). Disable if embedding the engine in
--- a/crates/engine/examples/abi_measure.rs
+++ b/crates/engine/examples/abi_measure.rs
@@ -103,10 +103,9 @@ mod measure {
         /// Points for an engine call whose response is a bare `()`: envelope encode, the host call
         /// and the length-prefixed free.
         bare_call: f64,
-        /// Points to bring a small response back as `InvokeResult(Value)`, over [`Self::bare_call`].
-        response_value: f64,
-        /// Points to convert that `Value` into the concrete type.
-        from_value: f64,
+        /// Points to decode an engine call's response into the concrete type, over
+        /// [`Self::bare_call`].
+        response_decode: f64,
     }
 
     fn probe_points() -> Fits {
@@ -168,15 +167,18 @@ mod measure {
         println!("One engine call with payloads of a few tens of bytes, by stage (guest points):");
         println!("  argument encode + host call + unit response     {bare_call:>8.0}   (EmitLog, empty message)");
         println!(
-            "  response arrives as InvokeResult(Value)         {:>8.0}   (CallerContextInvoke, 32-byte key)",
+            "  response decoded into the concrete type         {:>8.0}   (CallerContextInvoke, 32-byte key)",
+            floor - bare_call
+        );
+        println!("  total                                          {floor:>8.0}");
+        println!(
+            "  the same response built as a Value tree         {:>8.0}   (what a response no longer does)",
             raw - bare_call
         );
-        println!("  from_value into the concrete type              {:>8.0}", floor - raw);
-        println!("  total                                          {floor:>8.0}");
 
         let (state_get_typed, state_get_value, state_set) = state_fits(&mut test);
-        let from_value_fixed = state_get_typed.intercept - state_get_value.intercept;
-        let from_value_slope = state_get_typed.slope - state_get_value.slope;
+        let value_tree_fixed = state_get_value.intercept - state_get_typed.intercept;
+        let value_tree_slope = state_get_value.slope - state_get_typed.slope;
 
         println!(
             "\nOne argument costs {:.0} points before its first byte: the {noop}-point empty invocation is already \
@@ -197,8 +199,8 @@ mod measure {
             println!("{:<34} {:>14.0} {:>10.3} /{}", f.label, f.intercept, f.slope, f.unit);
         }
         println!(
-            "\nGuest-side `from_value` on a GetState response: {from_value_fixed:.0} points fixed, \
-             {from_value_slope:.3} points/byte"
+            "\nBuilding a Value tree from a GetState response rather than decoding it: {value_tree_fixed:.0} points \
+             fixed, {value_tree_slope:.3} points/byte"
         );
 
         Fits {
@@ -207,8 +209,7 @@ mod measure {
             hop3_arg,
             state_get_typed,
             bare_call,
-            response_value: raw - bare_call,
-            from_value: floor - raw,
+            response_decode: floor - bare_call,
         }
     }
 
@@ -247,7 +248,7 @@ mod measure {
         table("Hop 3: one SetState call", "state bytes", &set);
         (
             fit("hop 4: GetState into a type", "state byte", &typed),
-            fit("hop 4: GetState, response Value only", "state byte", &value_only),
+            fit("hop 4: GetState, response as a Value", "state byte", &value_only),
             fit("hop 3: SetState", "state byte", &set),
         )
     }
@@ -336,11 +337,13 @@ mod measure {
     /// encodes a realistic `SetState` argument, varying only how the destination buffer is obtained.
     fn buffer_strategies() {
         const COUNTS: [u32; 2] = [64, 512];
-        const STRATEGIES: [(&str, u32); 4] = [
+        const PAYLOADS: [u32; 4] = [64, 512, 4_096, 16_384];
+        const STRATEGIES: [(&str, u32); 5] = [
             ("encoded_len, exact buffer", 0),
             ("empty buffer, grown", 1),
             ("512-byte buffer", 2),
             ("1024-byte buffer", 3),
+            ("4096-byte buffer", 4),
         ];
 
         let mut test = TemplateTest::new(CRATE_PATH, [PROBE]);
@@ -348,14 +351,14 @@ mod measure {
 
         println!("\n== Guest points per engine-call argument encode ==");
         print!("{:<28}", "buffer strategy");
-        for size in [64u32, 512, 4096] {
+        for size in PAYLOADS {
             print!("{:>16}", format!("{size} B payload"));
         }
         println!();
         println!("{}", "-".repeat(76));
         for (label, strategy) in STRATEGIES {
             print!("{label:<28}");
-            for size in [64u32, 512, 4096] {
+            for size in PAYLOADS {
                 let low = method_points(&mut test, component, "encode_buffered", args![
                     COUNTS[0], size, strategy
                 ]);
@@ -442,9 +445,12 @@ mod measure {
         name: &'static str,
         wall_ns: u64,
         wasm_points: u64,
+        /// Whether the trials disagreed on points, i.e. the workload did not run the same
+        /// transaction each time.
+        points_vary: bool,
         census: abi_metrics::Census,
-        to_value: value_metrics::Phase,
-        from_value: value_metrics::Phase,
+        result_encode: value_metrics::Phase,
+        result_decode: value_metrics::Phase,
     }
 
     impl WorkloadCensus {
@@ -457,8 +463,8 @@ mod measure {
                     .ns
                     .saturating_sub(self.census.guest_alloc().ns) +
                 self.census.return_decode().ns +
-                self.to_value.ns +
-                self.from_value.ns
+                self.result_encode.ns +
+                self.result_decode.ns
         }
 
         fn engine_calls(&self) -> u64 {
@@ -517,6 +523,7 @@ mod measure {
         execute(test);
 
         let mut best: Option<WorkloadCensus> = None;
+        let mut all_points = Vec::with_capacity(CENSUS_TRIALS);
         for _ in 0..CENSUS_TRIALS {
             abi_metrics::reset();
             value_metrics::reset();
@@ -527,23 +534,37 @@ mod measure {
                 name,
                 wall_ns,
                 wasm_points: test.last_execution_points().wasm,
+                points_vary: false,
                 census: abi_metrics::snapshot(),
-                to_value: value_metrics::to_value(),
-                from_value: value_metrics::from_value(),
+                result_encode: value_metrics::encode(),
+                result_decode: value_metrics::decode(),
             };
+            all_points.push(candidate.wasm_points);
             if best.as_ref().is_none_or(|best| candidate.wall_ns < best.wall_ns) {
                 best = Some(candidate);
             }
         }
-        best.expect("at least one trial")
+        let mut best = best.expect("at least one trial");
+        // Every figure reported describes one trial, so a ratio drawn between two of them is a
+        // ratio over the same transaction. Points are deterministic per transaction, so trials
+        // that disagree mean the workload mutated state and ran a different transaction each time;
+        // that makes the points column incomparable across runs, which the flag says rather than
+        // the numbers hiding it.
+        best.points_vary = all_points.iter().any(|p| *p != all_points[0]);
+        best
     }
 
     fn report_workload(workload: &WorkloadCensus, timer_ns: f64) {
         println!("\n== Census: {} ==", workload.name);
         println!(
-            "Wall time {:.1} µs, {} WASM points, {} engine calls",
+            "Wall time {:.1} µs, {} WASM points{}, {} engine calls",
             workload.wall_ns as f64 / 1000.0,
             workload.wasm_points,
+            if workload.points_vary {
+                " (lowest; trials differ, so this workload is not a stable point comparison)"
+            } else {
+                ""
+            },
             workload.engine_calls()
         );
         println!();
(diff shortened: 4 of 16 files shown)


sdbondi (commented):
Reviewed at `6b2f257`. Nothing blocking — the wire-identity argument holds where it matters, and I checked the two places it could have gone wrong rather than taking the description's word for it.

## Verified

- **`unit()` keeps its bytes.** minicbor's `impl Encode for ()` is `e.array(0)?` (minicbor 2.3.0 `encode.rs:328`), i.e. `0x80`, the same byte `Value::Array(vec![])` emits. The 19 `InvokeResult::unit()` sites in `runtime/impl.rs` are unaffected, and `unit_keeps_its_wire_form` pins it.
- **`encoded_len` stays exact.** `RawCbor::cbor_len` returns `self.0.len()` and `encode` writes those bytes verbatim, so the `encoded_len(&resp)` in `wasm/process.rs:421` still sizes the guest allocation exactly. A mismatch here would have been a short write into linear memory; it isn't one.
- **`set_state` cannot move a substate.** The `to_value` drop in `component/manager.rs:100` looked like the one place the "f16/f32 widen, indefinite collapses" caveat could bite, since `set_state<T>` is generic over *user* state, not over `tari_template_lib` types. It can't: `ComponentAction::SetState` decodes the single arg into a `tari_bor::Value` (`runtime/impl.rs:1321` → `ComponentBody::set`, which stores `tari_bor::Value`), so the host normalises the argument exactly as `to_value` used to. Component state is byte-identical either way, including for a float-carrying template.
- **`GetState` likewise.** `runtime/impl.rs:1304` responds with `InvokeResult::encode(component.state())` over a `&Value`, so `RawCbor::from_encodable` writes precisely the bytes the old `to_value` + encode wrote.
- **`d.skip()` is iterative.** minicbor 2.3.0 `decode/decoder.rs:550` uses an explicit heap stack, so the comment at `raw.rs:77` is accurate — no recursion for nesting depth to drive.
- **Versions.** crates.io says `tari_bor` 0.15.0, `tari_template_lib` 0.31.1, `tari_template_abi` 0.19.1, `tari_engine` 0.40.0. The 0.15.1 patch is the only bump needed, and the breaking `from_value` signature does ride the 0.32/0.20/0.41 bumps already in the tree, as claimed.
- **`cargo test -p tari_bor -p tari_template_lib --lib`**: 37 + 9 pass, the new wire-form tests among them.
- `abi_probe` is referenced only by the example, so it costs the engine test suite nothing. CI runs clippy `--all-targets --all-features`, so the `#[cfg(feature = "abi-metrics")]` half of `abi_measure.rs` is type-checked rather than left to rot.

## Non-blocking

**1. `crates/tari_bor/src/raw.rs:5` — unused import in every no_std build.**
`ToString` is only used by the `#[cfg(test)]` module, which builds with std. Confirmed:

```
$ cargo check -p tari_bor --no-default-features --features alloc
warning: unused import: `string::ToString`
 --> crates/tari_bor/src/raw.rs:5:13
```

Every template compiles `tari_bor` with `default-features = false`, so this warning now shows up in every template build. Workspace clippy builds with std and won't catch it.

**2. Commit 2 left the measurement scaffolding describing the thing it removed.** The numbers are still right, but the words aren't, and the next person to re-run the probe reads the words:

- `examples/abi_measure.rs:106,108` — `response_value` is documented as "points to bring a small response back as `InvokeResult(Value)`" and `from_value` as "points to convert that `Value` into the concrete type". After the change, `raw` (via `caller_context_raw`) measures `into_value()`, a Value-tree decode, and `floor` measures a typed decode, so `from_value: floor - raw` (line 214) is the *difference between two decode strategies* — plausibly negative — not the cost of a conversion that no longer exists. The summary is unharmed because the two telescope to `floor - bare_call`, and the printlns at 169–178 are worded correctly; it's the field names and docs that are stale.
- `examples/abi_measure.rs:674` — "anything larger arrives as an `InvokeResult` and pays the `Value` detour". That detour is what this PR deleted.
- `tests/templates/abi_probe/src/lib.rs:193,222` — both say the difference "is the `InvokeResult` `from_value` conversion into a concrete type, which no change of codec removes". The change removes it.
- `crates/template_lib/Cargo.toml:32` — "Times the `tari_bor::Value` conversions an `InvokeResult` performs". There are none left; it times the `RawCbor` encode/decode.

**3. The description's build-time guarantee isn't one.** "Tests in `args/result.rs` pin byte identity … so a future float field fails the build rather than the network" — the tests enumerate `Value` variants, and every one of them survives the round trip by construction. The lossy direction is `to_value(&f32)` widening to f64, and nothing exercises it: adding an `f32` field to a response type would fail no test. Worth either an `assert_same_wire_form`-style case built from a typed value with an `f32` (which would fail today, documenting the boundary), or softening the claim. The practical exposure is nil — an already-published template decodes an f32 item into `Value::Float(f64)` with the same numeric value it got before — but the stated mechanism isn't there.

## Nits

- `crates/template_abi/src/abi/mod.rs:62` — "Determine the exact required buffer size to allocate" restates `encoded_len`, and the line is otherwise unrelated to this PR.
- `RawCbor::is_empty` can never return true for a value built through either constructor, since a complete CBOR item is at least one byte. Fine as clippy's `len`/`is_empty` pairing, just noting it's not a meaningful predicate.

Unrelated to this branch: `cargo test -p tari_template_lib` fails a doctest at `crates/template_lib/src/lib.rs:43` (`Component`/`Vault` unresolved in the `#[template]` example). That file is untouched here and the failure is in macro expansion, not in anything this PR moves.

sdbondi (commented):
Re-reviewed at `a47201c`. The only delta against `6b2f257` is `raw.rs`: the derived serde impls are replaced with hand-written ones that see through to the `Value`.

That's the right shape. `RawCbor`'s derived `Serialize` would have emitted the newtype's `Vec<u8>` as a JSON array of integers, so any field that swapped `Value` for `RawCbor` would have silently changed its JSON. Verified with `cargo test -p tari_bor --features serde`: 54 pass, `json_matches_the_value_it_holds` among them. Nothing in the tree serde-serialises an `InvokeResult` today, so this is about the type being safe to reuse rather than a bug fixed. `BorError`'s `Display` impl is ungated, so `Error::custom` is fine in no_std too.

Two things on the new commit:

**Test set omits the two variants that don't have a natural JSON form.** `raw.rs:166` covers null, bool, small int, bytes, text, array and text-keyed map. It skips `Value::Tag` and an `Integer` outside the i64/u64 range — precisely the cases `value_serde.rs` routes through the `@cbor` sentinel, and so the ones where a round trip is most likely to break. The CBOR-side test set at `result.rs` does include a tag. Adding `Value::Tag(42, ..)` and a non-text-keyed map to that loop would make the coverage match the claim in the doc comment.

**Deserialize is canonicalising, not byte-preserving.** JSON → `Value` → `from_value` re-encodes, so a `RawCbor` built from non-canonical CBOR won't come back byte-identical through a serde round trip, and `Serialize` is now fallible for any item `Value` can't represent (which `d.skip()` will happily have accepted on the way in). Both are fine for the intended JSON-only use and the doc says as much; noting it so the asymmetry with the CBOR path is on the record.

**Unchanged from the previous review** — none of these were touched by the force-push:

- `raw.rs:5` — `ToString` is still unused in no_std; `cargo check -p tari_bor --no-default-features --features alloc` still warns, and every template build compiles this crate that way.
- `abi_measure.rs:106,108,674`, `abi_probe/src/lib.rs:193,222`, `template_lib/Cargo.toml:32` still describe the `Value` detour commit 2 removed.
- The description's "a future float field fails the build" still has no test behind it.
- `template_abi/src/abi/mod.rs:62`'s stray comment.

Unrelated, and pre-existing on `development`: `cargo check -p tari_bor --no-default-features --features alloc,serde` fails with 29 errors — missing `alloc` prelude imports in `value_serde.rs`, `serde_codec.rs`, `adapters.rs` and `byte_counter.rs`. No file this PR touches is implicated, and the new `raw.rs` impls are clean in that config; the combination just isn't built by anything today.

sdbondi (commented):
Re-reviewed at `d278e10`. Every finding from the two previous rounds is resolved, and I re-checked each rather than taking the commit's word for it.

- `raw.rs:5` — `ToString` moved into the test module. `cargo check -p tari_bor --no-default-features --features alloc` is now clean, so template builds stop carrying the warning.
- `into_bytes` / `len` / `is_empty` dropped rather than kept unused. Confirmed nothing references them.
- `Fits::response_value` + `from_value` collapsed into `response_decode: floor - bare_call`. Identical arithmetic to the old sum, and the name now matches what it measures. `raw` is still live at `abi_measure.rs:164,176` for the printed comparison, so no dead binding for clippy to find.
- `op_points`, both `*_raw` probe docs and `template_lib/Cargo.toml:32` no longer describe the `Value` detour. `template_abi/src/abi/mod.rs` comment gone.
- `a_float_is_the_one_shape_a_value_does_not_reproduce` pins the widening it depends on, with the right byte counts (0xFA + 4 against 0xFB + 8; minicbor's `f64()` never shortens, and `Value`'s decode widens F16/F32 through `d.f64()`).
- `an_integer_beyond_cbor_range_cannot_be_held` is a good answer to the coverage gap I raised — it shows the `@cbor: int` sentinel is unreachable through a `RawCbor` rather than leaving the omission unexplained.
- The JSON set now carries `Value::Tag` twice, a non-text-keyed map and `u64::MAX`, which is the sentinel surface. `cargo test -p tari_bor --features serde`: 56 pass.
- The serde doc comment now states the canonicalising round trip and the fallible serialise.

One thing left, in the description rather than the code: it still says the `args/result.rs` tests mean "a future float field fails the build rather than the network". The new `tari_bor` test pins that `to_value` widens f32, so it fails if minicbor's behaviour changes — but adding an `f32` field to a response type still fails nothing. The code is now honest about the boundary; the PR body sentence isn't yet.

Nothing blocking. CI on this head is still running (clippy, test 1-3, fuzz, integration, ledger, docs) — worth waiting for it before merge, since none of what I ran locally covers the engine suites. Note that a `gh` approve won't stick on your own PR; this needs another maintainer's review to land.

sdbondi (commented):
Re-reviewed at `e77fbbc`. Delta is `Vec<u8>` → `Box<[u8]>` in `RawCbor`, nothing else. Everything resolved in `d278e10` stays resolved; `cargo check -p tari_bor --no-default-features --features alloc` is clean and `cargo test -p tari_bor --features serde` still passes 56.

The type change is sound — `as_bytes`, `decode_exact(&self.0)`, `PartialEq`, `Clone` and `Debug` all behave identically through the deref, and the `Decode` side is strictly better: `Box::from(bytes)` allocates exactly `len` where `to_vec` allocated a growable buffer. The guest, which only ever builds a `RawCbor` through `Decode`, gets the whole benefit.

One thing to weigh, non-blocking:

**`from_encodable` now pays a shrink on the host's hot path** (`raw.rs:31`). `minicbor::to_vec_with` is `Encoder::new(Vec::new())` (minicbor 2.3.0 `lib.rs:227`), so the buffer grows by doubling and `capacity > len` for essentially every response — a 40-byte response lands in a 64-byte buffer. `into_boxed_slice` then calls `shrink_to_fit`, i.e. a realloc per `InvokeResult::encode`. That is a new allocator call on precisely the path this PR set out to make cheaper, bought with one word of a struct that lives for the duration of one engine call. It doesn't touch the guest points in the results table, but it does land in the host serialisation figure the PR reports (15.3 → 12.4 µs).

If the word matters, `from_encodable` could size the buffer exactly instead of shrinking it — though that needs a `CborLen` bound it doesn't currently carry, which would propagate to `InvokeResult::encode`. Given `wasm/process.rs:421` already runs `encoded_len` over the finished `InvokeResult` one level up, the simpler read may be that `Vec<u8>` was fine and the eight bytes weren't worth an allocator round trip. Your call — it is small in either direction, I'd just rather it be a decision than a side effect.

Still open from the last round, description only: the body's "a future float field fails the build rather than the network" isn't what `a_float_is_the_one_shape_a_value_does_not_reproduce` does — it pins minicbor's widening, not the absence of float fields.

CI on this head is still running. Nothing blocking from me.

sdbondi (commented):
Re-reviewed at `feebf98`. Two changes since `e77fbbc`: the `points_vary` handling in `abi_measure.rs` and a doc line on `RawCbor`.

The `points_vary` catch is a good one — `create_funded_account` and `swap` mutate state, so each trial really is a different transaction and the points column was previously whichever one happened to win on wall time. Flagging it in the output beats silently reporting a number the reader would take as comparable.

One consequence worth a look, non-blocking: `wasm_points` now comes from the min-points trial while `census`, `result_encode`/`result_decode` and `wall_ns` still come from the fastest-wall-time trial (`abi_measure.rs:523-548`). When the trials genuinely differ those are two different transactions, so the "of WASM pts" percentage in the summary and the "of {} measured" line in `per_hop_table` divide one trial's estimated ABI points by another trial's total. The `points_vary` note tells the reader the column is unstable, so nothing is misreported — but the mismatch is new, where before both halves at least described the same run. Taking the whole `WorkloadCensus` from the min-points trial, or keeping the fastest-wall trial's own points and letting the flag carry the caveat, would keep numerator and denominator on the same transaction.

`Nothing mutates them afterwards` on `raw.rs:23` reads as the answer to the `Box<[u8]>` question I raised last round — taken as decided, and it is the right place to record why the buffer doesn't need to grow.

I haven't compiled the example locally (it needs the `abi-metrics` feature and pulls the engine with it); CI's `clippy --all-targets --all-features` covers it. Everything from the earlier rounds stays resolved, and the only outstanding item is still the description's float sentence. Nothing blocking.

sdbondi (commented):
Re-reviewed at `4af868e`. The min-points splice is gone, so `wasm_points`, `census`, `wall_ns` and the `result_*` phases all come from the one trial again and every ratio in the summary and `per_hop_table` is drawn over the same transaction. `all_points` is still live for the `points_vary` check and `best` still needs its `mut`, so nothing is left dangling. That closes the last item I raised.

Summary across the six rounds: the wire-identity argument is verified where it matters — `unit()`'s `0x80`, `RawCbor::cbor_len` keeping `encoded_len` exact for the write into linear memory, `GetState` encoding the same `&Value` it always did, and `SetState` unable to move a substate because the host decodes that argument into a `tari_bor::Value` regardless. Version bumps check out against crates.io. `cargo test -p tari_bor --features serde` and `-p tari_template_lib --lib` pass locally.

Nothing blocking from me. The engine suites are what actually exercise this and I have not run them, so CI is the gate — it is still spinning up on this head. A `gh` approve won't stick on your own PR, so this still needs another maintainer.

Only leftover: the description's "a future float field fails the build rather than the network". `a_float_is_the_one_shape_a_value_does_not_reproduce` pins that `to_value` widens f32, which is not the same guarantee.

PR #2546, opened 2026-09-08
refactor(bor)!: make encoded_len infallible

sdbondi:
Follow-up to #2540, on a point Stan raised after merge.

## The problem

`Transaction::encoded_size` resolved an unmeasurable size to `usize::MAX`, chosen so a size cap would reject rather than admit it. That only holds while every caller is a comparison — a caller sizing an allocation from the same value aborts the process. The opposite sentinel (`0`) silently passes the cap instead. No number is right for both.

That is a symptom. The cause is `tari_bor::encoded_len`, whose signature says it can fail when it cannot:

```rust
pub fn encoded_len_with<C, T: CborLen<C> + ?Sized>(val: &T, ctx: &mut C) -> Result<usize, BorError> {
    Ok(minicbor::len_with(val, ctx))   // unconditionally Ok
}
```

The `CborLen` bound is already enforced at compile time, so the length always exists. The `Result` was kept for API compatibility and against minicbor changing.

## Why it is worth the breaking change

Every call site resolves the impossible branch its own way — `?`, `unwrap`, `expect("ENCDLENFAIL")`, `unwrap_or(0)` — and `fee_estimate.rs:171` gave up and defined a private infallible shadow with the same name. Two of them size allocations from it:

```rust
let len = encoded_len(val).expect("ENCDLENFAIL");   // template_abi/src/abi/mod.rs:125
let ptr = internal_alloc(len);
let len = encoded_len(&input).unwrap();             // template_abi/src/abi/mod.rs:62
let mut encoded_input = Vec::with_capacity(len);
```

So the hazard behind the original review comment is real and already present in the codebase — it just has not fired, because the branch cannot be taken.

It also removes a **fail-open**: `calc_stealth_statement_weight` summed `encoded_len(..).unwrap_or(0)` over witness bytes, so an error there would have undercharged a transaction's weight at ingress rather than rejecting it. Same class of bug as the one that started this, in the opposite direction.

## What changed

`encoded_len` and `encoded_len_with` return `usize`. Callers lose a `?`, an `unwrap`, an `expect` or an `unwrap_or` each. **Behaviour is unchanged** — the branch that is gone could not be taken.

Kept fallible, because they genuinely are:
- `encoded_len_with_limit*` — errors when the length exceeds its limit
- `encoded_len_via_writer` — drives a real encoder

`Transaction::encoded_size` is back to returning `usize`, with no sentinel and no unwrap.

## Versioning

`tari_bor` 0.15.1 → 0.16.0, and the workspace pin with it. The rest of the cascade from `scripts/crate_versioning.py impact tari_bor --breaking` — workspace rollup to 0.42.0 plus 18 dependent bumps — is a release step and deliberately **not** in this PR: it touches every crate's version and would collide with anything else in flight. Worth doing before the next publish, not before this merge.

## Verification

```
cargo lints clippy -p tari_bor -p tari_template_abi -p tari_ootle_transaction -p tari_engine \
  -p tari_engine_types -p tari_ootle_transaction_validation -p tari_consensus -p tari_template_lib \
  -p tari_validator_node -p tari_indexer -p tari_ootle_walletd -p ootle_serde \
  -p tari_ootle_wallet_sdk -p tari_ootle_template_metadata -p tari_template_lib_types \
  -p tari_indexer_client --all-targets      # clean
cargo test -p tari_bor -p tari_template_abi -p tari_ootle_transaction \
  -p tari_ootle_transaction_validation -p tari_consensus --lib     # 58 + 17 + 20 + 43 + 63 pass
```

Not verified: the WASM guest side. `template_abi` is compiled into templates, and while the change is signature-only, nothing here has run a template end-to-end — `abi_probe`'s call site is migrated but the integration suite is CI's.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11205,7 +11205,7 @@ dependencies = [
 
 [[package]]
 name = "tari_bor"
-version = "0.15.1"
+version = "0.16.0"
 dependencies = [
  "borsh",
  "indexmap 2.13.0",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -145,7 +145,7 @@ tari_template_lib = { version = "0.32", path = "crates/template_lib" }
 tari_template_lib_types = { version = "0.32", path = "crates/template_lib_types", default-features = false }
 tari_template_abi = { version = "0.20", path = "crates/template_abi", default-features = false }
 tari_template_macros = { version = "0.23", path = "crates/template_macros" }
-tari_bor = { version = "0.15.1", path = "crates/tari_bor", default-features = false }
+tari_bor = { version = "0.16.0", path = "crates/tari_bor", default-features = false }
 
 ootle_byte_type = { path = "crates/ootle_byte_type", version = "0.13" }
 ootle_serde = { path = "crates/ootle_serde", version = "0.5", default-features = false }
--- a/crates/engine/src/runtime/working_state.rs
+++ b/crates/engine/src/runtime/working_state.rs
@@ -248,7 +248,7 @@ impl<TStore: StateReader> WorkingState<TStore> {
         if value.published_template().is_some() {
             return Ok(());
         }
-        let size = encoded_len(value)?;
+        let size = encoded_len(value);
         if size > limits::ENGINE_LIMITS.max_substate_size {
             return Err(LimitError::SubstateSizeExceeded { size }.into());
         }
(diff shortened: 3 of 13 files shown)


sdbondi (commented):
~~ACK~~ let's make encoded_len infallible, then ACK

sdbondi (commented):
ACK

PR #2547, opened 2026-09-08
fix(state_sync): serve state from the current committee, validate against the previous

sdbondi:
## Description

State sync used the previous epoch's committee for two unrelated jobs: the quorum that makes an `EpochCheckpoint` valid, and the set of peers asked to hand over the bytes. Only the first needs to be the previous committee. Using it for the second preferentially targets validators that may have left the register (`Idle`) or shut down, while continuing members of the same shard group are `Running` and hold identical state.

This introduces `SyncSource`, pairing the signing committee (prev epoch, used only by `validate_checkpoint`) with an ordered serving-peer list drawn from the current epoch's committees overlapping the target shard group:

1. continuing members
2. members that joined at this epoch (they synced from the same checkpoint)
3. departed prev-only members, as the fallback for full committee turnover

Each tier is shuffled so load still spreads. `sync_shard`'s retry pool is the serving list.

Unchanged: epoch-0 `NoCommittees(Epoch::zero())`, first-committee and sole-VN early-outs, and the literal error-string retry classification (converting that to a typed RPC error is a separate improvement).

## How Has This Been Tested?

`cargo lints clippy -p tari_rpc_state_sync --all-targets`. Integration tests via CI.

## Breaking Changes

- [x] None

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12533,6 +12533,7 @@ dependencies = [
  "futures 0.3.31",
  "log",
  "ootle-network",
+ "rand 0.10.1",
  "tari_consensus",
  "tari_consensus_types",
  "tari_epoch_manager",
--- a/applications/tari_indexer/src/network_state_sync/worker.rs
+++ b/applications/tari_indexer/src/network_state_sync/worker.rs
@@ -352,6 +352,7 @@ impl NetworkWideStateSync {
                         .get_checkpoints(rpc::GetCheckpointsRequest {
                             from_epoch: Some(from_epoch.into()),
                             num_to_return: 100,
+                            shard_group: None,
                         })
                         .await?;
 
--- a/applications/tari_validator_node/src/p2p/rpc/rpc_impl.rs
+++ b/applications/tari_validator_node/src/p2p/rpc/rpc_impl.rs
@@ -567,6 +567,24 @@ impl<TStateStore: StateStore + Clone + Send + Sync + 'static> ValidatorNodeRpcSe
             )));
         }
 
+        if let Some(shard_group) = msg.shard_group {
+            let shard_group = ShardGroup::decode_from_u32(shard_group)
+                .ok_or_else(|| RpcStatus::bad_request(format!("Invalid shard group {shard_group}")))?;
+            let checkpoint = self
+                .state_store
+                .with_read_tx(|tx| EpochCheckpoint::get_by_shard_group(tx, from_epoch, shard_group))
+                .optional()
+                .map_err(RpcStatus::log_internal_error(LOG_TARGET))?
+                .ok_or_else(|| {
+                    RpcStatus::not_found(format!(
+                        "No checkpoint for epoch {from_epoch} shard group {shard_group}"
+                    ))
+                })?;
+            return Ok(Response::new(GetCheckpointsResponse {
+                checkpoints: vec![checkpoint.into()],
+            }));
+        }
+
         if msg.num_to_return > 100 {
             return Err(RpcStatus::bad_request("num_to_return must be less than 100"));
         }
--- a/crates/p2p/proto/rpc.proto
+++ b/crates/p2p/proto/rpc.proto
@@ -213,6 +213,8 @@ message QuorumCertificates {
 message GetCheckpointsRequest {
   tari.ootle.common.Epoch from_epoch = 1;
   uint32 num_to_return = 2;
+  // When set, returns only the checkpoint for this shard group at from_epoch.
+  optional uint32 shard_group = 3;
 }
 
 message GetCheckpointsResponse {
--- a/crates/rpc_state_sync/Cargo.toml
+++ b/crates/rpc_state_sync/Cargo.toml
@@ -23,4 +23,5 @@ tari_validator_node_rpc = { workspace = true }
 anyhow = { workspace = true }
 futures = { workspace = true }
 log = { workspace = true }
+rand = { workspace = true }
 thiserror = { workspace = true }
--- a/crates/rpc_state_sync/src/error.rs
+++ b/crates/rpc_state_sync/src/error.rs
@@ -26,8 +26,8 @@ pub enum RpcStateSyncError {
     InvalidResponse(anyhow::Error),
     #[error("Block {block_id} failed SafeNode predicate")]
     BlockNotSafe { block_id: BlockId },
-    #[error("Failed to sync from all peers. The committee size is {committee_size}")]
-    SyncFailedAllPeers { committee_size: usize },
+    #[error("Failed to sync from all {num_peers} peer(s)")]
+    SyncFailedAllPeers { num_peers: usize },
     #[error("Proposal validation error: {0}")]
     ProposalValidationError(#[from] ProposalValidationError),
     #[error("State tree error: {0}")]
--- a/crates/rpc_state_sync/src/state_sync.rs
+++ b/crates/rpc_state_sync/src/state_sync.rs
@@ -147,8 +147,8 @@ where TConsensusSpec: ConsensusSpec<Addr = PeerAddress>
         };
 
         let num_returned = checkpoints.len();
-        // Most of the batch is checkpoints we did not ask for, so a malformed entry only disqualifies itself;
-        // the one we select is still checked against the signing committee's quorum.
+        // The batch can include checkpoints for other shard groups and later epochs, so a malformed entry
+        // only disqualifies itself; the one we select is still checked against the signing committee's quorum.
         for checkpoint in checkpoints {
             let checkpoint = match EpochCheckpoint::try_from(checkpoint) {
                 Ok(cp) => cp,
--- a/crates/rpc_state_sync/src/state_sync.rs
+++ b/crates/rpc_state_sync/src/state_sync.rs
@@ -21,7 +21,7 @@ use tari_ootle_common_types::{
     ShardGroup,
     VersionedSubstateId,
     VotePower,
-    committee::Committee,
+    committee::{Committee, CommitteeMember},
     optional::Optional,
     shard::Shard,
 };
@@ -448,11 +448,11 @@ where TConsensusSpec: ConsensusSpec<Addr = PeerAddress>
         Ok(())
     }
 
-    async fn get_sync_committees(
+    async fn get_sync_sources(
         &self,
         local_shard_group: ShardGroup,
         current_epoch: Epoch,
-    ) -> Result<HashMap<ShardGroup, Committee<PeerAddress>>, RpcStateSyncError> {
+    ) -> Result<HashMap<ShardGroup, SyncSource>, RpcStateSyncError> {
         // We are behind at least one epoch.
         // We get the current substate range, and we ask committees from previous epoch in this range to give us
         // data.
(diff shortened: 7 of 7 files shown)


sdbondi (commented):
Splitting "who signs the checkpoint" from "who serves the bytes" is the right call, and the mechanics are sound in the steady state: `validate_checkpoint` still sees only the previous committee (`state_sync.rs:562`), an empty current-epoch result degrades to the old behaviour because `validator_nodes_get_overlapping_shard_group` returns `Ok(empty)` rather than erroring, the epoch-0 / first-committee / sole-VN early-outs are preserved, and `Committee::shuffled()` really is a full shuffle (rand's `index::sample` documents "fully shuffled"), so the per-tier load spreading works as described.

The problem is the case the PR is aimed at — a boundary where the committee layout actually moves.

## Blocking

### 1. A current-epoch member is not a valid source for the whole previous shard group

`state_sync.rs:474-481` builds `current_members` from **every** current-epoch committee overlapping the *previous* shard group `X`, and `SyncSource::new` (`758-773`) then puts all of them (tiers 1 and 2) ahead of the departed prev-`X` members.

But a current-epoch member only holds the previous epoch's state for `its own current shard group ∩ X` — `sync_inner:726-735` syncs exactly `intersect_shard_group = shard_group ∩ local_shard_group`, nothing wider. So when the committee count grows at the boundary and `X` splits into `X1 | X2`, a joiner in `X1` is preferred over a departed prev-`X` member for shards in `X2` that it has never held.

Nothing downstream catches this early. The server deliberately serves a bounded `sync_state` request out of history with no membership check (`applications/tari_validator_node/src/p2p/rpc/rpc_impl.rs:608-616`), so that peer accepts the request, streams no transitions for the shard, and the client fails the root check at `state_sync.rs:247` with `StateRootMismatch` — recorded as `last_hard_error`. Every preferred peer in the wrong sub-range burns a full attempt, logs a `❌ State root mismatch`, and if the fallback tier also fails the terminal error is a state-root mismatch rather than something retryable.

Concretely: epoch N-1 has one committee over shards 1..64; epoch N splits it into 1..32 and 33..64; we are in 33..64. `current_members` contains both current committees, so every joiner in 1..32 is tried first for shards 33..64.

Fix suggestion: the current-epoch committee that covers `local_shard_group` — i.e. our own — covers by construction every shard this loop will request, so drawing `current_members` from `get_committees_overlapping_shard_group(current_epoch, local_shard_group)` gives the availability win without the coverage hole. Alternatively keep the per-current-shard-group map and skip a tier-2 member in `sync_shard` when its group does not contain `shard` (the global shard is exempt — it is reserved and replicated everywhere).

### 2. The checkpoint request has no shard-group selector, and tier-2 peers hold several

`GetCheckpointsRequest` carries only `from_epoch`/`num_to_return`, and the server answers it with `EpochCheckpoint::get_all_from_epoch` (`crates/state_store_rocksdb/src/reader.rs:1902-1915`) — an epoch-range scan, `.take(limit)`, no shard-group filter. With `num_to_return: 1` the peer returns whichever checkpoint at `prev_epoch` its index yields first.

A node that joined at the current epoch stores **one checkpoint per previous shard group it synced from** (`state_sync.rs:151`, called once per `sync_sources` entry plus once from `sync_global_shard`). So a tier-2 peer that synced from two overlapping prev groups can hand back the checkpoint for the other group. It fails `validate_checkpoint` (different signers), surfaces as `InvalidResponse`, and lands in `last_hard_error` — a hard, non-retryable classification for what is really "asked the wrong peer".

This could not happen before: the peer set was the prev committee, and each of its members holds exactly one checkpoint for that epoch, its own.

The same fix as (1) removes most of it. Independently, `get_or_fetch_valid_epoch_checkpoint` already computes the checkpoint's shard group at `state_sync.rs:143` and throws the value away — comparing it to what was asked for would turn this into a clear skip-and-try-next instead of a hard error. Note the comparison cannot be against the `for_shard_group` parameter as-is: `sync_global_shard` passes the `ShardGroup::all_shards(...)` sentinel (`state_sync.rs:715-718`), which never equals a real checkpoint's group.

## Non-blocking

- `state_sync.rs:626`: `SyncFailedAllPeers { committee_size: source.serving_peers.len() }` — the field name now describes something else, and the count includes ourselves even though `544-546` skips us. Either rename the field or exclude our own address when building `serving_peers`.
- `state_sync.rs:759-773`: three passes, each doing `Committee::contains`, which is a linear scan flagged `TODO(perf)` at `crates/common_types/src/committee.rs:88-91` — so this is O(n·m) on every sync attempt. A `HashSet<&PeerAddress>` of the signing committee's addresses, built once, makes all three passes linear.
- `state_sync.rs:776-780`: `[continuing, joined, departed].iter().flat_map(|tier| tier.shuffled().cloned().collect::<Vec<_>>())` clones every member and allocates a `Vec` per tier only to escape the borrow. `Committee` has `shuffle(&mut self)` and owns `IntoIterator`, so `[continuing, joined, departed].into_iter().flat_map(|mut t| { t.shuffle(); t })` is the same ordering with no clones.

## Nits

- `state_sync.rs:759-773`: the three tiers are `Committee` values but are only ever ordered lists — `Vec<CommitteeMember<PeerAddress>>` states that. `quorum_threshold()` on a "joined" tier would be meaningless, and the type currently invites it.
- Pre-existing, but adjacent: `self.valid_checkpoints` (`state_sync.rs:75`) is written at `152` and cleared at `1122`/`1130`, never read. The DB lookup at `120-123` is doing that job.

sdbondi (commented):
Re-reviewed at `f0d0c28`. Both blocking findings are resolved, and I checked the fixes rather than taking the commit message for them.

**(1) Serving-peer coverage** — `get_sync_sources` now takes `get_local_committee(current_epoch)` (`state_sync.rs:483`) instead of every current committee overlapping the previous shard group. That closes the hole, and the invariant holds in both directions:

- split (`local_shard_group ⊂ X`): the only source is `X`, requested shards are `X ∩ local = local`;
- merge (`local_shard_group ⊃ X1, X2`): sources are `X1` and `X2`, requested shards per source are `Xi`, and our committee's members hold prev-epoch state for all of `local ⊇ X1 ∪ X2`.

Either way, requested shards ⊆ `local_shard_group`, which is exactly the range every member of our current committee synced. A member reassigned across the boundary now lands in `joined` for the sub-range it did not previously hold, which is correct — it did sync that range at this epoch.

**(2) Checkpoint selection** — the request now pulls a batch and filters on `checkpoint.epoch() != prev_epoch || shard_group != for_shard_group` (`state_sync.rs:157`), and a non-match is a `continue` ending in `Ok(None)`, so it is classified as retryable rather than a hard `InvalidResponse`. The `ShardGroup::all_shards` sentinel is gone: `sync_global_shard` now passes the real `*sg` (`state_sync.rs:645`), so the filter compares like with like and the DB lookup at `:125` actually hits for the global-shard pass — which is also what retires the TODO that was there. `Shard::global()` is in every checkpoint's summary (`crates/consensus/src/hotstuff/common.rs:287,292`), so the replacement comment is accurate.

Also verified: `signing_addrs`/`local_addrs` borrows end before `signing_committee` is moved into `Self`; `SliceRandom::shuffle` is the right trait for `Vec` under rand 0.10; `serving_peers` now excludes us, so dropping the self-check in the `sync_shard` loop is right and `num_peers` is an honest count. The three non-blocking items and both nits from the previous pass are all addressed.

Two small things on the new code, neither blocking:

- `state_sync.rs:150-156`: a single undecodable or bad-shard-group entry anywhere in the batch aborts the whole request with `InvalidResponse`, even when the checkpoint we asked for is later in the same batch. Since the batch is peer-chosen and mostly entries we did not ask for, `continue` on a bad entry would be the more forgiving read; failing hard on a malformed response is defensible, so this is a judgement call rather than a defect.
- `MAX_CHECKPOINTS_PER_REQUEST = 32` (`state_sync.rs:68-70`): the reasoning is right — the server scans `from_epoch..Epoch::max()` ascending (`crates/state_store_rocksdb/src/reader.rs:1908`), so `prev_epoch`'s entries sort first and truncation only bites if one peer stored more than 32 checkpoints at that epoch, which needs a >32-way merge at the boundary. Worth knowing that when it does bite the failure is silent: the wanted checkpoint is simply absent, `Ok(None)`, and the node loops on `CheckpointNotAvailable` with nothing pointing at the cap. Separately the response now carries up to 32 checkpoints where one is needed, and a checkpoint's commit proof scales with committee size — against `RPC_MAX_FRAME_SIZE` of 6 MiB (`networking/rpc_framework/src/lib.rs:38`) that is fine at current sizes but is a ceiling worth remembering. A shard-group selector on `GetCheckpointsRequest` would remove both, if that proto change is ever worth making.

Holding approval only until CI reports green.

sdbondi (commented):
Re-reviewed `81cf9d6` — this is the durable fix I suggested rather than the workaround, and it removes both of the previous pass's caveats (the 32-checkpoint cap and the batch-scanning loop are gone). Checked the wire change in both directions:

- **old client → new server**: field 3 absent, `msg.shard_group` is `None`, falls through to the unchanged batch path. The indexer is explicitly on that path (`worker.rs:355`, `shard_group: None`, `num_to_return: 100`). ✅
- **new client → old server**: field 3 is an unknown field and is ignored; the old server honours `num_to_return: 1` and returns whichever checkpoint at `prev_epoch` its index yields first.

The protocol name is unchanged (`/tari/validator/1.0.0`, `crates/validator_node_rpc/src/rpc_service.rs:8`), so an optional field addition needs no negotiation — correct call.

Also verified: `ShardGroup::encode_as_u32`/`decode_from_u32` round-trip and reject out-of-range input (`crates/common_types/src/shard_group.rs:100-114`), matching how shard groups are already encoded elsewhere; the selector branch sits after the `from_epoch >= consensus_epoch` guard so it cannot serve the current epoch; `not_found` for a group the peer does not hold maps to `Ok(None)` on the client via the existing `err.is_not_found()` arm, which lands in `saw_unavailable_checkpoint` — the retryable classification, which is what you want for "this peer hasn't synced that group yet"; `ShardGroup` and `Optional` were already imported in `rpc_impl.rs`; and all three `GetCheckpointsRequest` construction sites are updated.

One non-blocking point, on the second bullet above. `state_sync.rs:152-157` now turns a shard-group/epoch mismatch into a hard `InvalidResponse`, which becomes `last_hard_error`. Against an un-upgraded peer the selector is silently ignored, so any peer holding more than one checkpoint at `prev_epoch` — a joiner, i.e. exactly the tier-2 peers this PR added — can trip it during a rolling upgrade. This is not a regression: `development` fails the same way (it accepts whatever comes back and then fails `validate_checkpoint`). But `Ok(None)` with a warn would cost nothing, skip that peer as retryable instead of poisoning the attempt, and still never accept a checkpoint for the wrong group. The only thing lost is a loud signal for a genuinely misbehaving upgraded peer, and the warn covers that.

Nothing blocking. Waiting on CI before approving.

sdbondi (commented):
Signing off on `81cf9d6`. All three blocking findings from the first pass are resolved and re-verified against the code at each head, not taken from the commit messages:

1. Serving peers now come from `get_local_committee(current_epoch)`, so every peer offered holds the previous epoch's state for the whole range the sync requests. Checked in both the split and merge directions.
2. Checkpoint selection is no longer ambiguous — `GetCheckpointsRequest` carries a `shard_group` selector, the server answers it from `EpochCheckpoint::get_by_shard_group`, and a group the peer does not hold comes back as `not_found`, which the client already maps to the retryable path.
3. The wire change is compatible both ways under the unchanged `/tari/validator/1.0.0` protocol name, and the indexer's batch call is explicitly updated to `shard_group: None`.

The non-blocking items and nits from all three passes were taken as well.

On CI: `Test Results (CI)` is red with 2 failures out of 2222, and both are network-setup flakes, not this change. `Scenario: Offline validator gets evicted` (`eviction.feature:8`) fails on its *first* step, `Given a network with spec`, with `Connection reset by peer` against `/epoch-manager/stats` while the network is still coming up. `Scenario: Claim and transfer stealth assets via wallet daemon` (`wallet_daemon.feature:54`) also fails on its first step, with `Validator exited with error: Networking service has shutdown`. Neither scenario reached a state-sync code path — both died before the first `When`. `development` shows the same class of flake independently (`concurrency.feature` failed on `a6d07e4`). Every other check passes: clippy, fmt, machete, all three test shards, and the integration-test job itself.

Note I cannot record a GitHub approval here — it is your own PR, so `gh` refuses `--approve`. This is a comment; another maintainer needs to click approve.

PR #2551, opened 2026-09-08
fix(engine)!: hash a receipt's exhaust_burn only from protocol version 1

sdbondi:
## Description

`SubstateHashMessage` now has a V1 shape. V0 writes `FeeReceipt` without `exhaust_burn`, V1 writes the whole struct. The per-type enum `SubstateValueHashMessage<R>` is shared by both versions (generic over the receipt message), so their variant tags cannot drift apart. Note that at the byte level V1 changes every substate's preimage, not only receipts: `SubstateHashMessage` is a borsh enum, so V1 writes a leading tag of `1` where V0 writes `0`. No leaf moves because the schema is selected from the substate's creation epoch, but at an activation epoch every substate created from then on hashes under the new tag. `ProtocolVersion::V1`'s doc names both things it changes.

## Motivation and Context

#2520 added `FeeReceipt::exhaust_burn` with `#[cbor(default)]` so receipts committed before it still decode. The substate hash preimage is borsh, and the derive writes every field, so a pre-#2520 receipt now re-hashes with an extra `0` that the committing node never covered. Any node that derives a receipt's hash from its value disagrees with the quorum-signed state roots: a validator or indexer syncing from scratch, and substate proof verification.

Found by resyncing a validator on a localnet swarm that had crossed the upgrade. Shard 0 matched; shard 1 mismatched at the first checkpoint after the first receipt. Walking both JMTs: 179 of 380 receipt leaves hashed differently, every other substate type identical, and the split fell exactly on the epoch the swarm was upgraded. A fresh validator could not sync at all.

Version 0 now writes the exact preimage the pre-#2520 binary produced, pinned by a golden hash captured by running `hash_substate` at `2cc729b95` (the parent of #2520) on the same synthetic receipt.

## Scope

This prevents the next occurrence; it cannot repair a network that already committed receipts under a post-#2520 binary while on V0. Those receipts were hashed with the field, and an epoch-granular activation cannot describe a mid-epoch cutover. Such a network needs a reset. Esmeralda has not run a post-#2520 binary, so it is covered as long as its V1 activation is scheduled at or after the epoch that binary is deployed.

The gap while esmeralda runs a post-#2520 binary on V0 is narrower than "an uncommitted consensus value": the burn is committed through `BlockHeader::total_accumulated_exhaust_burn`, which every voter re-derives (`on_ready_to_vote_on_local_block.rs`), so a node computing a different burn refuses to vote rather than forking. What V0 leaves out is only the per-receipt field in the state root.

## How Has This Been Tested?

`cargo test -p tari_engine_types` (85 tests) and `cargo lints clippy -p tari_engine_types --all-targets` clean.

- `version_0_reproduces_the_pre_exhaust_burn_hash` — the golden above, asserted through `ProtocolVersion::V0` directly, with esmeralda's V0 genesis stated as its own assertion. This is the whole guarantee.
- `version_1_hash_is_pinned` — a V1 golden; five networks launch at V1, so it is consensus-bound from their first block.
- `version_0_does_not_cover_exhaust_burn` / `version_1_covers_exhaust_burn`.
- `value_tags_are_stable_and_shared_by_both_versions` — pins the borsh tags 2,3,4,5,7,8 (vault, non-fungible, tombstone, receipt, fee pool, utxo) under both versions, and that V1 differs from V0 only in the leading version tag outside the receipt. That pins the rest transitively: any insertion except after `Utxo` shifts a pinned tag, and after `Utxo` only the final variant follows, where an append is the only move.

Run the tests. To re-derive the golden independently: check out `2cc729b95`, build the receipt in `substate_hasher::tests::receipt` (no `with_exhaust_burn`, which does not exist there) and print `hash_substate(Network::LocalNet, &value, 0, Epoch(3))` — LocalNet was V0-genesis at that commit.

## Breaking Changes

- [ ] None
- [ ] Requires hard fork
- [x] Other - Please specify

`SubstateHashMessage` gains a `V1` variant and `TransactionReceiptHashMessage` becomes `TransactionReceiptHashMessageV0`/`V1`. Networks launched at V1 hash receipts with `exhaust_burn`; V0 hashes are byte-identical to before.

### DIFF
--- a/crates/engine_types/src/fees.rs
+++ b/crates/engine_types/src/fees.rs
@@ -214,7 +214,7 @@ impl FeeReceipt {
 
     /// Writes the receipt as a protocol version 0 substate hash preimage: every field but
     /// `exhaust_burn`, which version 0 receipts do not carry.
-    pub fn borsh_serialize_v0<W: borsh::io::Write>(&self, writer: &mut W) -> borsh::io::Result<()> {
+    pub(crate) fn borsh_serialize_v0<W: borsh::io::Write>(&self, writer: &mut W) -> borsh::io::Result<()> {
         borsh::BorshSerialize::serialize(&self.total_fee_payment, writer)?;
         borsh::BorshSerialize::serialize(&self.total_fees_paid, writer)?;
         borsh::BorshSerialize::serialize(&self.total_fee_overcharge, writer)?;
--- a/crates/engine_types/src/fees.rs
+++ b/crates/engine_types/src/fees.rs
@@ -212,6 +212,15 @@ impl FeeReceipt {
         }
     }
 
+    /// Writes the receipt as a protocol version 0 substate hash preimage: every field but
+    /// `exhaust_burn`, which version 0 receipts do not carry.
+    pub fn borsh_serialize_v0<W: borsh::io::Write>(&self, writer: &mut W) -> borsh::io::Result<()> {
+        borsh::BorshSerialize::serialize(&self.total_fee_payment, writer)?;
+        borsh::BorshSerialize::serialize(&self.total_fees_paid, writer)?;
+        borsh::BorshSerialize::serialize(&self.total_fee_overcharge, writer)?;
+        borsh::BorshSerialize::serialize(&self.cost_breakdown, writer)
+    }
+
     pub fn to_cost_breakdown(&self) -> FeeCostBreakdown {
         FeeCostBreakdown {
             total_fees_charged: self.total_fees_charged(),
--- a/crates/engine_types/src/protocol_version.rs
+++ b/crates/engine_types/src/protocol_version.rs
@@ -21,8 +21,10 @@ pub enum ProtocolVersion {
     /// under.
     #[default]
     V0 = 0,
-    /// Block headers commit to their own protocol version. No network schedules this version in
-    /// [`Self::activations`]; scheduling it is a deliberate per-network deployment decision.
+    /// Block headers commit to their own protocol version, and a transaction receipt's substate hash
+    /// covers `FeeReceipt::exhaust_burn`. Networks without history launch at this version; a network
+    /// with V0 history reaches it through a scheduled activation in [`Self::activations`], a
+    /// deliberate per-network deployment decision.
     V1 = 1,
 }
 
--- a/crates/engine_types/src/substate_hasher.rs
+++ b/crates/engine_types/src/substate_hasher.rs
@@ -36,54 +36,29 @@ impl<'a> SubstateHashMessage<'a> {
     }
 }
 
+pub type SubstateValueHashMessageV0<'a> = SubstateValueHashMessage<'a, TransactionReceiptHashMessageV0<'a>>;
+
+/// Version 1 changes the transaction receipt's preimage, which covers `FeeReceipt::exhaust_burn`,
+/// and - through the leading `SubstateHashMessage` tag - the preimage of every substate.
+pub type SubstateValueHashMessageV1<'a> = SubstateValueHashMessage<'a, TransactionReceiptHashMessageV1<'a>>;
+
+/// The per-type preimage, shared by every protocol version so that the borsh tag of each variant is
+/// the same under all of them. Variant order is consensus-bound: a new variant goes at the end.
 #[derive(Debug, Clone, Copy, borsh::BorshSerialize)]
-pub enum SubstateValueHashMessageV0<'a> {
+pub enum SubstateValueHashMessage<'a, R> {
     Component(ComponentHashMessage<'a>),
     Resource(ResourceHashMessage<'a>),
     Vault(VaultHashMessage<'a>),
     NonFungible(NonFungibleContainerHashMessage<'a>),
     ClaimedOutputTombstone(ClaimedOutputTombstoneHashMessage<'a>),
-    TransactionReceipt(TransactionReceiptHashMessageV0<'a>),
+    TransactionReceipt(R),
     Template(PublishedTemplateHashMessage<'a>),
     ValidatorFeePool(ValidatorFeePoolHashMessage<'a>),
     Utxo(UtxoHashMessage<'a>),
     ConfidentialOutput(ConfidentialOutputHashMessage<'a>),
 }
 
-impl<'a> From<&'a SubstateValue> for SubstateValueHashMessageV0<'a> {
-    fn from(value: &'a SubstateValue) -> Self {
-        match value {
-            SubstateValue::Component(component) => Self::Component(component.into()),
-            SubstateValue::Resource(resource) => Self::Resource(resource.as_ref().into()),
-            SubstateValue::Vault(vault) => Self::Vault(vault.into()),
-            SubstateValue::NonFungible(nf) => Self::NonFungible(nf.into()),
-            SubstateValue::ClaimedOutputTombstone(tombstone) => Self::ClaimedOutputTombstone(tombstone.into()),
-            SubstateValue::TransactionReceipt(receipt) => Self::TransactionReceipt(receipt.into()),
-            SubstateValue::Template(template) => Self::Template(template.into()),
-            SubstateValue::ValidatorFeePool(pool) => Self::ValidatorFeePool(pool.into()),
-            SubstateValue::Utxo(utxo) => Self::Utxo(utxo.into()),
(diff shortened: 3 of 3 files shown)


sdbondi (commented):
Reviewed at `b6a9d4c30`. Nothing blocking. The fix is the right shape and the golden checks out.

## Verified

- **Golden provenance.** `git show 2cc729b95` confirms the preimage this reproduces: `SubstateHashMessage` already existed there with `V0` as its only variant, `SubstateValueHashMessageV0` had the identical variant order, and `FeeReceipt`'s four fields were in the same order `borsh_serialize_v0` writes them. So the leading `0` tag and the field sequence are byte-faithful, and the golden is pinning the thing it claims to pin.
- **No stragglers.** No reference to the old `TransactionReceiptHashMessage` name remains in the tree, and no other consensus preimage borsh-serializes a `FeeReceipt` — the receipt leaf really was the only site.
- **Epoch plumbing holds up.** This is the first change to make V0 and V1 substate hashes actually differ, so the "schema from the creation epoch" rule becomes load-bearing for the first time. Every call site passes a creation epoch: `substate_value_proof.rs:156` gets it from `substate.created().at_epoch` (`rpc_impl.rs:166,236`), `consensus_models/substate.rs:65` from `created.at_epoch`, and the `to_tree_change` paths from the epoch of the block being applied, which for an Up change is the creation epoch.
- **No version bump owed.** `scripts/crate_versioning.py impact tari_engine_types --breaking` reports every affected crate already on an unreleased bump.

## Non-blocking

**1. "the transaction receipt alone" is not true at the byte level** — `substate_hasher.rs:70-71` and the PR body.

`SubstateHashMessage` is a borsh enum, so the V1 arm writes a leading tag of `1` where V0 writes `0`. Every substate type hashes differently under V1, not just receipts. This is harmless — the schema is selected from the substate's creation epoch, so no leaf ever moves — but the comment will mislead whoever next reasons about the Esmeralda activation: they will expect only receipt leaves to change at the fork epoch. Worth saying that V1 changes the receipt schema *and*, through the version tag, the preimage of every substate.

**2. V1 is pinned by nothing** — `substate_hasher.rs:315-345`.

V0 gets a byte-exact golden; V1 gets two relative assertions. But mainnet, stagenet, nextnet, igor and localnet all launch at V1 genesis, so V1 is consensus-bound from their first block. Once any of them is live, an accidental reorder in `SubstateValueHashMessageV1` or a change to `FeeReceipt`'s derived borsh shape is a silent fork with no test failing. A second golden over `hash_at(ProtocolVersion::V1, &receipt(123))` costs one line and closes that.

**3. Two hand-maintained 10-variant enums that must agree forever** — `substate_hasher.rs:40-51` and `73-84`.

Adding a `SubstateValue` variant is a compile error in both `From` impls, which is the good case. Inserting it at a *different index* in one enum than the other compiles fine and silently shifts the tag of every variant after it — for V0, that is a consensus break with no error. The existing golden only covers `TransactionReceipt` at index 5, so a variant inserted at index 6 or later slips past it entirely. A test that constructs one value per `SubstateValue` variant and asserts `borsh::to_vec(&SubstateHashMessage::new(V0, &v))[1]` equals the expected index would pin V0's whole tag space, and asserting `borsh(V1) == [1] ++ borsh(V0)[1..]` for the non-receipt variants would pin the two enums against each other.

**4. The V0 gap is narrower than "Scope" makes it sound.**

Reading the PR body, an Esmeralda that runs a post-#2520 binary at V0 has `exhaust_burn` outside the state root, which sounds like an uncommitted consensus value. It isn't: `total_accumulated_exhaust_burn` lives in the block header and every voter re-derives it (`on_ready_to_vote_on_local_block.rs:513`), so a node computing a different burn fails to vote rather than forking quietly. Worth a sentence — it changes how urgent the Esmeralda V1 activation is.

## Nits

- `FeeReceipt::borsh_serialize_v0` (`fees.rs:217`) is `pub` on a type used all over the tree. It is a hashing-preimage detail and any caller outside `substate_hasher` reaching for "the v0 borsh encoding" is almost certainly making a mistake. `pub(crate)` covers the only caller.
- `version_0_reproduces_the_pre_exhaust_burn_hash` carries the golden through `hash_substate(Network::Esmeralda, .., Epoch(3))`, which couples an immutable constant to Esmeralda's current schedule. `hash_at(ProtocolVersion::V0, ..)` — already asserted equal on the next line — is the assertion that can never go stale; if the Esmeralda link is worth keeping, `assert_eq!(ProtocolVersion::at(Network::Esmeralda, Epoch(3)), ProtocolVersion::V0)` states it directly instead of by implication.

sdbondi (commented):
Re-reviewed at `7454d6881`. All six items are addressed, and the generic `SubstateValueHashMessage<'a, R>` is a better answer to item 3 than the test I suggested — sharing one enum makes tag drift between the versions unrepresentable rather than merely detectable. V0's preimage is untouched: the variant list and order are identical and borsh still writes the variant index, so the `061d83…` golden still stands for the same reason it did before.

Verified individually:

- **V0 unchanged** — the enum went generic but kept all ten variants in order, and `TransactionReceiptHashMessageV0` is byte-for-byte what it was.
- **V1 pinned** — `version_1_hash_is_pinned` at `substate_hasher.rs:310`. Self-captured, which is the right kind of golden for a version with no history behind it.
- **Doc claim corrected** in both the `SubstateValueHashMessageV1` alias doc and the PR description.
- **`borsh_serialize_v0` is `pub(crate)`**, and `substate_hasher` is the only caller.
- **Golden decoupled from Esmeralda's schedule** — `hash_at(V0, ..)` now carries it, with `ProtocolVersion::at(Esmeralda, Epoch(3)) == V0` asserted separately. The test now fails for the reason it names.
- **Scope reworded** — the `total_accumulated_exhaust_burn` point is in the body.

One thing on the new test, non-blocking:

`value_tags_are_stable_and_shared_by_both_versions` (`substate_hasher.rs:349`) says it "pins the borsh tag of every variant", but it covers five of ten: 2, 3, 4, 5, 7. Uncovered are `Component` (0), `Resource` (1), `Template` (6), `Utxo` (8) and `ConfidentialOutput` (9). Most insertion points are still caught transitively — anything inserted at index 0 through 7 shifts a pinned tag — but an insertion at 8 or 9 moves only `Utxo` and `ConfidentialOutput`, and nothing fails. The stated rule ("a new variant goes at the end") is what makes that acceptable, so this is really a wording issue: either say it pins the tags transitively and name the blind spot, or close it. `SubstateValue::Utxo(Utxo { output: None, is_frozen: false })` is a one-liner (both fields are `pub`) and shrinks the gap to the final variant, where an append is safe by construction anyway.

Nothing blocking. Waiting on CI — `test`, `clippy` and the rest are still running on this head. I can't approve my own PR, so this stays a comment.

sdbondi (commented):
Re-reviewed at `642f0aa36`. The `Utxo` tag is pinned and the test doc now states its coverage honestly — "pins the rest transitively: a variant inserted anywhere but after `Utxo` shifts one of these tags", with `ConfidentialOutput` named as the last slot where an append is the only move. That closes the last item from either round. Nothing outstanding from my side.

**On the red `Test Results (CI)`:** it is a bring-up flake, unrelated to this change. The aggregate reports 1 of 2233, and every job on this head passed on its own, `integration test` included (1h5m50s).

The failure is `Feature: Epoch change / Scenario: EndEpoch command is used on epoch change`, on its *first* `Given a network with registered validator VN and wallet daemon WALLET_D` — `ConnectionReset by peer` on `http://localhost:40791/`. The dumped `TariWorld` is entirely empty (`base_nodes: []`, `validator_nodes: []`, `wallets: []`), so nothing had started and no substate had been hashed when it failed. The scenario that ran immediately after it stood up a five-validator network and completed a full eviction run, and the 33 other scenarios cover account creation, template publishing, method calls and epoch changes end to end — every one of which hashes receipts. A broken receipt preimage does not present as a connection reset before the first process is up while everything downstream of it passes.

So: a rerun of the integration job should clear the aggregate. Worth confirming it goes green before merge, since a red required check will block it either way.

Everything I checked in the first two rounds still holds at this head: V0's preimage is byte-identical (same ten variants, same order, borsh still writing the variant index), the `061d83…` golden is asserted against `hash_at(V0, ..)`, V1 is pinned, and `crate_versioning.py impact tari_engine_types --breaking` still reports no bumps owed.

This stays a comment — GitHub won't let me approve my own PR, so it needs another maintainer.

PR #2554, opened 2026-09-09
feat(engine)!: allow a resource's auth hook to be replaced or removed

### LINKED ISSUE #2514: A resource's auth hook cannot be changed or removed after creation, so a broken hook permanently locks the resource
## Summary

A resource's authorization hook is fixed at creation and can never be changed or removed. Every other authorization field on a resource is updatable through `ResourceAction::UpdateAccessRule`, gated by a per-field `UpdateRule`. The auth hook has no equivalent, so a hook that stops behaving correctly makes the resource permanently unusable, with funds in its vaults unrecoverable.

## Detail

`Resource::auth_hook` (`crates/engine_types/src/resource.rs:62`) is set only by the two constructors, `Resource::new` (`:68`) and `Resource::load` (`:98`). There is no setter — the only accessor is `auth_hook()` at `:152` — and no `ResourceAction` variant updates it. The full set is `Create`, `Mint`, `Recall`, `UpdateNonFungibleData`, `GetTotalSupply`, `GetResourceInfo`, `GetNonFungible`, `UpdateAccessRule`, `SetVaultFreeze`, `StealthTransfer`, `SetStealthUtxosFreeze`, `StealthUtxoBurn`, `UpdateMetadata`, `SetConfidentialOutputsFreeze` (`crates/template_lib/src/args/types.rs:207`).

Contrast this with access rules, which are mutable by design: `ResourceAccessRules` carries eight updater fields (`mint_updater`, `burn_updater`, `recall_updater`, `withdraw_updater`, `deposit_updater`, `nft_data_updater`, `freeze_updater`, `metadata_updater`, `crates/template_lib_types/src/access_rules.rs:456-484`), each an `UpdateRule` of `Locked` / `Owner` / `AccessRule`. The pattern for "authorization field that may be changed under a stated rule" already exists; the auth hook simply isn't wired into it.

## Impact

The hook is invoked on essentially every resource action — Mint, Burn, Recall, Withdraw (five sites), Deposit, Freeze, UpdateMetadata, UpdateNonFungibleData (`crates/engine/src/runtime/impl.rs`, thirteen call sites) — and a failure inside it aborts the action. So a hook that panics, denies unconditionally, or fails to decode its arguments takes the whole resource offline: no transfers, no burns, no metadata changes. Any balance held in a vault of that resource becomes unspendable.

The ways a hook can end up in that state are ordinary ones: a bug on an upgrade path in the hook component's template, state the hook reads that becomes malformed, or an argument-shape change on the engine side. `check_resource_auth_hook` (`crates/engine/src/runtime/impl.rs:567`) validates the hook's signature at resource creation, but nothing constrains its behaviour afterwards.

## Recovery is narrow and can be absent entirely

Because the hook is `(component_address, method)` and the address is fixed, the only way to change what runs is to change the code behind that component — i.e. migrate the hook component to a new template via `UpdateComponentTemplate`. That requires:

1. the hook component's owner rule to permit migration, and
2. the replacement template to expose a migration function.

If the hook component was created with `OwnerRule::None`, there is no recovery at all. Nor is there any if the resource author pointed the hook at a component they don't control — nothing prevents a resource from binding a hook-shaped method on someone else's component.

There is also no way to *remove* a hook. Related: `invoke_resource_access_hook` skips the hook when the target component does not yet exist (`impl.rs:412-414`), so a hook aimed at an address that is never created silently never fires — but once that address is occupied, the hook becomes binding and permanent.

## Suggested direction

Add `ResourceAction::UpdateAuthHook` mirroring `UpdateAccessRule`, gated by a new `auth_hook_updater: UpdateRule` on `ResourceAccessRules`, defaulting to `UpdateRule::Locked`. That keeps today's immutability as the default — an author who wants a permanently binding hook still gets one — while letting an author opt into a hook they can repair or retire. `ResourceAccessRules::set_auth_hook_updater` would follow the existing builder shape.

Two things to weigh:

- It adds a field to `ResourceAccessRules`, which is a substate schema change, so it interacts with the `ProtocolVersion` activation question.
- Allowing the hook to be swapped is itself an authorization decision. Holders of a resource are relying on the hook's behaviour, so a mutable hook under `UpdateRule::Owner` lets the owner change the rules on existing holders. That is the same trade already accepted for `mintable`/`withdrawable` updaters, but it should be a deliberate choice rather than a side effect, which is why `Locked` is the right default.

Worth deciding before mainnet, since the failure mode is unrecoverable locked funds and the fix is a schema change.

sdbondi:
## Description

A resource's authorization hook was fixed at creation: no setter on `Resource`, no `ResourceAction` to change it. The hook runs on nearly every resource action, so a hook that panics, denies unconditionally, or fails to decode its arguments takes the whole resource offline — no transfers, no burns, no metadata changes — and the balances in its vaults become unspendable. Recovery meant migrating the hook component to a new template, which needs an owner rule that permits migration; a resource whose hook component was created with `OwnerRule::None`, or which points at a component the author does not control, had no recovery at all.

Every other authorization field on a resource is already mutable under a per-field `UpdateRule`. This wires the hook into the same pattern.

- `ResourceAccessRules` gains `auth_hook_updater: UpdateRule`, alongside the eight updaters that already exist, with `set_auth_hook_updater` following the existing builder shape.
- `ResourceAction::UpdateAuthHook` + `ResourceManager::set_auth_hook(Option<AuthHook>)` set or clear the hook, gated by that updater. `None` removes the hook entirely.
- The four `ResourceBuilder` variants get `.with_authorization_hook_updater(rule)`.
- The default is `UpdateRule::Locked` — stricter than deny-all, since not even the owner can change it. Today's immutability is preserved for every existing resource, and an author who wants a permanently binding hook still gets one by default.

Two deliberate calls:

- **The hook being replaced is not invoked.** A hook that denies or panics is the failure this action exists to repair, so asking it to approve its own removal would defeat the point. This matches `UpdateAccessRule`, which does not invoke the hook either.
- **A replacement hook is validated exactly like one supplied at creation** — same `check_resource_auth_hook` signature checks. Its `argument` field is now parameterised so a rejection names `UpdateAuthHookArg` rather than `CreateResourceArg` on this path.

## Protocol version

`ResourceAccessRules` is part of the resource substate, so a new field changes the resource's borsh hash preimage. This is bundled into **protocol version 1**, which is unlaunched everywhere:

`ResourceHashMessage` is split into `ResourceHashMessageV0`/`V1` the way `TransactionReceiptHashMessage` already is. V0 writes the access rules without `auth_hook_updater` (`ResourceAccessRules::borsh_serialize_v0`, destructured so a future field fails to compile rather than being silently dropped), V1 writes the whole struct. `SubstateValueHashMessage` takes a second type parameter for the resource; the variant tags are unchanged under both versions.

`version_0_reproduces_the_pre_auth_hook_updater_hash` pins a hash captured from `hash_substate` at 4da937665, so esmeralda's committed resources still re-derive their state roots. Every other network launches at V1, so no activation needs coordinating.

CBOR/serde decode of existing resources is covered by `#[cbor(default)]` / `#[serde(default)]` on the new field, which resolves to `UpdateRule::Locked`.

## Does this break existing templates?

No. `ResourceAccessRules`'s fields are private and it is only constructed through `new()` / `deny_all()` / the builder methods, so the new field is source-compatible. `ResourceAction` gains a variant at the end. Nothing in the workspace matched `ResourceAction` exhaustively outside the engine.

`scripts/crate_versioning.py impact tari_template_lib_types --breaking` reports no bumps outstanding — every affected crate is already on an unreleased breaking version.

## Motivation and Context

Closes #2514. The failure mode is unrecoverable locked funds and the fix is a schema change, so it wants deciding before mainnet.

## How Has This Been Tested?

`cargo test -p tari_engine --test access_rules` — 32 pass, including five new ones:

- `auth_hook_is_locked_by_default` — the update is denied on a resource built the old way.
- `a_denying_auth_hook_can_be_removed_by_the_owner` — a hook that panics blocks a deposit; after `set_auth_hook(None)` the same transfer succeeds.
- `a_denying_auth_hook_can_be_replaced_by_the_owner` — swapping in a permitting hook unblocks the resource.
- `a_replacement_auth_hook_must_have_a_hook_signature` — all six invalid-hook shapes are rejected on the update path.
- `badge_holder_can_update_an_auth_hook_gated_on_their_badge` — `UpdateRule::AccessRule`: the owner is denied, the badge holder succeeds.

`cargo test -p tari_engine_types` — 88 pass, including the pinned V0 resource hash, `version_0_does_not_cover_auth_hook_updater`, `version_1_covers_auth_hook_updater`, and the existing pinned V0/V1 receipt hashes.

`cargo lints clippy --all-targets` — clean.

Check that `ResourceAccessRules::borsh_serialize_v0` writes exactly the pre-change field order, then confirm `version_0_reproduces_the_pre_auth_hook_updater_hash` still passes — a transcription slip there changes esmeralda's committed resource hashes.

The other judgement call worth a second opinion: `UpdateAuthHook` deliberately skips the outgoing hook. That is what makes a broken hook recoverable, but it does mean whoever satisfies `auth_hook_updater` can retire a hook that holders were relying on. `Locked` by default is what keeps that opt-in.

### DIFF
--- a/bindings/src/types/ResourceAccessRules.ts
+++ b/bindings/src/types/ResourceAccessRules.ts
@@ -22,4 +22,14 @@ export type ResourceAccessRules = {
   freeze_updater: UpdateRule;
   update_metadata: AccessRule;
   metadata_updater: UpdateRule;
+  /**
+   * Who may replace or remove the resource's [`AuthHook`](crate::AuthHook). The hook itself is not an
+   * [`AccessRule`], so this updater stands alone rather than pairing with one.
+   *
+   * A hook runs on nearly every resource action, so one that panics or denies unconditionally takes the
+   * resource offline and strands the balances in its vaults. `Locked` — the default — keeps a hook binding
+   * for the life of the resource; anything else lets the hook be repaired or retired, at the cost of letting
+   * whoever satisfies the updater change the rules that existing holders are relying on.
+   */
+  auth_hook_updater: UpdateRule;
 };
--- a/bindings/src/types/ResourceAccessRules.ts
+++ b/bindings/src/types/ResourceAccessRules.ts
@@ -23,13 +23,14 @@ export type ResourceAccessRules = {
   update_metadata: AccessRule;
   metadata_updater: UpdateRule;
   /**
-   * Who may replace or remove the resource's [`AuthHook`](crate::AuthHook). The hook itself is not an
-   * [`AccessRule`], so this updater stands alone rather than pairing with one.
+   * Who may install, replace or remove the resource's [`AuthHook`](crate::AuthHook). The hook itself is not
+   * an [`AccessRule`], so this updater stands alone rather than pairing with one.
    *
    * A hook runs on nearly every resource action, so one that panics or denies unconditionally takes the
    * resource offline and strands the balances in its vaults. `Locked` — the default — keeps a hook binding
    * for the life of the resource; anything else lets the hook be repaired or retired, at the cost of letting
-   * whoever satisfies the updater change the rules that existing holders are relying on.
+   * whoever satisfies the updater change the rules that existing holders are relying on — up to and
+   * including giving a hook-free resource a hook.
    */
   auth_hook_updater: UpdateRule;
 };
--- a/crates/engine/src/runtime/actions.rs
+++ b/crates/engine/src/runtime/actions.rs
@@ -64,6 +64,7 @@ pub enum NativeAction {
     StealthUtxoSpend,
     UpdateComponentTemplate,
     UpdateResourceAccessRule(ResourceAuthAction),
+    UpdateResourceAuthHook,
 }
 
 impl Display for NativeAction {
@@ -76,6 +77,7 @@ impl Display for NativeAction {
             Self::StealthUtxoSpend => write!(f, "stealth_utxo.spend"),
             Self::UpdateComponentTemplate => write!(f, "component.update_template"),
             Self::UpdateResourceAccessRule(action) => write!(f, "resource.update_access_rule.{:?}", action),
+            Self::UpdateResourceAuthHook => write!(f, "resource.update_auth_hook"),
         }
     }
 }
--- a/crates/engine/src/runtime/impl.rs
+++ b/crates/engine/src/runtime/impl.rs
@@ -101,6 +101,7 @@ use tari_template_lib::{
         SpendContextAction,
         StealthTransferResourceArg,
         UpdateAccessRuleArg,
+        UpdateAuthHookArg,
         VaultAction,
         VaultCreateProofByFungibleAmountArg,
         VaultCreateProofByNonFungiblesArg,
@@ -578,7 +579,9 @@ impl<TStore: StateReader + Clone + 'static, TTemplateProvider: TemplateProvider<
         Ok(())
     }
 
-    fn check_resource_auth_hook(&mut self, hook: &AuthHook) -> Result<(), RuntimeError> {
+    /// Validates that `hook` names a method with an authorization hook's signature. `argument` names the engine
+    /// argument the hook arrived in, so that a rejection points at the call the caller actually made.
+    fn check_resource_auth_hook(&mut self, argument: &'static str, hook: &AuthHook) -> Result<(), RuntimeError> {
         let template_address = self
             .tracker
             .write_with(|state| state.get_template_for_component(hook.component_address))?;
@@ -586,26 +589,26 @@ impl<TStore: StateReader + Clone + 'static, TTemplateProvider: TemplateProvider<
         let func = template
             .get_function(&hook.method)
             .ok_or(RuntimeError::InvalidArgument {
-                argument: "CreateResourceArg",
+                argument,
                 reason: format!("Authorize hook '{}' not found", hook),
             })?;
 
         if func.is_mut {
             return Err(RuntimeError::InvalidArgument {
-                argument: "CreateResourceArg",
+                argument,
                 reason: format!("Authorize hook '{}' cannot be mutable", hook),
             });
         }
         if !matches!(func.output, Type::Unit) {
             return Err(RuntimeError::InvalidArgument {
-                argument: "CreateResourceArg",
+                argument,
                 reason: format!("Authorize hook '{}' must return unit", hook),
             });
         }
 
         if func.arguments.len() != 3 {
             return Err(RuntimeError::InvalidArgument {
-                argument: "CreateResourceArg",
+                argument,
                 reason: format!(
                     "Authorize hook '{}' must take 3 arguments (incl &self), but found {}",
                     hook,
@@ -619,7 +622,7 @@ impl<TStore: StateReader + Clone + 'static, TTemplateProvider: TemplateProvider<
             Some("ResourceAuthAction")
         ) {
             return Err(RuntimeError::InvalidArgument {
-                argument: "CreateResourceArg",
+                argument,
                 reason: format!("Authorize hook '{}' must take a ResourceAuthAction as argument 1", hook),
             });
         }
@@ -629,7 +632,7 @@ impl<TStore: StateReader + Clone + 'static, TTemplateProvider: TemplateProvider<
             Some("AuthHookCaller")
         ) {
             return Err(RuntimeError::InvalidArgument {
-                argument: "CreateResourceArg",
+                argument,
                 reason: format!("Authorize hook '{}' must take an AuthHookCaller as argument 2", hook),
             });
         }
@@ -1532,7 +1535,7 @@ where
 
                 // Check that auth hook is valid
                 if let Some(hook) = arg.authorize_hook.as_ref() {
-                    self.check_resource_auth_hook(hook)?;
+                    self.check_resource_auth_hook("CreateResourceArg", hook)?;
                 }
 
                 // Charge the initial mint's native verification cost against the payment-funded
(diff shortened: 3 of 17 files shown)


sdbondi (commented):
Reviewed at `7146240b5`.

## Blocking

**The new builder method was inserted inside the previous method's doc comment — in all four builder files.**

`crates/template_lib/src/resource/builder/fungible.rs:426-471` (and `confidential.rs:229-274`, `non_fungible.rs:194-239`, `stealth.rs:218-263`)

The new doc block was appended to the *existing* doc comment for `with_authorization_hook` rather than started after it. Two consequences in every one of the four files:

- `with_authorization_hook` (fungible.rs:472) is now **completely undocumented** — its "Specify a hook method that will be called to authorize actions on the resource / The signature of the method must be `fn(action: ResourceAuthAction, caller: CallerContext)` / The method should panic to deny the action" text and both of its examples are gone from it.
- `with_authorization_hook_updater` (fungible.rs:467) now carries that text as its own docs, so its rendered page opens by describing hook signatures and panics, contains two `## Examples` / `# Examples` headings, and shows three examples of which two set the hook rather than the updater.

`docs build` passes because nothing here is a broken link — it is a silent misattribution. The fix is one blank line and a `///`-block boundary before `/// Sets up who can replace or remove...`, repeated four times.

## Non-blocking

**1. `a_denying_auth_hook_can_be_replaced_by_the_owner` cannot distinguish "replaced" from "cleared".**

`crates/engine/tests/access_rules.rs:1402`

The test swaps a denying hook for `caller_gated_hook`, which permits in the happy path, then asserts the transfer succeeds. That assertion holds identically if `UpdateAuthHook` had dropped the hook entirely instead of installing the new one — which is precisely the neighbouring code path (`set_auth_hook(None)`). The test therefore covers "the old hook stopped denying", not "the replacement is in force". Installing a hook that *denies* (e.g. flip `allowed` after the swap, or point at a second component whose hook panics) and asserting `AccessDeniedAuthHook` names the new hook would pin the behaviour the method is named for.

**2. `set_auth_hook_updater` does not reject `caller_component` / `direct_caller_template` in an `UpdateRule::AccessRule`.**

`crates/template_lib_types/src/access_rules.rs:630`

Every sibling builder method runs `Self::assert_no_caller_requirement(&rule)` on its `AccessRule` (`mintable`, `burnable`, `recallable`, `freezable`, `withdrawable`, `depositable`, `update_non_fungible_data`, `update_metadata`), and the engine repeats the check on `ResourceAction::UpdateAccessRule` (`impl.rs:1863`). None of them checks the *updater's* inner rule, so this is a pre-existing gap — but `set_auth_hook_updater` is the first method whose only argument is an updater, so nothing at all is validated on this path. As the `rule!` macro doc itself states (access_rules.rs:754), `caller_component(..)` always evaluates to false on a resource rule, so `set_auth_hook_updater(rule!(caller_component(x)))` compiles, builds, and yields a hook that can never be changed — the exact bricking this PR exists to prevent, reached through the API added to prevent it. Guarding the `UpdateRule::AccessRule` arm here is three lines and worth doing even if the other updaters are left for a follow-up.

**3. The action installs, as well as replaces and removes.**

`crates/template_lib/src/resource/manager.rs:962`, `crates/template_lib_types/src/access_rules.rs:485`

`set_auth_hook(Some(..))` on a resource whose `auth_hook` is `None` adds a hook where there was none. That is coherent (the updater is `Locked` by default, so the creator opts in), but every doc string on the feature says "replace or remove", and the field doc frames the risk as "change the rules that existing holders are relying on" — understating it slightly, since a resource holders inspected and found hook-free can acquire one. One clause in `ResourceManager::set_auth_hook` and the field doc would close the gap.

**4. `version_1_covers_auth_hook_updater`'s second assertion cannot fail.**

`crates/engine_types/src/substate_hasher.rs:390`

`assert_ne!(hash_at(V0, Locked), hash_at(V1, Locked))` is satisfied by the leading `SubstateHashMessage` enum tag alone (byte 0 is `0` vs `1`), independently of anything in the resource preimage — the `variant_tags_are_stable_across_versions` test asserts exactly that byte. It would still pass if `ResourceHashMessageV1` had been wired to `borsh_serialize_v0`. The first assertion is the load-bearing one; either drop the second or compare `v0[1..]` / `v1[1..]` the way the tag test does.

## Nits

- `ResourceAccessRules::borsh_serialize_v0` (access_rules.rs:687) has to be `pub` to cross the crate boundary, which puts a consensus-internal preimage serializer on the public surface of a template-facing crate. `Resource::borsh_serialize_v0` is `pub(crate)` by contrast. `#[doc(hidden)]` would keep it out of the rendered API without changing anything else.
- `Resource::borsh_serialize_v0` has no `#[cfg(feature = "borsh")]` while the `ResourceAccessRules` one does; `engine_types` pulls `tari_template_lib` with `borsh` unconditionally (Cargo.toml:16) so it compiles either way, but the two halves of the same preimage guarding differently is a small inconsistency.

## Verified sound

- **V0 transcription (the check you asked for).** `ResourceAccessRules::borsh_serialize_v0` writes all sixteen pre-change fields in exactly declaration order (`mint`, `mint_updater`, … `update_metadata`, `metadata_updater`), and `Resource::borsh_serialize_v0` writes all eight in exactly declaration order (`resource_type`, `owner_rule`, `access_rules`, `metadata`, `total_supply`, `view_key`, `auth_hook`, `divisibility`), matching what the derived `BorshSerialize` emitted before. The destructuring guard does what it claims. Note that a pinned constant test can only ever confirm the constant matches the current code; the transcription check is the real verification, and it holds.
- **The "unlaunched everywhere" claim.** `ProtocolVersion::activations` gives V1 as genesis for mainnet/stagenet/nextnet/igor/localnet and leaves esmeralda at `[(Epoch(0), V0)]` with no scheduled activation, so no coordination is needed. `hash_substate` selects by creation epoch, so esmeralda's committed resources keep hashing at V0 even after a future activation.
- **No collateral breakage.** No exhaustive `ResourceAction` match exists outside the engine; no reference to the removed `ResourceHashMessage` name survives; `ResourceAccessRules` has no struct literal outside `new()`/`deny_all()`, both updated. The `#[cbor(default)]` + `#[serde(default)]` pair matches ~20 existing uses in the workspace and needs the new `Default` on `UpdateRule`, which is there.
- **Lock and auth ordering.** The `UpdateAuthHook` arm mirrors `UpdateAccessRule` exactly: write-lock, evaluate the updater, `AccessDenied` before anything else, mutate and unlock in a second `write_with`. Authorization is checked *before* hook validation, so an unauthorized caller learns nothing about the hook. Validating the replacement outside the closure while the lock is held matches what `UpdateMetadata` and `SetVaultFreeze` already do, and `get_template_for_component` takes no lock, so pointing the hook at the calling component (the common case, exercised by the tests) is safe.
- **Skipping the outgoing hook.** Agreed, and `Locked`-by-default is the right containment. The residual risk — whoever satisfies the updater can retire a hook holders relied on — is real but is the deliberate price, and is visible in the substate.
- **Bindings.** `ResourceAccessRules.ts` gets the field last, matching declaration order; `UpdateRule.ts` correctly unchanged; no `ResourceAction` binding exists to update. No generated artifacts or lockfiles swept in.
- **`crate_versioning.py` claim.** Consistent with the tier rollup already carried by the unreleased breaking versions.

## CI

`test (3/3)` fails on `tari_ootle_wallet_sdk_tests::transaction_service::finalization_notification_is_acted_on_without_waiting_for_the_poll` (poll count 2, expected 1) — unrelated to resources or auth hooks, and it cancelled the other two shards. All five new engine tests and all three new hasher tests, `version_0_reproduces_the_pre_auth_hook_updater_hash` included, passed across the shards before the cancel. Needs a rerun once the doc blocks are fixed.

sdbondi (commented):
**LGTM at `5ad02096c`** — no blocking findings remain. (GitHub will not let me record this as an approval on your own PR, so it is a comment; a second maintainer still needs to click approve.)

All five findings from the previous review are resolved. Re-verified individually rather than taken on the commit message:

- **Builder doc blocks** — `with_authorization_hook` sits directly after its own docs again in all four builders (`confidential.rs:254`, `fungible.rs:451`, `non_fungible.rs:219`, `stealth.rs:243`), with the updater block starting cleanly before `with_authorization_hook_updater`. Both methods now document themselves.
- **`a_replacement_auth_hook_is_in_force`** — inverted so the assertion is load-bearing: the resource starts on `caller_gated_hook` (transfer succeeds), swaps to `valid_auth_hook` with `allowed = false`, then expects `AccessDeniedAuthHook` on `Deposit`. A clear-instead-of-replace bug would let that second transfer succeed and fail the test, which the old shape could not detect. `a_denying_auth_hook_can_be_removed_by_the_owner` still covers the other half — that the outgoing denying hook is not consulted during the update — so no coverage was traded away.
- **`set_auth_hook_updater`** — rejects a `caller_component`/`direct_caller_template` requirement inside `UpdateRule::AccessRule`, pinned by `auth_hook_updater_rejects_a_caller_requirement`. All four `with_authorization_hook_updater` methods route through it, so the guard covers every entry point.
- **`version_1_covers_auth_hook_updater`** — now compares `v0[1..]` against `v1[1..]`, past the `SubstateHashMessage` tag, so wiring `ResourceHashMessageV1` to the version 0 preimage would fail here rather than passing on the tag alone.
- **Install-vs-replace** — stated on the field doc, `set_auth_hook_updater`, all four builders and `ResourceManager::set_auth_hook`. `#[doc(hidden)]` added to `borsh_serialize_v0`.

Still verified from the first pass and unchanged by this commit: the version 0 transcription writes all eight `Resource` and all sixteen `ResourceAccessRules` fields in exact declaration order; `ProtocolVersion::activations` leaves esmeralda at V0 with no scheduled activation, so bundling into V1 needs no coordination; no exhaustive `ResourceAction` match exists outside the engine; the `UpdateAuthHook` arm mirrors `UpdateAccessRule`'s lock and auth ordering, checking authorization before validating the replacement.

CI is green on `5ad02096c` — all 21 checks pass, including the wallet-SDK test (`finalization_notification_is_acted_on_without_waiting_for_the_poll`) that flaked on the previous head. `version_0_reproduces_the_pre_auth_hook_updater_hash`, `a_replacement_auth_hook_is_in_force`, `auth_hook_updater_rejects_a_caller_requirement`, `auth_hook_updater_defaults_to_locked` and `version_1_covers_auth_hook_updater` all pass across the three shards.

One leftover nit from last time, not worth another push: `Resource::borsh_serialize_v0` has no `#[cfg(feature = "borsh")]` while the `ResourceAccessRules` half does. It compiles either way since `engine_types` pulls `tari_template_lib` with `borsh` unconditionally.

PR #2555, opened 2026-09-09
feat(engine)!: derive caller identity from virtual badges

### LINKED ISSUE #2549: featreq(engine): derive caller identity from virtual badges rather than frame-context predicates
**Context**

Split out of the discussion in #2518. That issue proposed making `caller_component` walk up the call-frame stack to the nearest component ancestor, for parity with Radix's `GLOBAL_CALLER`. That proposal doesn't hold up — Radix's `global_caller` does not walk past function frames, it *replaces* the identity with `PackageBlueprint(blueprint_id)` (`radix-engine/src/system/system_modules/auth/auth_module.rs:368`, `radix-common/src/types/non_fungible_global_id.rs:90`) — and our existing `caller_component` / `direct_caller_template` pair is already a complete cover of Radix's caller semantics for an engine with no internal/owned object concept.

But the investigation surfaced a real gap, which is what this issue is about.

**Problem**

Caller identity in Ootle is a *predicate evaluated once, at method entry*, not a credential. `check_component_method_access_rules` builds `RuleContext::Caller(state.method_caller())` (`crates/engine/src/runtime/tracker_auth.rs:61`, `:147`); every other auth site passes `RuleContext::CurrentFrame`, where `CallerComponent` and `DirectCallerTemplate` return `false` unconditionally (`:319-324`). Consistently, both requirements are rejected at construction time in resource rules and covenants (`crates/template_lib_types/src/access_rules.rs:70`, tests at `:1005` / `:1012`).

Radix instead materialises caller identity as a virtual badge in the callee's auth zone at frame push (`AuthZone::new(..., direct_caller_package_address, global_caller, auth_zone_parent)`, `auth_module.rs:389`), backed by the two virtual resources in `radix-common/src/constants/native_addresses.rs`. Because it is a badge, it is re-checkable at every auth point for the lifetime of the frame — including resource-level checks.

We do have badges, virtual proofs and a per-frame `AuthorizationScope`. What we don't have is a *caller-derived* one: `AuthorizationScope::virtual_proofs` is an `Arc<IndexSet<NonFungibleAddress>>` fixed for the whole transaction, built once from the signer public keys (`crates/engine/src/runtime/auth.rs:21`, `applications/tari_ootle_app_utilities/src/transaction_executor.rs:140-146`). Nothing is stamped into it at frame-push time.

**Concrete consequence**

"Resource R may only be withdrawn while executing on behalf of component A."

Radix puts the rule on the resource:

```rust
let a_badge = NonFungibleGlobalId::global_caller_badge(a_component_address);
ResourceBuilder::new_fungible(OwnerRole::None)
    .withdraw_roles(withdraw_roles! {
        withdrawer => rule!(require(a_badge));
        withdrawer_updater => rule!(deny_all);
    })
```

We can't express that at all — `ResourceAccessRules::new().withdrawable(rule!(caller_component(a)))` is rejected at construction, and would evaluate to `false` anyway under `RuleContext::CurrentFrame`. The workaround is to gate a method on the holder instead:

```rust
// on B
ComponentAccessRules::new()
    .method("payout", rule!(caller_component(a_addr)));
// B::payout performs self.vault.withdraw(amount) internally
```

Two things are lost relative to the resource-level rule:

1. **Coverage.** The rule lives on B, not on R. Any other component C holding a vault of R withdraws with no caller check. The gate must be replicated on every holder, and cannot be retrofitted onto holders you don't control.
2. **Duration.** The check is consumed at the door. Inside `B::payout` nothing downstream can re-assert "we are acting for A" — the frame carries no residual credential, so a helper B calls cannot re-check it, and B cannot hand a narrowed form of it onward.

**Proposal**

Introduce two system virtual resources — mirroring the two Radix has — and stamp the corresponding badge into the callee's scope at frame push:

- a badge derived from a `ComponentAddress`, for the component caller (≡ `GlobalCaller::GlobalObject`);
- a badge derived from a `TemplateAddress`, for the direct caller's template (≡ `PACKAGE_OF_DIRECT_CALLER` / `GlobalCaller::PackageBlueprint`, which is what a static-function frame yields).

**Three constraints are load-bearing**

1. **Per-frame, re-stamped, never inherited.** The caller badge must live in a new per-frame field, not in the transaction-wide `Arc<IndexSet<..>>`. B's frame gets A's badge; C's frame gets B's, not A's. Merging it into the shared set would grant exactly the unbounded transitive authority that #2518's walk-up proposal would have — any static fn A transitively calls becoming a proxy for A.
2. **Not capturable as a `Proof`.** This is what stops B from capturing A's caller badge and forwarding it to C. It holds structurally today: proofs originate only from vaults and buckets (`crates/engine/src/runtime/impl.rs:2696`, `:3099`), with no virtual-proof→`Proof` path. Any future "create proof from virtual badge" API would silently break the containment, so this should be an explicit invariant with a test.
3. **Unforgeable resource address.** Same treatment as `PUBLIC_IDENTITY_RESOURCE_ADDRESS` / `STEALTH_TARI_RESOURCE_ADDRESS` (`crates/engine_types/src/substate.rs:359-368`): no template may create a resource at that address or mint those non-fungibles.

**Why this is a simplification, not just a feature**

Once caller identity is a badge, `CallerComponent` and `DirectCallerTemplate` are ordinary `NonFungibleAddress` checks through the existing `contains_badge` (`auth.rs:45`). That removes:

- `RuleContext` entirely (`tracker_auth.rs:30`) and the `Caller(..)` / `CurrentFrame` split threaded through `check_access_rule` / `check_ownership` / `check_requirement`;
- the special-cased `false` arms at `tracker_auth.rs:319-324`;
- `contains_caller_component_or_template` and the construction-time rejection of caller rules in resource rules and covenants (`access_rules.rs:70`).

Caller rules stop being unexpressible in resource contexts because there is no longer anything context-dependent about them.

**Open questions**

1. **Keep or drop `n(4)` / `n(5)`?** `RuleRequirement::CallerComponent` / `DirectCallerTemplate` could survive as sugar that desugars to the badge check, or be dropped in favour of `NonFungibleAddress`. Sugar is friendlier to template authors and keeps the CBOR indices stable; dropping them is the smaller engine.
2. **Consensus timing.** Free right now: #2503 (`bea8abb0b`, 2026-09-03) landed after `v0.40.0` (2026-09-02) and is in no tag, so `n(4)` / `n(5)` have never shipped. If this slips past a release it becomes a `ProtocolVersion`-gated change.
3. **Scope size / metering.** Ootle's `AuthorizationScope` is in-memory per frame rather than a substate, so this avoids the substate-size and fee constraints Radix hit (see the comment at `auth_module.rs:335`). Worth confirming there's no per-frame cost we care about.

**Not in scope**

A separate, smaller imprecision also surfaced in #2518: `direct_caller_template(T)` matches both a static function of `T` and a method on any component instance of `T` (as documented at `access_rules.rs:240`), where Radix can distinguish the two since its `PackageBlueprint` arm fires only for function callers. That's an independent change and should be its own issue if anyone needs it.

**References**

- #2518 — the discussion this came out of, including the corrected Radix mapping.
- #2503 — added `caller_component` / `direct_caller_template`.
- `radixdlt-scrypto`: `radix-engine/src/system/system_modules/auth/auth_module.rs:264-395`, `radix-engine/src/system/actor.rs:58`, `radix-common/src/types/non_fungible_global_id.rs:90`.

sdbondi:
Closes #2549.

## What changed

Caller identity was a **predicate evaluated once, at method entry**: `RuleContext::Caller` reached the
component-method check alone, and `CallerComponent` / `DirectCallerTemplate` returned `false` everywhere else
(and were rejected at construction on resource rules). So "resource R may only be withdrawn while executing on
behalf of A" was inexpressible — the gate had to be replicated on every holder of R, and was consumed at the
door.

Caller identity is now a **badge**, as in Radix. Two virtual resources are reserved:

| constant | badge id | Radix equivalent |
| --- | --- | --- |
| `CALLER_COMPONENT_RESOURCE_ADDRESS` (`resource_0100…0001`) | `U256(component address)` | `GLOBAL_CALLER_VIRTUAL_BADGE` |
| `DIRECT_CALLER_TEMPLATE_RESOURCE_ADDRESS` (`resource_0100…0002`) | `U256(template address)` | `PACKAGE_OF_DIRECT_CALLER_VIRTUAL_BADGE` |

`push_frame` stamps the pushing frame's component (when it has one) and its template into the callee's
`AuthorizationScope`, so the identity holds for the frame's lifetime and is checkable at every auth point —
method rules, owner rules, **resource rules**, spend conditions.

No template needs a reserved constant: `rule!(caller_component(a))` / `rule!(direct_caller_template(t))` gate
on a specific caller, `rule!(any_caller_component)` / `rule!(any_caller_template)` ask only whether there was
a caller of that kind, and `NonFungibleAddress::caller_component_badge(a)` builds the badge itself, mirroring
`from_public_key`.

Three load-bearing constraints:

1. **Never inherited.** A new per-frame `caller_badges` field, not the transaction-wide `Arc<IndexSet<..>>`,
   merged into neither parent nor child. B's frame carries A's badge; C's carries B's.
2. **Not capturable as a `Proof`.** Proofs originate only from vaults and buckets, and both badge resources
   are empty by invariant.
3. **Unforgeable address.** `is_system_reserved()` covers both badge resources plus the two genesis ones, and
   `ResourceAction::Create` rejects them.

Because caller identity is an ordinary badge, `RuleContext`, `MethodCaller`,
`contains_caller_component_or_template`, the construction-time rejections, and the
`check_ownership_in_current_frame` / `require_component_ownership` split all go away. `n(4)` / `n(5)` survive
as sugar for the badge check, keeping the CBOR indices stable.

## Three vulnerabilities fixed (`5edb7a814`)

- **Auth hook escalation — introduced by this PR.** A hook frame is pushed by the acting component, so it
  carried that component's caller badges with full write access. Any resource may bind a hook and anyone may
  deposit it into an account, so a hook could spend `caller_component(V)` against a third party's resource
  rules. Hook frames now run in `FrameWriteMode::OwnComponent`: no write locks, new substates, cross-template
  calls, buckets or proofs — only the component locked at push stays writable.
- **Sandbox escape (pre-existing).** A pushed frame now inherits its parent's write mode, so a spend-script
  frame can no longer reach a writable one through a hook.
- **Recall unbound from its resource (pre-existing).** `Recall` authorised against `resource_ref`'s rule but
  recalled from any vault, so `recallable(allow_all)` could drain a vault of any other resource. The vault's
  resource must now match.

## Input validation

Nothing in the validator chain looked at declared inputs, so naming a reserved substate was admitted,
gossiped and sequenced, only to abort at input resolution. `InputSubstateValidator` rejects it at ingress
(sender-fault: the address alone decides it).

Wired into the indexer's structural chain and the VN's `create_mempool_transaction_validator` — deliberately
**not** `create_node_transaction_validator`, which also backs `TariBlockTransactionValidator`. A rejection
rule in block validation is a consensus rule, for no gain: such a transaction already aborts
deterministically.

`SubstateId::is_virtual()` grows to cover the two badge resources; all three existing callers want the wider
set.

**A separate wallet bug rides along.** A template may store an engine-issued address in component state — an
`Account` approval keyed on a public-key identity. `add_component_dependencies` walked those into the
declared input set, and `insert_substate_if_absent` inserts with no network fetch on the `unversioned` path,
which every real call site takes. The wallet reported no error and the transaction aborted on chain. That is
a live break today, independent of this PR; it now applies the same `is_virtual()` filter
`to_referenced_substates` already does. (Code path verified; no affected live account demonstrated.)

## Compatibility

**No existing template breaks** — method and owner rules behave as before; the change is additive where
caller rules previously evaluated `false` or were rejected at construction.

**No `ProtocolVersion` gate.** `n(4)` / `n(5)` came from #2503, which landed after `v0.40.0` and is in no tag
(`git tag --contains bea8abb0b` is empty), so no committed substate can carry one.

**Validators must upgrade together**, as with any access-rule evaluation change. `crate_versioning.py impact
tari_template_lib_types --breaking` reports no bumps needed. Bindings unchanged: `RuleRequirement` gains no
variant, and the reserved addresses have no TS counterpart.

## Tests & docs

`crates/engine/tests/caller_badge.rs` — the motivating resource-gated-on-caller scenario (allowed via the
gate; denied for a top-level signer, an unrelated component, and through an intermediate frame),
`direct_caller_template` via a static function (the `component: None` path), neither badge resource yielding
a token, and the hook-escalation cases. Plus `recall.rs`, `spend_script.rs` and `access_rules.rs` coverage for
the three fixes.

`guides/authorization-and-access.mdx` moves from the predicate model to the badge model, adds a **Gating a
resource on the caller** section, and warns that calling out hands the callee a live badge for its whole
frame.

## Out of scope

`direct_caller_template(T)` still matches both a static function of `T` and a method on any instance of it,
where Radix's `PackageBlueprint` arm fires only for functions. Independent change, as the issue says.

### DIFF
--- a/applications/tari_validator_node/src/bootstrap.rs
+++ b/applications/tari_validator_node/src/bootstrap.rs
@@ -82,6 +82,7 @@ use tari_ootle_transaction_validation::{
     BasicValidations,
     BlobReferenceValidator,
     EpochRangeValidator,
+    InputSubstateValidator,
     PublishTemplateLimitValidator,
     SignatureLimitValidator,
     StealthTransactionLimitsValidator,
@@ -563,6 +564,14 @@ pub fn create_node_transaction_validator<TProvider: TemplateProvider>(
         .and_then(TemplateExistsValidator::new(template_manager))
 }
 
+/// Builds the validations applied when a transaction is admitted to this node's mempool: the node-local chain
+/// plus the two epoch-window rules and the ingress-only input check.
+///
+/// [`InputSubstateValidator`] is deliberately here rather than in [`create_node_transaction_validator`], which
+/// also backs `TariBlockTransactionValidator`. A rejection rule in block validation is a consensus rule: an
+/// upgraded validator would refuse a block that a non-upgraded one accepts. Refusing at ingress costs the
+/// sender and nothing else, and a transaction that slips past an un-upgraded node's mempool still aborts at
+/// input resolution as it does today.
 pub fn create_mempool_transaction_validator<TProvider: TemplateProvider>(
     network: Network,
     template_manager: TProvider,
@@ -571,7 +580,8 @@ pub fn create_mempool_transaction_validator<TProvider: TemplateProvider>(
     WithContext::<Epoch, Transaction, TransactionValidationError>::new()
         .map_context(
             |_| (),
-            create_node_transaction_validator(network, template_manager, constants),
+            create_node_transaction_validator(network, template_manager, constants)
+                .and_then(InputSubstateValidator::new()),
         )
         .and_then(EpochRangeValidator::new())
         .and_then(TransactionValidityWindowValidator::new(
--- a/crates/engine/src/runtime/auth.rs
+++ b/crates/engine/src/runtime/auth.rs
@@ -6,7 +6,7 @@ use std::fmt::Display;
 use indexmap::IndexSet;
 use tari_template_lib::{
     models::ProofId,
-    types::{NonFungibleAddress, ResourceAddress},
+    types::{ComponentAddress, NonFungibleAddress, ResourceAddress, TemplateAddress},
 };
 
 #[derive(Debug, Clone, Default)]
@@ -20,6 +20,13 @@ pub struct AuthorizationScope {
     /// the transaction signer public key
     virtual_proofs: IndexSet<NonFungibleAddress>,
 
+    /// System-issued badges naming the frame's immediate caller: the caller's component (when it has one) and its
+    /// template. Held per frame rather than in the transaction-wide `virtual_proofs` so that a callee only ever sees
+    /// the frame directly below it — the badges are re-derived at every push and never merged into a parent or child
+    /// scope. There is no path from a badge to a [`ProofId`], so a callee cannot forward the identity it was called
+    /// with.
+    caller_badges: IndexSet<NonFungibleAddress>,
+
     /// Resource-based proofs
     proofs: IndexSet<ProofId>,
 }
@@ -28,6 +35,7 @@ impl AuthorizationScope {
     pub fn new(virtual_proofs: IndexSet<NonFungibleAddress>) -> Self {
         Self {
             virtual_proofs,
+            caller_badges: IndexSet::new(),
             proofs: IndexSet::new(),
         }
     }
@@ -35,22 +43,33 @@ impl AuthorizationScope {
     pub fn empty() -> Self {
         Self {
             virtual_proofs: IndexSet::new(),
+            caller_badges: IndexSet::new(),
             proofs: IndexSet::new(),
         }
     }
 
-    pub fn virtual_proofs(&self) -> &IndexSet<NonFungibleAddress> {
-        &self.virtual_proofs
+    /// Stamps the identity of the frame that is pushing the frame this scope belongs to. The component badge is
+    /// omitted for a static function frame, which has no component identity.
+    pub(super) fn set_caller(&mut self, component: Option<ComponentAddress>, template: TemplateAddress) {
+        self.caller_badges.clear();
+        if let Some(component) = component {
+            self.caller_badges
+                .insert(NonFungibleAddress::caller_component_badge(component));
+        }
+        self.caller_badges
+            .insert(NonFungibleAddress::direct_caller_template_badge(template));
+    }
+
+    fn badges(&self) -> impl Iterator<Item = &NonFungibleAddress> {
+        self.virtual_proofs.iter().chain(&self.caller_badges)
     }
 
     pub fn contains_badge(&self, nf_address: &NonFungibleAddress) -> bool {
-        self.virtual_proofs.contains(nf_address)
+        self.virtual_proofs.contains(nf_address) || self.caller_badges.contains(nf_address)
     }
 
     pub fn contains_badge_of_resource(&self, resource_address: &ResourceAddress) -> bool {
-        self.virtual_proofs
-            .iter()
-            .any(|badge| badge.resource_address() == resource_address)
+        self.badges().any(|badge| badge.resource_address() == resource_address)
     }
 
     pub fn proofs(&self) -> &IndexSet<ProofId> {
@@ -77,8 +96,8 @@ impl AuthorizationScope {
 impl Display for AuthorizationScope {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(f, "Virtual: [")?;
-        for proof in &self.virtual_proofs {
-            write!(f, "{}", proof)?;
+        for badge in self.badges() {
+            write!(f, "{}", badge)?;
         }
         write!(f, "], Proofs: [")?;
         for proof in &self.proofs {
--- a/crates/engine/src/runtime/error.rs
+++ b/crates/engine/src/runtime/error.rs
@@ -214,6 +214,16 @@ pub enum RuntimeError {
     WriteInReadOnlyContext,
     #[error("Host operation '{operation}' is forbidden inside a read-only (spend-script) context")]
     ForbiddenInReadOnlyContext { operation: &'static str },
+    #[error("Write to {id} attempted in a resource auth hook, which may only modify its own component state")]
+    WriteOutsideOwnComponent { id: SubstateId },
+    #[error("Host operation '{operation}' is forbidden inside a resource auth hook")]
+    ForbiddenInAuthHookContext { operation: &'static str },
+    #[error("Recall on resource {resource_address} targeted vault {vault_id}, which holds resource {vault_resource}")]
+    RecallResourceMismatch {
+        vault_id: VaultId,
+        resource_address: ResourceAddress,
+        vault_resource: ResourceAddress,
+    },
     #[error("Spend script rejected the spend: {details}")]
     SpendScriptRejected { details: Box<RuntimeError> },
     #[error("Spend condition not met: {details}")]
--- a/crates/engine/src/runtime/impl.rs
+++ b/crates/engine/src/runtime/impl.rs
@@ -1573,7 +1573,7 @@ where
                     // to be unforgeable.
                     if resource_address.is_system_reserved() {
                         return Err(RuntimeError::InvalidArgument {
-                            argument: "address_allocation",
+                            argument: "resource_address",
                             reason: format!("Resource address {resource_address} is reserved by the system"),
                         });
                     }
--- a/crates/engine/src/runtime/impl.rs
+++ b/crates/engine/src/runtime/impl.rs
@@ -1387,7 +1387,7 @@ where
                     let component = state.get_component(&component_lock)?;
                     state
                         .authorization()
-                        .require_component_ownership(ComponentAction::SetAccessRules, component.as_ownership())?;
+                        .require_ownership(ComponentAction::SetAccessRules, component.as_ownership())?;
 
                     state.modify_component_with(&component_lock, |component| {
                         if access_rules == *component.access_rules() {
@@ -1567,6 +1567,16 @@ where
                         },
                         None => state_mut.id_provider()?.new_resource_address()?,
                     };
+
+                    // The system's resource addresses must stay under the system's control: the genesis resources
+                    // are created once, and the two caller-badge resources must stay empty for the engine's badges
+                    // to be unforgeable.
+                    if resource_address.is_system_reserved() {
+                        return Err(RuntimeError::InvalidArgument {
+                            argument: "address_allocation",
+                            reason: format!("Resource address {resource_address} is reserved by the system"),
+                        });
+                    }
 
                     let mut payload = Metadata::from_iter([("resource_type", resource.resource_type().to_string())]);
                     if let Some(symbol) = resource.metadata().get(TOKEN_SYMBOL) {
@@ -1855,14 +1865,6 @@ where
                         })?;
                 let UpdateAccessRuleArg { action, new_rule } = args.assert_one_arg()?;
 
-                if new_rule.contains_caller_component_or_template() {
-                    return Err(RuntimeError::InvalidArgument {
-                        argument: "new_rule",
-                        reason: "caller_component/direct_caller_template cannot be used on a resource access rule"
-                            .to_string(),
-                    });
-                }
-
                 let resource_lock = self.tracker.write_with(|state_mut| {
                     let resource_lock = state_mut.write_lock_substate(SubstateId::Resource(resource_address))?;
 
@@ -1871,9 +1873,7 @@ where
 
                     let authorized = match updater {
                         UpdateRule::Locked => false,
-                        UpdateRule::Owner => state_mut
-                            .authorization()
-                            .check_ownership_in_current_frame(resource.as_ownership())?,
+                        UpdateRule::Owner => state_mut.authorization().check_ownership(resource.as_ownership())?,
                         UpdateRule::AccessRule(rule) => state_mut.authorization().check_access_rule(rule)?,
                     };
 
@@ -1918,9 +1918,7 @@ where
 
                     let authorized = match updater {
                         UpdateRule::Locked => false,
-                        UpdateRule::Owner => state_mut
-                            .authorization()
-                            .check_ownership_in_current_frame(resource.as_ownership())?,
+                        UpdateRule::Owner => state_mut.authorization().check_ownership(resource.as_ownership())?,
                         UpdateRule::AccessRule(rule) => state_mut.authorization().check_access_rule(rule)?,
                     };
 
@@ -3817,7 +3815,7 @@ where
             let component = state.get_component(locked)?;
             state
                 .authorization()
-                .require_component_ownership(action, component.as_ownership())
+                .require_ownership(action, component.as_ownership())
         })
     }
 
--- a/crates/engine/src/runtime/impl.rs
+++ b/crates/engine/src/runtime/impl.rs
@@ -170,7 +170,7 @@ use crate::{
         error::{ArgumentValidationError, LimitError},
         locking::{LockError, LockedSubstate},
         pay_fee::PayFee,
-        scope::PushCallFrame,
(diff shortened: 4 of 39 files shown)


sdbondi (commented):
Reviewed at `e8e832cd7`. No blocking findings — the three load-bearing constraints hold under the checks below. Comments are suggestions and nits.

## Verified

- **Non-inheritance, both directions.** `AuthorizationScope::update_from_child` (`crates/engine/src/runtime/auth.rs:91`) merges only `proofs`, and `CallScope::update_from_parent` (`crates/engine/src/runtime/scope.rs:199`) is a no-op, so the badges genuinely cannot travel up or down. `set_caller` clears before inserting, which matters because a `ForComponent` frame starts from `AuthorizationScope::empty()` rather than a parent clone.
- **The refactor is behaviour-preserving.** `self.call_frames.last().is_none()` ⟺ the old `call_frame_depth() == 0` (`working_state.rs:1455`). The badge is derived at push from `call_frames.last()`, which is exactly the frame the old `method_caller()` read as `call_frames[len-2]` at check time; a frame's component lock and template are both fixed for its lifetime, so the two are identical. Dropping `RuleContext` is likewise a no-op for every other arm — `ScopedToComponent`/`ScopedToTemplate` never consulted it, and both call sites always passed the same current-frame `scope`.
- **No badge → `Proof` path.** The only two constructors that mint a proof out of a virtual badge (`impl.rs:1459` `GetOwnerProof`, `impl.rs:3634` `GetSignerProof`) both build `ResourceContainer::public_key(..)`, hardcoded to `PUBLIC_IDENTITY_RESOURCE_ADDRESS`; every other proof originates from a bucket or vault of a real resource. So invariant 2 rests on the badge resources being empty, and `ResourceAction::Create` (`impl.rs:1594`) is the only resource-creation path in the engine — genesis (`applications/tari_ootle_app_utilities/src/genesis_resources.rs`) creates only the identity and TARI resources.
- **Address space.** `0100…01` / `0100…02` collide with nothing in `constants.rs`.
- **Timing claim.** `git tag --contains bea8abb0b` is empty — #2503 is in no tag, so the "no committed substate or receipt carries an `n(4)`/`n(5)`" argument checks out.
- **Versioning claim.** `crate_versioning.py impact tari_template_lib_types --breaking` reports "No version changes required", as the description says.
- No leftover references to `MethodCaller`, `method_caller`, `check_ownership_in_current_frame`, `require_ownership_in_current_frame`, `require_component_ownership`, or `contains_caller_component_or_template`.

## Non-blocking

**1. The invariant-2 test asserts nothing about why it failed.** `crates/engine/tests/caller_badge.rs:129` — `caller_badge_resources_hold_no_tokens` calls `execute_expect_failure` and discards the reason. This is the one test carrying "there is no badge → `Proof` path", and it passes on *any* failure: an arg-decode panic, a scope error, a renamed function. Assert the reason the way the sibling tests do (`assert_reject_reason` on the substate-not-found for the badge resource), so it fails loudly if a resource ever does come to exist at either address.

**2. `direct_caller_template` has no end-to-end coverage.** The five new tests all exercise `caller_component`; `direct_caller_template` on a resource rule is only covered by the builder-level `resource_rule_allows_caller_requirements`. That leaves the other half of `set_caller` — the static-function frame, which stamps the template badge and *no* component badge — untested end to end. A `CallFunction` gated by `direct_caller_template(T)` would cover the `component: None` path.

**3. `contains_badge_of_resource` now matches the reserved resources.** `auth.rs:70` chains `caller_badges`, so `rule!(resource(CALLER_COMPONENT_RESOURCE_ADDRESS))` becomes a usable "called by any component" rule and `rule!(resource(DIRECT_CALLER_TEMPLATE_RESOURCE_ADDRESS))` an "any nested frame" rule. Both look harmless and arguably useful, but they are a capability the PR doesn't mention. Either document them as intended or exclude `is_caller_badge()` resources from `contains_badge_of_resource`.

**4. The `rule!` caution covers the callee hazard but not the new caller-side one.** `crates/template_lib_types/src/access_rules.rs:651-656` warns about proxy methods and package identity — both callee-gate concerns that predate this PR. The hazard this change introduces is the converse: calling any component now hands that component your identity as a live badge for the whole of its frame, usable at every auth point. Under the old model `caller_component` was consumed at the door and nothing deeper could re-assert it; that is precisely the property being traded away, and it's the one a template author needs to know about before calling out. Worth a sentence.

## Nits

**5. `impl.rs:1580` — wrong argument name on one of two paths.** The reject is `argument: "address_allocation"`, but the check also fires for the `None` branch where the address came from `id_provider().new_resource_address()`. `"resource_address"` covers both. (The check itself is unreachable today — allocations and generated addresses both derive from the transaction hash — which is fine as defence in depth, just noting there's no test and can't easily be one.)

**6. `crates/engine/tests/templates/caller_badge_resource/src/lib.rs:39`** — `balance()` is unused by any test.

**7. Description could note the second consensus-adjacent change.** The "no `ProtocolVersion` gate" argument is made only about `n(4)`/`n(5)`, but widening `SubstateId::is_virtual()` also changes `UnsignedTransactionV1::to_referenced_substates` (`crates/transaction/src/v1/unsigned.rs:211`). I checked the callers — they're all in `applications/tari_walletd/src/handlers/transaction.rs`, so it's wallet-side input auto-fill and not consensus. Worth saying so explicitly, since the diff makes it look like it could be.

sdbondi (commented):
Re-reviewed at `3a67fe585`. All seven items from the previous round are resolved; nothing outstanding from my side. LGTM.

Verified individually rather than taking the commit message for it:

- **Mint-attempt reason is now pinned.** `assert_reject_reason(reason, format!("{resource} not found"))` matches `RejectReason::SubstateNotFound(format!("{id} not found"))` (`crates/engine/src/runtime/error.rs:365`) as a substring, and dropping the args in favour of a fixed `[0u8; 32]` id removes the decode path that could previously have satisfied `execute_expect_failure` for the wrong reason.
- **The `component: None` branch of `set_caller` is now covered end to end.** In `resource_withdraw_rule_allows_a_static_function_of_the_gated_template`, `withdraw_from_static` runs as a top-level `CallFunction` — depth 0, so no badge is stamped on it — and the holder's depth-1 frame therefore carries only `direct_caller_template(caller_template)`. That is exactly the path the first round had no coverage for. `resource_withdraw_rule_denies_a_caller_from_another_template` gates on `holder_template` and calls in from a `Caller` component, pinning that the badge names the caller's template and not the callee's.
- **The `contains_badge_of_resource` consequence is documented** on both constants, and both readings check out against `auth.rs:70` + `working_state.rs:1525`: `resource(CALLER_COMPONENT_..)` is satisfied by any component caller and by no top-level instruction, and `resource(DIRECT_CALLER_TEMPLATE_..)` by every frame below the top level.
- **Caller-side caution** added at `access_rules.rs:659`, and it names the right thing — that the badge is live for the callee's whole frame rather than consumed at the method it entered through, which is the property the old model had and this one trades away.
- `argument: "resource_address"` now covers both the allocation and id-provider branches; `balance()` removed; the description now states the `is_virtual()` widening reaches only `applications/tari_walletd/src/handlers/transaction.rs`.

CI green on this head, integration tests included.

Approving in text only — GitHub won't take a formal approval from me on this PR. It still needs a maintainer approval before merge.

sdbondi (commented):
Reviewed `945908ab8` (docs) and `d860e296b` (input validation). The validator's premise checks out and the docs match the code. Three things below, none blocking — the first is the one worth a decision.

## Wiring

**`InputSubstateValidator` only reaches the structural chain, not the validator node's mempool.** It's added to `create_structural_transaction_validator` (`crates/transaction_validation/src/builder.rs:44`), whose consumers are `create_gossip_transaction_validator` and the indexer (`applications/tari_indexer/src/transaction_manager/mod.rs:123`). The validator node builds its own parallel chain in `create_node_transaction_validator` (`applications/tari_validator_node/src/bootstrap.rs:540`), which does not include it — and that chain feeds both `create_mempool_transaction_validator` (`:566`) and `TariBlockTransactionValidator` (`:378`). So a transaction submitted straight to a VN's JSON-RPC still gets sequenced and aborted at input resolution, which is the outcome the validator exists to prevent.

Keeping it out of `TariBlockTransactionValidator` looks right and I'd not change that — a new rejection rule in block validation makes upgraded and non-upgraded validators disagree on block validity, and anyone can craft a transaction naming `resource_0100…0001` today. But the two chains are otherwise identical modulo `TemplateExistsValidator`, so the *mempool* gap reads as accidental rather than chosen. Either wire it into the mempool on its own, or say why the indexer is the only gate.

Either way `create_structural_transaction_validator`'s docstring is now inaccurate: "The validator node composes the same validators plus the context-dependent ones (dry-run rejection, template existence, epoch range)" (`builder.rs:31-32`) — it no longer does.

## Tests

**The `Resource(PUBLIC_IDENTITY_RESOURCE_ADDRESS)` asymmetry is the subtlest line in the new match and nothing pins it.** `is_virtual()` returns true for a public-identity *non-fungible* but false for the *resource itself* (`crates/engine_types/src/substate.rs:324-329`), which is correct — genesis stores that resource (`applications/tari_validator_node/src/genesis_state.rs:63`) while its NFTs are issued from signatures. Nothing tests it, and getting it wrong would make every account transaction unsubmittable through an indexer.

**`validate(TARI_TOKEN.into())` in `it_accepts_a_real_substate` can't fail.** `all_inputs_iter` filters TARI_TOKEN out of `inputs()` and chains `XTR_REQUIREMENT` unconditionally (`crates/transaction/src/v1/transaction.rs:190-196`), so that assertion holds no matter what the validator does; the `ComponentAddress` line is carrying the test. Swapping the TARI_TOKEN assertion for `PUBLIC_IDENTITY_RESOURCE_ADDRESS` makes it non-vacuous and covers the point above in the same line.

## The wallet fix is a bug fix the description doesn't claim

Worth calling out in the PR body, because it stands on its own. `insert_substate_if_absent` inserts without any network fetch on the `unversioned` path (`crates/wallet/sdk/src/apis/substate.rs:291-292`), so an `Account` approval keyed on a public-key badge has been silently added to the declared input set. At consensus, `resolve_local_versions` calls `get_latest_version` on a substate that was never stored and aborts with `SubstateNotFound` (`crates/consensus/src/hotstuff/transaction_manager/manager.rs:210-216`). So those transactions fail today — which is both what makes the validator's "can never resolve" premise true and, separately, a live break in account approvals that `add_component_dependencies` now fixes.

It also settles the false-reject question in the validator's favour: because such transactions already abort at input resolution, this moves an already-failing transaction's rejection earlier and cheaper rather than rejecting anything that currently works. Only the *caller-badge* half of `is_virtual()` is new; the public-key-identity half is retroactive, and it's safe for that reason.

## Sound

- `calculate_id()` on the error path only, rather than eagerly like most of the sibling validators.
- The `is_virtual()` widening to `Resource(caller_badge)` is inert at its other five call sites — the two engine scope filters (`runtime/scope.rs:222,232`) and the three in `to_referenced_substates` — since nothing can exist at those addresses.
- Docs match the code: both `rule!(resource(..))` readings agree with `auth.rs:70` + `working_state.rs:1525`, the non-inheritance claim matches `update_from_child`, and the resource-gating example is the `caller_badge_resource` test template's rule.

PR #2556, opened 2026-09-09
fix(common_types)!: enforce shard bounds for fee pool addresses and shard groups

### LINKED ISSUE #2469: [Hardening] Fee-pool invariant regression test + two defense-in-depth suggestions
Hello Ootle team,

We are a security auditing team ([AFAP](https://github.com/AFAP)) doing proactive static review of the Ootle fee-pool path (wallet → engine authorization → consensus deduction → shard filtering). No exploitable issue to report — three hardening suggestions from our evidence chain (public code only, nothing probed).

### 1. Guard the "diffs never contain ValidatorFeePool Up/Down" invariant

Our review of `PendingSubstateStore::put_diff` (`crates/consensus/src/hotstuff/substate_store/pending_store.rs:237-301`) confirms that correctness currently relies on an implicit premise: fee-pool creation/deposits only ever happen through `apply_leader_fee_to_substate_store`'s `update_in_place` creator path, so diffs legitimately never carry fee-pool Up/Down changes (which `put_diff` skips). This invariant is enforced nowhere. If a future change lets templates or SDK code create pools through normal execution, engine-side and consensus-side state would silently diverge. Suggest a debug assertion / unit test: "any SubstateDiff containing a fee-pool Up/Down must be rejected".

### 2. `derive_fee_pool_address` panic → Result

`crates/common_types/src/fee_pool.rs:27` panics on `shard == 0`. Unreachable via normal committee construction (`Shard::first() = 1`), but any future caller that reaches it through `ShardGroup::new_unchecked` misuse turns this into a node-level DoS. A checked variant returning `Result` would make the invariant local instead of ecosystem-wide.

### 3. CI check for duplicated generated layouts

Prover-side and verifier-side each keep a copy of the generated `circuit_layout.rs`. We verified all current copies are byte-identical (SHA-256 across 7 machine families); a CI drift check would keep that true under refactors.

Happy to provide file:line references for anything above.

Best regards,
AFAP security research

sdbondi:
Closes #2469

## Description

Two shard-bound invariants that untrusted input could break.

### 1. `derive_fee_pool_address` panicked on the global shard

It panicked on `shard == 0` (`checked_sub(1).expect(...)`) and, for a shard above `num_preshards`, silently wrapped the shard index into another shard's prefix — returning a valid-looking address in the wrong shard.

Both are reachable from wallet daemon JSON-RPC input, not just from internal misuse:

- `validators.claim_fees` maps the caller's `shards: Vec<Shard>` straight into it (only emptiness is checked), so `shards: [0]` panics the handler. The release profile sets `panic = 'abort'`, so that takes down the daemon process.
- `validators.get_fees` iterates a caller-supplied `shard_group`, which reached the same panic.

The function now returns `Result<ValidatorFeePoolAddress, InvalidFeePoolShard>`, checking `1..=num_preshards` where the address is derived instead of leaving the invariant to every caller:

- wallet daemon: maps the error to an invalid-params JSON-RPC response.
- consensus (`apply_leader_fee_to_substate_store`): its shard always comes from the local committee's shard group, so the error maps to `HotStuffError::InvariantError`.

### 2. `ShardGroup` did not uphold `start <= end_inclusive` when deserialized

`ShardGroup` documents that ordering and its API takes it as given — `is_empty` is a const `false`, `len` underflows without it, and `shard_iter` was `iter::successors` walking until it *reached* `end_inclusive`. `new`, `new_checked`, `decode_from_u32` and `FromStr` all enforce it, but the derived `Deserialize`/`Decode` did not, so an inverted group like `{"start": 5, "end_inclusive": 2}` deserialized fine. Iterating one never terminated: it grew until the `shard + 1` counter overflowed (release builds have `overflow-checks = true`), after allocating ~4 billion shards. `validators.get_fees` did exactly that with a caller-supplied group.

`Deserialize` and `Decode` now go through `new_checked`, so an inverted group is a deserialization error rather than a value no accessor can describe, and `shard_iter` walks an inclusive range, which terminates for any bounds — a `new_unchecked` value (foreign proposals build one from unvalidated peer data before validation rejects it) now yields nothing instead of looping.

`new_checked` also bounds the end at `Shard::max()`. `len` adds one to `end_inclusive` before subtracting `start`, so a group ending at `u32::MAX` passes the ordering check and still overflows; `NumPreshards` tops out at 256, which `encode_as_u32`/`decode_from_u32` already assume, so requiring `end_inclusive <= Shard::max()` makes `len` total for every group the constructors can produce.

`state_store_rocksdb`'s `ShardGroupCodec` decodes the two shards separately and rebuilt the group with the panicking constructor, bypassing `Decode` entirely; it now returns `MalformedData`. Two remaining sites read a node's own SQLite columns the same way (`storage_sqlite/src/global/models/committee.rs:39`, `backend_adapter.rs:804`) — plumbing an error through those row mappings is left for a follow-up.

Reported by AFAP in #2469 (suggestion 2); the `ShardGroup` half came out of reviewing the first fix. Suggestions 1 and 3 in that issue do not hold — diffs do routinely carry fee-pool Up/Down changes, so the assertion suggested there would reject ordinary claim transactions, and no `circuit_layout.rs` exists in this repo. Both are answered on the issue, so merging this closes it.

No crate version bumps: `scripts/crate_versioning.py impact tari_ootle_common_types --breaking` reports every affected crate is already on an unreleased bump.

## Motivation and Context

An authenticated wallet daemon RPC call could crash the daemon (shard 0) or hang it while allocating billions of shard ids (inverted shard group), and an out-of-range shard could silently produce a fee pool address that does not live in the requested shard.

## How Has This Been Tested?

- New unit tests: `it_rejects_shards_outside_the_preshard_range` (shard 0, `num_preshards + 1` at two widths, `u32::MAX`), `it_rejects_inverted_bounds_when_deserializing` and `it_rejects_an_end_beyond_the_maximum_shard` (both JSON and CBOR), `it_round_trips_a_valid_group` (the hand-written `Decode` must match the still-derived `Encode`/`CborLen` framing) and `shard_iter_terminates_on_inverted_bounds`.
- Existing `tari_ootle_common_types` tests pass.
- `cargo lints clippy --all-targets` on `tari_ootle_common_types`, `tari_consensus`, `tari_ootle_walletd`, `consensus_tests`; `cargo check --all-targets` additionally on `tari_validator_node`, `tari_indexer`, `tari_ootle_p2p`, `tari_ootle_storage`.

Against a wallet daemon: `validators.claim_fees` with `shards: [0]` aborted the daemon before this change and now returns an invalid-params error; `validators.get_fees` with `shard_group: {"start": 5, "end_inclusive": 2}` hung allocating before, and is now rejected at deserialization.

## Breaking Changes

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

`derive_fee_pool_address` changes signature (public API of `tari_ootle_common_types`). No consensus or state changes: derived addresses for valid shards are unchanged, and `ShardGroup`'s encoding is untouched — only decoding of groups that were already outside the type's stated invariant now fails.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12005,6 +12005,7 @@ dependencies = [
  "prost-types 0.14.3",
  "rand 0.10.1",
  "serde",
+ "serde_json",
  "tari_bor",
  "tari_crypto",
  "tari_engine_types",
--- a/applications/tari_walletd/src/handlers/validator.rs
+++ b/applications/tari_walletd/src/handlers/validator.rs
@@ -68,9 +68,12 @@ pub async fn handle_get_validator_fees(
 
     let ids = shards
         .into_iter()
-        .map(|shard| derive_fee_pool_address(&claim_public_key, NUM_PRESHARDS, shard))
-        .map(SubstateId::from)
-        .collect::<Vec<_>>();
+        .map(|shard| {
+            derive_fee_pool_address(&claim_public_key, NUM_PRESHARDS, shard)
+                .map(SubstateId::from)
+                .map_err(|err| invalid_params("shard_group", Some(err)))
+        })
+        .collect::<Result<Vec<_>, _>>()?;
 
     let mut fees = HashMap::with_capacity(ids.len());
     const CHUNK_SIZE: usize = 20;
@@ -137,8 +140,11 @@ pub async fn handle_claim_validator_fees(
     let fee_pool_addresses = req
         .shards
         .iter()
-        .map(|shard| derive_fee_pool_address(&claim_public_key, NUM_PRESHARDS, *shard))
-        .collect::<Vec<_>>();
+        .map(|shard| {
+            derive_fee_pool_address(&claim_public_key, NUM_PRESHARDS, *shard)
+                .map_err(|err| invalid_params("shards", Some(err)))
+        })
+        .collect::<Result<Vec<_>, _>>()?;
 
     // build the transaction
     let max_fee = req.max_fee.max(1);
--- a/crates/common_types/Cargo.toml
+++ b/crates/common_types/Cargo.toml
@@ -33,6 +33,9 @@ serde = { workspace = true, default-features = true }
 ts-rs = { workspace = true, optional = true }
 thiserror = { workspace = true }
 
+[dev-dependencies]
+serde_json = { workspace = true }
+
 [package.metadata.cargo-machete]
 ignored = ["prost", "prost-types"] # false positive, used in OUT_DIR structs
 
--- a/crates/common_types/src/fee_pool.rs
+++ b/crates/common_types/src/fee_pool.rs
@@ -12,8 +12,8 @@ use crate::{NumPreshards, shard::Shard};
 /// grown — every existing pool then migrates to the lower-numbered child shard rather than fanning out by pk bits.
 /// Reserving two bytes lets `NumPreshards` grow to at most `2^16` without changing the layout.
 ///
-/// `shard` must be in `1..=num_preshards`: shard 0 is the global shard and has no index, and an index that does not
-/// fit in `shard_bits` would wrap into another shard's prefix, silently yielding an address in the wrong shard.
+/// `shard` must be in `1..=num_preshards`: shard 0 is the global shard and has no index, and an index outside that
+/// range does not fit in `shard_bits` and aliases another shard's prefix.
 pub fn derive_fee_pool_address(
     public_key_bytes: &RistrettoPublicKeyBytes,
     num_preshards: NumPreshards,
@@ -37,7 +37,7 @@ pub fn derive_fee_pool_address(
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, thiserror::Error)]
-#[error("Shard {shard} cannot hold a validator fee pool: expected a shard in 1..={}", .num_preshards.as_u32())]
+#[error("Shard {} cannot hold a validator fee pool: expected a shard in 1..={}", .shard.as_u32(), .num_preshards.as_u32())]
 pub struct InvalidFeePoolShard {
     pub shard: Shard,
     pub num_preshards: NumPreshards,
--- a/crates/common_types/src/fee_pool.rs
+++ b/crates/common_types/src/fee_pool.rs
@@ -11,26 +11,36 @@ use crate::{NumPreshards, shard::Shard};
 /// prefix encode `shard - 1`; the remaining prefix bits are zero so they are stable if `NumPreshards` is later
 /// grown — every existing pool then migrates to the lower-numbered child shard rather than fanning out by pk bits.
 /// Reserving two bytes lets `NumPreshards` grow to at most `2^16` without changing the layout.
+///
+/// `shard` must be in `1..=num_preshards`: shard 0 is the global shard and has no index, and an index that does not
+/// fit in `shard_bits` would wrap into another shard's prefix, silently yielding an address in the wrong shard.
 pub fn derive_fee_pool_address(
     public_key_bytes: &RistrettoPublicKeyBytes,
     num_preshards: NumPreshards,
     shard: Shard,
-) -> ValidatorFeePoolAddress {
+) -> Result<ValidatorFeePoolAddress, InvalidFeePoolShard> {
+    if shard.is_global() || shard.as_u32() > num_preshards.as_u32() {
+        return Err(InvalidFeePoolShard { shard, num_preshards });
+    }
     // For num_preshards = 256, log2(256) = 8
     let shard_bits = num_preshards.as_u32().trailing_zeros();
     // shift required to place the shard index in the top `shard_bits` of the 2-byte prefix
     let shift = u16::BITS - shard_bits;
     // shard 0 is global, so shard is count-based; convert to an index
-    let shard_index = shard
-        .as_u32()
-        .checked_sub(1)
-        .expect("derive_fee_pool_address: shard 0 is reserved for global");
+    let shard_index = shard.as_u32() - 1;
     let prefix = (shard_index << shift) as u16;
 
     let mut address = [0u8; 32];
     address[..2].copy_from_slice(&prefix.to_be_bytes());
     address[2..].copy_from_slice(&public_key_bytes[2..]);
-    ValidatorFeePoolAddress::from_array(address)
+    Ok(ValidatorFeePoolAddress::from_array(address))
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, thiserror::Error)]
+#[error("Shard {shard} cannot hold a validator fee pool: expected a shard in 1..={}", .num_preshards.as_u32())]
+pub struct InvalidFeePoolShard {
+    pub shard: Shard,
+    pub num_preshards: NumPreshards,
 }
 
 #[cfg(test)]
@@ -42,7 +52,7 @@ mod tests {
     use crate::SubstateAddress;
 
     fn assert_pool_lands_in_shard(pk: &RistrettoPublicKeyBytes, num_preshards: NumPreshards, shard: Shard) {
-        let fee_pool_address = derive_fee_pool_address(pk, num_preshards, shard);
+        let fee_pool_address = derive_fee_pool_address(pk, num_preshards, shard).unwrap();
         let addr = SubstateAddress::from_substate_id(&fee_pool_address.into(), 0);
         assert_eq!(addr.to_shard(num_preshards), shard);
     }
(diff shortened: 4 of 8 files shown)


sdbondi (commented):
Reviewed at 874be87. Nothing blocking.

## Verified

- **All call sites converted.** `derive_fee_pool_address` has exactly four callers (`consensus/src/hotstuff/common.rs:403`, `consensus_tests/src/consensus.rs:1622`, `tari_walletd/src/handlers/validator.rs:72` and `:144`); all handle the new `Result`. `pub use fee_pool::*` in `common_types/src/lib.rs:40` exports `InvalidFeePoolShard`, so downstream crates can match on it.
- **The bound is the right one.** With `1..=num_preshards`, `shard_index` is at most `num_preshards - 1` and `shift == 16 - log2(num_preshards)`, so `shard_index << shift < 2^16` — the `as u16` cast at `fee_pool.rs:31` is now provably lossless for every `NumPreshards` variant, including `P1` (`shift == 16`, index `0`).
- **No derivation change for valid shards.** The only change on the happy path is `checked_sub(1).expect(...)` → `- 1`, so the address bytes are identical. The "no consensus or state changes" claim in the description holds.
- **The consensus mapping is justified.** Both callers (`on_propose.rs:592`, `on_ready_to_vote_on_local_block.rs:527`) pass `local_committee_info.shard_group().start()`, which is always in range, so `InvariantError` is the correct classification rather than a validation error.
- **The walletd error actually surfaces as invalid-params.** `invalid_params` (`handlers/helpers.rs:151`) returns an `anyhow::Error` wrapping a `JsonRpcError`, and `resolve_any_error` (`server.rs:360`) downcasts it back and preserves the reason code. `InvalidFeePoolShard: Display` via `thiserror`, satisfying the `T: Display` bound.
- **`GetValidatorFeesRequest` is the only inbound `ShardGroup`** in `clients/wallet_daemon_client/src/types.rs:1264` — the indexer's `ShardGroup` fields are all response-side — so the two handlers touched here cover the full attack surface described.
- **Version claim checked.** `scripts/crate_versioning.py impact tari_ootle_common_types --breaking` reports "No version changes required" — every affected crate is already on an unreleased bump.

## Incidental fix worth noting in the description

`validator.rs:70` also closes an unbounded-loop DoS that the description doesn't mention. `ShardGroup::shard_iter` (`shard_group.rs:117`) is `iter::successors` that terminates only on `shard == end_inclusive`, so a deserialized inverted group — `{"start": 5, "end_inclusive": 2}` — never terminated: it grew `ids` until the `shard.as_u32() + 1` overflow panicked (release has `overflow-checks = true`), after allocating ~4 billion `SubstateId`s. Now `collect::<Result<Vec<_>, _>>()` short-circuits at shard 257, so the call returns invalid-params after a bounded number of iterations. That's a strictly better outcome than the OOM, and it's arguably a more severe pre-existing bug than the shard-0 abort — worth a line in the PR body since it's the sort of thing a reader would otherwise assume is still open.

## Nits

- `fee_pool.rs:40` — the error message stutters. `Shard`'s `Display` is `Shard({n})` (`shard.rs:114`), so the JSON-RPC response reads `Invalid param 'shards': Shard Shard(0) cannot hold a validator fee pool: expected a shard in 1..=256`. `{}` on `.shard.as_u32()` would render `Shard 0 cannot hold …`.
- `fee_pool.rs:15-16` — "an index that does not fit in `shard_bits` **would** wrap into another shard's prefix" is a counterfactual about the absent guard rather than a statement of the invariant, which the repo's comment convention rules out. It's a public-API precondition so the rationale earns more latitude than an inline comment would; "an index outside this range does not fit in `shard_bits` and aliases another shard's prefix" says the same thing in the present tense.

Test coverage is proportionate — `it_rejects_shards_outside_the_preshard_range` hits shard 0, `num_preshards + 1` at two different widths, and `u32::MAX`, which is the full set of interesting boundaries.

sdbondi (commented):
Re-reviewed at d152c67. Both nits from the previous round are addressed, and the new commit goes further than suggested — it moves the `ShardGroup` ordering invariant into `Deserialize`/`Decode` rather than relying on the fee-pool bound to contain it. That's the better fix. One non-blocking finding on the new code.

## Verified in the new commit

- **The mirror struct is wire-compatible.** `UncheckedShardGroup` keeps `#[n(0)]`/`#[n(1)]` over the same `Shard` types with no container attribute, matching the `Encode`/`CborLen` still derived on `ShardGroup`, so both sides use minicbor's default array encoding. `it_round_trips_a_valid_group` is a real check, not a tautology: `tari_bor::encode`/`decode` (`tari_bor/src/lib.rs:38,149`) delegate to `minicbor::to_vec_with`/`minicbor::decode`, so the test drives the derived `Encode` against the hand-written `Decode`. The JSON half covers the serde path separately.
- **The manual impls match the shapes the derives produced.** minicbor-derive emits `impl<'bytes, Ctx> Decode<'bytes, Ctx>` (`minicbor-derive-0.19.5/src/decode.rs:125`), which the `impl<'b, C> Decode<'b, C>` here matches, so `ShardGroup` still decodes under any context. The derive only overrides `nil()` on the `#[cbor(transparent)]` path (`decode.rs:511-542`), which a two-field struct never took — so omitting `nil()` here is not a behaviour change for parents that hold a `ShardGroup` in an array slot.
- **`Decoder::position` and `Error::message`/`at` exist as used** in minicbor 2.3.0 (`decode/decoder.rs:32`, `decode/error.rs:69,134`); the `Display`-taking `message` is the `alloc` variant, and the crate enables `alloc`.
- **`shard_iter` is now total.** `(start..=end).map(Shard::from)` yields nothing on inverted bounds instead of looping to the `+ 1` overflow. This also hardens `foreign_proposal.rs:450 shard_group_unchecked`, which builds a `ShardGroup` straight from a peer's block header and bypasses `Deserialize` entirely — the validation path does gate on `shard_group_checked` first (`on_receive_foreign_proposal.rs:456`), but the `shard_iter` change means an inverted group reaching one of the other `shard_group_unchecked` call sites can no longer hang the caller.
- **Lockfile delta is exactly the new dev-dep** — `serde_json` on `tari_ootle_common_types` and nothing else swept in.

## Non-blocking

`shard_group.rs:59` and `:68` — the error message is built eagerly on every decode, including every success:

```rust
let unchecked = UncheckedShardGroup::decode(d, ctx)?;
let message = unchecked.invalid_bounds_message();  // format! runs unconditionally
unchecked.check().ok_or_else(|| decode::Error::message(message).at(pos))
```

`ShardGroup` is a field of `BlockHeader` (`storage/src/consensus_models/block_header.rs:61`, which derives both `Deserialize` and `Decode`) and of `ProposalCertificate` (`consensus_types/src/certificates/proposal_certificate.rs:36`), so this is a heap allocation plus two integer formats on every block header and every QC decoded from the wire or read back from storage — on a path that previously allocated nothing.

It reads as written that way because `check(self)` consumes `self`, so the fields aren't available in the closure. Copying the two `Shard`s out first keeps the message lazy and drops the helper:

```rust
let unchecked = UncheckedShardGroup::decode(d, ctx)?;
let (start, end_inclusive) = (unchecked.start, unchecked.end_inclusive);
ShardGroup::new_checked(start, end_inclusive).ok_or_else(|| {
    decode::Error::message(format!(
        "invalid ShardGroup: start ({}) is greater than end_inclusive ({})",
        start.as_u32(),
        end_inclusive.as_u32()
    ))
    .at(pos)
})
```

## Nit

`shard_group.rs:109` and `:124` — with the invariant now enforced at every deserialization boundary, `is_empty` (hardcoded `false`) and `len` (which underflows and, with `overflow-checks = true`, panics) are the two remaining methods that still take the ordering on faith, and `new_unchecked` is the one way to reach them in a state they can't describe — `shard_iter` no longer is. Not for this PR, but the commit message's "uphold the invariant" framing invites the question of why those two keep their `WARN:` doc comments instead of the same treatment.

The doc comment at `fee_pool.rs:15-16` and the `InvalidFeePoolShard` message at `fee_pool.rs:40` both read correctly now — `Shard 0 cannot hold a validator fee pool: expected a shard in 1..=256`.

Nothing here blocks. CI is still running on this head (`clippy`, `test 1-3/3`, `integration test`, `ledger app`); I'll re-check once it settles.

sdbondi (commented):
Re-reviewed at dc6086a7. The perf finding from the last round is resolved, and the two commits after it are both sound. Nothing outstanding.

**94fd076 (lazy error message)** — `format!` now runs inside `ok_or_else` on both impls, so the success path allocates nothing. Destructuring `UncheckedShardGroup` into two `Copy` `Shard`s is what makes that possible without the borrow dance. The wire shape is untouched from d152c67, so the encoding-compatibility check from the previous round still stands.

**b9448a5 (rocksdb codec)** — this was the last decode boundary still reaching a panicking constructor. `MalformedData { operation: &'static str, details: String }` matches `state_store_rocksdb/src/error.rs:47` and maps to `StorageError::DataInconsistency` (`:98`), which is the right classification: the encode side only ever writes `value.start()`/`value.end()` off an already-valid `ShardGroup`, so an out-of-order record means corruption, not a reachable input. The `format!` is inside `ok_or_else` here too.

**dc6086a7 (upper bound)** — the one that needed the most care, since tightening `new_checked` narrows every caller including the panicking `new`. Checked:

- No production site constructs a group with `end_inclusive > 256`. `all_shard_groups_iter` (`num_preshards.rs:58`) caps at `num_preshards`, `substate_address.rs:182` is bounded because `num_committees` is clamped to `num_shards` (`:151`) so `shards_per_committee >= 1`, and the `epoch_oracles` sites are the literal `(1, 256)`. The existing tests all sit at or under `Shard::max()`, including `decode_from_u32(MAX_ENCODED_VALUE) == Some(ShardGroup::new(0x100, 0x100))`.
- The totality claim in the new doc comment holds for both methods, not just `len`: with `start <= end_inclusive <= 256`, `end_inclusive + 1` cannot overflow and `encode_as_u32`'s `start << 16` cannot either. `start` is bounded transitively by the ordering check, which is worth noting because it isn't bounded explicitly.
- It closes a real p2p gap that the commit message undersells. `decode_from_u32` guards only on `MAX_ENCODED_VALUE = (256 << 16) + 256`, which admits any `n` whose low half is up to `0xFFFF` — so a peer sending `shard_group = 0x0000FFFF` previously produced a well-formed `ShardGroup { 0, 65535 }` at `p2p/src/conversions/consensus.rs:305,342,497,760,928` and `tari_validator_node/src/p2p/rpc/rpc_impl.rs:571`, with `len() == 65536` and a `shard_iter` to match. Those call sites already handled `None`, so the tightening needs no changes there and rejects the value at the conversion boundary instead.

## Remaining nit, unchanged from the last round

`new_unchecked` is now the only way to reach `len` (`shard_group.rs:109`) and `is_empty` (`:124`) in a state they cannot describe, so "keeps `len` and `encode_as_u32` total" (`:78`) is true of every checked-constructed value but not literally of the type. That is a smaller gap than before — `shard_iter` is total now, and the sole production caller, `foreign_proposal.rs:450 shard_group_unchecked`, is gated on `shard_group_checked` in the validation path — so it stays a nit rather than something to fix here.

The scope growth across these five commits is coherent: the fee-pool bound, then the invariant at the serde/CBOR boundary, then the two remaining constructors that decode untrusted bytes. Worth folding the `ShardGroup` half into the PR description, since the title still reads as fee-pool only and a reader landing on the commit list won't know the p2p and rocksdb paths are in here.

Waiting on CI (`clippy`, `test 1-3/3`, `integration test`, `ledger app`) before calling it done. Note I can't approve this one — GitHub blocks approving your own PR, so it needs another maintainer.

PR #2562, opened 2026-09-09
perf(engine): drop thread-safe primitives from the single-threaded execution path

sdbondi:
## Description

Engine execution is strictly single-threaded — `Runtime` already asserts as much with a manual `unsafe impl Send/Sync` over a raw pointer — yet several types on the execution path used thread-safe primitives that nothing ever exercised. This removes them.

### `AuthorizationScope::virtual_proofs` / `AuthParams::initial_ownership_proofs`: `Arc<IndexSet<_>>` → `IndexSet<_>`

The `Arc` was never shared. It is built once per execution in `TariTransactionProcessor::execute`, moved through `AuthParams` into `TransactionProcessor`, destructured out and handed to the single `AuthorizationScope::new` call site. Its strong count only ever rose through the one clone in `WorkingState::push_frame`.

Owning it is also a small net win: `AuthorizationScope::empty()` ran `Arc::new(IndexSet::new())` — one heap allocation — on every `CallScope::new()`, which happens on every call-frame push including every cross-template call. `IndexSet::new()` allocates nothing. The clone in `push_frame` becomes a deep copy of what is typically a single `NonFungibleAddress`, so it more than pays for itself.

Semantics are unchanged: `push_frame` still installs the virtual proofs only at depth 0, so signer badges remain confined to the top-level frame and are not visible to nested calls.

I considered hoisting the virtual proofs out of `AuthorizationScope` entirely (they are immutable for the whole execution and `WorkingState` already stores them once in `initial_call_scope`). I did not, because the depth-0 gate would have to be reproduced by hand at every lookup, and getting that wrong silently grants signer authority to nested cross-template calls. Not worth the risk for one avoided memcpy of ~34 bytes per instruction.

### `RuntimeInterfaceImpl::blobs`: `Arc` → `Rc`

Created inside `execute()` from owned data; one clone goes into `RuntimeInterfaceImpl`, the original stays local for `process_instructions`. Both live on one stack in one synchronous call. A borrow would be better still, but `Box<dyn RuntimeInterface>` has no lifetime parameter.

### The runtime pointer: `AtomicPtr` → `NonNull`, end to end

The comment claimed the `AtomicPtr` existed "to make `RuntimeInterfaceImpl` Send + Sync to satisfy wasmer trait bounds". That is not what is happening: `RuntimeInterface` has no `Send`/`Sync` supertrait, so `Box<dyn RuntimeInterface>` is not `Send` and cannot propagate the requirement, and wasmer's `FunctionEnv<WasmEnv<Runtime>>` bound is satisfied by `Runtime`'s own manual impls. `StateTracker<TStore>` is not unconditionally `Send` either, since `TStore` carries no such bound. Nothing needed the atomic.

The pointer is also never null — it comes from `NonNull::from_mut` inside `Runtime::from_mut` — so the nullability was purely an artifact of round-tripping through `*mut` between `Runtime::as_pointer` and `set_runtime_pointer`. `RuntimeInterface::set_runtime_pointer` now takes a `NonNull`, and `Runtime::from_pointer`/`as_pointer` become `from_non_null`/`as_non_null`, which removes an unreachable null check in `get_call_runtime`.

### `PushCallFrame` no longer derives `Clone`

It is never cloned. `CallScope` and `CallFrame` keep theirs — `WorkingState::clone` needs them for the fee checkpoint.

## Motivation and Context

Removing synchronisation that the execution model does not need: fewer atomics on the hot path, one less allocation per call frame, and the type signatures now say what is actually true about the engine's threading.

## How Has This Been Tested?

`cargo lints clippy --all-targets` clean on `tari_engine`, `tari_template_test_tooling`, `tari_ootle_app_utilities`, `tari_validator_node` and `tari_ootle_walletd`. Behaviour is unchanged, so the engine and template test suites in CI are the real check.

Confirm the `Arc` removals are safe by checking the two claims the change rests on:

1. `AuthParams.initial_ownership_proofs` has exactly two construction sites (`transaction_executor.rs`, `template_test.rs`), both feeding a value that is moved straight through to the one `AuthorizationScope::new` call — no other holder of the old `Arc`.
2. Nothing requires `RuntimeInterfaceImpl: Send`. `grep` for a `Send` supertrait on `RuntimeInterface` (there is none) and note that the wasmer env holds `Runtime`, not `RuntimeInterfaceImpl`.

Then confirm the depth-0 gate in `WorkingState::push_frame` is untouched, which is what keeps virtual proofs out of nested frames.

## Breaking Changes

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify

`tari_engine`'s public API does change — `AuthorizationScope::new`, `AuthParams::initial_ownership_proofs`, `RuntimeInterfaceImpl::initialize`'s `blobs` parameter, and `RuntimeInterface::set_runtime_pointer` / `Runtime::from_pointer` / `Runtime::as_pointer` — but the in-tree 0.41.0 is an unreleased minor ahead of the published 0.40.0, so the pending release already carries it and no version bump is needed.

https://claude.ai/code/session_018N1AkKuPAJ1cfcHXch9NAX

### DIFF
--- a/applications/tari_ootle_app_utilities/src/transaction_executor.rs
+++ b/applications/tari_ootle_app_utilities/src/transaction_executor.rs
@@ -142,7 +142,7 @@ where TTemplateProvider: TemplateProvider<Template = LoadedTemplate>
             .map(|pk| NonFungibleAddress::from_public_key(*pk))
             .collect();
         let auth_params = AuthParams {
-            initial_ownership_proofs: Arc::new(initial_ownership_proofs),
+            initial_ownership_proofs,
         };
 
         let processor = TransactionProcessor::new(
--- a/crates/engine/src/runtime/auth.rs
+++ b/crates/engine/src/runtime/auth.rs
@@ -1,7 +1,7 @@
 //   Copyright 2023 The Tari Project
 //   SPDX-License-Identifier: BSD-3-Clause
 
-use std::{fmt::Display, sync::Arc};
+use std::fmt::Display;
 
 use indexmap::IndexSet;
 use tari_template_lib::{
@@ -11,21 +11,21 @@ use tari_template_lib::{
 
 #[derive(Debug, Clone, Default)]
 pub struct AuthParams {
-    pub initial_ownership_proofs: Arc<IndexSet<NonFungibleAddress>>,
+    pub initial_ownership_proofs: IndexSet<NonFungibleAddress>,
 }
 
 #[derive(Debug, Clone)]
 pub struct AuthorizationScope {
     /// Virtual proofs are system-issued non-fungibles that exist for no longer than the execution e.g. derived from
     /// the transaction signer public key
-    virtual_proofs: Arc<IndexSet<NonFungibleAddress>>,
+    virtual_proofs: IndexSet<NonFungibleAddress>,
 
     /// Resource-based proofs
     proofs: IndexSet<ProofId>,
 }
 
 impl AuthorizationScope {
-    pub fn new(virtual_proofs: Arc<IndexSet<NonFungibleAddress>>) -> Self {
+    pub fn new(virtual_proofs: IndexSet<NonFungibleAddress>) -> Self {
         Self {
             virtual_proofs,
             proofs: IndexSet::new(),
@@ -34,7 +34,7 @@ impl AuthorizationScope {
 
     pub fn empty() -> Self {
         Self {
-            virtual_proofs: Arc::new(IndexSet::new()),
+            virtual_proofs: IndexSet::new(),
             proofs: IndexSet::new(),
         }
     }
@@ -77,7 +77,7 @@ impl AuthorizationScope {
 impl Display for AuthorizationScope {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(f, "Virtual: [")?;
-        for proof in self.virtual_proofs.iter() {
+        for proof in &self.virtual_proofs {
             write!(f, "{}", proof)?;
         }
         write!(f, "], Proofs: [")?;
--- a/crates/engine/src/runtime/impl.rs
+++ b/crates/engine/src/runtime/impl.rs
@@ -20,7 +20,7 @@
 //   WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 //   USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-use std::sync::{Arc, atomic, atomic::AtomicPtr};
+use std::{ptr::NonNull, rc::Rc, sync::Arc};
 
 use log::{warn, *};
 use tari_bor::{MaybeTagged, decode_exact};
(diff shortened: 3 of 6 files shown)


sdbondi (commented):
Reviewed at `fc36cd2`. No blocking findings — the change is mechanical and the two claims it rests on hold up.

## Verified

- **`initial_ownership_proofs` has no other holder.** Two construction sites (`applications/tari_ootle_app_utilities/src/transaction_executor.rs:145`, `crates/template_test_tooling/src/template_test.rs:785`), both moved into `TransactionProcessor::new` and destructured in `execute` straight into the single `AuthorizationScope::new` at `crates/engine/src/transaction/processor.rs:166`. Nothing else names the field, and nothing ever mutated `virtual_proofs` (no `&mut` accessor on `AuthorizationScope`), so the old `Arc` sharing carried no observable aliasing — the deep copy is semantically identical.
- **The depth-0 gate is untouched.** `crates/engine/src/runtime/working_state.rs:1547-1555` still installs the base auth scope only at `call_frame_depth() == 0`; nested frames keep `AuthorizationScope::empty()` from `CallScope::new`. `update_from_parent` (`scope.rs:199`) is still a no-op, and `update_from_child` only merges `proofs`, never `virtual_proofs` — so signer badges still cannot leak up or down.
- **Nothing requires `RuntimeInterfaceImpl: Send`.** `trait RuntimeInterface` at `crates/engine/src/runtime/mod.rs:118` has no `Send`/`Sync` supertrait; the wasmer env holds `Runtime` (`mod.rs:302`, with its own `unsafe impl Send + Sync`), not `RuntimeInterfaceImpl`. `RuntimeInterfaceImpl` was already non-`Send` in practice — `StateTracker<TStore>`, `Arc<TTemplateProvider>` and `ModulesCollection<TStore>` carry no `Send` bound on their parameters — so the `Rc`/`NonNull` switch does not remove a property anything relied on. The whole `Rc` lifecycle is contained in the synchronous `TransactionProcessor::execute` frame; no thread or task boundary anywhere between `processor.rs:230` and the end of `execute`.
- **`Option<NonNull<_>>` matches `Runtime` itself**, which already stores `NonNull<Box<dyn RuntimeInterface>>` — the field type is now consistent with its only consumer, and `Option<NonNull<T>>` is niche-packed so the field is 8 bytes rather than the old `Option<AtomicPtr>`'s 16.
- **`PushCallFrame` is never cloned** — all four construction sites (`processor.rs:539,549,784,842,925`) pass by value through `push_call_frame` into `into_new_call_frame`.
- **Versioning claim checks out.** `scripts/crate_versioning.py impact tari_engine --breaking` reports "No version changes required" — 0.41.0 is pending against 0.40.0 on crates.io, so the whole tier-3 cohort is already covered.
- **No stray files** in the diff; imports that became unused were dropped and the remaining `Arc` uses in both touched applications are still live.

## Nits

1. `crates/engine/src/runtime/impl.rs:478-479` — the null branch is now dead. `ptr` is a `NonNull`, so `Runtime::from_pointer(ptr.as_ptr())` can never return `None` and `.expect("Runtime pointer is null")` is unreachable. Either add a `Runtime::from_non_null(NonNull<_>) -> Runtime` next to `from_pointer` (`mod.rs:320`) and drop the `expect`, or leave a note that the second unwrap is structurally infallible. Not worth blocking on, but a reader will pause at two consecutive `expect`s where only one is reachable.

2. `crates/engine/src/runtime/impl.rs:4181` — `NonNull::new(pointer)` silently folds a null pointer into `None`, so a null would surface later as `"BUG: Runtime pointer not set"` at the call site rather than `"Runtime pointer is null"` at the setter. Diagnostics only (no caller passes null; the only call is `runtime.as_pointer()` at `processor.rs:243`, derived from `NonNull::from_mut`), but `self.runtime_pointer = Some(NonNull::new(pointer).expect("BUG: null runtime pointer"))` would fail at the site that actually has the bug.

3. The "Breaking Changes" section lists `AuthorizationScope::new` and `AuthParams::initial_ownership_proofs` but not `RuntimeInterfaceImpl::initialize`, whose `blobs` parameter went `Arc` → `Rc`. It is re-exported (`runtime/mod.rs:27`) so it is part of the public surface. Same conclusion applies — covered by the pending 0.41.0 — just worth listing for completeness.

## On the perf reasoning

The net-win argument holds, and is worth stating precisely for anyone reading this later: `AuthorizationScope::empty()` allocated once per `CallScope::new()`, i.e. per frame push, including every nested cross-template call; that allocation is now gone entirely. The new deep copy is paid only at `push_frame` depth 0 — once per top-level instruction — and costs ~2 allocations (`IndexMap` keeps entries and indices separately) for a one-signer transaction. So nested frames are strictly better and depth 0 is roughly a wash. The counterweight not mentioned in the description is that `AuthorizationScope` grows from 8 to ~64 bytes for that field, which widens `CallScope`, `CallFrame` and every `Vec<CallFrame>` push and `WorkingState::clone` (`tracker.rs:670`, once per transaction). All small; the direction of the trade is right.

Agreed on not hoisting `virtual_proofs` out of `AuthorizationScope`. Reproducing the depth gate at each of the five `&AuthorizationScope` consumers in `tracker_auth.rs` would trade a memcpy for a silent-authority-escalation footgun.

Behaviour is unchanged, so CI's engine and template suites are the real check. `test (1/3)` and `integration test` were still pending when I looked; everything else is green.

*(Comment rather than an approving review — GitHub does not accept approvals on your own PR.)*

PR #2571, opened 2026-09-10
docs: add concepts section and optimize latest-release lookup

eccstartup:
Description
---

Adds a new **Concepts** section to the developer docs covering the protocol behind the playground —
architecture, consensus, state and execution, privacy, and tokenomics — and moves it ahead of the
Guides in the sidebar. Every factual claim cites the source path it comes from.

It also reworks `LatestReleaseTable.astro`, which backed the docs' "latest release" widget. It
previously fetched the full `/releases` list on every page render and filtered drafts client-side;
it now requests `?per_page=1` for the newest release, and only falls back to paging the list when
that single result is a draft (which only an authenticated caller can see), so an unauthenticated
build never renders an empty table.

Motivation and Context
---

The docs covered *how to write a template* but not *what the machine underneath is doing*. Readers
coming from the Guides hit terms — committee, shard group, three-chain rule, Pedersen commitment,
pledge, exhaust burn rate — with nowhere to look them up. Worse, older design vocabulary (Cerberus,
"braided" cross-shard ordering) was still circulating without ever being reconciled against the
implementation.

This PR adds that layer and makes the code the single source of truth for it. RFCs and TIPs predate
the current implementation and are **not** treated as authoritative — where they disagree with the
code, the code wins. Claims that could not be verified against a path in this repository were either
dropped or explicitly marked as non-official.

**What's included**

- **Ten new pages** under `src/content/docs/concepts/`: Overview, Architecture, Consensus,
  State and Execution, Privacy, Privacy in Applications, Stablecoins, Templates and Assets,
  Tokenomics, and Glossary.
- **Stablecoins** is new ground: how to build an issuance-backed asset from the native primitives —
  which resource type to choose, how issuance and redemption work, and exactly where the privacy
  holds and where it leaks.
- **Sixteen hand-authored inline SVG diagrams**, matching the existing `stealth-resources` style —
  no new diagramming dependency. They cover the two-layer seam, shard/committee derivation, the view
  round trip, the three-chain commit rule, substate address layout, `Up`/`Down` versioning, the two
  privacy layers, the stealth Diffie-Hellman exchange, the balance proof, the resource-type design
  space, the condition tree, the burn-to-mint bridge, and the fee split.
- **KaTeX** for the arithmetic in the privacy and consensus pages (`remark-math` + `rehype-katex`).
- **Sidebar reordered** to Getting Started → Concepts → Guides → Reference; "Getting Started" moves
  out of Guides to the top level.
- `public/llms.txt` and the landing page updated for the new structure.

**Housekeeping**

- `XTR` → `TARI_TOKEN` in the *Play the Guessing Game* and *Testing Templates* guides. `XTR` still
  resolves as a deprecated alias; the docs now use the current identifiers.

How Has This Been Tested?
---

- `npm run build` completes without errors.
- Each new page was fetched from the dev server and asserted to return 200 with a non-empty body.
- Every inline SVG was verified to render as raw markup (not escaped) and to keep all text and
  shapes inside its `viewBox`.
- Every constant, type, and API cited in the prose and in the diagrams was grepped against the
  repository before being written — e.g. `DEFAULT_DEVNET_COMMITTEE_SIZE`,
  `committee_size_per_shard_group`, `all_shard_groups_iter`, `exhaust_burn_rate`,
  `STEALTH_TARI_RESOURCE_ADDRESS`, `ResourceBuilder::stealth()`, `burn_with_value_proofs`, `to_shard`.

---

From `docs/developer-docs`:

```
npm install
npm run dev
```

1. Confirm the sidebar order is Getting Started → Concepts → Guides → Reference, and that
   "Getting Started" is no longer inside Guides.
2. Visit each page under `/concepts/` and confirm it renders. Spot-check diagrams against the prose:
   `/concepts/architecture/` (shard group derivation), `/concepts/consensus/` (view round trip,
   three-chain rule), `/concepts/state-and-execution/` (address layout, `Up`/`Down`),
   `/concepts/privacy/` (stealth DH exchange, balance proof), `/concepts/stablecoin/`.
3. Check the math renders — `/concepts/privacy/` and `/concepts/consensus/` should show KaTeX, not
   raw `$$`.
4. Skim `crates/` alongside the prose. The inline source paths are the point of the section, so a
   claim that does not match the code is a bug worth reporting.
5. `npm run build` should complete cleanly. Run it with `astro dev` stopped — they share
   `node_modules/.vite`, and building while the dev server is running invalidates its SSR dep cache.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (commented):
Thanks for this — it's a big, genuinely useful addition, and the "cite the source path" discipline is the right call for a Concepts section. Most of what I spot-checked holds up: `NumPreshards`, the 3-committee split test, committee sizes 40/7, `pacemaker_block_time` 10s, the three-chain snippet in `block_store.rs`, the ten `SubstateId` variants, `SubstateAddress::LENGTH`, `to_shard`, `SubstateDiff`, the BP+ aggregation factors 1–16, `EncryptedData`'s 8/32/0–255 layout, `exhaust_burn_rate` at 500 bps, the claim/tombstone path, and the genesis TARI resource's `deny_all` burn plus disabled supply tracking. All ~30 internal `basePath` links resolve.

The findings below are the places where the code and the prose disagree. Since the whole premise of the section is that the code wins, I've flagged those first.

I haven't run `npm run build` or `npm run dev` — I only checked the sources the prose cites.

## Blocking — factually wrong

**1. `architecture.mdx` — "How a validator lands in a shard group" documents the dev-only oracle, and the "no VRF, no sortition" Aside is wrong for real networks.**

`calculate_shard_key` lives in `crates/epoch_oracles/src/configured/config.rs` — the config-file oracle used for local swarms. On the base-layer path the shard key arrives from L1 (`crates/epoch_manager/src/epoch_event_oracle/event.rs:120`, `add.shard_key`), where Minotari computes it as `generate_shard_key(public_key, block_hash)` — a Blake2b over the VN key **and a base-layer block hash** — and re-rolls it periodically via `does_require_new_shard_key(pk, epoch, interval)`. So it is not "a function of its own registration … not of a network-wide randomness beacon"; base-layer entropy is exactly what it mixes in. As written the Aside would mislead anyone reasoning about committee stacking.

**2. `architecture.mdx` — the P2P section inverts the topic design.**

The page says topics are "prefixed per shard group … so a validator can subscribe to the exact slice of the network it is responsible for", and that "consensus traffic is confined to a committee, while transaction gossip is fanned out to whichever committees hold the inputs". Both topics are deliberately single and network-wide, and the source comments say why:

- `crates/p2p/src/gossip.rs:48` — "All transactions are gossiped on a single network-wide topic. Using one topic (rather than a topic per shard group) keeps the gossipsub mesh stable across epoch boundaries…"
- `applications/tari_validator_node/src/p2p/services/consensus_gossip/service.rs:36` — the same wording for `consensus`.

The prefix routing in `networking/core/src/message_mode.rs:403` is a router fallback, not a subscription strategy.

**3. `stablecoin.mdx` — `rule!(require(mint_badge))` does not compile.**

`__rule_requirement!` (`crates/template_lib_types/src/access_rules.rs:840`) has arms for `resource`, `non_fungible`, `public_key`, `component`, `template`, `caller_component`, `direct_caller_template` and `any_caller_*` — there is no `require`. For a badge it should be `rule!(resource(mint_badge))`. It's the page's headline snippet.

**4. `templates-and-assets.mdx` — `ResourceBuilder::fungible()` doesn't exist.**

The builder is `public_fungible()` (`crates/template_lib/src/resource/builder/mod.rs:47`). Every other page in these docs already uses `public_fungible`.

**5. `consensus.mdx` — `NewViewMessage` has no `high_tc`.**

The fields are `high_pc`, `last_vote: Option<ProposalVote>` and `timeout: TimeoutVote` (`crates/consensus/src/messages/new_view.rs:11`). "carrying their `high_pc` … and their `high_tc` (the highest timeout certificate)" describes a field that isn't there.

**6. `consensus.mdx` — the cross-shard stage table stops one stage short.**

`LocalAccept` is described as "All foreign pledges are in; execute and commit", but `Command` (`crates/storage/src/consensus_models/command.rs:90`) continues to `AllAccept` / `SomeAccept`, which are where the committee agrees the transaction accepted or aborted everywhere. As written a reader takes `LocalAccept` to be terminal.

## Should fix

**7. `architecture.mdx` — remainder distribution is stated two ways.** The elided snippet comment correctly says the remainder "spreads … across the leading groups", then the worked example and the Summary both say "the remainder goes to the **first** group". `all_shard_groups_iter` gives one extra shard to each of the first `remainder` groups (256/6 → four groups of 43, two of 42). It only looks like "the first group" because the cited 3-committee test has remainder 1.

**8. `package.json` — top-level `katex@^0.18.7` doesn't match the renderer.** The lock resolves `rehype-katex@7.0.1` against its own nested `katex@0.16.47`, so the markup is generated by 0.16 while `customCss: ["katex/dist/katex.min.css"]` ships 0.18's stylesheet (`@types/katex` is also 0.16.8). Worth pinning `katex` to the `^0.16` line that `rehype-katex` resolves, so CSS and markup share a major.

**9. `package.json` — `@astrojs/markdown-remark` is added as a direct dependency but never imported.** It also duplicates: 7.3.1 top-level against `@astrojs/mdx`'s 7.2.4. Probably droppable.

**10. `privacy.mdx` — "SHA-256 appears only as an interoperability option for hashlocks".** Templates can call it directly: `EngineArgs::HASH_SHA256 = 0x0021`, alongside `HASH_SHA512` and `HASH_KECCAK256` (`crates/template_lib_types/src/engine_args.rs:91-94`).

**11. `stablecoin.mdx` — `[minimum_value_proof, 2^64)`.** The field is `minimum_value_promise` (e.g. `crates/wallet/ootle-rs/src/stealth/spec.rs:182`). Also `2^64` should be `2^64 − 1`, here and in privacy.mdx's "bounds the largest single value at $2^{64}$".

**12. `state-and-execution.mdx` — `WasmMeteringRate` is cited as `crates/engine/src/transaction/`.** It's defined at `crates/engine/src/fees/fee_table.rs:205`.

**13. `overview.mdx` / `glossary.mdx` — "liveness borrows from the base layer" / proof-of-work as "a liveness backstop".** Nothing in the code supports this. The base layer supplies the epoch clock and the membership register; a stalled committee is unstuck by the pacemaker plus the next epoch's re-formed membership. The overview's own Aside says this correctly two paragraphs later, so the "backstop" framing contradicts it.

## Nits

- `architecture.mdx`: "The number of **transactions** a transaction may remain valid for is bounded … by `max_transaction_validity_epochs`" — should be *epochs*.
- `privacy.mdx`: `stealth_owner_dh` is private (`fn`, not `pub fn`), and the derivation also domain-separates on `Network`; the table presents it as a callable public API.
- `glossary.mdx`: "Balance proof" sorts after "Base node".
- `LatestReleaseTable.astro`: the comment says the fallback "pag[es] the list", but it fetches one default page (30) and doesn't follow `Link` headers — either page it or reword to "fetching the default page". The `?per_page=1` optimisation itself is sound and behaviour-preserving for unauthenticated builds.
- Branch hygiene: the PR head is the fork's `development` branch and carries an upstream merge commit, which will make follow-up syncs awkward for you.

sdbondi (commented):
_Posted by Claude._

Thanks for turning these around so quickly — all thirteen points from the last pass are addressed, and I checked the two dependency ones against the lockfile rather than just the manifest: `katex` now resolves to a single `0.16.47` with no nested duplicate under `rehype-katex`, and the top-level `@astrojs/markdown-remark` is gone. The remainder, `NewViewMessage`, `AllAccept`/`SomeAccept`, `public_fungible()`, `rule!(resource(...))`, `minimum_value_promise`, `fee_table.rs`, the epoch-clock rewrite and the glossary sort are all correct now.

Two things the rewrites introduced, and one broader note.

## 1. `architecture.mdx` — the six-committee test doesn't exist

The remainder explanation is right now, but it's attributed to a test that isn't there:

> The code's own tests pin this both ways: three committees split 256 into 86 + 85 + 85 (remainder 1), and six committees into 43 + 43 + 43 + 43 + 42 + 42 (remainder 4).

`crates/common_types/src/num_preshards.rs` has exactly two tests: `it_calculates_all_shard_groups` (`num_committees = 3`) and `total_shard_group_lengths_equal_num_preshards` (`num_committees = 234`, which only asserts the lengths sum to 256 — it never checks the per-group split). The six-committee arithmetic is correct; the citation isn't. Either present it as worked arithmetic rather than as a test, or add the assertion to `it_calculates_all_shard_groups` — it'd be a good test to have either way.

## 2. `architecture.mdx` — propagation is still described as committee-scoped

> The topics being network-wide does **not** mean the traffic reaches everyone: consensus gossip is handled only by committee members, and a transaction is propagated only to the committees that hold its inputs.

This is the previous claim in softer wording. The source says the opposite in two places:

- `applications/tari_validator_node/src/p2p/services/mempool/gossip.rs:97` — "Since transactions are published on a single network-wide topic, gossipsub delivers the transaction to every validator with one publish; replicas in the involved shard groups pick it up directly and no per-shard-group forwarding is required."
- `.../mempool/service.rs:314` — "a single publish reaches every validator (including all involved shard groups)".

Every validator receives the transaction. `is_involved` at `service.rs:298` decides who acts on it — uninvolved nodes log "🙇 Not in committee for transaction" and drop it. Propagation is network-wide; processing is committee-scoped, and the sentence describes propagation.

Worth saying *why*, because the single topic is a deliberate design choice and the page is a good place to record it. The source comment gives the mesh-stability reason, but understates it: a per-shard-group topic doesn't just mean a resubscribe, it effectively clears the mesh every time the network reshards, and connectivity is degraded until each new topic's mesh re-forms. On top of that, one network-wide mesh keeps the whole validator set well connected rather than splitting it into per-group islands — groups have to communicate across shard boundaries anyway — and indexers subscribe to the same transaction topic (`applications/tari_indexer/src/transaction_gossip/service.rs:103`), so a per-shard-group topic would mean an indexer subscribing to all 256 or missing transactions. Something like:

> Both topics are single and network-wide. Scoping them per shard group would tear the gossipsub mesh down every time the network reshards: validators unsubscribe and resubscribe at the epoch boundary, and each new topic's mesh has to be rebuilt from scratch — leaving the network poorly connected until it re-forms, exactly when a fresh epoch's traffic starts flowing. One network-wide topic survives resharding untouched, and it keeps the whole validator set in a single well-connected mesh — which matters because shard groups have to communicate across shard boundaries anyway. It also means indexers, which subscribe to the same transaction topic, see every transaction without tracking shard assignments.
>
> Network-wide propagation is not the same as network-wide *work*: every validator receives a gossiped transaction, and each one checks whether it holds any of the inputs (`is_involved`, `mempool/service.rs:298`) before passing it to consensus. Nodes outside the involved committees drop it.

Small related nit: the Summary still ends "libp2p with **topic-scoped** gossip", which now reads against the section's own "single and network-wide".

## 3. Some of the rewrites narrate the revision rather than describe the system

A few of the new sentences read as replies to review feedback. A reader who never saw the review doesn't need the contrast, and it dates the page:

- `privacy.mdx` — "`stealth_owner_dh` itself is a private helper — callers go through the address/secret functions above rather than the shared-secret step directly", plus "(private helper)" in the table. The `Network` domain-separation sentence beside it is the part worth keeping; that an internal helper is internal isn't something the page needs to argue.
- `architecture.mdx` — "That is the code you will hit when you run a swarm locally, and it is *not* how a live network assigns committees."
- `architecture.mdx` — "Both definitions carry that reasoning in a comment, which is worth reading before you consider scoping these topics by shard." This is addressed to someone thinking of changing the code, not to a reader of the docs.
- `architecture.mdx` — "The topics being network-wide does **not** mean the traffic reaches everyone … but that is a routing fallback, not a per-shard-group subscription strategy."
- `consensus.mdx` — "It is the pair of those last two, not `LocalAccept`, that makes the shared decision final in every group."
- `overview.mdx` — "and liveness is not borrowed from the base layer either." The "either" carries an argument the reader never heard.

Each of these states the right fact somewhere nearby — it's the contrast clause doing the work of a review reply. Stating the behaviour positively covers it.

To be clear, this doesn't apply to the Cerberus and "braided" Asides. Reconciling vocabulary that's still circulating is a stated goal of the section and belongs on the page; those are about the reader's likely priors, not about what this PR changed.

sdbondi (approved):
Thanks!

Issue #2568, opened 2026-09-09
Remove epoch_end_spread_blocks: dead since #2164's next-epoch hash ratification


sdbondi:
`epoch_end_spread_blocks` has been unable to affect a vote since #2164 introduced next-epoch hash ratification. It should be removed, or the liveness gap it was meant to cover should be closed where it actually lives.

## The two gates

Voting for an `EndEpoch` block requires both:

1. `can_propose_epoch_end` — `crates/consensus/src/hotstuff/on_receive_local_proposal.rs:328`

```rust
let can_propose_epoch_end =
    em_epoch > current_epoch || self.epoch_manager.is_within_epoch_end_spread(current_epoch).await?;
```

2. Hash ratification — `crates/consensus/src/hotstuff/on_ready_to_vote_on_local_block.rs:444`, where `expected_next_epoch_hash` is `get_epoch_hash(current_epoch + 1)`. `None` is a terminal no-vote (`NoVoteReason::EndOfEpochHashNotObserved`).

## Why the spread is dead

`get_epoch_hash(E + 1)` reads the `epochs` row for `E + 1`. That row is written by `EpochManager::insert_current_epoch` (`crates/epoch_manager/src/service/epoch_manager.rs:200-212`), which in the same call sets `MetadataKey::EpochManagerCurrentEpoch` and `set_current_epoch(epoch)`. The only caller is `activate_epoch`, driven by `EpochEvent::EpochChanged` from the oracle.

So gate 2 holds **iff** `em_epoch > current_epoch` — precisely gate 1's first disjunct. The spread disjunct can never turn a no-vote into a vote. A node whose scanner has not crossed the boundary no-votes at any spread value, including values larger than the epoch itself.

The history matches: leeway voting landed in #2051, ratification in #2164, and #2164 subsumed it without anyone noticing.

## Secondary symptom

`is_within_epoch_end_spread` tests `last_scanned_height + spread >= (current_epoch + 1) * epoch_length` (`crates/epoch_oracles/src/base_layer/oracle.rs:891-910`). Since `last_scanned_height >= current_epoch * epoch_length` always holds, any `spread >= epoch_length` makes it unconditionally true. Mainnet's `10` becomes exactly that once the L1 shortens `vn_epoch_length` from 60 to 10; TESTNET's `5` against the 10-block epochs on igor/nextnet/stagenet already covers the back half of every epoch. This is a symptom — retuning the value fixes nothing while gate 2 dominates.

## Proposed change

[...]

PR #2572, opened 2026-09-10
fix(consensus): ratify an EndEpoch hash from an observed boundary, not an activated epoch

sdbondi:
## Problem

Voting for an `EndEpoch` block needs the voter's own view of the next epoch's boundary hash (`expected_next_epoch_hash`). It was read with `get_epoch_hash(current_epoch + 1)`, which reads the `epochs` table. That row is written by `EpochManager::insert_current_epoch`, and the same call does `set_current_epoch(epoch)` — so "I know epoch E+1's hash" was exactly "I have activated epoch E+1".

The scanner stores the boundary header before the resulting `EpochChanged` event is applied. During a catch-up that event can sit behind several epoch activations, each running `assign_validators_for_epoch` and busting the committee cache. A node in that window has genuinely scanned the boundary block and can check the proposed hash against it, but no-voted with `NoVoteReason::EndOfEpochHashNotObserved`.

## Change

- `EpochEventOracle::observed_epoch_boundary_hash(epoch)` — new trait method, default `None`. The base layer oracle answers it from the header store it already keeps, via the existing `get_first_block_header_in_epoch`. A stored header only counts as the boundary when its height is exactly `epoch * epoch_length`, so a scan range opening mid-epoch does not ratify against the wrong hash (the same guard `reset_last_epoch_hash_to_boundary` already applies).
- The store handle and the epoch length live on `BaseLayerOracle`, not in `inner`. `poll_next_event` moves `inner` into the scan task for the whole of each batch, so a lookup that went through `inner` would answer `None` for the duration of every gRPC fetch — the exact catch-up window this PR targets. The epoch length is a shared `AtomicU64`; zero is the pre-constants state the old `Option` spelled `None`.
- `EpochManagerReader::get_observed_epoch_hash(epoch)` — the activated hash when the epoch has been activated (it is the one the self-healing correction maintains and `lock_epoch` freezes), otherwise the observed boundary. Precedence lives in the epoch manager, where both sources are, so consensus makes one call.
- `can_propose_epoch_end` accepts the observed boundary too. Without this the fix is inert: a node in the window would clear ratification and then no-vote at `NotEndOfEpoch`.
- A store failure is an error, not `None` — folding it in made an unreadable store look like a node that had not scanned yet, so the epoch stalled with nothing pointing at the store. It surfaces as `EpochManagerError::EpochEventOracleError`, which is how the activated-hash path beside it already behaves. An epoch whose boundary height overflows reports `EpochHeightOverflow` for the same reason: epochs are `height / epoch_length`, so that is a corrupt epoch, not a large chain.

No change to what a node will ratify: it still only votes for a hash it has seen itself, never one taken on the leader's word.

## Scope: this helps the committee, not the lagging node's own recovery

A node that votes from an observed boundary commits the EOE while `get_epoch_hash(next_epoch)` is still `None`, so `process_end_of_epoch` defers rather than stamping the next genesis and calling `lock_epoch`. `try_resume_pending_end_of_epoch` fires from `EpochManagerEvent::EpochChanged`, published by `on_scanning_complete` on `EpochEvent::DoneForNow` — emitted only once the scan reaches the lagged tip. So on a multi-batch catch-up the node votes and commits promptly but does not open the next epoch until the whole catch-up finishes, even though the activation it is waiting for lands moments later in the queue.

That is the intended scope: the committee gets its quorum instead of stalling on a member that could verify the boundary all along. Accelerating the lagging node's own transition is a separate change to the resume trigger.

Deferring `lock_epoch(next_epoch)` to the resume also widens the window in which `activate_epoch`'s self-healing correction could store a hash other than the one the committee committed. It needs a reorg landing between the queued activation and the resume, so I do not think it is reachable, but it is wider than it was.

## Note for #2568

The `EpochOracle` enum dispatcher in `crates/epoch_oracles/src/lib.rs` does not forward `is_within_epoch_end_spread`, so every variant falls through to the default `false`. That is a second, independent reason `epoch_end_spread_blocks` cannot affect a vote today — the knob never reaches the base layer oracle at all. Deliberately not fixed here, since #2568 proposes removing it; the new method is forwarded through the enum.

## Test plan

- [x] `cargo test -p tari_epoch_oracles --lib --features base_layer` — the boundary is readable with `inner` taken away (the scan-in-flight regression), a first-stored-header that is not the boundary height yields `None`, and so does the pre-constants state
- [x] `cargo test -p consensus_tests epoch_change` — 7 passed, including the new `epoch_change_observed_boundary_is_voted_before_activation`, which fails when the `can_propose_epoch_end` term is removed
- [ ] CI

https://claude.ai/code/session_01HHLL5F3fKuzsMcdJdquy6R

### DIFF
--- a/crates/epoch_oracles/src/base_layer/oracle.rs
+++ b/crates/epoch_oracles/src/base_layer/oracle.rs
@@ -898,10 +898,6 @@ impl<TStore: EpochOracleStore + BaseLayerBlockHeaderStore, TClient: BaseNodeClie
         }
     }
 
-    /// Returns the boundary block hash we have stored for `epoch`, or `None` if we have not scanned
-    /// it. The stored header only counts as the boundary when its height is exactly the epoch's first
-    /// height: a scan range that starts mid-epoch stores a first header that is not the boundary, and
-    /// ratifying against that would compare the wrong hash.
     /// Returns true when our lagged scanner position is within `epoch_end_spread_blocks` of the
     /// next epoch boundary. Used by consensus to accept `EndEpoch` proposals speculatively when
     /// peers' oracles have already crossed and ours is almost there.
--- a/crates/epoch_oracles/src/base_layer/oracle.rs
+++ b/crates/epoch_oracles/src/base_layer/oracle.rs
@@ -889,7 +889,7 @@ impl<TStore: EpochOracleStore + BaseLayerBlockHeaderStore, TClient: BaseNodeClie
     /// it. The stored header only counts as the boundary when its height is exactly the epoch's first
     /// height: a scan range that starts mid-epoch stores a first header that is not the boundary, and
     /// ratifying against that would compare the wrong hash.
-    fn observed_epoch_boundary_hash(&self, epoch: Epoch) -> anyhow::Result<Option<FixedHash>> {
+    fn observed_epoch_boundary_hash(&self, epoch: Epoch) -> Result<Option<FixedHash>, BaseLayerOracleError> {
         // The epoch length is only known once a scan has obtained the L1 constants; until then we
         // cannot say which height opens the epoch.
         let Some(epoch_length) = self.cached_epoch_length else {
@@ -898,7 +898,11 @@ impl<TStore: EpochOracleStore + BaseLayerBlockHeaderStore, TClient: BaseNodeClie
         let Some(boundary_height) = epoch.as_u64().checked_mul(epoch_length) else {
             return Ok(None);
         };
-        let Some(boundary) = self.store.get_first_block_header_in_epoch(epoch)? else {
+        let Some(boundary) = self
+            .store
+            .get_first_block_header_in_epoch(epoch)
+            .map_err(BaseLayerOracleError::StoreError)?
+        else {
             return Ok(None);
         };
         Ok((boundary.height == boundary_height).then_some(boundary.block_hash))
@@ -1043,7 +1047,7 @@ impl<TStore: EpochOracleStore + BaseLayerBlockHeaderStore + Send + 'static, TCli
         // `inner` is briefly None while a scan task is in flight; the voter then falls back to the
         // activated epoch hash alone.
         match self.inner.as_deref() {
-            Some(inner) => inner.observed_epoch_boundary_hash(epoch),
+            Some(inner) => Ok(inner.observed_epoch_boundary_hash(epoch)?),
             None => Ok(None),
         }
     }
--- a/crates/epoch_oracles/src/base_layer/oracle.rs
+++ b/crates/epoch_oracles/src/base_layer/oracle.rs
@@ -889,11 +889,19 @@ impl<TStore: EpochOracleStore + BaseLayerBlockHeaderStore, TClient: BaseNodeClie
     /// it. The stored header only counts as the boundary when its height is exactly the epoch's first
     /// height: a scan range that starts mid-epoch stores a first header that is not the boundary, and
     /// ratifying against that would compare the wrong hash.
-    fn observed_epoch_boundary_hash(&self, epoch: Epoch) -> Option<FixedHash> {
-        let epoch_length = self.cached_epoch_length?;
-        let boundary_height = epoch.as_u64().checked_mul(epoch_length)?;
-        let boundary = self.store.get_first_block_header_in_epoch(epoch).ok()??;
-        (boundary.height == boundary_height).then_some(boundary.block_hash)
+    fn observed_epoch_boundary_hash(&self, epoch: Epoch) -> anyhow::Result<Option<FixedHash>> {
+        // The epoch length is only known once a scan has obtained the L1 constants; until then we
+        // cannot say which height opens the epoch.
+        let Some(epoch_length) = self.cached_epoch_length else {
+            return Ok(None);
+        };
+        let Some(boundary_height) = epoch.as_u64().checked_mul(epoch_length) else {
+            return Ok(None);
+        };
+        let Some(boundary) = self.store.get_first_block_header_in_epoch(epoch)? else {
+            return Ok(None);
+        };
+        Ok((boundary.height == boundary_height).then_some(boundary.block_hash))
     }
 
     /// Returns true when our lagged scanner position is within `epoch_end_spread_blocks` of the
@@ -1031,10 +1039,13 @@ impl<TStore: EpochOracleStore + BaseLayerBlockHeaderStore + Send + 'static, TCli
             .unwrap_or(false)
     }
 
-    fn observed_epoch_boundary_hash(&self, epoch: Epoch) -> Option<FixedHash> {
+    fn observed_epoch_boundary_hash(&self, epoch: Epoch) -> anyhow::Result<Option<FixedHash>> {
         // `inner` is briefly None while a scan task is in flight; the voter then falls back to the
         // activated epoch hash alone.
-        self.inner.as_deref()?.observed_epoch_boundary_hash(epoch)
+        match self.inner.as_deref() {
+            Some(inner) => inner.observed_epoch_boundary_hash(epoch),
+            None => Ok(None),
+        }
     }
 }
 
@@ -1405,11 +1416,11 @@ mod tests {
 
         // Epoch 3 opens at height 15, which we have scanned.
         assert_eq!(
-            inner.observed_epoch_boundary_hash(Epoch(3)),
+            inner.observed_epoch_boundary_hash(Epoch(3)).unwrap(),
             Some(hash_header(NETWORK, &chain[15]))
         );
         // Epoch 4 opens at height 20, above our lagged scan position.
-        assert_eq!(inner.observed_epoch_boundary_hash(Epoch(4)), None);
+        assert_eq!(inner.observed_epoch_boundary_hash(Epoch(4)).unwrap(), None);
     }
 
     #[tokio::test]
@@ -1431,7 +1442,7 @@ mod tests {
             }])
             .unwrap();
 
-        assert_eq!(inner.observed_epoch_boundary_hash(Epoch(2)), None);
+        assert_eq!(inner.observed_epoch_boundary_hash(Epoch(2)).unwrap(), None);
     }
 
     #[tokio::test]
--- a/crates/epoch_oracles/src/base_layer/oracle.rs
+++ b/crates/epoch_oracles/src/base_layer/oracle.rs
@@ -5,6 +5,10 @@ use std::{
     collections::{HashMap, VecDeque},
     future::{Future, poll_fn},
     pin::Pin,
+    sync::{
+        Arc,
+        atomic::{AtomicU64, Ordering},
+    },
     task::{Context, Poll},
 };
 
@@ -81,6 +85,11 @@ type ScanTask<TStore, TClient> = Pin<Box<dyn Future<Output = TaskOutput<TStore, 
 #[allow(clippy::struct_excessive_bools)]
 pub struct BaseLayerOracle<TStore, TClient = GrpcBaseNodeClient> {
     inner: Option<Box<BaseLayerOracleInner<TStore, TClient>>>,
+    /// Handles for the epoch-boundary lookup, held here rather than in `inner` because `inner` is
+    /// owned by the scan task for the whole of each batch — the boundary must stay readable while a
+    /// catch-up is in flight, which is when a lagging voter needs it.
+    store: TStore,
+    epoch_length: Arc<AtomicU64>,
     is_initialized: bool,
     is_done: bool,
     has_more: bool,
@@ -105,16 +114,20 @@ struct BaseLayerOracleInner<TStore, TClient> {
     // instead of being held for the whole batch.
     header_buf: Vec<BlockHeaderModel>,
     network: Network,
-    /// Epoch length in base-layer blocks, cached from consensus constants on the first scan
-    /// that obtains them. `None` until the first scan completes.
-    cached_epoch_length: Option<u64>,
+    /// Epoch length in base-layer blocks, cached from consensus constants on the first scan that
+    /// obtains them. Shared with the owning `BaseLayerOracle`, which reads it while this inner state
+    /// is away in a scan task. Zero until the first scan completes.
+    epoch_length: Arc<AtomicU64>,
 }
 
-impl<TStore: EpochOracleStore + BaseLayerBlockHeaderStore + 'static, TClient: BaseNodeClient>
+impl<TStore: EpochOracleStore + BaseLayerBlockHeaderStore + Clone + 'static, TClient: BaseNodeClient>
     BaseLayerOracle<TStore, TClient>
 {
     pub fn new(store: TStore, base_node_client: TClient, config: BaseLayerEpochOracleConfig, network: Network) -> Self {
+        let epoch_length = Arc::new(AtomicU64::new(0));
         Self {
+            store: store.clone(),
+            epoch_length: epoch_length.clone(),
             inner: Some(Box::new(BaseLayerOracleInner {
                 config,
                 store,
@@ -128,7 +141,7 @@ impl<TStore: EpochOracleStore + BaseLayerBlockHeaderStore + 'static, TClient: Ba
                 pending_events: VecDeque::new(),
                 header_buf: Vec::new(),
                 network,
-                cached_epoch_length: None,
+                epoch_length,
             })),
             is_initialized: false,
             is_done: false,
@@ -221,7 +234,7 @@ impl<TStore: EpochOracleStore + BaseLayerBlockHeaderStore, TClient: BaseNodeClie
                         .base_node_client
                         .get_consensus_constants(tip.height_of_longest_chain)
                         .await?;
-                    self.cached_epoch_length = Some(constants.epoch_length());
+                    self.epoch_length.store(constants.epoch_length(), Ordering::Relaxed);
                     let lagged_height = tip.height_of_longest_chain.saturating_sub(self.config.height_lag);
                     let epoch = constants.height_to_epoch(lagged_height);
                     // If no progress has been made since restarting, we still need to tell the epoch manager that
@@ -463,7 +476,7 @@ impl<TStore: EpochOracleStore + BaseLayerBlockHeaderStore, TClient: BaseNodeClie
             .base_node_client
             .get_consensus_constants(tip.height_of_longest_chain)
             .await?;
-        self.cached_epoch_length = Some(constants.epoch_length());
+        self.epoch_length.store(constants.epoch_length(), Ordering::Relaxed);
 
         // Acquire and scan this batch's blocks. The validator node needs every header (e.g. for
         // burn-proof validation), so it streams and stores the full contiguous range. The indexer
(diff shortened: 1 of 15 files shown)


crates/epoch_oracles/src/base_layer/oracle.rs
```diff
@@ -885,6 +885,17 @@ impl<TStore: EpochOracleStore + BaseLayerBlockHeaderStore, TClient: BaseNodeClie
         }
     }
 
+    /// Returns the boundary block hash we have stored for `epoch`, or `None` if we have not scanned
+    /// it. The stored header only counts as the boundary when its height is exactly the epoch's first
+    /// height: a scan range that starts mid-epoch stores a first header that is not the boundary, and
+    /// ratifying against that would compare the wrong hash.
+    fn observed_epoch_boundary_hash(&self, epoch: Epoch) -> Option<FixedHash> {
```
sdbondi: This trait function should be fallible (db error returns None)

sdbondi: Agreed, fixed in d042c496d.

`observed_epoch_boundary_hash` now returns `anyhow::Result<Option<FixedHash>>`, and the epoch manager surfaces a store failure as `EpochManagerError::EpochEventOracleError` rather than folding it into `None`. The `Option` still means "no boundary observed"; the error means "could not determine", which is what the activated-hash path beside it already does via `.optional()?`.

Worth spelling out why it mattered: swallowing the error made a node with an unreadable header store abstain from every EndEpoch vote with `EndOfEpochHashNotObserved` — an epoch stall whose logs point at consensus rather than at the store.

sdbondi (commented):
Reviewed at c4bd0e7. No blocking findings.

I traced the claim chain rather than the hunks: the boundary-height formula (`epoch * epoch_length`) matches `reset_last_epoch_hash_to_boundary` (`base_layer/oracle.rs:390`), and the hash a stored boundary header carries is the same value `scan_full_headers` puts into `EpochEvent::EpochChanged` (`oracle.rs:611-630`), so the observed fallback and the activated row can only disagree across a reorg — which is exactly the case the "activated wins" precedence is chosen for. `get_first_block_header_in_epoch` is lowest-height-in-epoch in both the sqlite adapter (`storage_sqlite/.../backend_adapter.rs:942`) and the in-memory test store, so the height guard does what the doc comment says. Broadening `can_propose_epoch_end` adds nothing that was not already reachable via `em_epoch > current_epoch`: `insert_current_epoch` writes the `epochs` row and `set_current_epoch` in the same call, so an activated E+1 implies `em_epoch > E` — the new term only admits the observed-boundary case, and it is `None` on non-EOE blocks so nothing else moves. The two implementors of `EpochManagerReader` and all three of `EpochEventOracle` are updated, and `tari_common_types` is a non-optional dep of `tari_epoch_oracles`, so the `#[cfg(not(feature = "base_layer"))]` arm of the enum dispatcher still builds. `epoch_change_unobserved_next_hash_is_not_voted` still exercises the no-vote path, since `set_oracle_visible_epoch` caps `get_epoch_hash`, which the test `get_observed_epoch_hash` delegates to.

## Non-blocking

**1. The fallback is inert while a scan task is in flight — the window this PR targets.** `crates/epoch_oracles/src/base_layer/oracle.rs:1034` reads the header store through `self.inner`, which `poll_next_event` takes for the duration of each scan batch (`oracle.rs:1004`). During a catch-up, wall-clock is split between draining queued events (inner present) and fetching the next batch over gRPC (inner absent), so an EOE proposal arriving in the latter half gets `None` and the node no-votes exactly as before. The header rows are durable and the lookup needs nothing from `inner` but the store handle and `cached_epoch_length`; `GlobalDb` is `Clone`, so hoisting a `store: TStore` clone (and the epoch length) onto `BaseLayerOracle` would make this hold across the whole scan. The same limitation already applies to `is_within_epoch_end_spread`, but that one is dead for an unrelated reason, so this is the first place it costs anything.

**2. `.ok()?` turns a store error into "not observed".** `oracle.rs:895` — a transient sqlite error on the vote path produces a silent `None`, and the node then no-votes with `EndOfEpochHashNotObserved`, which in the log is indistinguishable from the legitimate "I haven't scanned it yet" case. `reset_last_epoch_hash_to_boundary` propagates via `.optional()?`; the trait signature here can't, but a `warn!` on the error arm would keep an operator from chasing a scanner lag that isn't there.

**3. The consensus half of the change has no consensus-level coverage.** `crates/consensus_tests/src/support/epoch_manager.rs:525` defines `get_observed_epoch_hash` as `get_epoch_hash(...).optional()`, so the harness cannot express "boundary observed, epoch not activated" — the exact state the new `can_propose_epoch_end` branch exists to serve. The two new tests cover the oracle accessor, not the behaviour change. A `set_observed_epoch_hash_override` on `TestEpochManager` would let a test cap `current_epoch()` at Epoch(1), leave the observed hash visible, and assert the validator now votes for the EOE.

**4. Two test comments are now false.** `crates/consensus_tests/src/epoch_change.rs:334-337` says capping `current_epoch()` "makes the vote-time check `em_epoch > current_epoch` fail for this validator — reproducing the no-vote path", and "ensures that *if* it ever did receive the EOE proposal it would no-vote". After this change `set_oracle_current_epoch_cap` no longer implies a no-vote: it does not touch `get_epoch_hash`, so `expected_next_epoch_hash` is `Some` and `can_propose_epoch_end` is true. Both tests still pass only because the validator is also network-offline. `epoch_change_multi_epoch_lag_escalates_on_future_qc` (:459) inherits the same assumption. Either add `set_oracle_visible_epoch(Epoch(1))` alongside the cap to keep the stated mechanism real, or reword the comments to say the offline state is what reproduces the wedge.

**5. The deferral path becomes the normal case, and its resume trigger is coarse.** A node that votes from an observed boundary commits the EOE while `get_epoch_hash(next_epoch)` is still `None`, so `process_end_of_epoch` defers (`on_receive_local_proposal.rs:538-557`) instead of stamping the genesis and calling `lock_epoch`. `try_resume_pending_end_of_epoch` fires from `on_epoch_manager_event` (`worker.rs:786`), and `EpochManagerEvent::EpochChanged` is published only from `on_scanning_complete` (`epoch_manager_service.rs:342`), which runs on `EpochEvent::DoneForNow` — emitted only once the scan reaches the lagged tip (`oracle.rs:524`). So on a multi-batch catch-up the node votes and commits promptly but cannot open the next epoch until the whole catch-up finishes, even though the activation it was waiting for lands moments later in the queue. That is still a clear improvement — the committee gets its quorum instead of stalling — but the lagging node's own recovery is not accelerated by this PR, which is worth saying out loud in the description since the problem statement is framed around that node. Secondary: deferring `lock_epoch(next_epoch)` to the resume widens the window in which `activate_epoch`'s self-healing correction can store a hash other than the one the committee committed and the genesis will be stamped with. It needs a reorg landing between the queued activation and the resume, so I don't think it is reachable in practice — but the window is wider than it was, and the two hashes diverging is an `InvalidEpochHash` wedge for that node (`on_message_validate.rs:327` validates against `epoch_state.epoch_hash`, sourced from the epoch manager).

## Nit

`observed_boundary_hash_rejects_a_first_header_that_is_not_the_boundary` sets `cached_epoch_length` by hand and never scans, so it verifies the guard against a hand-built store rather than against a scan range that actually opened mid-epoch (`config.start_height` above a boundary would produce the real shape). Fine as a unit test of the predicate; just not the end-to-end version of the case the doc comment describes.

The PR's note about `is_within_epoch_end_spread` not being forwarded through the `EpochOracle` enum checks out — `crates/epoch_oracles/src/lib.rs` has only `next_epoch_event` and the new method, so every variant falls through to the default `false`. Agreed it belongs with #2568 rather than here.

crates/epoch_oracles/src/base_layer/oracle.rs
```diff
@@ -885,6 +885,25 @@ impl<TStore: EpochOracleStore + BaseLayerBlockHeaderStore, TClient: BaseNodeClie
         }
     }
 
+    /// Returns the boundary block hash we have stored for `epoch`, or `None` if we have not scanned
+    /// it. The stored header only counts as the boundary when its height is exactly the epoch's first
+    /// height: a scan range that starts mid-epoch stores a first header that is not the boundary, and
+    /// ratifying against that would compare the wrong hash.
+    fn observed_epoch_boundary_hash(&self, epoch: Epoch) -> anyhow::Result<Option<FixedHash>> {
```
sdbondi: Why not use BaselayerOracleError?

sdbondi: Done for the inherent method, in 2ac61bf7c — it now returns `Result<Option<FixedHash>, BaseLayerOracleError>` and maps the store failure into `StoreError`, which is how `reset_last_epoch_hash_to_boundary` already reports the very same lookup.

The `EpochEventOracle` trait method stays `anyhow::Result`: `BaseLayerOracleError` lives in `tari_epoch_oracles` and the trait lives in `tari_epoch_manager`, and the dependency runs oracles -> epoch_manager, so the trait cannot name it. The alternatives are an associated error type on the trait for a single method, or having oracles construct the consumer’s `EpochManagerError`. `anyhow` at that boundary matches `BaseLayerBlockHeaderStore`, which returns `anyhow::Result` for the same reason. The outer `BaseLayerOracle` impl converts there, so the typed error survives everywhere it is meaningful.

sdbondi (commented):
Re-reviewed d042c49 against c4bd0e7. The store-error fix is correct, and propagating rather than logging is the better call than what I suggested — I checked the premise in the commit message and it holds: `bootstrap.rs:294` and `:307` hand the same `GlobalDb` clone to the oracle and the epoch manager, and `SqliteGlobalDbAdapter` is one `Arc<Mutex<SqliteConnection>>` (`storage_sqlite/.../backend_adapter.rs:92`), so the header-store read cannot lose a lock race with the scanner's writes. A failure there is a real DB fault, which is exactly what `get_epoch_hash` beside it already propagates.

Verified in the new code: `is_not_found_error`'s `_ => false` arm (`epoch_manager/src/error.rs:62`) keeps `EpochEventOracleError` from being converted back to `None` by an upstream `.optional()`, which was the way this could have quietly regressed; the `let ... else` rewrite preserves the previous `None` semantics for an unknown epoch length, an overflowing boundary height, and an absent header; `anyhow` is already a dependency of `tari_epoch_manager`; and all four implementors plus both enum arms carry the new signature.

Worth being explicit about the consequence, since it is heavier than a log line: `HotStuffError::EpochManagerError` is unclassified in `handle_hotstuff_error`, so it falls to the `_ => return Err(err)` arm (`worker.rs:1480`), back out through `on_unvalidated_message` (`worker.rs:493`), and `Running::on_enter` treats it as "HotStuff crashed" (`state_machine/running.rs:51`). A node whose global DB is unreadable stops consensus instead of abstaining. Given the shared connection that is the same outcome the activated-hash path already produces, so I read it as intended — flagging only so it is a decision rather than a side effect.

Nit: `err.to_string()` at `epoch_manager_service.rs:543` renders only the outermost `anyhow` context and drops the source chain, so the diesel error underneath — the part that says *why* the store was unreadable — does not reach the log. `format!("{err:#}")` keeps it, which is the whole point of the commit.

Still open from the previous review, all non-blocking and none touched by this commit: the fallback being inert while `inner` is taken by an in-flight scan (item 1 — the largest of them, since it overlaps the catch-up window the PR targets); no consensus-level test for the new `can_propose_epoch_end` branch (item 3); the two now-false comments at `consensus_tests/src/epoch_change.rs:334-337` and `:459` (item 4); and the deferral/`lock_epoch` observations (item 5).

crates/epoch_oracles/src/base_layer/oracle.rs
```diff
@@ -885,6 +885,29 @@ impl<TStore: EpochOracleStore + BaseLayerBlockHeaderStore, TClient: BaseNodeClie
...
+    /// it. The stored header only counts as the boundary when its height is exactly the epoch's first
+    /// height: a scan range that starts mid-epoch stores a first header that is not the boundary, and
+    /// ratifying against that would compare the wrong hash.
+    fn observed_epoch_boundary_hash(&self, epoch: Epoch) -> Result<Option<FixedHash>, BaseLayerOracleError> {
+        // The epoch length is only known once a scan has obtained the L1 constants; until then we
+        // cannot say which height opens the epoch.
+        let Some(epoch_length) = self.cached_epoch_length else {
+            return Ok(None);
+        };
+        let Some(boundary_height) = epoch.as_u64().checked_mul(epoch_length) else {
```
sdbondi: An overflow here is breaking an invariant of the system. This should error or even panic - do you agree?

crates/epoch_oracles/src/base_layer/oracle.rs
```diff
@@ -885,6 +885,29 @@ impl<TStore: EpochOracleStore + BaseLayerBlockHeaderStore, TClient: BaseNodeClie
...
     }
 
+    /// Returns the boundary block hash we have stored for `epoch`, or `None` if we have not scanned
+    /// it. The stored header only counts as the boundary when its height is exactly the epoch's first
+    /// height: a scan range that starts mid-epoch stores a first header that is not the boundary, and
+    /// ratifying against that would compare the wrong hash.
+    fn observed_epoch_boundary_hash(&self, epoch: Epoch) -> Result<Option<FixedHash>, BaseLayerOracleError> {
+        // The epoch length is only known once a scan has obtained the L1 constants; until then we
+        // cannot say which height opens the epoch.
+        let Some(epoch_length) = self.cached_epoch_length else {
```
sdbondi: The cached_epoch_length should also be populated before calling this as an invariant. Check under what conditions this is reachable and decide if error or panic is more appropriate

crates/epoch_oracles/src/base_layer/oracle.rs
```diff
@@ -885,6 +885,29 @@ impl<TStore: EpochOracleStore + BaseLayerBlockHeaderStore, TClient: BaseNodeClie
...
+            return Ok(None);
+        };
+        let Some(boundary) = self
+            .store
+            .get_first_block_header_in_epoch(epoch)
+            .map_err(BaseLayerOracleError::StoreError)?
+        else {
+            return Ok(None);
+        };
+        Ok((boundary.height == boundary_height).then_some(boundary.block_hash))
```
sdbondi: nit:
```suggestion
       self
            .store
            .get_first_block_header_in_epoch(epoch)
            .map_err(BaseLayerOracleError::StoreError)
            .map(|b| b
       		.filter(|boundary| boundary.height == boundary_height))
       		.map(|boundary| boundary.block_hash))
```

crates/epoch_oracles/src/base_layer/oracle.rs
sdbondi: Agreed, and done — it now returns `BaseLayerOracleError::EpochHeightOverflow { epoch, epoch_length }`.

Error rather than panic: epochs are derived as `height / epoch_length`, so reaching an overflow needs an epoch around 1.8e18, i.e. a corrupt value rather than a large chain. The call site is the consensus vote path, which already has a path for an oracle error, and the named error says which epoch and which length — a panic would take the node down for the same information.

crates/epoch_oracles/src/base_layer/oracle.rs
sdbondi: Taken, thanks — the whole tail is now `.filter(|boundary| boundary.height == boundary_height).map(|boundary| boundary.block_hash)`, which drops the `let ... else`.

crates/epoch_oracles/src/base_layer/oracle.rs
sdbondi: I checked the reachability and I think this one is legitimately `None` rather than a broken invariant, so I have left it as `Ok(None)` with a comment saying why.

Consensus does wait for the first scan (`state_machine/initialising.rs:58` -> `wait_for_initial_scanning_to_complete`), but that waiter is released by `on_scanning_complete` on `EpochEvent::DoneForNow`, and `sync_blockchain` has early `return Ok(false)` paths above the `get_consensus_constants` call — the lagged tip being below `start_height`, and no progress past the last scanned height. A validator started against a base layer that has not yet reached its configured `start_height` therefore completes initial scanning with the epoch length still unset.

In exactly that state the node has also crossed no epoch boundary, so the header store is empty for any epoch and "not observed" is the truthful answer, not a violated invariant — erroring or panicking would turn a fresh-network start into a fault. Note this is also where the old `Option<u64>`/`None` and the new `0` sentinel mean the same thing.

The overflow next to it is a different matter and is now an error, per your other comment.

sdbondi (commented):
Re-reviewed 29aa1a6. Items 1, 3, 4 and the `{err:#}` nit are addressed, and correctly. One thing to fix.

**Orphaned doc comment.** `crates/epoch_oracles/src/base_layer/oracle.rs:901-904` — moving `observed_epoch_boundary_hash` out of the inner impl left its doc comment behind, so it now sits directly above `is_within_epoch_end_spread`'s own doc and reads as one comment:

```rust
/// Returns the boundary block hash we have stored for `epoch`, or `None` if we have not scanned
/// it. The stored header only counts as the boundary when its height is exactly the epoch's first
/// height: a scan range that starts mid-epoch stores a first header that is not the boundary, and
/// ratifying against that would compare the wrong hash.
/// Returns true when our lagged scanner position is within `epoch_end_spread_blocks` of the
/// next epoch boundary. ...
fn is_within_epoch_end_spread(&self, current_epoch: Epoch) -> bool {
```

The text describes a lookup that function does not perform, and it is already carried (reworded) on `lookup_epoch_boundary_hash` at :1046. Delete the first four lines.

**Verified on the rest:**

The `store` clone plus `Arc<AtomicU64>` on `BaseLayerOracle` is the right shape: `new` is the only method in the impl block that gained the `Clone` bound, and `BaseLayerOracle::new` has exactly two callers, both updated (`tari_validator_node/src/bootstrap.rs:652`, `tari_indexer/src/bootstrap.rs:554`) — `HybridEpochOracle` takes an already-constructed oracle, so it needed nothing. `Relaxed` is sufficient: a stale zero and a header write that has not landed both degrade to `None`, and the epoch length is stored before each batch's headers rather than after, so the reader can never see a length newer than the headers it implies.

`Option<u64>` → `0`-as-unknown preserves `is_within_epoch_end_spread` exactly (it already had its own `epoch_length == 0` guard). The overflow case turning into `EpochHeightOverflow` rather than `Ok(None)` is unreachable by construction — epochs are `height / epoch_length`, so `epoch * epoch_length <= height` for any legitimately derived epoch — and an error is the honest answer for a value that cannot be interpreted.

`epoch_change_observed_boundary_is_voted_before_activation` is the right test and it does fail without the change: with `current_epoch()` capped, `em_epoch > current_epoch` is false and `TestEpochManager::is_within_epoch_end_spread` returns `false`, so the assertion that *all four* validators reach Epoch(2) can only pass through the new `expected_next_epoch_hash.is_some()` branch. Adding `set_oracle_visible_epoch` to both wedge tests restores the mechanism their comments claim, and both clear it on recovery.

Item 5 from the first review (the deferral path becoming routine, and `EpochManagerEvent::EpochChanged` only being published on `DoneForNow`) is still untouched. It was informational — nothing here needs to change for it — but the description is still framed around the lagging node recovering, and on a multi-batch catch-up that node votes promptly and then waits for the whole scan to finish before it can open the next epoch.

sdbondi (commented):
Reviewed, makes sense

sdbondi (commented):
Re-reviewed 6393c2c. The four orphaned doc lines are gone and nothing else moved. CI is green on this head, integration test included.

Every finding raised across this review is resolved:

- store errors no longer read as "boundary not observed" — they propagate as `EpochEventOracleError`, with the sqlite message carried through (d042c49, 2ac61bf)
- the boundary lookup no longer depends on `inner`, so it answers while a scan task owns it — the catch-up window the PR exists for (29aa1a6)
- the new `can_propose_epoch_end` branch has a consensus-level regression test, and both wedge tests now establish the state their comments claim (29aa1a6)
- the doc comment left behind by the move is deleted (6393c2c)

Only the item 5 observations remain, and they need no change here: `EpochManagerEvent::EpochChanged` is published from `on_scanning_complete`, which runs on `DoneForNow`, so on a multi-batch catch-up a node that votes and commits the EOE still waits for the scan to reach the lagged tip before it can open the next epoch. The committee gets its quorum either way, which is what this PR is for — worth a line in the description so the lagging node's own recovery is not read as fixed too.

I cannot approve my own PR (GitHub refuses `addPullRequestReview` with APPROVE), so this needs another maintainer's click. From my side it is ready.

PR #2573, opened 2026-09-10
fix(consensus): apply foreign proposal updates when proposing

sdbondi:
## Problem

CI run [34367513887](https://github.com/tari-project/tari-ootle/actions/runs/34367513887/job/102519956391) timed out in `consensus::foreign_shard_group_decides_to_abort` — not an assertion, a stall. The committee wedged at `high_qc = NodeHeight(6)` and never formed another QC, spinning dummy blocks until nextest's 600s kill (14 GB of log).

Every block from height 7 on drew zero votes:

```
❌ NO VOTE LocalAccept: transaction fee disagreement tx 38608720… in block [NodeHeight(7) …].
   Leader proposed 5, we calculated 0
```

Block 7 carried two commands for the same transaction: `ForeignProposal(…)` (the foreign committee's `LocalAccept(…, Abort)`) and `LocalAccept(tx)`.

## Cause

A replica evaluates a block as: newly justified block, then commands in block order — foreign proposals first, which `Command`'s ordering explicitly arranges ("Foreign proposals should come first in the block so that they are processed before commands") — all against **one** change set.

`on_propose` did the opposite. It processed foreign proposals, then called `take_all_transaction_executions`, whose `drain(..)` emptied the change set and discarded every pending pool update and foreign pledge it had just recorded, and only then processed the justified block. `apply_transaction_update` in the command loop was left with nothing to apply.

So the leader's record still read `Commit`, the `is_abort()` early return in `local_accept_transaction` was skipped, it executed the transaction and committed to `transaction_fee: 5`. Replicas applied the foreign abort and computed `0`. Nobody voted — including the proposer. The next leader rebuilt the identical block from the same still-pending inputs, so the stall is stable, not transient.

The `TODO` asserting "the system currently never proposes foreign proposals affecting a transaction in the same block for LocalPrepare/LocalAccept" was never enforced — `fetch_next_proposal_batch` picks foreign proposals and the transaction batch independently.

## Fix

Make the proposer's sequence match the replica's: process the justified block first, then the foreign proposals, then harvest executions without clearing the rest of each transaction's change. Commands are now derived from the same state the replica will evaluate them against.

This also restores the foreign pledges arriving in the block being built, which `execute_transaction` reads via `change_set.get_foreign_pledges` and the drain was throwing away.

## Compatibility

No format change — no header field, no encoding, no `hash_substate` input moves; not a `ProtocolVersion` matter and needs no activation entry.

The only blocks whose content differs are ones no honest replica votes for, and a block that never gathers a QC is never committed. Every block that can carry a QC is unchanged, so:

- **No reset, no re-sync.** Indexers consume committed blocks and substates; neither changes.
- **Mixed fleet is safe.** The replica evaluation path is untouched and was always the correct one. An old-code replica votes for a new-code leader's proposal; a new-code replica still refuses an old-code leader's stale one. Liveness returns proportionally as leaders upgrade — no coordinated rollout.

A chain currently wedged this way recovers through ordinary leader-failure/TC recovery once an upgraded node takes a turn as leader.

## Testing

- New unit test `taking_executions_keeps_the_pending_transaction_update` — fails on the old `drain(..)`, passes on the fix (verified both ways).
- `cargo test -p consensus_tests` — 67 passed, 0 failed.
- `cargo test -p tari_consensus --lib` — 22 passed, 0 failed.
- `cargo lints clippy -p tari_consensus --all-targets` — clean.

No deterministic reproduction of the original stall is included. Landing the two commands in one block depends on the foreign proposal arriving inside the window where the transaction is already `LocalPrepared`-ready, and `foreign_shard_group_decides_to_abort` only hits it occasionally — it is what caught this, but it is not a reliable guard. The unit test pins the specific invariant that broke instead.

## Note

Unrelated to #2569, which was merely the PR whose CI run surfaced it. That diff is one line in the LocalNet activation schedule; `ProtocolVersion` is not read anywhere on this path, 712 of 713 tests passed on the same shard under it, and the `drain(..)` dates to #1447 (June 2025).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11874,6 +11874,7 @@ dependencies = [
  "tari_state_tree",
  "tari_template_lib_types",
  "thiserror 2.0.18",
+ "time",
  "tokio",
 ]
 
--- a/crates/consensus/Cargo.toml
+++ b/crates/consensus/Cargo.toml
@@ -30,3 +30,6 @@ log = { workspace = true }
 serde = { workspace = true, default-features = true }
 thiserror = { workspace = true }
 tokio = { workspace = true, default-features = false, features = ["sync", "time", "macros", "rt"] }
+
+[dev-dependencies]
+time = { workspace = true }
--- a/crates/consensus/src/hotstuff/block_change_set.rs
+++ b/crates/consensus/src/hotstuff/block_change_set.rs
@@ -324,13 +324,16 @@ impl ProposedBlockChangeSet {
     /// The rest of each transaction's change — the pending pool update, foreign pledges, evidence — stays in
     /// the change set. Those describe state that later commands in the same block are evaluated against, so
     /// they must outlive the executions that were produced alongside them.
-    pub fn take_transaction_executions(&mut self) -> impl Iterator<Item = (TransactionId, TransactionExecution)> + '_ {
-        self.transaction_changes.iter_mut().filter_map(|(tx_id, change)| {
-            change
-                .execution
-                .take()
-                .map(|execution| (*tx_id, execution.into_transaction_execution()))
-        })
+    pub fn take_transaction_executions(&mut self) -> Vec<(TransactionId, TransactionExecution)> {
+        self.transaction_changes
+            .iter_mut()
+            .filter_map(|(tx_id, change)| {
+                change
+                    .execution
+                    .take()
+                    .map(|execution| (*tx_id, execution.into_transaction_execution()))
+            })
+            .collect()
     }
 
     pub fn add_transaction_execution(
@@ -656,8 +659,9 @@ mod tests {
 
     use tari_consensus_types::Decision;
     use tari_engine_types::commit_result::AbortReason;
-    use tari_ootle_common_types::{Epoch, NumPreshards};
-    use tari_ootle_storage::consensus_models::TransactionPoolStage;
+    use tari_ootle_common_types::{Epoch, NumPreshards, VersionedSubstateId};
+    use tari_ootle_storage::consensus_models::{SubstatePledge, TransactionPoolStage};
+    use tari_template_lib_types::ComponentAddress;
 
     use super::*;
 
@@ -673,15 +677,26 @@ mod tests {
 
         let mut aborted = local_prepared_record();
         aborted.set_local_decision(Decision::Abort(AbortReason::ExecutionFailure));
+        let transaction_id = *aborted.id();
         change_set.set_next_transaction_update(aborted).unwrap();
+        change_set.add_foreign_pledges(&transaction_id, ShardGroup::all_shards(NumPreshards::P256), vec![
+            SubstatePledge::Output {
+                substate_id: VersionedSubstateId::new(SubstateId::Component(ComponentAddress::from_array([1; 32])), 0),
+            },
+        ]);
 
-        assert_eq!(change_set.take_transaction_executions().count(), 0);
+        assert_eq!(change_set.take_transaction_executions().len(), 0);
 
         let mut committing = local_prepared_record();
         change_set.apply_transaction_update(&mut committing);
         assert!(
             committing.current_decision().is_abort(),
             "pending update must survive execution harvesting so later commands in the block see it"
+        );
+        assert_eq!(
+            change_set.get_foreign_pledges(&transaction_id).count(),
+            1,
+            "foreign pledges must survive execution harvesting so the transaction can still be executed"
         );
     }
 
--- a/crates/consensus/src/hotstuff/block_change_set.rs
+++ b/crates/consensus/src/hotstuff/block_change_set.rs
@@ -319,14 +319,17 @@ impl ProposedBlockChangeSet {
             .and_then(|change| change.execution.take())
     }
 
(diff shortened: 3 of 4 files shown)


sdbondi (commented):
Reviewed at `c68e8a6`. No blocking findings — the reorder is the right shape and I traced the claims below rather than taking them from the description.

## Verified

- **Ordering now matches the replica.** `on_ready_to_vote_on_local_block.rs:158` runs `process_newly_justified_block`, then `decide_what_to_vote` iterates `block.commands()` (`:291`) in `Ord` order. The proposer at `on_propose.rs:423` → `:425` → `:445` reproduces that sequence against one change set.
- **Moving the harvest after `process_newly_justified_block` does not change what lands in `executed_transactions`.** `common.rs:451-591` only mutates evidence and calls `set_next_transaction_update`; it never calls `add_transaction_execution`, so the harvested set is identical to before — only pool updates and pledges are additionally retained.
- **Retaining `transaction_changes` has no persistence effect here.** `change_set` in `build_next_block` is never `save()`d (it is built on a read transaction) and `add_transaction_execution` is not called on it after the reorder, so the now-execution-less retained entries cannot trigger `TransactionAlreadyExecuted`.
- **No `unreachable!()` exposure from the newly-effective `apply_transaction_update` at `on_propose.rs:504`.** The only pending stages a foreign proposal can set are `New`, `LocalPrepared`, `LocalAccepted` (`foreign_proposal_processor.rs:324,353,498,526,538`), all of which `transaction_pool_record_to_command` (`on_propose.rs:292-329`) dispatches; the `unreachable!()` arm stays unreachable.
- **Restored pledges cannot double-count.** `manager.rs:264-278` collects `local_pledges ∪ foreign_pledges` into a `HashMap<SubstateRequirement, Substate>`, so a pledge present both in the DB and in the change set collapses.
- **Abort-atom parity.** Proposer `local_accept_transaction` early-returns on `is_abort()` (`on_propose.rs:949-952`); the replica's abort branch does nothing (`on_ready_to_vote_on_local_block.rs:1186`). Both then emit the pool record's `transaction_fee`, which is what the fee-disagreement no-vote at `:1188` was comparing.
- **The test is a real regression guard.** `IndexMap::drain(..)` removes its range when the `Drain` is dropped regardless of consumption, and `.count()` drives it — so the old code empties the map and the `is_abort()` assert fails. Arg order in `TransactionPoolRecord::load` matches `transaction_pool.rs:468-486`.
- **Cargo.lock delta is only the `time` dev-dep**, and `time = "0.3.47"` is already in `[workspace.dependencies]`. Nothing generated swept in.
- **Compatibility claim holds** as far as I can check: no header field, no encoding, no `hash_substate` input change; only the command content of blocks that cannot gather a QC.

## Non-blocking

1. **`on_propose.rs:426-441` — the failed foreign proposal is still proposed.** The `ForeignProposal` commands are inserted into `commands` at `:398-406`, before this loop. When `process_foreign_block` fails, the warn says "Not proposing..." but the command is already in the block and `continue` leaves whatever partial updates the failed call wrote in the change set — which, after this PR, now flow into `apply_transaction_update` at `:504` and shape the emitted atoms. The block is doomed either way (the replica returns `ForeignProposalProcessingFailed`), so this is not a new liveness hazard, but the log is untrue and the half-state is now load-bearing rather than discarded. Removing the atom from `commands` (or bailing on the whole proposal) would make the message match the behaviour.

2. **`on_propose.rs:428` — leaf-block asymmetry that the PR's own premise argues against.** `process_foreign_block` is given `high_qc_certificate.as_leaf_block()` (the justify block), while the replica passes the block under evaluation, and the change set itself is anchored at `start_of_chain_block.as_leaf()` (`:414`). `TransactionPoolRecord::get` → `transaction_pool_get_for_blocks(to_block_id, …)` resolves pending updates along the chain *up to that block*, so the DB fallback inside `process_foreign_block` sees state as of the justify block. In steady state justify == leaf and this is invisible; it diverges when the leader extends an unjustified leaf above the justify. Pre-existing, and out of scope for a fix here, but worth a follow-up given the rest of the diff is about closing exactly this class of gap.

3. **`block_change_set.rs:327` — `take_transaction_executions` is lazy over `&mut self`, `drain` was not.** `drain` removed its range on drop whether or not the iterator was consumed; the `iter_mut().filter_map()` version only takes what a caller pulls. The single caller uses `.extend(...)`, so it is correct today, but a partially-consumed or dropped iterator now silently leaves executions behind. `#[must_use]`, or collecting into a `Vec` before returning, removes the footgun for the next caller.

## Nits

- `on_propose.rs:419` — "foreign proposals first, see `Command`'s ordering" is not literally what `CommandOrdering` says: `EvictNode` sorts first (`storage/src/consensus_models/command.rs:119-126`). It does not weaken the argument (evict commands carry no transaction change), but "before the transaction commands" would be accurate.
- The test pins the pending pool update; the description also claims the foreign pledges are restored, and that is the half that feeds `execute_transaction`/`prepare` via `get_foreign_pledges`. One extra `add_foreign_pledges` + `get_foreign_pledges().count()` assert would pin both halves of the contract the doc comment states.

sdbondi (commented):
Reviewed, makes sense

PR #2575, opened 2026-09-10
fix(engine)!: keep a callee's substates, buckets and proofs out of the caller's scope

sdbondi:
Part of the engine audit wave 1 — the five PRs that close every fund-theft and validator-kill finding. They ship as one coordinated validator + indexer release.

## The bug

`CallScope::include_owned_in_scope` seeded a callee frame's `owned` set with every substate reachable from the component's state — vault ids included — and `update_from_child_scope` then extended the caller's `owned` with the callee's whole set when the frame popped.

So any template could:

1. call a method the victim permits (`Account::get_balances` is `allow_all`),
2. and from that point have the victim's vault in its own scope, so `Vault::for_test(victim_vault).withdraw_all()` passes `check_component_scope`.

TARI's withdraw rule is `AllowAll` (`ResourceAccessRules::new()`), so the withdraw itself is unguarded. The comment above `check_component_scope` claimed owned references are not propagated up the call stack, which was the opposite of what `update_from_child_scope` did. The same propagation also let an attacker component add a victim's vault id to its own state permanently.

The existing `shenanigans.rs` regression tests never call into the victim first, so they passed throughout.

## The fix

Substates reachable through a component's state now live in a separate `component_owned` set. It is in scope for that frame alone and is dropped when the frame is popped; orphans that get attached to component state land there too. Only substates the frame created and still holds loosely cross back to the caller.

Buckets and proofs no longer merge upward either: a frame hands back only those named in its return value, and one it created that is still live and unreturned fails the call with `UnreturnedBuckets` / `UnreturnedProofs`. Buckets and proofs the caller lent as arguments — or that came from the transaction's base auth scope — remain the caller's and are exempt, so `account.deposit(bucket)` and the workspace-proof flow are unaffected.

Which proofs a frame *holds* is now tracked in `CallScope::proof_scope`, separately from the `auth_scope` subset that is currently authorizing an action. The two were one set, which was a live bug of its own: dropping a `ProofAccess` guard also gave up the proof, so `Proof::drop` after `Proof::authorize` failed with `ProofNotFound`, and a proof authorized before being returned never reached the caller.

## Tests

`crates/engine/tests/shenanigans.rs`:

- `it_does_not_leak_a_callees_vaults_into_the_callers_scope` — the attack above. On `development` this transaction **succeeds** and drains the victim's vault into the attacker's account.
- `it_rejects_a_bucket_that_is_neither_consumed_nor_returned_by_a_call`
- `it_rejects_a_proof_that_is_neither_dropped_nor_returned_by_a_call`
- `a_proof_outlives_the_authorization_taken_from_it` — cover for the authorize-then-drop and authorize-then-return patterns the `proof_scope` split fixes.

All 43 `tari_engine` test binaries pass.

### DIFF
--- a/crates/engine/src/runtime/auth.rs
+++ b/crates/engine/src/runtime/auth.rs
@@ -87,10 +87,6 @@ impl AuthorizationScope {
     pub fn contains_proof(&self, proof_id: &ProofId) -> bool {
         self.proofs.contains(proof_id)
     }
-
-    pub(super) fn update_from_child(&mut self, child: AuthorizationScope) {
-        self.proofs.extend(child.proofs);
-    }
 }
 
 impl Display for AuthorizationScope {
--- a/crates/engine/src/runtime/error.rs
+++ b/crates/engine/src/runtime/error.rs
@@ -320,6 +320,10 @@ pub enum RuntimeError {
     LockError(#[from] LockError),
     #[error("{count} substate locks were still active after call")]
     DanglingSubstateLocks { count: usize },
+    #[error("Bucket(s) {} were neither consumed nor returned by the call", .bucket_ids.iter().map(ToString::to_string).collect::<Vec<_>>().join(", "))]
+    UnreturnedBuckets { bucket_ids: Vec<BucketId> },
+    #[error("Proof(s) {} were neither dropped nor returned by the call", .proof_ids.iter().map(ToString::to_string).collect::<Vec<_>>().join(", "))]
+    UnreturnedProofs { proof_ids: Vec<ProofId> },
     #[error("No active call frame")]
     NoActiveCallFrame,
     #[error("Max call depth {max_depth} exceeded")]
--- a/crates/engine/src/runtime/impl.rs
+++ b/crates/engine/src/runtime/impl.rs
@@ -356,7 +356,12 @@ impl<TStore: StateReader + Clone + 'static, TTemplateProvider: TemplateProvider<
                 let _ignore = state.get_bucket(*bucket_id)?;
             }
 
+            // `get_bucket` scope-checks; `get_proof` does not, so the frame's own scope is checked here to keep a
+            // returned proof to the same rule as a returned bucket.
             for proof_id in value.proof_ids() {
+                if !state.current_call_scope()?.is_proof_in_scope(proof_id) {
+                    return Err(RuntimeError::ProofNotInScope { proof_id: *proof_id });
+                }
                 let _ignore = state.get_proof(*proof_id)?;
             }
 
--- a/crates/engine/src/runtime/impl.rs
+++ b/crates/engine/src/runtime/impl.rs
@@ -38,7 +38,7 @@ use tari_engine_types::{
     events::Event,
     fees::FeeReceipt,
     hashing::hash_template_code,
-    indexed_value::IndexedValue,
+    indexed_value::{IndexedValue, IndexedWellKnownTypes},
     instruction_result::InstructionResult,
     limits,
     lock::LockFlag,
@@ -3882,8 +3882,8 @@ where
         Ok(())
     }
 
-    fn pop_call_frame(&mut self) -> Result<(), RuntimeError> {
-        self.tracker.pop_call_frame()?;
+    fn pop_call_frame(&mut self, returned: &IndexedWellKnownTypes) -> Result<(), RuntimeError> {
+        self.tracker.pop_call_frame(returned)?;
         Ok(())
     }
 
--- a/crates/engine/src/runtime/mod.rs
+++ b/crates/engine/src/runtime/mod.rs
@@ -61,7 +61,7 @@ use tari_engine_types::{
     component::Component,
     confidential::{ClaimBurnOutputData, MinotariBurnClaimProof},
     fees::FeeReceipt,
-    indexed_value::IndexedValue,
+    indexed_value::{IndexedValue, IndexedWellKnownTypes},
     lock::LockFlag,
     published_template::TemplateBlob,
 };
@@ -219,7 +219,7 @@ pub trait RuntimeInterface {
     fn validate_return_value(&self, value: &IndexedValue) -> Result<(), RuntimeError>;
 
     fn push_call_frame(&mut self, frame: PushCallFrame) -> Result<(), RuntimeError>;
-    fn pop_call_frame(&mut self) -> Result<(), RuntimeError>;
+    fn pop_call_frame(&mut self, returned: &IndexedWellKnownTypes) -> Result<(), RuntimeError>;
     fn publish_template(
         &mut self,
         template: TemplateBlob,
--- a/crates/engine/src/runtime/scope.rs
+++ b/crates/engine/src/runtime/scope.rs
@@ -24,7 +24,20 @@ use crate::runtime::{
 pub struct CallScope {
     orphans: IndexSet<SubstateId>,
     owned: IndexSet<SubstateId>,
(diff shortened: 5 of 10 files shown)


sdbondi (commented):
Reviewed, makes sense, covered well by existing and new tests

sdbondi (commented):
_Posted by Claude._

Reviewed at `9f8172a` against `5faf1b5` (merge-base with `development`). I traced the frame lifecycle (`push_frame` → `include_owned_in_scope`/`include_refs_in_scope` → `pop_frame` → `update_from_child_scope`) and every `push_call_frame`/`pop_call_frame` pair in `processor.rs` (4/4, symmetric, and the auth-hook path re-enters via `invoke_component_method` so it is covered too).

What holds up:

- The core fix is right. Vault ids can only reach `owned` through `move_node_to_owned` (orphans only) or `add_substate_to_scope_unchecked` (root ids only, and `SubstateId::is_root` excludes vaults), so after this change nothing puts a callee's vault into the caller's `owned`. The `set_state` route is closed independently by `check_all_substates_in_scope` in `validate_component_state`, so "add a victim's vault id to its own state permanently" is genuinely gone.
- The `proof_scope` / `auth_scope` split is correct: `DropAuthorize` touches only `auth_scope`, `drop_proof` goes through `remove_proof_from_scope` and clears both, and `is_proof_in_scope` now reads `proof_scope`. `authorize_then_drop_proof` / `authorize_then_return_proof` are the right cover for it.
- `update_from_child_scope` filtering returned buckets through `child.bucket_scope` is load-bearing in a way worth noting: `validate_return_value` scope-checks buckets (`get_bucket`) but not proofs (see below), so for proofs the `child.proof_scope.contains` filter is the only thing stopping a callee from naming an id it never held.
- Consuming a lent bucket in the callee is handled: the caller keeps a stale `bucket_scope` entry, but the `self.buckets.contains_key` filter in `pop_frame` means it does not become a false `UnreturnedBuckets`.

## Blocking

**Address allocations still merge upward wholesale** — `scope.rs:271`

```rust
self.address_allocation_scope.extend(child.address_allocation_scope);
```

This is the one upward merge the PR leaves untouched, and it is the same shape as the bug being fixed. `use_allocated_address` (`working_state.rs:1214`) gates on `is_address_allocation_in_scope`, so an allocation a callee created and did not consume becomes usable by its caller: attacker calls victim, victim's frame allocates an address (e.g. `Component::with_public_key_address`) and returns before using it, the allocation lands in the attacker's scope and the attacker creates a component at that address. `validate_finalized`'s dangling-allocation check does not catch it — the attacker consuming the allocation is exactly what satisfies that check.

The same return-value filter you apply to buckets and proofs works here without breaking anything I can find: an allocation consumed in its own frame is already removed by `use_allocated_address`, and a factory that hands one back names it in the return value, which `include_refs_in_scope` already indexes via `component_address_allocations()` / `resource_address_allocations()`. If there is a flow that needs the wholesale merge, the reason belongs in a comment next to it, because it now reads as an oversight.

## Non-blocking

**1. `pop_frame` is stricter about empty buckets than `validate_finalized`** — `working_state.rs:1583` vs `:541`

`validate_finalized` deliberately tolerates leftover empty buckets (`.any(|(_, b)| !b.is_empty())`); the new per-frame check does not, since it filters only on `self.buckets.contains_key`. A template that splits a bucket with `take` and leaves the now-empty original un-`drop_empty`'d used to reach the end of the transaction and pass; inside a call it now aborts with `UnreturnedBuckets`. The builtin account template does call `drop_empty()` (`account/src/lib.rs:132`) so nothing in-tree hits it, but third-party templates may. Either exempt empty buckets here, or tighten `validate_finalized` to match so there is one rule rather than two.

**2. `validate_return_value` scope-checks buckets but not proofs** — `impl.rs:360`

`get_bucket` checks `is_bucket_in_scope`; `get_proof` checks nothing. A callee can therefore return a proof id it does not hold and pass validation. I could not build an escalation out of it — the `child.proof_scope` filter in `update_from_child_scope` stops it entering the caller's scope, and at the top level the only live proofs are the submitter's own — but the asymmetry means the scope filter is doing work the validator should also be doing. `if !state.current_call_scope()?.is_proof_in_scope(proof_id) { … }` alongside the existing lookup costs nothing and doesn't affect the legitimate return-a-proof case.

**3. `proofs_owed`'s doc comment says "auth scope"** — `scope.rs:281`

> The proofs this frame must account for before it is popped: those in its auth scope that its caller did not lend it.

It iterates `proof_scope`, not `auth_scope`. Given the whole point of this commit is that those two sets differ, the comment says the wrong thing about the exact case the split exists for — a proof whose `ProofAccess` was dropped is out of `auth_scope` but still owed.

**4. `attach_node_to_component`'s comment overstates what it does** — `scope.rs:188`

It only acts on addresses currently in `orphans`. An address already in `owned` — a `ComponentAddress` the frame created, or a resource `take_bucket` added at `working_state.rs:631` — is left in `owned` when attached to component state, so it still crosses to the caller on pop. That matches the old behaviour and I don't think it's exploitable (component and resource actions are gated by access rules, not scope), but "stays with the component when the frame is popped" isn't true for it.

**5. The regression test only covers the static-frame attacker** — `shenanigans.rs:439`

`call_then_steal_from_vault` is a function, so with the fix it lands on `NotInComponentContext` rather than on the scope check itself. It does still fail if the leak returns, so it guards the regression — but the attacker-*component* variant (calls the victim from a component frame, then `Vault::for_test(victim_vault).withdraw_all()`) is the one that exercises `check_component_scope`'s `SubstateNotOwned` path, and that path is what the `component_owned` split actually protects. Worth adding next to it.

sdbondi (commented):
_Posted by Claude._

Re-reviewed at `04cde46`. Every item from the previous round is resolved, verified against the code rather than the commit message:

- **Address allocations** (`scope.rs:274-285`) — now filtered through `child.address_allocation_scope` by what the return value names, on the same terms as buckets and proofs. An allocation consumed in its own frame is already removed by `use_allocated_address`, and `include_refs_in_scope` indexes both `component_address_allocations()` and `resource_address_allocations()`, so a factory handing one back is unaffected. `it_does_not_leak_a_callees_address_allocation_into_the_callers_scope` is a real regression test: `address_allocation_id` starts at 0 per transaction (`working_state.rs:199`), so the id the attacker names is deterministically the victim's, and on `development` the attacker creates at the victim's reserved address.
- **Empty buckets** (`working_state.rs:1585`) — `is_some_and(|bucket| !bucket.is_empty())` matches `validate_finalized`'s rule, so there is one rule for a leaked empty bucket rather than two.
- **`validate_return_value`** (`impl.rs:359-364`) — the proof scope check is safe at all five call sites: `wasm/process.rs:589` and the four in `processor.rs` run inside the returning frame, while `PutLastInstructionOutput` and `put_on_workspace` run against the base scope, which `update_from_child_scope` has already populated with the returned proof. No legitimate return path is blocked.
- **Component-frame attack test** — `it_does_not_leak_a_callees_vaults_into_a_calling_components_scope` lands on `SubstateNotOwned { requested_owner: attacker }`, which is the branch `component_owned` actually guards, alongside the existing static-frame case.
- Both doc comments now say what the code does.

CI is green on this head: `clippy`, `fmt`, `test (1-3/3)`, `integration test`, `ledger app`, all four fuzz targets.

No outstanding blocking items from me. I can't record an approval on this PR — GitHub does not accept an approving review from the PR author's own account — so this stands as a comment and it still needs a second maintainer to approve.

sdbondi (commented):
Makes sense, tests cover changes and preexisting tests pass

PR #2576, opened 2026-09-10
fix(engine)!: track spent inputs in WorkingStateStore

sdbondi:
Part of the engine audit wave 1 (with #2575). They ship as one coordinated validator + indexer release.

## The bug

`WorkingStateStore` kept no record of what the transaction had already spent that its own reads consulted:

- `exists()` answered from the immutable input snapshot, which the backing store leaves untouched until commit, so a downed UTXO still read as present;
- `try_lock` then let it be locked again and `load_and_cache` reloaded it;
- `down_utxo` inserts into an `IndexSet`, so the second down was a no-op.

Three spends followed.

**One UTXO listed n times in a single statement.** `stealth::transfer` folds the inputs positionally into the excess, so a spender — who knows the mask — can build a valid balance proof for n times the UTXO's value. The spend downs the one UTXO once and mints n·v.

**k statements over one UTXO in one transaction.** Up to `max_transfers_per_transaction` (64) in the main intent plus one in the fee intent, each with its own fresh outputs. One down, k output sets. A per-statement dedup alone does not close this.

**Spend then `StealthUtxoBurn`.** The resulting diff carried both a down and an up@v+1 of the same address.

The same shape existed for `down_confidential_output`.

## The fix

`exists`, `try_lock`, `load_and_cache` and `insert` now refuse any id in `downed_utxos` / `downed_confidential_outputs`. Nothing new is persisted and no new substate is introduced — the two sets the transaction already keeps become the spent set that every presence and load path consults.

The checks sit ahead of `down_utxo`, whose collapse-versus-down decision is unchanged: a UTXO mutated earlier in the transaction (unfrozen, say) moves from `loaded_substates` into `new_substates` without becoming new, so it consults the unmodified store to decide, and only a UTXO that did not exist before the transaction may collapse.

Two more, so the rule does not rest on the store alone:

- `stealth::basic_validations` rejects duplicate input commitments outright, which holds for every caller of `validate_transfer`.
- `validate_and_spend_stealth_utxos` validates the statement before downing anything, so a statement rejected on its own terms leaves its inputs unspent.

## Tests

`crates/engine/tests/stealth.rs` — all three of these **succeed** on `development`:

- `duplicate_inputs_in_one_statement_are_rejected`
- `two_statements_may_not_spend_the_same_utxo`
- `a_utxo_spent_in_this_transaction_cannot_also_be_burnt`

All 43 `tari_engine` test binaries plus `tari_engine_types` pass.

### DIFF
--- a/crates/engine/src/runtime/error.rs
+++ b/crates/engine/src/runtime/error.rs
@@ -76,6 +76,8 @@ pub enum RuntimeError {
     WorkspaceError(#[from] WorkspaceError),
     #[error("Substate '{id}' not found or is not a transaction input")]
     SubstateNotFound { id: SubstateId },
+    #[error("Substate '{id}' was already spent earlier in this transaction")]
+    SubstateAlreadySpent { id: SubstateId },
     #[error("Root substate '{id}' not found")]
     RootSubstateNotFound { id: SubstateId },
     #[error("Referenced substate '{id}' not found")]
--- a/crates/engine/src/runtime/state_store.rs
+++ b/crates/engine/src/runtime/state_store.rs
@@ -53,6 +53,9 @@ impl<TStore: StateReader> WorkingStateStore<TStore> {
     }
 
     pub fn try_lock(&mut self, id: SubstateId, lock_flag: LockFlag) -> Result<LockId, RuntimeError> {
+        if self.is_spent(&id) {
+            return Err(RuntimeError::SubstateAlreadySpent { id });
+        }
         if !self.exists(&id)? {
             return Err(RuntimeError::SubstateNotFound { id: id.clone() });
         }
@@ -141,7 +144,22 @@ impl<TStore: StateReader> WorkingStateStore<TStore> {
         })
     }
 
+    /// Whether `id` names a UTXO or confidential output this transaction has already spent. The backing store is
+    /// untouched until the transaction commits, so `downed_utxos` and `downed_confidential_outputs` are the whole
+    /// record of what this transaction has spent; every presence and load path consults them so that a spent output
+    /// reads as gone for the rest of the transaction.
+    fn is_spent(&self, id: &SubstateId) -> bool {
+        match id {
+            SubstateId::Utxo(address) => self.downed_utxos.contains(address),
+            SubstateId::ConfidentialOutput(address) => self.downed_confidential_outputs.contains(address),
+            _ => false,
+        }
+    }
+
     pub fn exists(&self, id: &SubstateId) -> Result<bool, RuntimeError> {
+        if self.is_spent(id) {
+            return Ok(false);
+        }
         let exists = self.new_substates.contains_key(id) ||
             self.loaded_substates.contains_key(id) ||
             self.state_store.exists(id)?;
@@ -149,6 +167,11 @@ impl<TStore: StateReader> WorkingStateStore<TStore> {
     }
 
     pub fn insert(&mut self, id: SubstateId, value: SubstateValue) -> Result<(), RuntimeError> {
+        // A spent address reads as absent to `exists`, so this is the check that keeps the substate diff to one
+        // record per address: a down or an up, never both.
+        if self.is_spent(&id) {
+            return Err(RuntimeError::SubstateAlreadySpent { id });
+        }
         if self.exists(&id)? {
             return Err(RuntimeError::DuplicateSubstate { address: id });
         }
@@ -157,6 +180,9 @@ impl<TStore: StateReader> WorkingStateStore<TStore> {
     }
 
     fn load_and_cache(&mut self, id: SubstateId) -> Result<&SubstateValue, RuntimeError> {
+        if self.is_spent(&id) {
+            return Err(RuntimeError::SubstateAlreadySpent { id });
+        }
         if let Some(s) = self.new_substates.get(&id) {
             return Ok(s);
         }
--- a/crates/engine/src/runtime/working_state.rs
+++ b/crates/engine/src/runtime/working_state.rs
@@ -381,6 +381,10 @@ impl<TStore: StateReader> WorkingState<TStore> {
         stmt: &StealthTransferStatement,
         view_key: Option<&RistrettoPublicKey>,
     ) -> Result<ValidatedStealthTransfer, RuntimeError> {
+        // The statement is checked in full before anything is downed, so a statement rejected on its own terms
+        // (duplicate inputs among them) leaves the ledger untouched.
+        let valid_transfer = stealth::validate_transfer(stmt, view_key)?;
+
         for input in &stmt.inputs_statement.inputs {
             let address = UtxoAddress::new(resource_address, input.commitment.into());
             let lock_id = self.store.try_lock(address.clone().into(), LockFlag::Write)?;
@@ -400,7 +404,6 @@ impl<TStore: StateReader> WorkingState<TStore> {
             }
         }
 
-        let valid_transfer = stealth::validate_transfer(stmt, view_key)?;
         Ok(valid_transfer)
     }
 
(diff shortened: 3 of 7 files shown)


sdbondi (commented):
_Posted by Claude._

Reviewed at `e19929a` against `upstream/development` (now carrying #2575). I traced both stealth spend paths end to end — `ResourceAction::StealthTransfer` → `stealth_transfer` (`impl.rs:4030`) and `VaultAction::PayFee` (`impl.rs:2676`) — plus `StealthUtxoBurn`, the freeze paths, and diff generation.

What holds up:

- The spent set is consulted on every route into the store that matters: `exists`, `try_lock`, `load_and_cache`, `insert`. `try_lock` is what closes spend-then-burn, since `StealthUtxoBurn` reaches the UTXO through `write_lock_substate`; `insert` is what closes a re-created input commitment producing a down and an up at one address.
- `down_utxo`'s collapse-versus-down decision is genuinely untouched, and `generate_substate_diff` reads downed addresses through `get_unmodified_substate`, which deliberately bypasses `is_spent` — so the down records still resolve their version. A mutated-then-downed UTXO still resolves through `state_store` because `get_for_mut` moved it out of `loaded_substates` and `down_utxo` shift-removed it from `new_substates`.
- The fee checkpoint is a whole-`WorkingState` clone (`tracker.rs:111`), so `downed_utxos` is captured and restored with it. A fee-intent spend survives a main-intent revert, and a main-intent spend of a UTXO the fee intent already spent is rejected before the revert — which is the "plus one in the fee intent" case in the description.
- `basic_validations` is the first statement in `validate_transfer`, so the dedup does reach every caller of it, wallet-side included.
- Duplicate *outputs* in one statement were already rejected — `new_substate` hits `add_substate_to_scope`'s `DuplicateSubstate` on the second one — so the asymmetry with inputs is not a gap.
- `PedersenCommitmentBytes` is `Copy + Ord + Hash`, so the `BTreeSet` is fine and deterministic.

Nothing blocking. Three things worth a look:

## 1. The dedup runs after the covenant fold, not before it

`SpendScriptExecution::covenant_balanced` folds `self.inputs` positionally into `input_commitments` (`spend_script_execution.rs:124-131`) and hands them to `validate_covenant_balance_proof`. That runs inside `verify_input_authorizations`, which on both spend paths executes *before* `execute_stealth_transfer` — `impl.rs:4063` for `stealth_transfer`, `impl.rs:2701` for `PayFee`. So a statement with a repeated input commitment reaches the covenant sub-balance proof with the duplicate counted twice, and only fails afterwards in `basic_validations`.

I do not think it is exploitable as it stands: the transaction aborts before anything is written, and predicate frames run `FrameWriteMode::ReadOnly` so a satisfied covenant buys nothing. But this is a second positional fold over the same inputs, and it is guarded only transitively — by the statement failing later — rather than by the check. The PR body's "holds for every caller of `validate_transfer`" is accurate, and the covenant proof is precisely not one of those callers. Hoisting the dedup to the top of `stealth_transfer` / the `PayFee` arm, or into `verify_input_authorizations`, would put every fold behind it and make the invariant hold by construction rather than by ordering.

## 2. `is_spent`'s doc claims more than the set holds

`state_store.rs:147-150`:

> `downed_utxos` and `downed_confidential_outputs` are the whole record of what this transaction has spent

They are not, quite. The collapse branch (`state_store.rs:231-240`) removes a UTXO created earlier in this same transaction from `new_substates` without adding it to `downed_utxos` — that UTXO is spent and in neither set. Nothing breaks today, because `exists()` answers false for it anyway once it is out of `new_substates`, so a re-spend fails with `SubstateNotFound`.

The visible consequence is that a collapsed address can be created again later in the same transaction: `insert` sees `is_spent == false` and `exists == false`, so create → spend → create passes. I worked through the accounting and it is value-neutral — the same commitment implies the same value and mask, and each statement balances on its own — with the surviving substate carrying the last creation's `UtxoOutput`. So this is a comment accuracy point, not a bug, but the `insert` comment at `state_store.rs:170-171` ("the substate diff to one record per address: a down or an up, never both") reads as if it covers the collapse case, and it does not. Either record collapses in the set as well, or narrow both comments so the next reader does not build on the stronger claim.

## 3. Validating before downing moves the cost of a bad input

Hoisting `validate_transfer` above the down loop means a statement naming a nonexistent, frozen or burnt input now pays for the full balance proof and range proof verification before that is discovered, where it previously failed on the first `try_lock`. That is sound — `charge_native_execution(transfer_native_points(..))` at `impl.rs:4049` charges the whole statement's verification up front, so nothing is extracted unpaid — but it is a behaviour change the description presents purely as a safety improvement, and the reason it is safe (the up-front charge) is worth a sentence in the commit message.

## Tests

The three regression tests are well chosen and each targets a distinct mechanism — `basic_validations`, `try_lock` across statements, `write_lock_substate` for the burn. All the imports the new burn test needs are already in the file. One note: `two_statements_may_not_spend_the_same_utxo` and `a_utxo_spent_in_this_transaction_cannot_also_be_burnt` both assert on the same `"was already spent earlier in this transaction"` string, which they reach through different call sites (`get_stealth_utxo_spend_auth`'s read lock in the first, `write_lock_substate` in the second). That is fine, but neither would notice if one of those two sites lost its check, since the other error would still surface first.

sdbondi (commented):
_Posted by Claude._

Correction to the last paragraph of my review — the tests note is wrong and can be ignored.

I said the two tests reach the error "through different call sites". They do not: `get_stealth_utxo_spend_auth`'s read lock and `StealthUtxoBurn`'s `write_lock_substate` both funnel into the same `WorkingStateStore::try_lock`, so there is one check, not two, and removing it turns both tests into `SubstateNotFound` failures. The tests pin it correctly. The three numbered points above stand.

sdbondi (commented):
Makes sense, red-green tested. Just waiting for the updates that checks for duplicate inputs on the wallet side

sdbondi (commented):
_Posted by Claude._

Re-reviewed the second commit, `9ed324c` (`fix(wallet): reject a stealth input listed twice when building a transfer statement`). It is sound, and it makes the engine test stronger rather than weaker.

- `create_transfer_statement` (`wallet/crypto/src/stealth.rs:174-184`) rejecting rather than deduplicating is the right call, and the comment gives the reason that matters: the balance proof signs the aggregate mask difference, so a silently deduplicated input set would produce a statement that fails at the validator as an unbalanced proof, with nothing pointing back at the caller. `WalletCryptoError::InvalidArgument` already exists with those fields, and the reachable callers — `ootle_wasm/core/src/stealth/transfer.rs:52`, `wallet/crypto/src/api.rs:74`, the local key provider — all propagate it. `crate_versioning.py impact tari_ootle_wallet_crypto` says nothing needs a bump: 0.43.0 is pending against 0.42.0 on crates.io, so the release already carries it.
- Moving the engine test off the builder is the right consequence. `spend_first_input_twice` (`template_test_tooling/src/support/stealth.rs:126-146`) is correctly constructed: with one 100 UTXO duplicated and 200 revealed out and no stealth outputs, the excess reduces to `2·mask·G`, and re-signing with `agg_input_mask = mask + mask` against an empty `agg_output_mask` matches it. So the statement the engine sees is balanced and internally valid, and the only thing wrong with it is the repeated commitment — which is exactly what the test should be asserting on. The previous version leaned on the builder being willing to emit an invalid statement, which it no longer is.

Worth noting: this test now walks straight through the ordering point from my first review. `spend_first_input_twice` clones the input's witness, so both copies go through `verify_input_authorizations` before `basic_validations` ever sees the statement. On this test's input that is harmless (a key-path spend), but it is the same traversal a covenant-rooted input would take into `covenant_balanced`'s positional fold at `spend_script_execution.rs:124-131`. The dedup landing before `verify_input_authorizations` rather than inside `validate_transfer` would close that ordering by construction.

My other two points (the `is_spent` / `insert` comments overstating what the collapse branch records, and the up-front `charge_native_execution` being the reason validate-before-down is safe) are unchanged and still non-blocking.

CI is mid-run on this head — `clippy`, the three `test` shards, `integration test`, `ledger app` and the four fuzz targets are all still pending. I'll pick it up when it settles.

PR #2577, opened 2026-09-10
fix(engine)!: scope-check proof access; match the resource on vault freeze

sdbondi:
Part of the engine audit wave 1 (with #2575 and #2576). They ship as one coordinated validator + indexer release.

> **Stacked on #2575.** This branch carries that commit, because the proof scope check needs the `proof_scope` / `auth_scope` split introduced there. Review the second commit; merge after #2575.

Two unrelated authority checks were missing.

## Proof access was unscoped

`ProofAction::Authorize` asked only whether the proof existed in the transaction-wide `proofs` map, and the read-only getters (`GetAmount`, `GetResourceAddress`, `GetResourceType`, `GetNonFungibles`) asked nothing at all. Proof ids come from a counter shared by the whole transaction, so a component called by a frame that holds a proof could name that proof by id and authorize with it — a badge it was never handed — or read its amount, resource and non-fungible ids.

All five now require the proof to be in the calling frame's scope: created there, passed in as an argument, or handed back by a callee. `Proof::authorize` on an out-of-scope proof returns `NotAuthorized`; the getters return `ProofNotInScope`.

Note on what this deliberately does *not* change: the base auth scope a top-level instruction inherits still carries the transaction's own proofs. That is what `DropAllProofsInWorkspace` exists for — a proof the submitter puts on the workspace authorizes their following instructions, including a call into a third-party component. Removing it breaks the documented workspace-proof flow (three `access_rules` tests are its spec). Placing a proof on the workspace and then calling a stranger stays the submitter's choice to make, in the same way that passing a `Proof` as an argument authorizes the callee.

## `SetVaultFreeze` froze any vault of any resource

`ResourceAction::SetVaultFreeze` authorized `Freeze` against `resource_address` and then write-locked `arg.vault_id` without checking what that vault holds. So anyone could publish a resource with `freezable(rule!(allow_all))` and freeze any vault on the network — including a TARI vault, leaving its owner unable to withdraw or to pay fees. Network-wide, and cheap.

It now requires the vault's resource to match, mirroring `Recall`, and fails with `FreezeResourceMismatch`.

## Tests

- `shenanigans.rs::it_refuses_to_authorize_a_proof_the_frame_does_not_hold` — a victim component holds a proof across a call into an attacker component, which is handed nothing and guesses the id. On `development` the attacker gets the badge.
- `shenanigans.rs::it_refuses_to_read_a_proof_the_frame_does_not_hold`
- `freeze.rs::it_refuses_to_freeze_a_vault_of_another_resource`

All 43 `tari_engine` test binaries pass.

### DIFF
--- a/crates/engine/src/runtime/error.rs
+++ b/crates/engine/src/runtime/error.rs
@@ -220,6 +220,12 @@ pub enum RuntimeError {
     WriteOutsideOwnComponent { id: SubstateId },
     #[error("Host operation '{operation}' is forbidden inside a resource auth hook")]
     ForbiddenInAuthHookContext { operation: &'static str },
+    #[error("Freeze on resource {resource_address} targeted vault {vault_id}, which holds resource {vault_resource}")]
+    FreezeResourceMismatch {
+        vault_id: VaultId,
+        resource_address: ResourceAddress,
+        vault_resource: ResourceAddress,
+    },
     #[error("Recall on resource {resource_address} targeted vault {vault_id}, which holds resource {vault_resource}")]
     RecallResourceMismatch {
         vault_id: VaultId,
--- a/crates/engine/src/runtime/impl.rs
+++ b/crates/engine/src/runtime/impl.rs
@@ -2077,6 +2077,18 @@ where
 
                 self.tracker.write_with(|state_mut| {
                     let vault_lock = state_mut.write_lock_substate(arg.vault_id.into())?;
+
+                    // The freeze rule that authorized this action belongs to `resource_address`, so it may only
+                    // reach vaults holding that resource.
+                    let vault_resource = *state_mut.get_vault(&vault_lock)?.resource_address();
+                    if vault_resource != resource_address {
+                        return Err(RuntimeError::FreezeResourceMismatch {
+                            vault_id: arg.vault_id,
+                            resource_address,
+                            vault_resource,
+                        });
+                    }
+
                     state_mut.set_vault_freeze(&vault_lock, arg.flags)?;
                     let payload =
                         Metadata::from_iter([("vault_id", arg.vault_id.to_string()), ("flags", arg.flags.to_string())]);
@@ -3305,7 +3317,7 @@ where
                 })?;
                 args.assert_no_args("Proof.GetAmount")?;
                 self.tracker.write_with(|state| {
-                    let proof = state.get_proof(proof_id)?;
+                    let proof = state.get_proof_in_scope(proof_id)?;
                     Ok(InvokeResult::encode(&proof.amount())?)
                 })
             },
@@ -3316,7 +3328,7 @@ where
                 })?;
                 args.assert_no_args("Proof.GetResourceAddress")?;
                 self.tracker.write_with(|state| {
-                    let proof = state.get_proof(proof_id)?;
+                    let proof = state.get_proof_in_scope(proof_id)?;
                     Ok(InvokeResult::encode(proof.resource_address())?)
                 })
             },
@@ -3329,7 +3341,7 @@ where
                 args.assert_no_args("Proof.GetResourceType")?;
 
                 self.tracker.write_with(|state| {
-                    let proof = state.get_proof(proof_id)?;
+                    let proof = state.get_proof_in_scope(proof_id)?;
                     Ok(InvokeResult::encode(&proof.resource_type())?)
                 })
             },
@@ -3342,7 +3354,7 @@ where
                 args.assert_no_args("Proof.GetNonFungibles")?;
 
                 self.tracker.write_with(|state| {
-                    let proof = state.get_proof(proof_id)?;
+                    let proof = state.get_proof_in_scope(proof_id)?;
                     let nfts = proof.non_fungible_token_ids();
                     Ok(InvokeResult::encode(&nfts)?)
                 })
@@ -3355,7 +3367,10 @@ where
                 args.assert_no_args("Proof.CreateAccess")?;
 
                 self.tracker.write_with(|state| {
-                    if !state.proof_exists(proof_id) {
+                    // A proof id is a sequential counter shared by the whole transaction, so authority has to come
+                    // from the frame's own scope: a proof it created, was passed as an argument, or a callee handed
+                    // back. Every other live proof in the transaction belongs to someone else.
+                    if !state.proof_exists(proof_id) || !state.current_call_scope()?.is_proof_in_scope(&proof_id) {
                         return Ok(InvokeResult::encode(&Err::<(), _>(NotAuthorized))?);
                     }
                     state.current_call_scope_mut()?.auth_scope_mut().add_proof(proof_id);
--- a/crates/engine/src/runtime/impl.rs
+++ b/crates/engine/src/runtime/impl.rs
@@ -3385,15 +3385,13 @@ where
                 args.assert_no_args("Proof.DropAuthorize")?;
 
                 self.tracker.write_with(|state| {
-                    // Scope before existence, so an id this frame does not hold answers the same whether or not a
-                    // proof is live at it: the ids are a dense counter, and a frame that could tell the two apart
-                    // could enumerate every proof in the transaction.
-                    if !state.current_call_scope()?.is_proof_in_scope(&proof_id) {
-                        return Err(RuntimeError::ProofNotInScope { proof_id });
-                    }
-                    if !state.proof_exists(proof_id) {
-                        return Err(RuntimeError::ProofNotFound { proof_id });
-                    }
+                    // Giving up an authorization only shrinks this frame's own auth scope, so it succeeds for any
+                    // id: an id the frame never authorized is already in the state being asked for. That makes it
+                    // answerless by construction, which is what keeps it from reporting whether a proof is live at
+                    // an id the frame does not hold — the ids are a dense counter, so an answer would enumerate
+                    // every proof in the transaction. `ProofAccess::drop` is the only route here, and a `Drop` has
+                    // nowhere to report a failure, so a rejection would abort the transaction from a drop point
+                    // the template author never wrote.
                     state.current_call_scope_mut()?.auth_scope_mut().remove_proof(&proof_id);
 
                     Ok(InvokeResult::unit())
--- a/crates/engine/src/runtime/impl.rs
+++ b/crates/engine/src/runtime/impl.rs
@@ -3385,6 +3385,12 @@ where
                 args.assert_no_args("Proof.DropAuthorize")?;
 
                 self.tracker.write_with(|state| {
+                    // Scope before existence, so an id this frame does not hold answers the same whether or not a
+                    // proof is live at it: the ids are a dense counter, and a frame that could tell the two apart
+                    // could enumerate every proof in the transaction.
+                    if !state.current_call_scope()?.is_proof_in_scope(&proof_id) {
+                        return Err(RuntimeError::ProofNotInScope { proof_id });
+                    }
                     if !state.proof_exists(proof_id) {
                         return Err(RuntimeError::ProofNotFound { proof_id });
                     }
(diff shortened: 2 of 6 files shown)


sdbondi (commented):
_Posted by Claude._

Reviewed at `8dc2923`, rebased onto `development` (now carrying #2575 and #2576) and down to a single commit — the stacked-on-#2575 note in the description is stale and can go.

Both fixes are correct. Nothing blocking. Two things worth a look, and some notes on what I checked so it is not re-derived later.

## Verified

- **`SetVaultFreeze` was the only freeze action that could reach across resources.** `SetStealthUtxosFreeze` (`impl.rs:2137`) and `SetConfidentialOutputsFreeze` (`impl.rs:2216`) both construct `UtxoAddress::new(resource_address, ..)` / `ConfidentialOutputAddress::new(resource_address, ..)` from the authorized resource, so a cross-resource target is not expressible there. `SetVaultFreeze` took a caller-supplied `VaultId`, which is not namespaced by anything — hence the hole, and hence the fix being sufficient. The new check mirrors `Recall`'s exactly, including leaving the vault lock held on the mismatch path; that is consistent and the transaction aborts regardless.
- **The invariant the unscoped `get_proof` calls rely on still holds.** `tracker_auth.rs:239` and `:253` iterate `auth_scope.proofs()` and look the proofs up without a scope check, which is only safe while `auth_scope.proofs() ⊆ proof_scope`. After this change it still is: `add_proof_to_scope` writes both sets, `remove_proof_from_scope` clears both, `set_auth_scope` seeds both, `Authorize` now adds to `auth_scope` only *after* checking `proof_scope`, and `DropAuthorize` only ever shrinks `auth_scope`. Worth keeping in mind if either set grows another writer.
- **The base-scope carve-out in the description is accurate.** `set_auth_scope` is reached only for the first frame of a top-level instruction (`working_state.rs:1540`); a nested frame starts with an empty auth scope plus caller badges. So the submitter's workspace proofs authorize a top-level callee and nothing deeper, which is what the three `access_rules` tests specify.
- `Authorize` returning `NotAuthorized` as an `Ok` value while the getters return a hard error is the right split, not an inconsistency — `try_authorize` is a `Result` in template code and the getters are infallible accessors, so a template that reads a proof it does not hold should trap.
- I also checked the residual freeze surface: an attacker can still get a vault of their own `freezable(allow_all)` resource into any account (`deposit` is `allow_all` by default, `account/src/lib.rs:62`, and creates a vault per resource) and freeze that. It denies the victim nothing they wanted, and it cannot touch their TARI vault or their fees, so the fix is complete for everything that matters.

## 1. `DropAuthorize` is the one proof action left unscoped

`impl.rs:3380-3395` still checks only `proof_exists`. Six of the seven proof actions now require frame scope; this one does not.

The practical effect is nil — it can only remove a proof from the calling frame's own auth scope, and a frame that does not hold the proof has nothing there to remove. What it leaves behind is an existence oracle: `DropAuthorize(id)` returns `ProofNotFound` for an id that is not live and `Ok(())` for one that is, so a frame can still enumerate which proof ids exist in the transaction. That is weak, since the ids are a dense counter and an attacker can already bound it — but it is the last probe this PR leaves open, and the asymmetry with the other six will read as an oversight to the next person.

I looked for a legitimate `DropAuthorize` on an out-of-scope proof and could not construct one: `ProofAccess` carries a `ProofId` in its CBOR, so one handed back by a callee is already indexed as a returned proof and enters the caller's scope through `update_from_child_scope`. Adding `is_proof_in_scope` here looks free.

## 2. Both proof tests are silently vacuous if the hard-coded id is ever wrong

`hold_proof_and_call` passes `args![0u32]`. That is right today: `object_ids` is one per-transaction counter shared between buckets and proofs (`working_state.rs:1492` hands out `&self.object_ids`), and neither test transaction carries fee instructions or workspace proofs, so the victim's `create_proof()` is genuinely id 0 and both tests fail on `development` as claimed.

The problem is what happens if that ever stops being true. If either transaction gains a fee instruction, or the harness starts allocating a bucket first, id 0 becomes something else and **both tests still pass**: `try_authorize` returns `Err` for a nonexistent proof exactly as it does for an out-of-scope one, and `get_proof_in_scope` returns `ProofNotInScope` before it ever consults existence. Nothing in either test would notice that it had stopped testing anything.

There is no public `ProofId → u32` accessor (only `From<u32>` and `Display`), so handing the attacker the real id needs a small `template_lib` addition. Failing that, `hold_proof_and_call` asserting its own `proof.amount()` still works after the call would at least pin that a proof with that id was live, which is the part that can rot.

sdbondi (commented):
Makes sense

sdbondi (commented):
_Posted by Claude._

Re-reviewed `3f4ed43`. Both points are resolved, and the second one is fixed better than what I suggested.

**`DropAuthorize`** — scope before existence is the right order and closes the oracle: an id the frame does not hold now answers `ProofNotInScope` whether or not a proof is live at it. I checked it does not break the legitimate guard path. `ProofAccess` is only produced by `authorize`/`try_authorize`, which already require scope, and Rust's reverse-declaration drop order means the guard drops before the `Proof` it came from — so the natural pattern never reaches `DropAuthorize` after `drop_proof` has cleared `proof_scope`. The inverted order was already a hard error before this commit (`ProofNotFound`), so nothing that worked stops working.

**The test brittleness** — threading `proof_id` through `hold_proof_and_call` and asserting `try_authorize().is_ok()` in the victim's frame first is a better fix than the `ProofId → u32` accessor I floated, because it needs no public API change and it fails in words. I traced the consequences for all three tests and the guard holds in each: if id 0 ever stops being the victim's proof, the assert's panic message replaces the reject reason, so `it_refuses_to_authorize_a_proof_the_frame_does_not_hold` fails on the `Amount` decode, and both `assert_reject_reason` tests fail on the mismatch. None of them can go quietly vacuous any more. The temporary `ProofAccess` from `try_authorize()` drops at the end of the assert, which removes the proof from `auth_scope` but not `proof_scope`, so the trailing `proof.drop()` still succeeds — that is the #2575 split doing exactly what it was built for.

**The new oracle test** is well constructed: id 0 live and held by the victim, id 7 with no proof at all, both drawing `ProofNotInScope`. Building the guard by hand as `ProofAccess { id }` is the right way to reach `DropAuthorize` — the field is `pub`, so any template can forge one, which is precisely why the check has to be there rather than relying on the RAII type.

One small thing, non-blocking and adjacent rather than new: `drop_proof` (`working_state.rs:833`) returns `ProofNotFound` for an out-of-scope proof, not `ProofNotInScope`. It is oracle-free for the same reason — scope is checked first, before `self.proofs` is consulted — so there is no soundness gap. But now that its sibling action reports the same condition as `ProofNotInScope`, the two disagree on what to call it, and `ProofNotFound` reads as "no such proof" when the proof usually does exist and simply is not this frame's.

CI is mid-run on this head. I'll pick it up when it settles.

sdbondi (commented):
_Posted by Claude._

Following up on the two-check shape in `DropAuthorize` — I think it can collapse to nothing, and end up stronger for it.

Both checks exist to police an operation that cannot hurt anyone. `DropAuthorize` only ever reaches `current_call_scope_mut().auth_scope_mut()`, so it shrinks the calling frame's own auth scope and nothing else. An id the frame never authorized is already in the state the caller is asking for, and an id whose proof has since been dropped elsewhere is too. There is no outcome to deny.

The scope check was only earning its place because the existence check was there to leak — remove the existence check and the oracle goes with it, by construction rather than by ordering:

```rust
ProofAction::DropAuthorize => {
    let proof_id = proof_ref.proof_id().ok_or_else(|| RuntimeError::InvalidArgument {
        argument: "proof_ref",
        reason: "DropAuthorize proof action requires a proof id".to_string(),
    })?;
    args.assert_no_args("Proof.DropAuthorize")?;

        // Giving up an authorization only shrinks this frame's own auth scope, so it succeeds for
        // any id: the frame reaches the state it asked for either way. `ProofAccess::drop` is the
        // only route here, and a `Drop` has no way to report a failure.
        state.current_call_scope_mut()?.auth_scope_mut().remove_proof(&proof_id);
        Ok(InvokeResult::unit())
    })
},
```

Two things make this better than either of the checked versions rather than merely shorter.

**`ProofAccess::drop` is the only production caller** (`template_lib/src/models/proof.rs:180`), and a `Drop` impl has nowhere to put an error — it does `unwrap_or_else(|_| panic!(..))`, so every failure mode here is an abort fired from an implicit drop point the template author did not write. `ProofNotFound` is reachable there today: a frame creates a proof, authorizes it, passes it to a component that drops it, and then the guard going out of scope aborts the transaction. That is the caller being punished for a callee's choice, at a point in the source where nothing appears to happen.

**The property worth testing changes into a better one.** "Live and absent draw the same error" is a statement about error text. What actually matters is that one frame cannot un-authorize another frame's proof — so `hold_proof_and_call(attacker, "drop_authorize_proof", 0)` should *succeed*, and the victim should then find its own proof still authorized afterwards. That asserts the isolation directly instead of asserting that two rejections are spelled the same way, and it would have caught a `DropAuthorize` that reached into the wrong scope, which the current test would not.

I checked nothing else depends on the current behaviour: `ProofAccess::drop` is the only route in-tree, and the only test touching it is the one added in this PR.

Worth noting this deletes a check that predates the PR, so it is slightly wider than the fix as scoped — but it is the same four lines this commit is already rewriting, and leaving a half-guarded version behind seems worse than either finishing it or reverting it.

sdbondi (commented):
_Posted by Claude._

Re-reviewed `8bed96c`. The simplification landed as intended, and the `drop_proof` rename that came with it is safe — I checked the three ways an error rename usually breaks something silently:

- No test asserts `ProofNotFound`, from `drop_proof` or anywhere else.
- All four `drop_proof` callers (`impl.rs:3407`, `:3459`, `:3483`, `tracker.rs:676`) use `?`. None goes through `.optional()`.
- That last one matters because `ProofNotFound` is in `IsNotFoundError` (`error.rs:417`) and `ProofNotInScope` is not, so an out-of-scope drop is no longer classified as not-found. With no `.optional()` caller the reclassification is inert, but it is the part worth re-checking if a caller ever starts tolerating a missing proof.

`DropAuthorize` being infallible also removes the last reachable `ProofNotFound` from a `Drop` impl, which was the concrete failure I was worried about.

One thing the applied test cannot see, and I do not think it is worth contorting it to fix:

`it_confines_dropping_an_authorization_to_the_calling_frame` asserts the victim's proof is still *held* after the attacker's call, not still *authorized*. Those are different sets after #2575, and the one `DropAuthorize` touches is the other one. The assert that pins the id — `Proof::from_id(proof_id).try_authorize().is_ok()` — builds a `ProofAccess` temporary that drops at the end of that statement, so the victim has already given up its own authorization before it calls the attacker. Proof 0 is out of the victim's `auth_scope` by the time the attacker runs. The trailing `proof.drop()` then reads `proof_scope`, which `DropAuthorize` never writes.

So a `DropAuthorize` that reached into the parent frame's `auth_scope` instead of its own would pass this test. What the test does pin is real and worth keeping — the attacker's call does not abort, and the victim's proof survives as held — it just is not the isolation property its name claims.

Observing the real property needs the victim to hold the guard across the call and then do something gated on the badge, since there is no engine query for `auth_scope` membership. That is a fair amount of scaffolding for a frame-local `swap_remove`, so I would leave it and adjust the name and doc comment to say what is actually asserted.

CI is mid-run on this head.

sdbondi (commented):
_Posted by Claude._

`71f3f3a` settles my last point. The name and doc now state what is pinned — the holder's `proof_scope` entry surviving the call, and the call answering at all — and separate that from the `auth_scope` half, which is structural via `current_call_scope_mut` being `call_frames.last_mut()`. That is the right split: the structural property does not need a test, and the doc says why rather than leaving the gap for someone to rediscover.

Also worth retracting something from my side. I called the engine comment's "`ProofAccess::drop` is the only route here" imprecise, on the grounds that the guard is forgeable. That was wrong — forging `ProofAccess { id }` still reaches the action through `ProofAccess::drop`, so the claim holds exactly as written, and the sentence only needs the caller to be a `Drop` impl for the "nowhere to report a failure" argument to stand. The template-side comment now says both halves plainly, which is better than either of us had it.

Nothing outstanding from me on the code. One leftover: the PR description still opens with

> **Stacked on #2575.** This branch carries that commit, because the proof scope check needs the `proof_scope` / `auth_scope` split introduced there. Review the second commit; merge after #2575.

The branch is rebased to a single commit on `development` and #2575 is merged, so all three sentences are now misleading — worth dropping before this merges, since the body is what ends up in the squash message. The description also predates the `DropAuthorize` and `drop_proof` changes, which are not mentioned in it.

CI has restarted on this head; `integration test` was the only job left outstanding on the previous one and it had not reported.

PR #2578, opened 2026-09-10
fix(engine)!: gate account ownership on the signer badge

sdbondi:
Closes audit finding **C5** (fund theft).

An account's component address is derived from its public key, so which rules a component may be created at that address under is a decision for that key. Two routes answered it in the caller's favour instead.

### The `CreateAccount` instruction

`CreateAccount { public_key, owner_rule, access_rules }` derived the canonical address for any public key and forwarded a caller-supplied `owner_rule`/`access_rules` to the constructor. A squatter could create the victim's account ahead of them, name itself the owner, and collect every subsequent deposit — including the `CreateAccount { victim_pk, bucket }` deposits the victim's own senders make.

The signer badge for `public_key` is now required in the transaction's base auth scope, but only when either rule is supplied. Creating an account for someone else on the default rules stays permissionless, which is what lets a sender deposit to an account that does not exist yet.

### The account template's constructor

`Account::create` was also reachable as a plain function — from a `CallFunction` instruction, and from another template through `TemplateManager::call`, which never touches the instruction. Both land in `TransactionProcessor::call_function`, which now refuses that one function by name (`ACCOUNT_CONSTRUCTOR_FUNCTION`, the same constant `create_account` looks it up with). Any other function added to the account template later stays callable.

That leaves `CreateAccount` as the only way to create an account, and the gate above as the only place the decision is made:

- `create_account` invokes the constructor directly rather than through `call_function`, so the instruction is unaffected.
- The template's methods go through `call_method` and are unaffected.
- Squatting from an attacker's own template is already impossible: `CreateComponentAllocation` derives under `state.current_template()`, so only a frame already executing as the account template can allocate an account address.

### Scope note

A template can no longer create an account on a user's behalf. Given the address derivation, a template doing so *is* the squat; the `CreateAccount` instruction remains available at the transaction level.

### Tests

`crates/engine/tests/account.rs`:

- `custom_ownership_of_another_keys_account_is_refused` — custom owner rule, then custom access rules, over the `CreateAccount` route
- `the_account_constructor_is_not_callable_as_a_function` — the `CallFunction` route
- `the_account_constructor_is_not_reachable_by_a_cross_template_call` — the `TemplateManager::call` route, via a new `Shenanigans::create_account_for`
- `an_account_for_another_key_may_still_be_created_on_the_default_rules` — the permissionless deposit flow still works
- `attempt_to_overwrite_account` — retargeted at the new rejection

Each of the three attack tests was first run against the unfixed code: all three transactions **succeed** and hand the attacker ownership of the victim's account.

Verified with the full `cargo test -p tari_engine` sweep (43 binaries, 0 failures), `cargo test -p tari_template_builtin`, and `cargo lints clippy -p tari_engine --all-targets`.

### Breaking

Rejects transactions that were previously accepted. No template recompile — the builtin account binary is unchanged.

### DIFF
--- a/crates/engine/src/runtime/impl.rs
+++ b/crates/engine/src/runtime/impl.rs
@@ -3847,6 +3847,16 @@ where
             .read_with(|state| state.authorization().check_current_component_access_rules(method))
     }
 
+    fn check_signer_badge_in_scope(&self, public_key: RistrettoPublicKeyBytes) -> Result<(), RuntimeError> {
+        self.tracker.read_with(|state| {
+            let badge = NonFungibleAddress::from_public_key(public_key);
+            if !state.base_call_scope().auth_scope().contains_badge(&badge) {
+                return Err(RuntimeError::SignerBadgeNotInScope { public_key });
+            }
+            Ok(())
+        })
+    }
+
     fn check_component_ownership(&self, action: ActionIdent) -> Result<(), RuntimeError> {
         self.tracker.read_with(|state| {
             let locked = state
--- a/crates/engine/src/runtime/mod.rs
+++ b/crates/engine/src/runtime/mod.rs
@@ -106,6 +106,7 @@ use tari_template_lib::{
         NonFungibleAddress,
         TemplateAddress,
         ValidatorFeePoolAddress,
+        crypto::RistrettoPublicKeyBytes,
         engine_args::IntrinsicId,
         stealth::StealthTransferStatement,
     },
@@ -213,6 +214,10 @@ pub trait RuntimeInterface {
     fn check_component_access_rules(&self, method: &str) -> Result<(), RuntimeError>;
     /// Checks whether the current execution context has owner permission of the given component.
     fn check_component_ownership(&self, action: ActionIdent) -> Result<(), RuntimeError>;
+
+    /// Asserts that the signer badge for `public_key` is in the transaction's base auth scope, i.e. that the
+    /// transaction is signed by that key.
+    fn check_signer_badge_in_scope(&self, public_key: RistrettoPublicKeyBytes) -> Result<(), RuntimeError>;
 
     fn update_component_template(&mut self, new_template: TemplateAddress) -> Result<(), RuntimeError>;
 
--- a/crates/engine/src/transaction/error.rs
+++ b/crates/engine/src/transaction/error.rs
@@ -124,6 +124,8 @@ pub enum TransactionErrorKind {
     NotAMigrationFunction { name: String },
     #[error("Migration functions cannot be called directly: {name}")]
     CannotCallMigrationFunctionDirectly { name: String },
+    #[error("The account constructor cannot be called directly. Use the CreateAccount instruction.")]
+    CannotCallAccountConstructor,
     #[error("Invalid CreateAccount operation for component {component_address}: {details}")]
     InvalidCreateAccount {
         component_address: ComponentAddress,
--- a/crates/engine/src/transaction/processor.rs
+++ b/crates/engine/src/transaction/processor.rs
@@ -701,6 +701,13 @@ where
                 address: ACCOUNT_TEMPLATE_ADDRESS,
             })?;
 
+        // The derived address belongs to `public_key_address`, so anything that departs from the rules that key
+        // would get requires that key's signature. Creating the account on the default rules stays permissionless
+        // so that a sender can deposit to an account that does not exist yet.
+        if owner_rule.is_some() || access_rules.is_some() {
+            runtime.interface().check_signer_badge_in_scope(*public_key_address)?;
+        }
+
         let account_address = derive_component_address_from_public_key(&ACCOUNT_TEMPLATE_ADDRESS, public_key_address);
 
         let maybe_existing_account = runtime
--- a/crates/engine/src/transaction/processor.rs
+++ b/crates/engine/src/transaction/processor.rs
@@ -820,6 +820,13 @@ where
         function: &str,
         args: Vec<InstructionArg>,
     ) -> Result<InstructionResult, TransactionErrorKind> {
+        // An account lives at an address derived from its public key, so which rules a component may be created
+        // there under is that key's decision. `CreateAccount` is the sole route to the constructor and is where
+        // that decision is enforced.
+        if *template_address == ACCOUNT_TEMPLATE_ADDRESS && function == ACCOUNT_CONSTRUCTOR_FUNCTION {
+            return Err(TransactionErrorKind::CannotCallAccountConstructor);
+        }
+
         let template = template_provider
             .get_template(template_address)
             .map_err(|e| TransactionErrorKind::FailedToLoadTemplate {
--- a/crates/engine/tests/account.rs
+++ b/crates/engine/tests/account.rs
@@ -6,7 +6,7 @@ use tari_crypto::{keys::PublicKey, ristretto::RistrettoPublicKey};
 use tari_engine::runtime::{ActionIdent, RuntimeError};
 use tari_ootle_transaction::{Epoch, Transaction, args};
 use tari_template_builtin::ACCOUNT_TEMPLATE_ADDRESS;
-use tari_template_lib::types::{Amount, access_rules::ComponentAccessRules, constants::TARI_TOKEN, rule};
+use tari_template_lib::types::{Amount, OwnerRule, access_rules::ComponentAccessRules, constants::TARI_TOKEN, rule};
 use tari_template_test_tooling::{
     TemplateTest,
     support::assert_error::{assert_access_denied_for_action, assert_reject_reason},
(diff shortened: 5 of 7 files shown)


sdbondi (commented):
_Posted by Claude._

Reviewed at `b729c63`. The fix is correct and nothing blocks. The branch is still based on `c7d59d0`, so it wants a rebase onto `development` now that #2575–#2577 are in — it touches `impl.rs`, `mod.rs` and `processor.rs`, all of which those three moved.

## The red check is a flake, not this PR

`Test Results (CI)` reports `Stealth transfer to account that does not previously exist` failing, which reads alarmingly like the exact flow this PR carves out. It is not. The junit artifact shows the scenario dying on its *first* step:

```
Given a network with registered validator VN and wallet daemon WALLET_D
Step panicked: Wallet MINOTARI_WALLET did not have at least 20000000 uT after 100 seconds
  (balance: 0 uT, pending: 203090087002 uT)
```

Base-layer coinbase maturity timing out during setup — the scenario never reached a transfer, let alone an account creation. The `integration test` job itself passed; only the aggregator surfaced it. A re-run should clear it.

## Verified

- **Why leaving the pk-derived allocation open is actually safe.** The description justifies it as "the same derivation the permissionless deposit flow needs", which undersells it. The real reason is that the derivation is template-namespaced: `CreateComponentAllocation { public_key }` derives from `state.current_template()` (`impl.rs:3744`), and `Instruction::AllocateAddress` carries no public key at all — so an attacker template can only ever reach `derive(ITS_OWN_TEMPLATE, victim_pk)`. The only code that can produce `derive(ACCOUNT_TEMPLATE_ADDRESS, pk)` is `Account::create` itself, via `with_public_key_address`, which is exactly what this PR gates. Worth putting in the description, because as written a reader has to take the claim on trust.
- **The signer check means what it says.** `initial_ownership_proofs` is built from `transaction.signers_iter()` (`transaction_executor.rs:140`), and the base scope's `caller_badges` is always empty (`set_caller` only runs on pushed frames), so `base_call_scope().auth_scope().contains_badge(..)` is exactly "this key signed the transaction".
- **The rollout claim holds.** `ACCOUNT_TEMPLATE_ADDRESS` is the constant `[0u8; 32]` (`template_builtin/src/lib.rs:27`), not a code hash, and builtins load from the binary rather than a `PublishedTemplate` substate (`wasm/cache.rs:279`). No substate moves; only execution results change.
- **The `.drop()` is load-bearing, not tidiness.** `GetSignerProof` mints a real `ProofId` through `new_proof` (`impl.rs:3708`), so without it the frame would fail its pop with `UnreturnedProofs` under #2575. Easy to lose in a later refactor — worth a word in the comment above it.
- **No in-tree caller breaks.** Every account-creation path uses `create_account` / `create_account_with_bucket`, both `owner_rule: None, access_rules: None`. The SDK's `generic_builder` reaches `create_account_custom` only when an owner rule or bucket is asked for, and still passes `owner_rule: None` in the bucket-only case — so the deposit-to-a-nonexistent-account flow does not trip the new check.

## 1. The two checks are not the same check

`check_signer_badge_in_scope` reads `base_call_scope()` (`impl.rs:3826`); the template-side `GetSignerProof` reads `current_call_scope()` (`impl.rs:3701`). For `Instruction::CreateAccount` they agree, because the first frame's auth scope is a clone of the base. They diverge for a nested `CallFunction(ACCOUNT_TEMPLATE_ADDRESS, "create", ..)` from inside another template: a nested frame carries no virtual proofs, so custom-rule account creation becomes impossible from a template even for the key's own owner.

That is the safe direction and I am not asking for it to change, but it is a real capability removal that the description presents as "the template being the chokepoint for the direct `CallFunction` route" — the direct route is only one of the two it closes. No test covers the nested case either way.

## 2. `check_signer_badge_in_scope` is named for a scope it does not read

It says "in scope" but reads the base scope specifically. Any future caller invoking it from inside a frame gets a check against the transaction's signers rather than that frame's authority — weaker than the name implies, and it is now a public `RuntimeInterface` method so the next caller may not be this one. The doc comment says the right thing ("i.e. that the transaction is signed by that key"); naming it for that, e.g. `check_transaction_signed_by`, would make the trait surface say it too.

## 3. The refusal test cannot tell "victim did not sign" from "nobody signed"

`custom_ownership_of_another_keys_account_is_refused` passes `vec![]` as the auth proofs, so the base scope holds no badges at all. The test would pass just as well against a check that merely required *some* badge. The payer's proof is already bound as `_payer_proof` — passing `vec![payer_proof]` makes it a signed transaction whose signer is not the victim, which is the case that actually distinguishes the fix from a weaker one.

## 4. Nothing asserts the property that makes the open path safe

`an_account_for_another_key_may_still_be_created_on_the_default_rules` reads `all_accounts().keys().next()` and checks a balance. It never asserts that the account sits at `derive(ACCOUNT_TEMPLATE_ADDRESS, victim_pk)` or that its owner rule is `OwnerRule::ByPublicKey(victim_pk)` — and the owner rule is the whole reason permissionless creation is not squatting. If the template's default owner ever changed, this test would still pass while the security argument quietly stopped holding.

## 5. The same shape is open to every other template

`with_public_key_address` (`template_lib/src/component/instance.rs:55`) is public API, and the fix here lives inside `Account::create` rather than in the engine. Any third-party template that gives each user a component addressed by their public key has the identical squatting exposure the account template just closed, with nothing to warn the author. Out of scope for this PR, but a doc warning on `with_public_key_address` would stop the next template inheriting the bug.

sdbondi (commented):
_Posted by Claude._

Re-reviewed at `8644cfd` (rebased onto `development`, two commits). This is a better fix than the one I reviewed, and it dissolves my first finding rather than answering it: there is now one gate on one route, the builtin account WASM is untouched, and no template author has to write anything.

The thing I'd have worried about when the template-side check was deleted — that removing it reopens the bypass — is covered, and covered by a test. Worth recording why the single guard is sufficient, since the two entry points look independent:

- `invoke_template_function` (`impl.rs:525`) is not a parallel implementation; it delegates to the same `TransactionProcessor::call_function`. So the `CallFunction` instruction and `TemplateManager::call` from inside a template both land on the one check, and `the_account_constructor_is_not_reachable_by_a_cross_template_call` pins the second route rather than assuming it.
- The other three `invoke_template` sites are accounted for: `create_account` (`processor.rs:801`, the sanctioned route), `call_method` (`:960`, methods, unaffected), and the migration path (`:574`). Migration cannot reach `create` — it is not a migration function, so `get_function`/`NotAMigrationFunction` rejects it before invocation.
- `create` is indeed the account template's only function; every other item takes `&self`/`&mut self`.

The rewritten description is accurate on all the points I checked, including the `CreateComponentAllocation` namespacing argument and the scope note about templates no longer being able to create accounts. Good that the latter is stated outright — it is a real capability removal, and "a template doing so *is* the squat" is the right framing.

## Still open from my previous review

Three of the four non-blocking points carry over unchanged, and two of them read differently now:

**The naming of `check_signer_badge_in_scope`** (`impl.rs:3850`) matters more than it did. It is no longer one of two checks — it is *the* check, and it reads `base_call_scope()` while being named for scope generally. A future caller reaching it from inside a frame would get a check against the transaction's signers rather than that frame's authority. `check_transaction_signed_by` says what it does.

**`with_public_key_address` is now the odd one out** (`template_lib/src/component/instance.rs:55`). The account is protected by a special case naming `ACCOUNT_TEMPLATE_ADDRESS` in the processor — nothing a third-party template can benefit from. Any template giving each user a component addressed by their public key has the exposure this PR just closed for accounts, and now there is visibly no general mechanism. A hazard note on that method is the cheap version.

**The two test points** are unchanged: `custom_ownership_of_another_keys_account_is_refused` still passes `vec![]`, so it cannot distinguish "the victim did not sign" from "nobody signed" — `_payer_proof` is already bound and passing it would make the test assert the case that matters. And `an_account_for_another_key_may_still_be_created_on_the_default_rules` still checks only that a deposit landed in *some* account, never that it sits at `derive(ACCOUNT_TEMPLATE_ADDRESS, victim_pk)` under `OwnerRule::ByPublicKey(victim_pk)` — which is the entire reason that path is safe to leave open.

## New: `attempt_to_overwrite_account` now duplicates another test

Retargeted, it asserts exactly what `the_account_constructor_is_not_callable_as_a_function` asserts, and its name no longer describes what it does. The property it used to pin is unreachable now, so nothing is lost by deleting it — but there is a better home for it. The existing-account branch of `create_account` (`processor.rs:731`) rejects `owner_rule`/`access_rules` with `InvalidCreateAccount`, and nothing tests it. With the signer gate now sitting ahead of that branch, it is reachable only by the key's own owner, which makes it exactly the kind of thing that rots unnoticed.

## Nit

The guard keys on the template address while the message names the function and prescribes the instruction (`processor.rs:816`). Correct today, since `create` is the only function — but if the account template ever gains a second one, the rejection will tell its caller to use `CreateAccount`, which will not be the answer.

sdbondi (commented):
Makes sense

PR #2579, opened 2026-09-10
fix(engine)!: remove attacker-reachable panics from execution

sdbondi:
Part of the engine audit wave 1 (with #2575, #2576, #2577 and #2578). They ship as one coordinated validator + indexer release.

The release profile builds with `panic = 'abort'` and `overflow-checks = true`, and nothing on the execution path catches unwinds. Each of the following was a way for a submitter to stop every validator that executed their transaction.

## Validator kills

**`schnorr_verify` on `PublicKey::Zero`.** `RistrettoSchnorrBlake2bVerifier::verify` unwrapped the `PublicKey` and `SignaturePayload` variants with `expect`. `PublicKey::Zero` is a valid CBOR variant and the `#[default]`, reachable from the intrinsic on any caller-supplied key and from any template that verifies one. A variant this verifier cannot use is now a failed verification.

**Guest-controlled pointer arithmetic at publish.** `WasmEnv::load_template_def` computed `ptr as u32 + WASM_PTR_SIZE as u32` on the guest's `_ABI_TEMPLATE_DEF` global, and it runs *before* `validate_instance`. A ~50-byte module declaring that global as `-1` overflowed the `u32`, so a `PublishTemplate` instruction killed every validator that reached `validate_code`. The sum is now checked and yields `MemoryPointerOutOfRange`.

**Resource balance overflow.** `ResourceContainer::deposit` and the re-adds in `unlock` used the raw `Amount` operators, which are a plain `u128` add. A resource with total-supply tracking disabled keeps no running total to catch it, so minting past `Amount::MAX` and depositing overflowed. Both now use a checked add and return `ResourceError::BalanceOverflow`.

**`FeeBreakdown::add`** accumulated a `u64` raw. It saturates rather than wrapping, which is safe because saturation can only pin the total at `u64::MAX` and no `u64` payment can meet that — so an overflowing breakdown rejects the transaction rather than letting a submitter under-pay. The breakdown *is* what affordability is decided against: `FeeState::is_paid_in_full` compares `total_payments()` against `fee_charges.get_total()`.

**A container's whole balance is now bounded, not just its unlocked field.** A container holds its balance in two fields, unlocked and locked, and `Amount::MAX` bounds their sum. Bounding the deposit against the unlocked field alone let the pair exceed it: lock part of a vault, deposit up to `Amount::MAX` into what is left unlocked, and the container holds more than the maximum. The six lock sites then add the unlocked field onto the locked one with a raw `u128` add, which aborts the executor. A confidential resource with total-supply tracking disabled has nothing else in the way, so the mint reaches the container unchallenged. Deposit is the only operation that raises the sum — locking and unlocking move value between the fields — so it is bounded against the pair and every one of those moves is in range.

## Fund inflation, same file

`lock_all`'s `Confidential` arm copied the revealed amount where it moves it, so the amount was added to `locked_revealed_amount` while remaining in `revealed_amount`. `unlock` then returned it a second time, so a proof taken over a confidential vault and dropped again **doubled** the vault's revealed balance — reachable by any template calling `create_proof()` on a confidential vault, in one transaction, with no privileges. The `Fungible` and `Stealth` arms already moved it with `mem::take`.

This one is pre-existing and is not an audit finding; it sits in the function whose `Stealth` arm is fixed above.

**The `expect`s in `finalize_fees_and_refunds`** return `InvariantError`. They are guarded by `FeeState`, but the guard and the unwrap are far apart, and `FeePaymentInMainIntent` is the only thing standing between them.

## Functional break, same file

`ResourceContainer::lock_all`'s `Stealth` arm returned a `public_fungible` container. `unlock` rejects the type mismatch, so **every** proof taken over a stealth vault or bucket — TARI included — aborted the transaction, leaving an undroppable proof behind. `Account::create_proof_for_resource(TARI_TOKEN)` followed by `DropAllProofsInWorkspace` was enough to hit it.

## Tests

Each of these panics or aborts on `development`:

- `signature.rs::check_signature_with_a_zero_public_key`
- `publish_template.rs::publish_template_with_an_out_of_range_abi_pointer` (adds `wat` as a dev-dependency to write the module)
- `fungible.rs::minting_past_the_maximum_vault_balance_is_rejected`
- `account.rs::a_proof_over_a_stealth_resource_can_be_created_and_dropped`
- `fungible.rs::minting_past_the_maximum_is_rejected_while_part_of_the_vault_is_locked`
- `fungible.rs::a_proof_over_a_confidential_vault_preserves_the_balance` (reports `80` where the vault holds `40`)

Each of the last two was also verified against the rest of this branch with only its own fix reverted, so neither passes by way of the other.

All 43 `tari_engine` test binaries plus `tari_engine_types` and `tari_ootle_common_types` pass.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12021,6 +12021,7 @@ dependencies = [
  "thiserror 2.0.18",
  "wasmer",
  "wasmer-middlewares",
+ "wat",
 ]
 
 [[package]]
--- a/crates/common_types/src/engine_signature.rs
+++ b/crates/common_types/src/engine_signature.rs
@@ -50,10 +50,14 @@ impl RistrettoSchnorrBlake2bVerifier {
 
 impl Verifier for RistrettoSchnorrBlake2bVerifier {
     fn verify(&self, domain: &[u8], message: &[u8], public_key: &PublicKey, signature: &SignaturePayload) -> bool {
-        let sig = signature
-            .ristretto_schnorr_blake2b()
-            .expect("Expected Ristretto Schnorr signature");
-        let ristretto_public_key = public_key.ristretto25519().expect("Expected Ristretto PublicKey");
+        // A caller chooses both the key and the payload, and `PublicKey`/`SignaturePayload` are open enums whose
+        // other variants are valid encodings, so a variant this verifier cannot use is a failed verification.
+        let Some(sig) = signature.ristretto_schnorr_blake2b() else {
+            return false;
+        };
+        let Some(ristretto_public_key) = public_key.ristretto25519() else {
+            return false;
+        };
 
         let Ok(pk) = ristretto_public_key.try_from_byte_type() else {
             return false;
--- a/crates/engine/Cargo.toml
+++ b/crates/engine/Cargo.toml
@@ -69,6 +69,7 @@ criterion = { version = "0.8.1", features = ["html_reports"] }
 minicbor = { workspace = true, default-features = false, features = ["alloc", "derive"] }
 serde_json = { workspace = true }
 tempfile = { workspace = true }
+wat = "1.258"
 
 
 [[bench]]
--- a/crates/engine/src/runtime/working_state.rs
+++ b/crates/engine/src/runtime/working_state.rs
@@ -1925,11 +1925,16 @@ impl<TStore: StateReader> WorkingState<TStore> {
         let mut total_fee_overcharge = 0;
         // First collect fees that cannot be refunded (we have to take all fees even if they exceed the required amount)
         for resx in self.fee_state.non_refundable_fee_payments_mut_iter() {
-            // PANIC: this is checked by FeeState
             let paid_amount = resx
                 .unlocked_amount()
                 .to_u64_checked()
-                .expect("invalid fee entry in fee state");
+                .ok_or_else(|| RuntimeError::InvariantError {
+                    function: "finalize_fees_and_refunds",
+                    details: format!(
+                        "Non-refundable fee payment {} does not fit in a u64",
+                        resx.unlocked_amount()
+                    ),
+                })?;
 
             debug!(
                 target: LOG_TARGET,
@@ -1956,11 +1961,16 @@ impl<TStore: StateReader> WorkingState<TStore> {
                     "Collecting {} of refundable fees", resx.unlocked_amount()
                 );
 
-                // PANIC: this is checked by FeeState
-                let paid_amount = resx
-                    .unlocked_amount()
-                    .to_u64_checked()
-                    .expect("invalid fee entry in fee state");
+                let paid_amount =
+                    resx.unlocked_amount()
+                        .to_u64_checked()
+                        .ok_or_else(|| RuntimeError::InvariantError {
+                            function: "finalize_fees_and_refunds",
+                            details: format!(
+                                "Refundable fee payment {} does not fit in a u64",
+                                resx.unlocked_amount()
+                            ),
+                        })?;
 
                 // Withdraw only what is needed
                 let amount_to_withdraw = cmp::min(paid_amount, remaining_fees);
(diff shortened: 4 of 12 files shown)


sdbondi (commented):
_Posted by Claude._

Reviewed at 34010f9. Traced each panic site into its callers and checked the fix against the surrounding invariants.

## Sound

- `RistrettoSchnorrBlake2bVerifier::verify` (`crates/common_types/src/engine_signature.rs:53-60`) — both `let ... else` arms return `false`, which is the only correct answer for a variant this verifier cannot consume; the rest of the function already failed closed the same way.
- `load_template_def` (`crates/engine/src/wasm/environment.rs:262-268`) — `module.rs:148` does run before `validate_instance` at `module.rs:153`, so the ordering claim in the description holds, and `WasmExecutionError` maps to `TransactionErrorKind`'s `"Load template error: {0}"` (`crates/engine/src/transaction/error.rs:115`) which the test asserts on.
- `lock_all`'s `Stealth` arm (`crates/engine_types/src/resource_container.rs:692`) — `unlock` type-checks the container before it matches (`:706`), so the old `public_fungible` return could never round-trip. `Self::stealth(address, amount)` produces `locked_amount: Amount::zero()` and `unlocked_amount()` reads `revealed_amount`, which is exactly what the `Stealth` unlock arm at `:801` consumes. `:638` (the `Fungible` arm) was already correct; only the second site was wrong.
- The `finalize_fees_and_refunds` conversions — `RuntimeError::InvariantError { function, details }` matches `crates/engine/src/runtime/error.rs:326`, and the merged `.and_then(as_vault_mut)` still distinguishes nothing the caller acts on. Propagating instead of aborting is strictly better here; `tracker.rs:530` was already fallible.
- No crate version bumps are outstanding: `scripts/crate_versioning.py list` reports `tari_engine_types` 0.41.0 and `tari_ootle_common_types` 0.41.0 both `pending`, so the new `ResourceError::BalanceOverflow` variant rides the unreleased minor.

## Blocking

**`deposit` bounds only the unlocked field, so the lock paths can still abort a validator.**
`crates/engine_types/src/resource_container.rs:283`, `:319`, `:329`

`checked_add` is applied to `amount` / `revealed_amount` alone, but a container's balance is `unlocked + locked`, held in two fields. The lock paths that move value between them are still raw `+=` on a `u128` (`Amount`'s `AddAssign` at `crates/template_lib_types/src/amount/ops.rs:14-18` delegates straight to `u128::add_assign`, so it panics under `overflow-checks = true`):

- `lock_all`: `:637`, `:670`, `:691`
- `lock_by_amount`: `:866`, `:912`, `:930`

That is reachable with the same primitive this PR's own test uses. Take the `Fungible` test template: create a proof over the confidential vault first (`lock_by_amount` moves some `n > 0` into `locked_revealed_amount`, leaving `revealed_amount == 0`), then `confidential_mint_more(mint_revealed(Amount::MAX))` — `deposit` at `:319` sees `0 + MAX` and accepts. The vault now holds `MAX + n`. The next `create_proof_for_resource` on that vault hits `:912` with `n + MAX` and aborts the executor. Same class of kill the PR is closing, one step further along.

The narrow fix is `checked_add` on the six lock sites too, but the sturdier one is to bound `deposit` against the container's total (`unlocked_amount() + locked_amount()`); then the lock adds are provably in range because they only redistribute a total that is already `<= Amount::MAX`, and the invariant lives in one place rather than six.

## Non-blocking

**The PR description's justification for the `FeeBreakdown::add` saturation is wrong.**
`crates/engine_types/src/fees.rs:423-426`

"affordability is decided elsewhere" — it is not. `FeeState::is_paid_in_full` (`crates/engine/src/runtime/fee_state.rs:126`) compares `total_payments()` against `total_charges()`, which is `fee_charges.get_total()` (`:129-131`), i.e. the breakdown. Saturating is still the right call, but for a different reason: saturation can only ever pin the total at `u64::MAX`, and no `u64` payment can meet that, so an overflowing breakdown rejects rather than letting a submitter under-pay. Worth correcting in the body, since the stated reason is the thing a future reader will rely on when touching `get_total`.

## Nits

**`MemoryPointerOutOfRange { size: 0, .. }`** — `crates/engine/src/wasm/environment.rs:264`. The variant renders as `"memory pointer out of range: memory size is {size}, ..."` (`crates/engine/src/wasm/error.rs:43`), so this reports a memory size of 0 that is not the instance's actual size. Every other construction of this variant in the file (`:271`, `:300`, `:311`, `:320`) passes the real `view.data_size()`. Either pass it here too, or use a variant that does not claim a size it does not have.

**The branch is ~40 commits behind `development`** (merge-base `c7d59d006`), including #2577, which also touches proof access on vaults. Worth a rebase before merge so the stealth-proof test runs against the scope checks it now has to coexist with.

PR #2580, opened 2026-09-10
refactor(consensus)!: remove epoch_end_spread_blocks

### LINKED ISSUE #2568: Remove epoch_end_spread_blocks: dead since #2164's next-epoch hash ratification

sdbondi:
Closes #2568.

## Why it is dead

Voting for an `EndEpoch` block needs two things: `can_propose_epoch_end`, and this node's own view of the next epoch's boundary hash to ratify against. The spread relaxed the first. The second was `get_epoch_hash(current_epoch + 1)`, whose row is written by `EpochManager::insert_current_epoch` — the same call that does `set_current_epoch` — so it asked the same question as `em_epoch > current_epoch` and terminally no-voted with `EndOfEpochHashNotObserved` when it failed. The spread's disjunct could never be the one that mattered.

Leeway voting landed in #2051, hash ratification in #2164, and #2164 subsumed it. On a validator it was doubly dead: the `EpochOracle` enum dispatcher never forwarded `is_within_epoch_end_spread`, so every variant fell through to the default `false` and the configured value never reached the base layer oracle.

Separately, mainnet's `10` becomes exactly one whole epoch once the L1 shortens `vn_epoch_length` to 10, at which point the predicate is unconditionally true — a symptom of the value never having been exercised.

## Why now

#2572 gave `can_propose_epoch_end` a term that does what the spread was reaching for, and that ratification can honour: a node that has scanned the boundary block votes on the strength of that observation, whether or not it has applied the `EpochChanged` that activates the epoch. The leeway is now sized by what the node can actually verify instead of by a block count, so the knob has nothing left to express.

## Change

Removes the constant from all four network profiles and the field from `ConsensusConstants`, the `||` disjunct in `on_receive_local_proposal.rs`, `BaseLayerEpochOracleConfig::epoch_end_spread_blocks`, `EpochEventOracle::is_within_epoch_end_spread` and its base-layer and hybrid implementations, `EpochManagerReader::is_within_epoch_end_spread`, `EpochManagerRequest::IsWithinEpochEndSpread` and its service handler, the `epoch_metrics` pass-through, and the test-harness and bootstrap wiring.

It also comes out of `ConsensusConstantsFile`. That struct is `deny_unknown_fields`, so a devnet constants file that still sets the field now fails to load rather than being accepted and ignored — devnets author this file per network, so a stale one gets edited rather than migrated.

Marked `!` for that and because `ConsensusConstants` is public, though no consensus behaviour changes — the removed branch could not alter a vote.

## Test plan

- [x] `cargo test -p consensus_tests epoch_change` — 7 passed, including both wedge tests and `epoch_change_observed_boundary_is_voted_before_activation`, which is what now covers the case the spread was written for
- [x] `cargo lints clippy --all-targets` across consensus, epoch_manager, epoch_oracles (with `base_layer`), consensus_tests, validator_node, indexer, app_utilities, swarm_daemon
- [x] CI

### DIFF
--- a/applications/tari_indexer/src/bootstrap.rs
+++ b/applications/tari_indexer/src/bootstrap.rs
@@ -563,7 +563,6 @@ async fn create_base_layer_epoch_oracle<TStore: EpochOracleStore + BaseLayerBloc
                 sync_headers: false,
                 sync_validator_node_changes: true,
             },
-            epoch_end_spread_blocks: consensus_constants.epoch_end_spread_blocks,
         },
         config.network,
     ))
--- a/applications/tari_ootle_app_utilities/src/consensus_constants_file.rs
+++ b/applications/tari_ootle_app_utilities/src/consensus_constants_file.rs
@@ -7,13 +7,10 @@ use std::{
     time::Duration,
 };
 
-use log::warn;
 use serde::{Deserialize, Serialize};
 use tari_consensus::consensus_constants::ConsensusConstants;
 use tari_engine_types::fees::{ExhaustBurnRate, MAX_EXHAUST_BURN_RATE_BPS};
 use tari_ootle_transaction::Network;
-
-const LOG_TARGET: &str = "tari::ootle::consensus_constants_file";
 
 /// Overrides for a LocalNet's consensus constants, applied over the network's built-in values.
 ///
@@ -41,10 +38,6 @@ pub struct ConsensusConstantsFile {
     pub max_block_validation_execution_points: Option<u64>,
     pub exhaust_burn_rate_bps: Option<u16>,
     pub max_transaction_validity_epochs: Option<u64>,
-    /// Accepted and ignored, so that an existing constants file that sets it still loads:
-    /// `deny_unknown_fields` would otherwise reject the whole file.
-    #[serde(skip_serializing)]
-    pub epoch_end_spread_blocks: Option<u64>,
 }
 
 impl ConsensusConstantsFile {
@@ -72,13 +65,6 @@ impl ConsensusConstantsFile {
             max_block_validation_execution_points,
             max_transaction_validity_epochs,
         );
-
-        if self.epoch_end_spread_blocks.is_some() {
-            warn!(
-                target: LOG_TARGET,
-                "⚠️ 'epoch_end_spread_blocks' in the consensus constants file is ignored and can be removed."
-            );
-        }
 
         if let Some(secs) = self.pacemaker_block_time_secs {
             constants.pacemaker_block_time = Duration::from_secs(secs);
--- a/applications/tari_ootle_app_utilities/src/consensus_constants_file.rs
+++ b/applications/tari_ootle_app_utilities/src/consensus_constants_file.rs
@@ -12,6 +12,8 @@ use tari_consensus::consensus_constants::ConsensusConstants;
 use tari_engine_types::fees::{ExhaustBurnRate, MAX_EXHAUST_BURN_RATE_BPS};
 use tari_ootle_transaction::Network;
 
+const LOG_TARGET: &str = "tari::ootle::consensus_constants_file";
+
 /// Overrides for a LocalNet's consensus constants, applied over the network's built-in values.
 ///
 /// Consensus constants have to agree across a network, so this is authored once and handed to every
@@ -38,6 +40,11 @@ pub struct ConsensusConstantsFile {
     pub max_block_validation_execution_points: Option<u64>,
     pub exhaust_burn_rate_bps: Option<u16>,
     pub max_transaction_validity_epochs: Option<u64>,
+    /// Accepted and ignored. The leeway this configured could never withhold or grant a vote:
+    /// ratifying an `EndEpoch` independently requires having observed the next epoch's boundary.
+    /// The field is kept because `deny_unknown_fields` would otherwise reject a whole constants
+    /// file that still sets it.
+    #[serde(skip_serializing)]
     pub epoch_end_spread_blocks: Option<u64>,
 }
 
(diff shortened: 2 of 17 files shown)


sdbondi (commented):
Reviewed at 3c7a9cf. No blocking findings.

Verified the deadness argument rather than taking it from the description:

- `on_ready_to_vote_on_local_block.rs:444-468` is the real gate — an `EndEpoch` command no-votes with `EndOfEpochHashNotObserved` whenever `expected_next_epoch_hash` is `None`. Since `expected_next_epoch_hash.is_some()` is already a disjunct of `can_propose_epoch_end`, the removed third disjunct could only ever move a no-vote from `NotEndOfEpoch` to `EndOfEpochHashNotObserved`. No vote outcome changes.
- `can_propose_epoch_end` has no other consumer: the only other use is the `|= justified_block.is_epoch_end()` majority signal at `on_ready_to_vote_on_local_block.rs:165`, which is deliberately not allowed to bypass the local-oracle ratification.
- Confirmed on `development` that nothing forwarded `is_within_epoch_end_spread` through the `EpochOracle` dispatcher, so on a validator the base-layer implementation was unreachable regardless.

Removal is complete: no references to `epoch_end_spread` or `IsWithinEpochEndSpread` survive anywhere in the tree, including toml/md/ts. `last_scanned_height`, `epoch_length` and the `Ordering` import in `base_layer/oracle.rs` all still have other users, and both bootstrap sites still use `consensus_constants` for `base_layer_confirmations`, so nothing is left dangling.

The `ConsensusConstantsFile` treatment is the right call — `deny_unknown_fields` would harden a stale devnet file into a load failure, and `apply_to` is only reached on LocalNet, so the warning fires exactly where a file could still set it (`consensus_constants_file.rs:76-82`). `#[serde(skip_serializing)]` keeps it out of anything the struct writes back.

Nits, take or leave:

- `consensus_constants_file.rs:44-48`: the doc comment justifies the field by what the removed knob could not do. A future reader of this file has no removed knob to contrast against — "Accepted and ignored so that an existing constants file that sets it still loads; `deny_unknown_fields` would otherwise reject the whole file." says everything they need.
- `consensus_constants_file.rs:76`: the crate imports nothing from `log` today; `use log::warn;` alongside the other imports would match how the consensus and epoch_manager crates call it, but the fully-qualified form is fine for one call site.

Holding approval only until `integration test` finishes; every other check is green.

sdbondi (commented):
All changes expected

PR #2582, opened 2026-09-10
fix(consensus): resume a deferred end-of-epoch on the periodic tick

### LINKED ISSUE #2581: Deferred end-of-epoch resume waits for the whole base-layer catch-up, not for the epoch it needs
A validator that defers its end-of-epoch processing waits for a signal that only arrives when its entire base-layer catch-up finishes, even though the data it is waiting for landed much earlier. It sits out the new epoch for the remainder of the scan.

This was noted during the review of #2572 and deliberately left out of scope there.

## The chain

1. `process_end_of_epoch` (`crates/consensus/src/hotstuff/on_receive_local_proposal.rs:539-557`) needs `get_epoch_hash(next_epoch)` to assign the next epoch's committee. If the local oracle has not activated that epoch yet, it stores a `PendingEndOfEpoch` and returns — no next-epoch genesis, no `lock_epoch`.
2. The retry is `try_resume_pending_end_of_epoch`, called at commit time (`worker.rs:749`) and from `on_epoch_manager_event` on `EpochManagerEvent::EpochChanged` (`worker.rs:786`).
3. That event is published from exactly one place: `on_scanning_complete` (`epoch_manager/src/service/epoch_manager_service.rs:342`).
4. `on_scanning_complete` runs on `EpochEvent::DoneForNow`, which the base-layer oracle emits only when the scan has reached the lagged tip (`epoch_oracles/src/base_layer/oracle.rs:536`) or when there was no progress at all (`:242`). A multi-batch catch-up emits nothing until the last batch.

The mismatch is that the *data* the deferral waits on is written far earlier than the *notification* that triggers the retry. `activate_epoch` -> `insert_current_epoch` writes the `epochs` row as soon as the queued `EpochChanged` for that epoch is applied, which during a catch-up happens between batches. The row is there; nothing tells consensus to look.

## Impact

Since #2572 a lagging node votes for the `EndEpoch` off an observed boundary, so it now commits the EOE promptly — which makes this the normal path rather than a rare one, and moves the cost from the committee to the node itself. Between committing the EOE and the end of its catch-up the node has no genesis in the new epoch: it cannot propose or vote there, and it is not on the leaf its peers are building on, despite being current on consensus and holding the epoch row it needs.

Batches are up to 1000 headers, so the wait scales with how far behind the scanner is, not with how far behind consensus is.

## Options

- **Retry on the data, not the event.** The resume condition is a single DB read. Re-checking a pending EOE on the existing block cadence (or on each applied `EpochChanged`, before `on_scanning_complete`) would resume within one batch instead of one catch-up.
- **Publish the epoch activation separately from scan completion.** `activate_epoch` knows the epoch became available; `on_scanning_complete` currently batches that into one event per scan via `has_epoch_changed`. A narrower "epoch N is now available" signal would let consensus react without waiting for `DoneForNow`. Worth weighing against the churn during a long catch-up, where each activation already reassigns committees and clears the committee cache.

The first is the smaller change and needs no new event.

## Related

Deferring `lock_epoch(next_epoch)` to the resume also widens the window in which `activate_epoch`'s self-healing hash correction could store a hash other than the one the committee committed and the genesis will be stamped with. It needs a reorg landing between the queued activation and the resume, so it does not look reachable in practice, but shortening the deferral shortens that window too.

sdbondi:
Fixes #2581.

## Problem

`process_end_of_epoch` defers when the local oracle has not yet observed `next_epoch`, and the retry was driven by `EpochManagerEvent::EpochChanged`. That event is published from exactly one place — `on_scanning_complete`, on `EpochEvent::DoneForNow` — which the base-layer oracle emits only when the scan reaches the lagged tip. A multi-batch catch-up emits nothing until the last batch.

The data the deferral waits on is written far earlier: `activate_epoch` writes the epoch row, assigns the validator set and clears the committee cache as soon as the queued `EpochChanged` for that epoch is applied, which during a catch-up happens between batches. The row is there; nothing told consensus to look.

Since #2572 a lagging node votes for the `EndEpoch` off an observed boundary and commits the EOE promptly, making this the normal path. Between committing the EOE and finishing its scan the node has no genesis in the new epoch — it cannot propose or vote there, and is not on the leaf its peers are building on. The wait scales with the scanner's lag (batches are up to 1000 headers), not with consensus lag.

## Change

Retry on the data, not the event (option 1 in the issue): the worker polls `try_resume_pending_end_of_epoch` on its existing 10s periodic tick, so a deferred EOE resumes within one batch instead of one catch-up. The `EpochChanged` and startup retries are unchanged; the shared check/log/propagate is factored into one worker helper.

No new event, no epoch-manager change. Resume errors (`NeedsSync`, `NotRegisteredForCurrentEpoch`) propagate from the tick exactly as they did from the event arm.

### DIFF
--- a/crates/consensus/src/hotstuff/on_receive_local_proposal.rs
+++ b/crates/consensus/src/hotstuff/on_receive_local_proposal.rs
@@ -85,15 +85,6 @@ pub struct OnReceiveLocalProposalHandler<TConsensusSpec: ConsensusSpec> {
     pending_end_of_epoch: Option<PendingEndOfEpoch>,
 }
 
-/// Whether `process_end_of_epoch` was entered from the commit that produced the EOE or from a
-/// retry of a deferral. Only the caller can tell: the retry takes the pending state before
-/// re-entering, so the callee sees an empty slot either way.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-enum IsRetry {
-    No,
-    Yes,
-}
-
 #[derive(Debug, Clone)]
 struct PendingEndOfEpoch {
     eoe_block: Block,
@@ -500,7 +491,7 @@ impl<TConsensusSpec: ConsensusSpec> OnReceiveLocalProposalHandler<TConsensusSpec
             // genesis state.
             let commit_qc = valid_block.justify().clone();
             let next_genesis_state_merkle_root = *valid_block.block().state_merkle_root();
-            self.process_end_of_epoch(eoe_block, commit_qc, next_genesis_state_merkle_root, IsRetry::No)
+            self.process_end_of_epoch(eoe_block, commit_qc, next_genesis_state_merkle_root)
                 .await?;
         }
 
@@ -520,7 +511,6 @@ impl<TConsensusSpec: ConsensusSpec> OnReceiveLocalProposalHandler<TConsensusSpec
         eoe_block: Block,
         commit_qc: ProposalCertificate,
         next_genesis_state_merkle_root: FixedHash,
-        is_retry: IsRetry,
     ) -> Result<(), HotStuffError> {
         let _timer = TraceTimer::debug(LOG_TARGET, "process-end-of-epoch");
         let prev_epoch = eoe_block.epoch();
@@ -553,23 +543,12 @@ impl<TConsensusSpec: ConsensusSpec> OnReceiveLocalProposalHandler<TConsensusSpec
             .optional()?
             .is_none()
         {
-            // Only the caller knows whether this is the first defer or a retry finding the oracle
-            // still behind — the pending state is taken before the retry re-enters here, so it
-            // cannot be read off `self`. Retries run on a tick for the length of the catch-up, so
-            // only the first defer is news.
-            match is_retry {
-                IsRetry::No => warn!(
-                    target: LOG_TARGET,
-                    "⏳ EOE block {} committed but local oracle has not observed {next_epoch}. \
-                     Deferring next-epoch genesis until oracle catches up.",
-                    eoe_block.id()
-                ),
-                IsRetry::Yes => debug!(
-                    target: LOG_TARGET,
-                    "⏳ EOE block {} still deferred: local oracle has not observed {next_epoch}.",
-                    eoe_block.id()
-                ),
-            }
+            warn!(
+                target: LOG_TARGET,
+                "⏳ EOE block {} committed but local oracle has not observed {next_epoch}. \
+                 Deferring next-epoch genesis until oracle catches up.",
+                eoe_block.id()
+            );
             self.pending_end_of_epoch = Some(PendingEndOfEpoch {
                 eoe_block,
                 commit_qc,
@@ -716,7 +695,6 @@ impl<TConsensusSpec: ConsensusSpec> OnReceiveLocalProposalHandler<TConsensusSpec
             pending.eoe_block,
             pending.commit_qc,
             pending.next_genesis_state_merkle_root,
-            IsRetry::Yes,
         )
         .await?;
         Ok(true)
--- a/crates/consensus/src/hotstuff/on_receive_local_proposal.rs
+++ b/crates/consensus/src/hotstuff/on_receive_local_proposal.rs
@@ -534,8 +534,8 @@ impl<TConsensusSpec: ConsensusSpec> OnReceiveLocalProposalHandler<TConsensusSpec
 
         // We still need the local oracle to have observed `next_epoch` to assign its committee /
         // validator set. If it has not yet (rare, given the base-layer scan lag keeps the boundary
-        // block buried), defer; `try_resume_pending_end_of_epoch` retries from `on_epoch_manager_event`
-        // or worker startup.
+        // block buried), defer; the worker retries `try_resume_pending_end_of_epoch` on its periodic
+        // tick, on `EpochManagerEvent::EpochChanged` and at startup.
         if self
             .epoch_manager
             .get_epoch_hash(next_epoch)
@@ -677,9 +677,9 @@ impl<TConsensusSpec: ConsensusSpec> OnReceiveLocalProposalHandler<TConsensusSpec
         self.pending_end_of_epoch.is_some()
     }
 
-    /// Retry deferred end-of-epoch processing. Called by the worker when the local oracle
-    /// observes the new epoch (via `EpochManagerEvent::EpochChanged`). If the oracle is still
-    /// not ready, `process_end_of_epoch` will re-defer.
+    /// Retry deferred end-of-epoch processing. The worker polls this so that the resume happens
+    /// as soon as the oracle has written the epoch row, without waiting for the base-layer scan
+    /// to finish. If the oracle is still not ready, `process_end_of_epoch` will re-defer.
     pub async fn try_resume_pending_end_of_epoch(&mut self) -> Result<bool, HotStuffError> {
         let Some(pending) = self.pending_end_of_epoch.take() else {
             return Ok(false);
--- a/crates/consensus/src/hotstuff/on_receive_local_proposal.rs
+++ b/crates/consensus/src/hotstuff/on_receive_local_proposal.rs
@@ -85,6 +85,15 @@ pub struct OnReceiveLocalProposalHandler<TConsensusSpec: ConsensusSpec> {
     pending_end_of_epoch: Option<PendingEndOfEpoch>,
 }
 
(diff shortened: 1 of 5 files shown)


sdbondi (commented):
Makes sense, waiting on regression test

sdbondi (commented):
Reviewed at e573e80. Nothing blocking.

The invariant the change rests on holds, and it is worth stating because polling the row is only safe if it does: `activate_epoch` writes the epoch row, runs `assign_validators_for_epoch` and clears the committee cache in one awaited call inside the epoch manager's event branch (`epoch_manager_service.rs:298-305`), and the service handles requests only between events — so no `GetEpochHash` can observe the row without the assignment behind it. That matters because `process_end_of_epoch` reaches for `get_num_committees(next_epoch)` and `get_our_validator_node(next_epoch)` immediately after the check it polls on, and those read the assignment.

Also verified: the re-defer round-trips (the handler `take`s the pending state and the defer branch re-stores it, `on_receive_local_proposal.rs:556`); the tick path not setting `next_epoch` is harmless, since `next_epoch` only gates `can_propose_epoch_end` (`worker.rs:1012`), i.e. proposing the *following* EOE, which needs an event regardless; and the description's claim about error propagation is accurate — the arm this mirrors (`Ok(event) = epoch_manager_events.recv()`, `worker.rs:466`) is also a bare `?`, so `NeedsSync` and `NotRegisteredForCurrentEpoch` reach `Running::on_enter` the same way. Both differ from the other fallible arms in skipping `on_failure`, but that is pre-existing and consistent with what was copied.

I traced the consensus test against the base branch rather than trusting its docstring, and it does fail without the change: the harness publishes `EpochManagerEvent::EpochChanged` only from `set_current_epoch` (`support/epoch_manager.rs:157`), `clear_oracle_visible_epoch` publishes nothing, and taking the node offline before clearing closes the probe→`NeedsSync`→sync route that would otherwise decide the outcome. Without the tick there is no third retry, so `wait_for_leaf_at_epoch` times out. The harness `get_observed_epoch_hash` mirrors production precedence correctly.

## Non-blocking

**1. Both log lines were written for a one-shot and are now a poll.** `on_receive_local_proposal.rs:549` warns "⏳ EOE block … Deferring next-epoch genesis until oracle catches up" and `:692` logs "▶️ Resuming deferred end-of-epoch processing" at info. With the retry on a 10s tick, a node deferring across a long catch-up emits both every 10 seconds for the whole scan — roughly 180 warn-level lines for a 30-minute catch-up, all identical, during exactly the incident an operator would be reading the log to understand. Worth logging the defer only when the pending state is newly set (it is `None` on entry to the defer branch in the first-defer case and `Some` on a re-defer, so the distinction is already available), and dropping the repeat resume line to debug.

**2. The tick arm resumes before it ticks.** `worker.rs:521` runs the resume, then `:523` runs `on_task_tick(&epoch_state)` with the `epoch_state` captured at the top of the iteration. A successful resume ends in `pacemaker.set_epoch(next_epoch)` (`on_receive_local_proposal.rs:641`), so that `on_task_tick` re-requests timed-out foreign proposals against `local_committee_info()` for the epoch the node has just left; `epoch_state` is only refreshed at the top of the next iteration (`worker.rs:427`). Swapping the two statements avoids it for nothing.

## Nit

The new oracle test's doc (`epoch_oracles/src/base_layer/oracle.rs:1431-1439`) explains both halves of the ordering — the oracle's per-batch `EpochChanged` and the epoch manager publishing only from `on_scanning_complete` — but the test asserts only the first. That is the half that lives in this crate, so it is the right scope; the second half is context, and someone reading the assertions will not find it guarded here. Worth a word saying so, since it is the half a future change would most likely move.

sdbondi (commented):
_Posted by Claude._

Re-reviewed 3b82ad3. The tick reordering and the test-doc note are right; the log change does not do what it says.

## Blocking

**The re-defer branch is unreachable, so the warn still fires every tick.** `on_receive_local_proposal.rs:551` distinguishes first defer from re-defer with `if self.pending_end_of_epoch.is_none()`, but the retry path clears that field before it gets there: `try_resume_pending_end_of_epoch` does `let Some(pending) = self.pending_end_of_epoch.take()` (`:697`) and only then calls `process_end_of_epoch`. So the field is `None` on every entry to the defer branch — the commit path (nothing pending yet) and the retry path (just taken) alike. `warn!` fires on all of them and the `debug!` arm at `:558` is dead code.

Grepped every mutation of the field to be sure there is no third path: `:138` init, `:565` the defer branch itself, `:682` `set_pending_end_of_epoch`, `:697` the take. The startup recovery route also ends in a `take()` before `process_end_of_epoch` runs, so it lands the same way.

The state that distinguishes the two cases lives in the caller, not the callee. Either pass it down (`process_end_of_epoch(..., is_retry: bool)`), or move the log out to the two call sites — the commit path warns, `try_resume_pending_end_of_epoch` debugs — which also removes the need for the callee to reason about how it was reached.

## Verified

`worker.rs:517-528` now runs `on_task_tick` before the resume, so a resume that ends in `pacemaker.set_epoch` no longer leaves the same tick re-requesting foreign proposals against the committee of the epoch just left. The added comment states the reason rather than the change.

`try_resume_pending_end_of_epoch`'s own line is now `debug!` and reworded to "Attempting", which is accurate — it logs before knowing whether the resume will succeed or re-defer.

## Nit

The new paragraph at `epoch_oracles/src/base_layer/oracle.rs:1440-1444` says the epoch manager's publish-on-`DoneForNow` side is "guarded by `epoch_change_deferred_eoe_resumes_without_scan_completion` in `consensus_tests`". That test runs against `TestEpochManager`, which publishes only from `set_current_epoch` and never exercises `EpochManagerService::on_scanning_complete` — so it guards the consequence (a deferred EOE resumes without a scan-completion event) but not the publishing policy. Change `on_scanning_complete` to publish per activation and no test in either crate fails. Worth saying which of the two it pins.

crates/consensus/src/hotstuff/on_receive_local_proposal.rs
```diff
@@ -546,21 +556,22 @@ impl<TConsensusSpec: ConsensusSpec> OnReceiveLocalProposalHandler<TConsensusSpec
...
         {
-            // A re-defer means a retry found the oracle still behind, which happens on every tick for
-            // the length of the catch-up. Only the first defer is news; the rest are the same line.
-            if self.pending_end_of_epoch.is_none() {
-                warn!(
+            // Only the caller knows whether this is the first defer or a retry finding the oracle
+            // still behind — the pending state is taken before the retry re-enters here, so it
+            // cannot be read off `self`. Retries run on a tick for the length of the catch-up, so
+            // only the first defer is news.
+            match is_retry {
```
sdbondi: New single use enum just for a log level? just keep the warn on retries

sdbondi: _Posted by Claude._

Agreed, and already reverted in 4cb3c247b (force-pushed after this thread was written, so it is against the stale commit). The defer branch is back to the unconditional `warn!` and the enum is gone. A node deferring across a long catch-up repeats the warning, which is fine — it is an ongoing degraded state.

Only the line that needed no plumbing kept its change: `try_resume_pending_end_of_epoch`'s own line is `debug!` and says "Attempting", since it fires before the outcome is known.

sdbondi (commented):
_Posted by Claude._

Re-reviewed 4cb3c24. Nothing outstanding from me.

Dropping `IsRetry` resolves what I raised, just not the way I proposed, and the reasoning holds: a node deferring across a catch-up is in an ongoing degraded state, and repeating the warning is a reasonable way to say so. What I actually objected to was the dead `debug!` arm and a commit claiming a quietening it did not deliver — both are gone, and the code no longer claims anything it does not do. The per-attempt line stays at `debug!`, which was the half that needed no plumbing. `use log::*` in that file, so dropping the `info!` leaves no unused import.

I checked the force-push rather than assuming it was a clean rebase: against the current base (`77f172a`) the branch is exactly its five commits over five files, with none of `CLAUDE.md`, `SECURITY.md` or the other newly-merged base content appearing in the PR's own diff. The production hunks are byte-identical to what I reviewed at `e573e80` and `3b82ad3`, so the earlier verification still stands — in particular the invariant the change rests on: `activate_epoch` writes the epoch row, assigns validators and clears the committee cache in one awaited call, so a poll that sees the row cannot race ahead of the assignment `process_end_of_epoch` needs immediately after.

CI on this head: everything green except `integration test`, still running. All three `test` shards pass, which is where `epoch_change_deferred_eoe_resumes_without_scan_completion` runs.

I cannot approve my own PR (GitHub refuses `addPullRequestReview` with APPROVE), so this needs another maintainer.

PR #2585, opened 2026-09-10
fix(consensus): drop and reject a foreign proposal that fails to process while proposing

sdbondi:
_Posted by Claude._

Raised in review of #2573 and confirmed by a consensus audit. Third and last instance of the class #2573 and #2574 closed: the proposer committing to state a replica will not reproduce.

## Problem

`Command::ForeignProposal` atoms are inserted into `commands` (`on_propose.rs:399-410`) *before* the proposals are processed. The failure arm logged and `continue`d:

```rust
warn!(target: LOG_TARGET, "Failed to process foreign proposal: {}. Not proposing...", err);
// TODO: should mark as invalid?
continue;
```

The atom had already shipped, so the log was untrue. Three consequences, each needing its own handling:

**The partial change set.** `process_foreign_block` writes transaction updates and pledges as it iterates (`add_transaction_execution`, `set_next_transaction_update`) and can return an error part-way. Before #2573 the drain discarded that; since #2573 the transaction loop reads it via `apply_transaction_update`, so half-applied state now shapes the emitted atoms.

**The atom.** The rejected proposal's command sits in a block whose commands were derived from state the block does not describe.

**The record.** This is the one that turns a wasted slot into a stall. `foreign_proposals_get_all_new` (`state_store_rocksdb/src/reader.rs:579-624`) excludes a proposal only when `ProposedInBlockIndex` has an entry against a block in the pending chain. That entry is written by `set_foreign_proposal_proposed_in`, which the replica calls only on success; the failure arm returns `NoVoteReason::ForeignProposalProcessingFailed`. And `ProposedBlockChangeSet::save` (`block_change_set.rs:442-454`) early-returns on `no_vote_reason`, writing the diagnostic and nothing else. The leader's own change set is never saved either.

So no block that fails this way records the proposal as proposed. It stays in the unconfirmed index, the next leader's batch selects it again, the same deterministic failure repeats, and another unvotable block ships. The record clears only when a block containing it commits — exactly what cannot happen. There is no self-clearing path, and the epoch guard does not provide one: `get_all_new` breaks only on `epoch > locked.epoch`, so an older-epoch proposal stays selectable across a rollover.

## Fix

- **Snapshot and restore.** Clone the change set before any foreign proposal is applied; restore it on failure so no partial state reaches the transaction commands. One clone per proposal, not per foreign proposal.
- **Drop the atoms.** Remove every `ForeignProposal` command when one is rejected. The restored change set carries none of their effects, so none may be proposed. They keep their `New` status and are proposed again next round.
- **Reject the record.** Return the rejected block ids to the caller — which holds the write transaction, the same pattern already used for `lock_conflicts` and `executed_transactions` — and mark them `Invalid`. `Invalid` is not `is_unconfirmed`, so `foreign_proposals_set_status` deletes the `UnconfirmedIndex` entry and the batch query stops returning it.

Only a **validation** failure condemns a proposal: that is a property of the proposal itself, so every replica reaches it too. A storage or epoch-manager error says nothing about the proposal and must not permanently discard a valid one, so it propagates and the next round retries. This mirrors `on_receive_local_proposal.rs:245-257`, which already discriminates on `err.validation_error()` for the same decision.

`process_foreign_block` only ever *reads* the substate store (`foreign_proposal_processor.rs:212` is its sole use), so no store rollback is needed.

## Scope

Marking `Invalid` is a local decision, so each leader that selects a bad proposal rejects it once. The stall is bounded by (bad proposals x committee size) instead of unbounded, since one leader condemns one proposal per round.

One narrower path stays open and is out of scope here: foreign proposals are only processed when `batch.transactions` is non-empty, so a leader with an empty batch still proposes them unprocessed and never learns they are bad. That block is still unvotable, but the loop breaks as soon as any leader has transactions in its batch — the same condition under which the stall was described. Closing it means processing proposals even with no transactions to propose, which changes what work a proposal costs and deserves its own diff.

## Testing

- `cargo test -p consensus_tests` — 68 passed, 0 failed.
- `cargo test -p tari_consensus --lib` — 22 passed, 0 failed.
- `cargo lints clippy -p tari_consensus -p tari_ootle_storage --all-targets` — clean.
- New test `a_rejected_proposal_leaves_the_unconfirmed_set` pins the property the third part depends on: `Invalid` must not be `is_unconfirmed`, since that is exactly what makes `foreign_proposals_set_status` delete the index entry that `get_all_new` iterates.

No end-to-end test. Reaching the failure arm needs a foreign proposal that passes receipt validation and then fails processing, which the harness has no way to inject; a test that faked it would exercise the fake and not the path. The unit test pins the storage property, and the rest of the change is verified by reading.

## Compatibility

Proposer-side only. Replica evaluation is untouched, no format or hash inputs move, and the only blocks whose content changes are ones no honest replica votes for. Mixed fleet needs no coordination.

### DIFF
--- a/crates/consensus/src/hotstuff/on_propose.rs
+++ b/crates/consensus/src/hotstuff/on_propose.rs
@@ -11,7 +11,15 @@ use std::{
 use log::*;
 use ootle_byte_type::ToByteType;
 use tari_common_types::types::FixedHash;
-use tari_consensus_types::{Decision, HighPc, HighestSeenBlock, LeafBlock, ProposalCertificate, TimeoutCertificate};
+use tari_consensus_types::{
+    BlockId,
+    Decision,
+    HighPc,
+    HighestSeenBlock,
+    LeafBlock,
+    ProposalCertificate,
+    TimeoutCertificate,
+};
 use tari_crypto::tari_utilities::epoch_time::EpochTime;
 use tari_engine_types::commit_result::RejectReason;
 use tari_epoch_manager::EpochManagerReader;
@@ -37,6 +45,7 @@ use tari_ootle_storage::{
         EvictNodeAtom,
         ForeignProposal,
         ForeignProposalRecord,
+        ForeignProposalStatus,
         LockedEpoch,
         PendingShardStateTreeDiff,
         TransactionAtom,
@@ -85,6 +94,9 @@ struct NextBlock {
     foreign_proposals: Vec<ForeignProposal>,
     executed_transactions: HashMap<TransactionId, TransactionExecution>,
     lock_conflicts: TransactionLockConflicts,
+    /// Foreign proposals this node rejected while proposing. Recorded by the caller, which holds the write
+    /// transaction, so that they leave the unconfirmed index and are not selected into a later block.
+    invalid_foreign_proposals: Vec<BlockId>,
 }
 
 #[derive(Debug, Clone)]
@@ -185,9 +197,21 @@ where TConsensusSpec: ConsensusSpec
                     foreign_proposals,
                     executed_transactions,
                     lock_conflicts,
+                    invalid_foreign_proposals,
                 } = next_block;
 
                 lock_conflicts.save_for_block(tx, next_block.id())?;
+
+                // Invalid leaves the unconfirmed index, so a proposal rejected here is not selected into a
+                // later block. Selection is what makes such a failure repeat: the batch query returns every
+                // proposal still in that index, and a validation failure is deterministic across leaders.
+                for block_id in invalid_foreign_proposals {
+                    warn!(
+                        target: LOG_TARGET,
+                        "⚠️ Marking foreign proposal {block_id} invalid: it failed validation while proposing"
+                    );
+                    ForeignProposalRecord::set_status_by_id(tx, &block_id, ForeignProposalStatus::Invalid, None)?;
+                }
 
                 // Add executions for this block
                 if !executed_transactions.is_empty() {
@@ -413,6 +437,7 @@ where TConsensusSpec: ConsensusSpec
 
         // NOTE: the block for the change set is not used.
         let mut change_set = ProposedBlockChangeSet::new(start_of_chain_block.as_leaf());
+        let mut invalid_foreign_proposals = Vec::new();
 
         // No need to include evidence from justified block if no transactions are included in the next block
         if !batch.transactions.is_empty() {
@@ -423,27 +448,49 @@ where TConsensusSpec: ConsensusSpec
             // TODO: we dont need to process transactions here that are not in the batch
             process_newly_justified_block(tx, &justify_block, high_qc_id, local_committee_info, &mut change_set)?;
 
+            // A failed `process_foreign_block` leaves whatever it wrote before the error behind, and the
+            // transaction commands below are derived from the change set, so that partial state would shape
+            // atoms no replica reproduces. Restoring this snapshot is what keeps a rejected foreign proposal
+            // from reaching them.
+            let change_set_before_foreign_proposals = change_set.clone();
+
             for fp in &batch.foreign_proposals {
                 // Resolves pending transaction pool records along the chain up to this block, so it must be
                 // the anchor the substate store this call also writes to is built on: the justify block
                 // under a dummy chain, the extended leaf otherwise. A replica passes the block it is
                 // evaluating, whose parent chain runs back through any dummies to the justify block.
-                if let Err(err) = process_foreign_block(
+                let Err(err) = process_foreign_block(
                     tx,
                     &state_anchor_leaf,
                     fp,
                     local_committee_info,
                     &mut substate_store,
                     &mut change_set,
-                ) {
-                    warn!(
-                        target: LOG_TARGET,
-                        "Failed to process foreign proposal: {}. Not proposing...",
-                        err
-                    );
-                    // TODO: should mark as invalid?
+                ) else {
                     continue;
-                }
+                };
+
+                // A validation failure is a property of the proposal itself, so every replica reaches it too
+                // and the block cannot be voted for. Anything else (storage, epoch manager) says nothing
+                // about the proposal and must not condemn it — abandon the proposal and let the next round
+                // retry.
+                let Some(validation_err) = err.validation_error() else {
+                    return Err(err);
+                };
+
+                warn!(
+                    target: LOG_TARGET,
+                    "⚠️❌ Foreign proposal {} failed validation while proposing: {validation_err}. Marking it \
+                     invalid and dropping every foreign proposal from this block.",
+                    fp.to_atom().block_id,
+                );
+
+                invalid_foreign_proposals.push(fp.to_atom().block_id);
+                change_set = change_set_before_foreign_proposals;
+                // The proposals already applied to the discarded change set are not re-applied, so none of
+                // them may be proposed here. They keep their `New` status and are proposed again next round.
+                commands.retain(|cmd| !matches!(cmd, Command::ForeignProposal(_)));
+                break;
             }
 
             // Add all (ABORT) executions that may have resulted from foreign proposals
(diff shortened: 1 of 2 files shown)


sdbondi (commented):
_Posted by Claude._

The three-part diagnosis holds up and I verified the load-bearing claims: `process_foreign_block` really does only read the substate store (`foreign_proposal_processor.rs:212` is its sole use), so no store rollback is needed; `ProposedBlockChangeSet` is `derive(Clone)` over owned collections, so the snapshot is a genuine deep copy; the snapshot is taken after `process_newly_justified_block`, and `BlockHeader::create_unsigned` hashes the post-`retain` `commands`, so the header and the emitted atoms agree with the restored change set; and the index mechanics are as described (`state_store_rocksdb/src/writer.rs:592-600` drops the `UnconfirmedIndex` entry when the status leaves the unconfirmed set, which is what `reader.rs:596-620` iterates).

One blocking problem, in the error discrimination rather than in the snapshot/restore.

## Blocking

### `return Err(err)` on a non-validation error puts the validator to sleep, and two of the reachable ones are decided by remote content

`on_propose.rs:477-479` propagates anything that is not a `ProposalValidationError`. Follow that error out: `build_next_block` → `on_propose::handle` → `worker.rs:472-475` calls `on_failure`, which stops the pacemaker and publishes `HotstuffEvent::Failure` → `state_machine/worker.rs:112-115` transitions to `ConsensusState::Sleeping`. This is not a skipped slot; consensus stops on this node.

Two errors reachable from `process_foreign_block` are not `ProposalValidationError` but *are* pure functions of the foreign proposal's content:

- `HotStuffError::ForeignNodeOmittedTransactionPledges` — `foreign_proposal_processor.rs:668`, raised when the foreign shard group's block pledge omits pledges the local stage requires.
- `HotStuffError::InvariantError` for a missing justify QC — `foreign_proposal_processor.rs:448-454` (and the sibling at `:176-182`), whose own message says "this should have been rejected by validations".

Before this diff both were `warn!` + `continue`. Now a foreign shard group that ships a proposal with omitted pledges — bug or malice — parks the local validator. Worse, it does not self-clear: the record is marked neither `Invalid` nor `Proposed`, so on Resume the batch selects it again and the node sleeps again. That is the same unbounded loop the PR sets out to close, with a strictly worse symptom than the one it replaces.

The precedent cited in the description does not carry the decision. `on_receive_local_proposal.rs:245-257` discriminates on `validation_error()` to choose between *no-voting* and propagating on a block that has already arrived. The propose-side analogue of "no-vote" is "do not ship the atom", not "leave consensus". And the replica path this proposal must stay symmetrical with, `on_ready_to_vote_on_local_block.rs:1597-1616`, deliberately no-votes on *every* error from `process_foreign_block`, with an explicit `TODO: split validation errors from HotStuff errors so that we can selectively crash or not vote` — the split this diff relies on does not exist yet in the error type.

Suggested shape: make the drop-and-restore unconditional on any error from `process_foreign_block` (that half is safe for every error class, and it is the part that actually keeps unvotable blocks off the wire), and gate only the `Invalid` marking on `err.validation_error()`. A proposal that hits `ForeignNodeOmittedTransactionPledges` then costs its foreign proposals one round instead of costing the node its consensus loop, and the block still ships and is still votable. If some error class genuinely must propagate, it should be an explicit allowlist of local-resource failures, not "everything that is not `ProposalValidationError`".

For what it is worth, `ForeignNodeOmittedTransactionPledges` looks misfiled — it is a statement about the proposal, so moving it under `ProposalValidationError` would make both this call site and the replica's future split fall out correctly. That is a larger change than this PR needs, so the allowlist is the cheaper fix here.

## Non-blocking

### The condemned proposal is still broadcast to the whole local committee

`on_propose.rs:695` returns `batch.foreign_proposals` unchanged, and `broadcast_local_proposal` (`:264`) attaches all of them to the `ProposalMessage`. Two consequences. First, the block now contains no `ForeignProposal` command referencing any of them, so they are dead weight on the wire. Second, on any replica that has not already recorded the proposal, `on_receive_local_proposal.rs:227-256` runs `validate_and_save` and stores it as `New` — receipt validation passes, since it is *processing* that fails — which puts it straight into that replica's unconfirmed index. The leader condemns the proposal locally and re-seeds it to the committee in the same message. It works against the committee-size bound the Scope section argues for. Returning only the proposals whose atoms survived in `commands` costs a filter.

### The snapshot clone is unconditional

`on_propose.rs:455` deep-copies the whole change set on every propose that has transactions, including the common case where `batch.foreign_proposals` is empty and nothing can ever restore it. The change set at that point holds whatever `process_newly_justified_block` wrote, and the memory guards in `block_change_set.rs:47-53` put the ceiling at 100k substate locks and 10k block-diff changes. `let snapshot = (!batch.foreign_proposals.is_empty()).then(|| change_set.clone());` keeps the one-clone-per-proposal property and skips it entirely when there is nothing to protect.

## Nits

- `on_propose.rs:462-471`: `let Err(err) = … else { continue; };` inverts the usual reading for no gain — the block ends in `break`, so `if let Err(err) = … { …; break; }` is the same control flow in the surrounding idiom.
- `foreign_proposal.rs:511-525`: the test asserts variant properties in the same file as the enum, but the behaviour its doc comment describes — `foreign_proposals_set_status` deleting the `UnconfirmedIndex` entry — lives in `state_store_rocksdb/src/writer.rs:592-600` and stays uncovered. Fine as a tripwire against someone adding `Invalid` to `is_unconfirmed`; the comment claims more than the assertions check.
- `break` at `on_propose.rs:493` condemns only the first failing proposal per round, so *n* bad proposals take *n* rounds per leader. The Scope section's bound is really (bad proposals x committee size), not committee size.

## Sound

- Restricting `Invalid` to `validation_error()` is right for the two sites that can produce one: `foreign_proposal_processor.rs:168` and `:441` are both `ForeignInvalidPledge` on "no evidence for its own shard group", a pure function of the proposal, so every leader reaches the same verdict. `get_or_sequence_transaction` returns `Ok(None)` rather than an error for the local-state cases (unsequenced, finalized, unknown), so a node that is merely behind cannot condemn a good proposal.
- Condemning is not destructive: `set_status_by_id` keeps the record, and `on_ready_to_vote_on_local_block.rs:1588` only rejects `Confirmed`, so if another leader does get the same proposal into a committed block this node still processes it from there.
- Restoring the snapshot discards the successful proposals' effects along with the failed one, and `retain` drops all their atoms to match. Correct, and the alternative (keeping the prefix) would need a per-proposal clone.

sdbondi (commented):
Makes sense

sdbondi (commented):
_Posted by Claude._

Re-reviewed at `a12eb98`. All findings resolved; CI green (clippy, all three test shards, integration test, fmt, ledger app, web ui build).

**Blocking item — fixed.** `return Err(err)` is gone from `on_propose.rs`. Every error class from `process_foreign_block` now takes the same path — restore the snapshot, `retain` the `ForeignProposal` commands out, `break` — and only `err.validation_error().is_some()` pushes onto `invalid_foreign_proposals`. A `ForeignNodeOmittedTransactionPledges` or a missing-justify-QC `InvariantError` from a remote committee now costs this block its foreign proposals instead of routing through `on_failure` into `ConsensusState::Sleeping`, and the block that ships is still votable because it carries no atom for them.

**Attached proposals — fixed.** `dropped_foreign_proposals` gates `NextBlock.foreign_proposals` to `Vec::new()`. Since the drop removes *every* `ForeignProposal` command, clearing the whole vector is the exact complement — nothing the block references is withheld, and a proposal this node just condemned is no longer handed to replicas for `validate_and_save` to store as `New`.

**Conditional snapshot — fixed.** `(!batch.foreign_proposals.is_empty()).then(|| change_set.clone())` skips the deep copy on the common path. The `.take().expect(...)` is sound: the snapshot is `Some` whenever the loop body can run at all, and the `break` immediately after means it is taken at most once.

Both nits applied. The remaining one — `break` condemning only the first bad proposal per round — is unchanged and fine as is.

The properties I checked on the first pass still hold at this head: `process_foreign_block` only reads the substate store, so no store rollback is needed; the snapshot sits after `process_newly_justified_block` and `BlockHeader::create_unsigned` hashes the post-`retain` `commands`, so header and atoms agree; and `Invalid` is reachable only from the two `ForeignInvalidPledge` sites, both pure functions of proposal content, with the record kept rather than deleted so another leader's committed block can still resolve it.

PR #2586, opened 2026-09-10
fix(consensus): classify sync-class and byzantine-triggerable errors correctly

sdbondi:
Three independent fixes to how consensus errors are classified, plus the test-harness change that makes the resulting failures legible. Follows on from #2573/#2574/#2585, but does not depend on them.

### 1. Sync-class errors are not consensus failures

`on_failure` published `HotstuffEvent::Failure` and stopped the pacemaker for *every* error leaving the HotStuff worker loop, including `NeedsSync`, `FallenBehind` and `ProposalValidationError::FutureEpoch`. Those three are a designed transition out of the Running state — `state_machine::running` classifies exactly them as `NeedSync` — so subscribers saw a consensus failure emitted immediately before the state machine logged `⚠️ Behind peers, starting sync`. Both halves appear back-to-back in CI logs.

Note this is *not* the `handle_hotstuff_error` classifier, which already returns `Ok(())` for these shapes; the `on_failure` call sites never reach it.

Adds `HotStuffError::is_sync_required()` and a `HotstuffEvent::SyncRequired` variant, and routes the sync-class arm through it. `pacemaker.stop()` is deliberately kept — the state machine re-enters sync regardless.

This also removes the ability of `epoch_change::epoch_change_hash_divergence_self_heals` to fail on a spurious `Failure` event, which it did intermittently.

### 2. `EvictNode` for an unknown public key no longer aborts consensus

`ValidatorConsensusStats::get_by_public_key` returns `NotFound` for a key with no stats in the epoch. That error reached `handle_hotstuff_error`'s fatal catch-all, so a byzantine leader proposing an `EvictNode` for an arbitrary key could drive a replica into a `Failure` → `Sleeping` → 5s → re-init loop on demand.

A key with no recorded stats has missed no proposals and is never eligible for eviction, so the missing row is treated as zero missed proposals and takes the existing `ShouldNotEvictNode` no-vote path.

### 3. Test harness timeout is now a wall-clock bound

`on_block_committed` wrapped each individual `on_hotstuff_event()` await in `tokio::time::timeout` from inside its loop, so every ignored event reset the timer and the declared budget was never a bound. A consensus livelock that keeps emitting non-commit events therefore never tripped it: the `foreign_shard_group_decides_to_abort` livelock burned nextest's 600s slow-timeout and produced a 14 GB job log instead of panicking at its declared 60s with the pool/block dumps the code already emits. Hoists a deadline out of the loop and uses `timeout_at`.

### Compatibility

Replica evaluation of well-formed proposals is untouched, no block format or hash input moves, and no network coordination is required. `HotstuffEvent::SyncRequired` is a new variant on an internal event enum; all external consumers match non-exhaustively.

### Testing

`cargo test -p consensus_tests` — 68 passed, run green on each commit individually.

### DIFF
--- a/crates/consensus/src/hotstuff/error.rs
+++ b/crates/consensus/src/hotstuff/error.rs
@@ -131,6 +131,18 @@ impl HotStuffError {
             _ => None,
         }
     }
+
+    /// True for errors that mean this node is behind its peers and must sync. These are a designed transition out
+    /// of the Running state, not a consensus failure, and must be classified as such wherever consensus errors are
+    /// reported. The state machine (see `state_machine::running`) maps exactly these variants to `NeedSync`.
+    pub fn is_sync_required(&self) -> bool {
+        matches!(
+            self,
+            Self::NeedsSync { .. } |
+                Self::FallenBehind { .. } |
+                Self::ProposalValidationError(ProposalValidationError::FutureEpoch { .. })
+        )
+    }
 }
 
 impl From<EpochManagerError> for HotStuffError {
--- a/crates/consensus/src/hotstuff/error.rs
+++ b/crates/consensus/src/hotstuff/error.rs
@@ -134,7 +134,7 @@ impl HotStuffError {
 
     /// True for errors that mean this node is behind its peers and must sync. These are a designed transition out
     /// of the Running state, not a consensus failure, and must be classified as such wherever consensus errors are
-    /// reported. The state machine (see `state_machine::running`) maps exactly these variants to `NeedSync`.
+    /// reported. `state_machine::running` maps them to `NeedSync`.
     pub fn is_sync_required(&self) -> bool {
         matches!(
             self,
--- a/crates/consensus/src/hotstuff/event.rs
+++ b/crates/consensus/src/hotstuff/event.rs
@@ -14,6 +14,9 @@ pub enum HotstuffEvent {
     },
     #[error("Consensus failure: {message}")]
     Failure { message: String },
+    /// This node is behind its peers and is leaving consensus to sync. A designed transition, not a failure.
+    #[error("Sync required: {message}")]
+    SyncRequired { message: String },
     #[error("Leader timeout: height {height}")]
     LeaderTimeout { height: NodeHeight },
     #[error("Block {block} has been parked ({num_missing_txs} missing, {num_awaiting_txs} awaiting execution)")]
--- a/crates/consensus/src/hotstuff/state_machine/running.rs
+++ b/crates/consensus/src/hotstuff/state_machine/running.rs
@@ -6,7 +6,6 @@ use log::*;
 use crate::{
     hotstuff::{
         HotStuffError,
-        ProposalValidationError,
         WorkerExitReason,
         state_machine::{
             check_sync::CheckSync,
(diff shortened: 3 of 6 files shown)


sdbondi (commented):
Makes sense

sdbondi (commented):
_Posted by Claude._

Reviewed at `c0b0dfa`. No blocking findings. Verified rather than assumed:

- `is_sync_required()` (`crates/consensus/src/hotstuff/error.rs:138`) matches exactly the three arms `state_machine::running::on_enter` maps to `NeedSync` (`running.rs:43-45`) — the description's claim holds.
- All six `on_failure` call sites (`worker.rs:473,480,494,512,687,695`) return the error afterwards, so the sync-class arm reaches `running.rs` as before; `pacemaker.stop()` staying outside the branch is right.
- Every out-of-crate `HotstuffEvent` consumer matches non-exhaustively (`node.rs:137` let-else, `consensus_gossip/service.rs:75`, `state_sync_task.rs:244`, `mempool/service.rs:125`), so the new variant is additive.
- `.optional()` on `get_by_public_key` maps only `StorageError::NotFound`/`NotFoundDbAdapter` (`storage/src/error.rs:79`); genuine storage errors still propagate.
- The `unwrap_or(0)` default is safe against the threshold: `missed_proposal_evict_threshold` is 10 on every network profile (`consensus_constants.rs:155,203,243,297`), so an unknown key always lands on the `ShouldNotEvictNode` no-vote path. It would only mis-evict if the threshold were ever set to 0 — worth keeping in mind if that constant becomes tunable.
- `on_block_committed` is the only place in the harness that wrapped a per-iteration await in `timeout` — no sibling instances left (`harness.rs:270`).

### Non-blocking

1. `worker.rs:824` — `self.hooks.on_error(err)` still fires for sync-class errors, so the validator's `consensus_errors` counter (`applications/tari_validator_node/src/consensus/metrics.rs:145`) increments on every routine sync transition. That's the same mis-classification the PR fixes for the event stream, just in the metrics path; operators alerting on that counter will still see the noise. Moving `on_error` into the `else` arm (or adding a separate hook) would finish the job.

2. `error.rs:138` vs `running.rs:43-45` — the two lists are now duplicated and the doc comment asserts they stay identical. Having `running.rs` call `err.is_sync_required()` in a guard would make that structural instead of a comment. The `err @` bindings make it slightly awkward, but a single `Err(err) if err.is_sync_required()` arm reads fine.

3. `worker.rs:825` — `"⚠️ Behind peers ({context}): {err}"` and `running.rs:46`'s `"⚠️ Behind peers, starting sync ({err})"` will now both emit for the same error, back to back. Not wrong, but it's the same back-to-back pair the description cites as the symptom; consider dropping this one to `debug!` since the state machine already logs the transition at `info!`.

### Nit

`on_ready_to_vote_on_local_block.rs:400-401` — "A proposal naming one is byzantine and must no-vote rather than abort consensus" defines the code against the behaviour it replaced. The first sentence ("A key with no stats recorded for this epoch has missed no proposals, so it is never eligible for eviction") already carries the invariant a future reader needs; the contrast belongs in the commit message.

One risk worth flagging on the harness change: the deadline now bounds the whole wait for a commit rather than the gap between events, so any test that legitimately absorbs a long run of ignored events (e.g. many offline-node events) before a commit gets less headroom than it had. 68 green tests suggest nothing is close to the line today, but a future flake there would trace back to this.

PR #2587, opened 2026-09-10
refactor!: remove node eviction

sdbondi:
Eviction is disabled on every network (`enable_eviction_proposal = false` in the shipped validator node preset) and is not a mechanism we want to carry into mainnet: it is a consensus-level punishment with no recovery path, gated on a missed-proposal counter that conflates liveness faults with equivocation. This removes it ahead of mainnet.

## What goes

- The `EvictNode` command and its atom, and its proposal / voting / commit paths.
- Eviction proof generation (`generate_eviction_proofs`) and submission to layer one. That was the only writer of the global DB's layer-one transaction table, so `BaseLayerDb`, `DbLayerOnePayloadType` and the epoch manager's `LayerOneSubmitter` plumbing go with it. The validator node's JSON-RPC registration/exit submission uses its own `FileLayerOneSubmitter` and is untouched.
- The `evicted_node` rocksdb column family and its reader/writer methods.
- The `missed_proposal_evict_threshold` consensus constant and the `enable_eviction_proposal` HotStuff config flag.
- Leader-view skipping of evicted validators, which collapses `get_leader_for_view` back to a plain leader-strategy lookup, and with it the `vote_to_skip_next` NEWVIEW-on-vote path.
- The eviction cucumber feature, the `single_shard_node_goes_down_and_gets_evicted` consensus test and the eviction-proof fixture test.

Missed-proposal statistics are still tracked (`missed_proposal_suspend_threshold` / `missed_proposal_recovery_threshold` are untouched); they simply no longer drive eviction.

## Compatibility

Chain state is unaffected: no live network ever committed an `EvictNode` command, so no stored block contains one.

A command's hash is its Borsh encoding, and the base layer recomputes that hash from `tari_sidechain::Command` when verifying an end-of-epoch inclusion proof. `EndEpoch` therefore has to keep the discriminant it has there. Rather than parking a placeholder variant in the `EvictNode` slot, `Command` now states its discriminants explicitly (`#[borsh(use_discriminant = true)]`, `#[repr(u8)]`) — so the variant is simply gone and 6 is absent, with nothing behind it renumbered. The slot can be reclaimed once `tari_sidechain` and minotari drop the variant.

Two tests hold that correspondence rather than leaving it to argument, in `crates/storage/src/consensus_models/command.rs`:

- `discriminants_match_the_sidechain_enum` serialises every variant alongside its `tari_sidechain::Command` counterpart and asserts the tag bytes are equal.
- `an_end_epoch_command_hashes_identically_on_both_sides` asserts the full `EndEpoch` command hash matches across the two enums.

Both fail loudly if a discriminant drifts (verified by deliberately renumbering `EndEpoch` to 6).

Config is deliberately **not** shimmed: `ConsensusConfig::enable_eviction_proposal` is removed outright, and since the struct is `deny_unknown_fields` a node config still setting the key now fails to load and needs the line dropped. The shipped preset no longer writes it. `ConsensusConstantsFile::missed_proposal_evict_threshold` is likewise removed, so a localnet override file setting it needs the line dropped.

The rocksdb `KeyPrefix::EvictedNodes = 13` slot is left unallocated; any stale keys under it in an existing DB are inert. Proto field 8 and the `EvictNodeAtom` message are retained and marked do-not-reuse; the Rust side rejects the field on decode.

## Follow-up

Remove `EvictionProof` / `EvictNodeAtom` from `tari_sidechain` and the eviction proof handling from minotari, then reclaim the reserved `Command` slot.

## Verification

`cargo lints clippy --all-targets` is clean. `cargo test -p tari_consensus -p tari_ootle_storage -p tari_ootle_p2p --lib` passes. The full consensus and integration test suites are left to CI.

https://claude.ai/code/session_01VuuEEPtveDwcSnX1zN9Dk5


sdbondi (commented):
_Posted by Claude._

Read the whole change at 8f8ac42 against its base. The removal is thorough and self-consistent: I grepped the tree at head for every eviction identifier (`evicted_nodes_*`, `suspended_nodes_is_evicted`, `get_nodes_to_evict`, `EvictNodeAtom`, `missed_proposal_evict*`, `enable_eviction_proposal`) and the only survivors are the deliberate ones — the `ConsensusConfig` shim, the `Command::Reserved` doc comment, and the `.proto` field.

I verified the central compatibility claim independently: `tari_sidechain::Command` (5.7.0-pre.5, `src/command.rs:16`) is `LocalOnly, LocalPrepare, LocalAccept, AllAccept, SomeAccept, ForeignProposal, EvictNode, EndEpoch`, and `Command::hash` on both sides is `command_hasher().chain(self)` over the Borsh enum encoding. So `EndEpoch` really is ordinal 7 on both sides, and dropping the variant outright really would have shifted it. Keeping the slot is the right call and the reasoning in the description holds.

Also checked: the deleted `fixtures/block.json` was only consumed by the deleted `eviction_proof.rs` — `block_with_dummies.json`, `committee.json` and `state.wasm` all still have live consumers. No stray generated artifacts or lockfile churn in the diff.

## Blocking

1. **`cargo-machete` fails: `tari_sidechain` is now an unused dependency in two crates.** After this change there are zero references to `tari_sidechain` under `applications/tari_watcher/src` and `crates/epoch_manager/src`, but both still declare it — `applications/tari_watcher/Cargo.toml:19` and `crates/epoch_manager/Cargo.toml:16`. Drop both lines. (`crates/consensus`, `crates/storage`, `crates/consensus_tests`, `integration_tests` and `tari_swarm_daemon` all still use it legitimately.)

2. **`fmt` fails.** Run `cargo +nightly-2025-12-05 fmt --all`. The likely culprits are the reflowed doc comments left over-long by the in-place edits — `utilities/tari_vn_bench/src/report.rs:8-9`, `utilities/tari_vn_bench/src/execution.rs:8-9`, `clients/validator_node_client/src/types.rs:285-286` — where a sentence was deleted mid-paragraph without rewrapping.

Both are mechanical; nothing in the logic blocks. `clippy` and the test jobs were still pending when I looked, so I have not confirmed those.

## Non-blocking

3. **`Command::is_reserved()` has no callers** (`crates/storage/src/consensus_models/command.rs:201`). It replaced `evict_node()`, which did have callers. Either delete it or wire it into block validation, so the rejection is enforced at the validator boundary rather than only in the vote path at `on_ready_to_vote_on_local_block.rs:372`. As it stands it is a public API surface that nothing exercises.

4. **The reserved `EvictNodeAtom evict_node = 8;` field in `crates/p2p/proto/consensus.proto:124` carries no marker.** The Rust side rejects it (`crates/p2p/src/conversions/consensus.rs:684`), but a future edit to the proto has nothing telling it that field 8 and the `EvictNodeAtom` message are load-bearing-by-absence and must not be reused. A `// deprecated: rejected on decode, do not reuse` comment on both would cost nothing.

5. **`Command::Reserved` → `proto::Command { command: None }`** (`consensus.rs:663`) surfaces on the peer as `"Command is missing"` (`consensus.rs:674`), which points at a malformed message rather than at the real cause. Unreachable today since nothing proposes `Reserved`, so this is only about the diagnostic if the invariant ever breaks.

## Nits

6. `crates/consensus_tests/src/catch_up_rewind.rs:70` — the deleted comment ("The whole point is recovery — never evict the isolated node for missed proposals") was the only explanation for `missed_proposal_suspend_threshold = 50`, and that line survives. The threshold now reads as an arbitrary number. Worth keeping a trimmed version that speaks to suspension.

7. `applications/tari_ootle_app_utilities/config_presets/c_validator_node.toml:29` — `[validator_node.consensus]` is now an empty section header followed by a blank line. Harmless to TOML, but it reads like something was cut out (it was).

8. The description's non-breaking argument is scoped to live networks, which is right for chain state. Worth noting for anyone hitting it locally: `#[n(6)] Reserved` is a minicbor unit variant where the stored form had a field, so an existing **localnet/swarm** state DB that did commit an `EvictNode` command will fail to decode that block rather than ignore it. That is an acceptable dev-only cost, but it is not covered by "no live network ever committed one".

sdbondi (commented):
All expected changes

PR #2588, opened 2026-09-11
feat(template_builtin): add proof-taking account methods

sdbondi:
Groundwork for the wave-3 proof-scope change (boundary checks, Radix-style), shipped ahead of it so call sites can migrate before the protocol flips.

## Why

A resource's withdraw and deposit rules are evaluated in the frame that touches the vault — for an account, the account's own frame (`impl.rs` `check_resource_access_rules`, reached from `VaultAction::Withdraw`, `Deposit` and all three `CreateProof*`). That frame carries the transaction's signer badge and nothing else of its own, so a rule naming another badge is satisfied only by handing that badge in, and a `Proof` argument is how a badge reaches a frame.

Today a top-level caller's workspace proofs are copied into the callee's frame, so the account picks up a badge proof without being given one. Wave 3 removes that copy. These methods are the explicit route that survives it.

## What

A `_with_auth` variant of each affected method, taking the proof that authorizes the rule for the duration of the call:

- `withdraw_with_auth`, `withdraw_all_with_auth`
- `withdraw_non_fungible_with_auth`, `withdraw_many_non_fungibles_with_auth`
- `withdraw_confidential_with_auth`
- `deposit_with_auth`
- `create_proof_for_resource_with_auth`, `create_proof_by_non_fungible_ids_with_auth`, `create_proof_by_amount_with_auth`

The proof remains the caller's — neither consumed nor dropped, so the caller's `DropAllProofsInWorkspace` still accounts for it.

Not included: `join_confidential` and `pay_fee`/`pay_fee_stealth`. The first composes from the two variants above; the second moves TARI, whose withdraw rule is `AllowAll`.

## Honest limitation

**These methods are inert under the current scope model.** With the proof on the workspace, the plain method already succeeds — verified by stripping the `authorize_with` out of `withdraw_with_auth` and watching the test still pass. They become load-bearing only when wave 3 stops frames inheriting the caller's proofs.

So the test pins what is provable now: the methods exist, accept a proof, and drive a badge-restricted withdraw/deposit/create-proof end to end; and the same withdraw with no proof anywhere fails with `AccessDenied`.

## Compatibility

Additive. No existing signature changes, so every current manifest and SDK call site is unaffected, and no wallet needs updating — `ResourceAccessRules::new()` sets `withdraw`/`deposit` to `AllowAll` and TARI uses those defaults, so nothing wallets move today is badge-restricted.

The builtin account binary changes, so validators and indexers roll together; the template address is the `[0u8; 32]` constant and no substate moves.

## Tests

`access_rules.rs::the_account_takes_a_badge_restricted_resource_when_handed_the_badge` — a resource whose withdraw *and* deposit rules both name a badge, exercised through `withdraw_with_auth`, `deposit_with_auth` and `create_proof_by_amount_with_auth`, with the no-proof withdraw as the negative control.

`cargo nextest run -p tari_engine`: 423 passed, 0 failed. `cargo lints clippy -p tari_engine --all-targets` clean.

### DIFF
--- a/crates/engine/tests/access_rules.rs
+++ b/crates/engine/tests/access_rules.rs
@@ -675,6 +675,115 @@ mod resource_access_rules {
         );
     }
 
+    /// A resource's withdraw and deposit rules are evaluated in the account's own frame, and a `Proof` argument
+    /// is how a badge reaches a frame. The account holds the transaction's signer badge and nothing else of its
+    /// own, so a rule naming some other badge is satisfied only by handing that badge in.
+    #[test]
+    fn the_account_takes_a_badge_restricted_resource_when_handed_the_badge() {
+        let mut test = TemplateTest::new(CRATE_PATH, ["tests/templates/access_rules"]);
+
+        let (owner_proof, _, owner_key) = test.create_owner_proof();
+        let (user_account, user_proof, user_key) = test.create_empty_account();
+
+        let access_rules_template = test.get_template_address("AccessRulesTest");
+
+        let result = test.execute_expect_success(
+            Transaction::builder_localnet(Epoch(1))
+                .call_function(access_rules_template, "using_resource_rules", args![])
+                .build_and_seal(&owner_key),
+            vec![owner_proof.clone()],
+        );
+
+        let access_rules_component = result.finalize.execution_results[0]
+            .decode::<ComponentAddress>()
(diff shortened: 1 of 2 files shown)


sdbondi (commented):
Reviewed

PR #2589, opened 2026-09-11
fix(engine)!: validate template modules before instantiating them

sdbondi:
Wave 2 of the engine audit: the WASM loader and the windows around a template call. Closes C8, H2, H4, L7, L8, I1 and I2 from the audit report.

## Validate before instantiating

`finalize_loaded_module` instantiated an untrusted module first and validated it afterwards, and the legacy ABI path read a template definition out of the instance's linear memory through a guest-controlled `_ABI_TEMPLATE_DEF` global. Both are on the `PublishTemplate` path, so an attacker chose what got parsed.

The ABI now comes only from the `tari_tdef` custom section (before it fell back on _ABI_TEMPLATE_DEF for compat reasons) — part of the module, no instance and no guest memory involved — and the exports are checked off `wasmer::Module` before `Instance::new`. The legacy embedding is dropped from the loader and from the static extractor used by the wallet daemon.

**Compatibility:** every binary the current `#[template]` macro produces carries the section, and all four builtin templates do (checked). A template published before the macro emitted the section would no longer load — `tari_tdef` has been shipping long enough that this is not expected to affect anything live, testnet included.

## Start sections

A start function runs on every instantiation — at publish validation and on every call — unmetered and outside any invocation. Templates have no use for one; modules declaring one are rejected.

## Table limits (C8)

`LimitingTunables` clamped linear memory and passed table types through untouched. `VMTable::grow` bounds a grow only by the table's declared maximum, which rustc's wasm32 output leaves unset, so `table.grow` with a delta of `0xFFFFFFFF` asked the host for tens of gigabytes at a cost of two metering points. Tables are now given a maximum and validated against `WASM_LIMITS.max_table_elements` (16384, well above `max_functions`).

## Unbilled windows (H2)

The engine drives three pieces of template code per call: `tari_alloc` to stage the `CallInfo`, the template function, and `tari_free` on the returned pointer. Only the middle one was metered — the allocator ran on the compile-time per-call ceiling and its consumption was discarded, and the free ran after the call had already been charged. The metering window now spans all three. The window in which a template may call the engine is unchanged, and an out-of-gas in either allocator call is reported against whatever authorized the compute.

## Bounds on what comes back (H4, L7, L8, I1, I2)

- The value a template returns is capped at `max_call_size`, like the arguments passed in.
- `tari_debug` is capped at `max_log_size_bytes`, budgeted per instance, and goes through `log` instead of `eprintln!`.
- `tari_debug` and `on_panic` are answered only during an invocation, so neither can be driven from `tari_alloc` — a template could otherwise plant a panic message for the next call that traps.
- `on_panic`'s length goes through `u32::try_from`.
- An operator the metering table does not price costs 1000 rather than 1, so a wasmer bump cannot silently make a new operator nearly free.

## Cache

`ENGINE_FINGERPRINT` v4 → v5: compiled artifacts change.

## Tests

`crates/engine/tests/wasm_loader.rs` builds each case as a hand-written WAT module — the macro cannot express a start section, an oversized table or a malformed ABI section. Thirteen tests: the rejection paths above, plus three that execute a hand-written template (a `table.grow` past the cap refuses instead of allocating, an oversized return value is rejected, and points spent in `tari_alloc` reach `wasm_execution_points`). The `buggy` template's ABI moves to a hand-written `tari_tdef` section. Three seeds added for the fuzz target that covers `validate_code`.

`cargo nextest run -p tari_engine`: 436 passed.

https://claude.ai/code/session_01PQL5pB4JHzijVcc3BZRc62


sdbondi (commented):
_Posted by Claude._

Read every changed file at `3324e63` against `development`, traced the new checks into wasmer/wasmparser, and verified the ABI-section claims against the committed binaries. **No blocking findings** — the ordering is right (`load_template_def_from_custom_section` → `validate_template_abi_version` → `validate_functions` → `validate_module_exports` → `Instance::new`), `reject_start_section` is reached on the publish path via `validate_code` → `load_template_from_code`, and the metered-window rework does not move charging: `record_wasm_execution` still runs before the outcome is inspected, and the new `Err` arm returns after it.

Things I checked that hold up:

- **Compatibility claim.** All four `crates/template_builtin/compiled/*.wasm` carry `tari_tdef` and none carry `_ABI_TEMPLATE_DEF`; `crates/template_macros/src/template/template_def.rs:96` emits only the section, so no current-macro output depends on the deleted path.
- **Unpriced-operator fallback.** Spot-checked the enabled `Features` set against `cost_function` — sign-extension, sat-trunc, bulk-memory, reference-types and table ops are all named, so `_ => 1_000` (`metering.rs:511`) is genuinely unreachable today and changes no existing template's gas.
- **Window invariant.** `run_metered` has no early return between `enter_template_invocation` and `exit_template_invocation`, and a refusal recorded from the `CallInfo` `tari_alloc` is still drained by `take_refused_engine_call` before the outcome match.
- **Table cap.** `(table 1 funcref)` → adjusted max 16384 → `VMTable::grow` refuses. `validate_table` rejects both an over-limit declared maximum and an over-limit minimum, which the test covers both ways.

## Non-blocking

1. **The table cap is per-table, and a module may declare 100 of them.** `limiting_tunable.rs:85` bounds each table independently; wasmparser's `MAX_WASM_TABLES` is 100, so a module can declare 100 tables of 16384 `funcref`s and force ~13–26 MB of host `Vec` at every instantiation — including publish validation — against a 2 MiB linear-memory ceiling. That is a far smaller lever than the unbounded `table.grow` this closes, but it leaves tables as the largest per-instance allocation a template controls. A table *count* limit (rustc emits exactly one `__indirect_function_table`) or a total-elements budget would close it; `Tunables` sees one table at a time, so it would have to live in a module-level check rather than here.

2. **`crates/consensus_tests/fixtures/state.wasm` is now a binary the engine would reject.** It carries `_ABI_TEMPLATE_DEF` and no `tari_tdef`. `multishard_publish_template` (`crates/consensus_tests/src/consensus.rs:1550`) only survives because it mocks execution via `add_execution_at_destination`, so nothing ever calls `validate_code` on it. Worth regenerating with the current macro so the fixture isn't quietly a legacy artifact waiting to bite whoever makes that test execute for real.

3. **`ABI_TEMPLATE_DEF_GLOBAL_NAME` (`crates/template_abi/src/lib.rs:59`) has no remaining non-test referent.** The macro doesn't emit it, the loader doesn't read it, the static extractor doesn't fall back to it. Since this PR is already `!`, that's the cheap moment to delete the const and its doc block rather than leave a deprecated name pointing at a path that no longer exists.

4. **`validate_module_exports` checks the entrypoint's signature but not the allocator pair's.** `module.rs:349` now reads exports off `wasmer::Module`, which is exactly where `tari_alloc: (i32) -> i32` and `tari_free: (i32) -> ()` could be required too. As it stands a published template missing or mistyping either loads fine and fails on its first call, in `mem_alloc_func()` rather than at admission. Small, but it is the same class of check the PR is centralising.

5. **Gating `on_panic` on the invocation window also drops genuine panics from `alloc_response`.** `process.rs:524` suspends the window around the `tari_alloc` that stages an engine response, so a template whose allocator legitimately panics there now traps with no message attached — `take_last_panic_message` returns `None` and the generic `RuntimeError` is reported. That is the correct security trade (the plant is worse than the lost diagnostic), but the doc comment on `on_panic_handler` frames the suspended window only as an attack surface; saying that a real panic raised from the response allocator is reported without its message would save the next reader the trace.

## Nits

- `debug_handler` budgets against `ENGINE_LIMITS.max_logs` (`environment.rs:80`), the same constant that bounds transaction logs. The doc comment explains why debug output is budgeted separately from the transaction's logs, then spends the transaction's log constant to do it — a `max_debug_messages` of its own would make each limit independently tunable and say what it means.
- `wasm_loader.rs:44` puts the return blob at offset 16 and `wasm_loader.rs:323` puts the oversized one at 1048, both just outside the fixed region `tari_alloc` hands out at 1024. The relationship is load-bearing for the test and only the second one says so.

sdbondi (commented):
_Posted by Claude._

Re-reviewed at `378d9a3`. All 22 checks pass. Nothing blocks; I can't submit an approval on your own PR, so this stands in for one.

Verified each change against the previous head:

- **Table count (`module.rs:398`, `limits.rs:16`).** `max_tables: 4` read off the table section's declared count, which is the right place — imported tables can't exist here, since only the three `env` functions resolve at instantiation. Paired with the per-table element cap this bounds a module's table storage at ~65k elements, and `rustc`'s wasm32 output declares one.
- **Allocator signatures (`module.rs:355`, `module.rs:387`).** `tari_alloc: (i32) -> i32` and `tari_free: (i32) -> ()` match `WasmAllocFn`/`WasmFreeFn`, so the admission check and the call-time lookup agree. Checked for compatibility: every committed binary — the four builtins, `compute_bench`, and all seven `buggy` variants — exports both under those names. The `no_template_def` and `unexpected_export_function` variants still fail earlier (missing section, unexpected export), so those assertions keep testing what they name.
- **`max_debug_messages: 256`.** Separate constant, and the `allow_debug_message` doc no longer explains a separation it then didn't make.
- **`ABI_TEMPLATE_DEF_GLOBAL_NAME`** gone, and `TEMPLATE_DEF_CUSTOM_SECTION`'s doc no longer describes the global as a parallel embedding.
- **`on_panic_handler`** now says a panic raised inside `alloc_response`'s allocation reaches the transaction without its message.

`rejects_a_missing_or_mistyped_allocator` covers both the absent export and the wrong parameter type, and `rejects_more_tables_than_the_limit` sits one over the cap — both fail for the reason they assert rather than on an earlier check.

One item from the first pass is still open, and is fine to leave: `crates/consensus_tests/fixtures/state.wasm` carries only `_ABI_TEMPLATE_DEF` and no `tari_tdef`, so it is now a binary this engine would reject. `multishard_publish_template` never loads it — execution is mocked through `add_execution_at_destination` — so nothing fails today. Worth regenerating whenever that area is next touched.

sdbondi (commented):
Make sense

PR #2590, opened 2026-09-11
fix(consensus): anchor the whole proposal at the state anchor

sdbondi:
## Problem

When a leader fills a timeout gap with a dummy chain, the candidate extends from the **justify block**, not from the highest seen block. The dummies are empty blocks carrying the justify block's `accumulated_data` and state forward (see `calculate_last_dummy_block`), so the candidate's parent chain runs back through them to the justify block and never through a locally stored fork above the high QC.

`accumulated_data`, the substate store, the pending state tree diff lookup and propose-time foreign proposal processing already followed that anchor. The **proposal batch**, the transaction pool query, foreign proposal *selection* (`get_all_new`), the change set and command generation still read at the highest seen block.

That block is an orphan the candidate abandons. Its pool records carry stages and decisions no replica reproduces, so the leader can skip a transaction that is still pending on the chain it is actually building, or emit a command for a stage the candidate's ancestors never reached.

## Change

`state_anchor_leaf` — the justify block when `dummy_block.is_some()`, the extended leaf otherwise — now drives all of it. `start_of_chain_block` survives only as the non-dummy arm of that conditional, so this is a **strict no-op when `dummy_block.is_none()`**.

This is distinct from #2574, which moved one argument: foreign proposal *processing*. This moves *selection* and command generation. #2574 changed how records resolve; this changes which transactions get proposed.

## Scope

Like #2573 and #2574, this is **proposer-side only**:

- replica evaluation is untouched,
- no block format or hash input moves,
- the only blocks whose content changes are ones no honest replica votes for.

## Test

`crates/consensus_tests/src/dummy_fill_anchor.rs` withholds the first command-bearing proposal from one of three committee members. The block is stored and evaluated by the rest of the committee — so it is the highest seen block — but two of three votes are one short of quorum, so it never gathers a QC and the pacemaker fills its height with a dummy. The proposal built on that dummy must repeat the orphan's command.

Verified to fail on `development` with only `on_propose.rs` reverted:

```
assertion `left == right` failed: proposal on the dummy chain must repeat the orphan's command.
  left: None
 right: Some("LocalPrepare(fe1ea88a…, Commit)")
```

Without the change, the post-dummy proposal carries no command for the transaction at all — the batch was selected against the orphan, where the record already looked handled.

- `cargo test -p consensus_tests` — 68 passed, 0 failed.
- `cargo lints clippy -p tari_consensus --all-targets` / `-p consensus_tests --all-targets` — clean.

### DIFF
--- a/crates/consensus/src/hotstuff/on_propose.rs
+++ b/crates/consensus/src/hotstuff/on_propose.rs
@@ -302,7 +302,7 @@ where TConsensusSpec: ConsensusSpec
     /// Returns Ok(None) if the command cannot be sequenced yet due to lock conflicts.
     fn transaction_pool_record_to_command<TTx: StateStoreReadTransaction>(
         &self,
-        start_of_chain_id: &LeafBlock,
+        state_anchor: &LeafBlock,
         locked_epoch: &LockedEpoch,
         pool_tx: TransactionPoolRecord,
         local_committee_info: &CommitteeInfo,
@@ -313,7 +313,7 @@ where TConsensusSpec: ConsensusSpec
     ) -> Result<Option<Command>, HotStuffError> {
         match pool_tx.current_stage() {
             TransactionPoolStage::New => self.prepare_transaction(
-                start_of_chain_id,
+                state_anchor,
                 locked_epoch,
                 pool_tx,
                 local_committee_info,
@@ -325,7 +325,7 @@ where TConsensusSpec: ConsensusSpec
             // Leader thinks all foreign PREPARE pledges have been received (condition for LocalPrepared stage to be
             // ready)
             TransactionPoolStage::LocalPrepared => self.local_accept_transaction(
-                start_of_chain_id,
+                state_anchor,
                 local_committee_info,
                 change_set,
                 pool_tx,
@@ -336,7 +336,7 @@ where TConsensusSpec: ConsensusSpec
             // Leader thinks that all foreign ACCEPT pledges have been received and, we are ready to accept the result
             // (COMMIT/ABORT)
             TransactionPoolStage::LocalAccepted => {
-                self.accept_transaction(start_of_chain_id, &pool_tx, local_committee_info, substate_store)
+                self.accept_transaction(state_anchor, &pool_tx, local_committee_info, substate_store)
             },
             // Not reachable as there is nothing to propose for these stages. To confirm that all local nodes
             // agreed with the Accept, more (possibly empty) blocks with QCs will be
@@ -369,7 +369,6 @@ where TConsensusSpec: ConsensusSpec
     ) -> Result<NextBlock, HotStuffError> {
         let high_qc_id = high_qc_certificate.calculate_id();
         let justify_block = Block::get_justified_block(tx, &high_qc_certificate, epoch)?;
-        let start_of_chain_block = highest_seen_block;
         let parent_block = dummy_block.unwrap_or_else(|| highest_seen_block.as_leaf());
         let highest_seen_block = Block::get(tx, highest_seen_block.block_id())?;
         let is_end_of_epoch_in_chain = highest_seen_block.is_epoch_end_proposed_in_chain(tx)?;
@@ -383,8 +382,8 @@ where TConsensusSpec: ConsensusSpec
         };
 
         let mut total_leader_fee = 0u64;
-        // The block the candidate extends from, and therefore the point every part of this proposal is read
-        // at. When filling a timeout gap with a dummy chain that is justify_block, not the highest seen block:
+        // The block the candidate extends from, and the point at which every speculative state and pool read
+        // for this proposal is taken. When filling a timeout gap with a dummy chain that is justify_block:
         // the dummies are empty blocks carrying justify_block's accumulated_data and state forward (see
         // `calculate_last_dummy_block`), so the candidate's parent chain runs back through them to
         // justify_block and never through a locally-stored fork above the high QC. A validator recomputes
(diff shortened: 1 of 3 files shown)


sdbondi (commented):
_Posted by Claude._

No blocking findings. Verified the premise rather than taking the description's word for it.

## Verified sound

- `transaction_pool_get_many_ready` (`crates/state_store_rocksdb/src/reader.rs:1299`) reads the base pool record and overlays `transaction_pool_state_update` rows along `get_pending_chain_ordered(block_id)`. Passing the orphan therefore merges updates written by blocks the candidate's parent chain does not contain — exactly the "stage no replica reproduces" failure the PR describes. `justify_block` is the correct anchor, and the dummies contribute no updates, so nothing is lost by stopping there.
- `foreign_proposals_get_all_new` (`reader.rs:576`) uses `get_pending_chain_with_commands_between(block_id)` to skip proposals already proposed in the chain. Same reasoning: the orphan's `ProposedInBlockIndex` entries are not on the candidate's chain, so selection at the orphan silently drops a foreign proposal the candidate is entitled to carry.
- Both readers assert `blocks_exists(block_id)`, so the dummy leaf could not have been passed even if it were the "true" parent — `justify_block` is the only valid stored anchor. Good.
- `transaction_pool_record_to_command` → `prepare` → `fetch_execution` → `BlockTransactionExecution::get_pending_for_block` is another chain-scoped lookup, now consistent with the substate store and change set built on the same leaf (`on_propose.rs:411`, `:436`, `:568`).
- `ProposedBlockChangeSet::new(state_anchor_leaf)` and the command generation now agree with `process_foreign_block`'s anchor from #2574; previously the change set and the foreign-proposal processing disagreed under a dummy fill.
- The no-op claim holds: `state_anchor_leaf` reduces to `start_of_chain_block.as_leaf()` when `dummy_block.is_none()`, which is the old `start_of_chain_block` in every moved call site. Nothing else in `build_next_block` changed.
- Header epoch hash (`on_propose.rs:685`) is validated against the replica's own `epoch_state.epoch_hash()` (`validations/common.rs:75`), not against the parent, so leaving it at `highest_seen_block` is not an anchor divergence.
- Test quorum arithmetic checks out: `max_failures = (3-1)/3 = 0`, so quorum is 3 of 3 and two votes genuinely cannot certify the withheld block (`common_types/src/committee.rs:256`).

## Non-blocking

1. `on_propose.rs:405-409` — `accumulated_data` repeats the `dummy_block.is_some()` conditional instead of deriving from `state_anchor_leaf`. Two conditionals that must always agree is exactly the coupling this PR is removing everywhere else. Since `justify_block` is already in scope and `highest_seen_block` is the non-dummy anchor, this could read off the same branch (or the anchor could carry the header) so a future edit cannot desync them.

2. `on_propose.rs:503-505` — `locked_epoch` is built from `highest_seen_block.epoch()`, but the replica computes `block.to_locked_epoch()` from the candidate (`on_ready_to_vote_on_local_block.rs:569`, `block.rs:373`), i.e. from the `epoch` argument. The `epoch_hash` component matches by construction (the header takes it from `highest_seen_block` too), but `epoch` can differ from `highest_seen_block.epoch()` across a boundary. Pre-existing and out of scope, but the comment at `:395` asserts "every read for the candidate follows this anchor", which this and `is_end_of_epoch_in_chain` / `is_epoch_end()` at `:374-382` do not. Either narrow the claim or move them.

3. `on_propose.rs:374-382` — `is_end_of_epoch_in_chain` and `is_epoch_end()` are still read at the orphan. The effect is conservative (commands suppressed, no fork risk) and self-correcting once the candidate becomes the highest seen block, so not a bug — but it is the one remaining read that contradicts the new comment.

4. `dummy_fill_anchor.rs:169-176` — the test only catches the bug when the leader at the orphan's height + 1 is one of the two validators that *saw* the orphan. If that leader is the starved validator, its `highest_seen_block` is already the justify block, `state_anchor_leaf` is a no-op, and the assertion passes on unfixed code. You verified it fails on `development`, so the harness lands on the catching case today, but the selection is sensitive to which height first carries the command. Asserting on the proposer of `post_dummy` (or skipping to the first post-dummy block proposed by a non-starved validator) would make the regression coverage independent of that.

## Nits

- `on_propose.rs:372` — `start_of_chain_block` is now a single-use alias for `highest_seen_block`, read only at `:402`. It exists to capture the value before the shadowing at `:374`; worth either inlining or renaming to say that, since the name no longer describes anything the code does.
- `on_propose.rs:305` — `transaction_pool_record_to_command`'s parameter is still `start_of_chain_id` while its only caller now passes `state_anchor_leaf`. Same for `prepare_transaction`'s `parent_block` at `:774`, which is not the parent under a dummy fill.
- `on_propose.rs:396-398` — the comment enumerates the six reads that follow the anchor. That list will rot on the next read added; the invariant ("every read for the candidate is taken at this block") carries the same weight without the roll call.
- `consensus_tests/src/lib.rs:13-14` — duplicated `#[cfg(test)]` above `mod last_voted_persistence`. Pre-existing, adjacent to your hunk, trivial to drop while you are here.

sdbondi (commented):
_Posted by Claude._

Re-reviewed `296490d3d` against each earlier finding. All four non-blocking items and all four nits are addressed; nothing new blocks.

**Re-verified, not taken on faith:**

- (1) `accumulated_data` now derives from the same `state_anchor` binding as `state_anchor_leaf` (`on_propose.rs:394-402`), so the two can no longer desync. The no-op claim still holds after the refactor: the non-dummy arm changed from `HighestSeenBlock::as_leaf` to `Block::as_leaf`, and those build the same `LeafBlock` from the same four fields (`last_seen_block.rs:60`, `block_header.rs:426`) — `HighestSeenBlock` is persisted from `Block::as_highest_seen` (`block.rs:310`), so `block_id`/`height`/`epoch`/`shard_group` are identical by construction.
- (3) The epoch-boundary reads are now documented as deliberately at `highest_seen_block` rather than claimed to follow the anchor.
- (4) The test now resolves each dummy extension's proposer through the epoch manager and skips the starved validator (`dummy_fill_anchor.rs:187-210`), so the regression coverage no longer depends on which validator happens to lead the height after the orphan. `get_validator_node_by_public_key(Epoch, RistrettoPublicKeyBytes)` matches `epoch_manager/src/traits.rs:91` and `block.proposed_by()` returns that type. The panic message names the reason the case is unusable, which is the right failure mode — an unresolvable run is a skipped assertion, not a pass.
- Renames, the single-use alias, the comment's read enumeration, and the duplicated `#[cfg(test)]` are all gone.

**One nit on the new comment** (`on_propose.rs:394-396`): "The epoch-boundary checks above and the locked epoch below are read at `highest_seen_block` instead. They gate whether commands are proposed at all, so reading them a block early only ever suppresses commands." That justification is accurate for `is_end_of_epoch_in_chain` / `is_epoch_end()`, but `locked_epoch` does not gate anything — it is written into the pool record by `update_locked_epoch` and feeds execution. Its `epoch_hash` half is covered by the following clause, but its `epoch` half is read at `highest_seen_block.epoch()` while the replica derives it from the candidate's own header (`block.rs:373`, called at `on_ready_to_vote_on_local_block.rs:569`), i.e. from the `epoch` argument. Those are the same block today; the comment reads as if the suppression argument covers `locked_epoch` too, and it does not. Splitting the sentence — gates at `highest_seen_block` because reading early only suppresses; `locked_epoch` there because the header takes its epoch hash from the same block — would say what is actually true of each.

Holding off on a verdict until `clippy` and the three `test` shards report; everything else on this head is green so far.

sdbondi (commented):
Makes sense

PR #2591, opened 2026-09-11
fix(engine)!: bind the write-mode chokepoint, substate size and event size

sdbondi:
Engine audit wave 4, first of four. Three limits the engine documents but does not enforce on every path. Independent of the other wave-4 PRs.

## Write-mode chokepoint

`check_write_allowed` is documented as *the* single chokepoint for `FrameWriteMode`, but only `write_lock_substate` and `new_substate` called it. Five sites took a write lock without it — `validate_and_spend_stealth_utxos` and `spend_confidential_outputs` through `store.try_lock`, and the NFT lock in `burn_bucket`, the vault lock in `drop_proof` and `withdraw_fees_from_pool_up_to` through the private `lock_substate`.

The check now lives in a `try_lock` helper that every lock passes through, so the mode binds on the lock flag rather than on which helper a caller reached for.

**This is defence in depth, not a live exploit.** I went looking for a reachable path and did not find one: `stealth_transfer` is reachable only from the instruction processor, and the confidential-spend and burn paths need a bucket or a vault write lock that a `ReadOnly` or `OwnComponent` frame is already refused. The point is that the invariant is now structural — the next site to take a write lock gets the check for free instead of having to remember it.

## Substate size on mutation

`enforce_substate_size_limit` ran only in `new_substate`, so `max_substate_size` bound creation and nothing else. A substate can be grown through any of the `&mut SubstateValue` handles the working state hands out (`modify_component_with`, `get_vault_mut`, `get_resource_mut`, `get_non_fungible_mut`), and one of them accumulates without any call-size bound above it: a vault's non-fungible id set grows one deposit at a time, across transactions, with nothing measuring it.

The limit now binds where it means something — a sweep of `mutated_substates()` in `validate_finalized`, which is the one place every persisted substate passes through. The error names the offending substate.

## Event payload size

`push_event` capped the event count at 256 but not the payload, unlike `push_log`. Every event lands in the transaction receipt, which is persisted as a substate like any other but is built after fees settle, so it cannot be rejected for being oversized.

New `max_event_size_bytes` of 2 KiB, enforced against the whole encoded event rather than the payload alone, since it is the whole event the receipt carries. 256 × 2 KiB = 512 KiB, which leaves the receipt room for its diff summary (up to `max_substate_outputs` entries) and fee breakdown inside the 1 MiB substate budget.

## Compatibility

The last two change which transactions succeed, so validators and indexers need to upgrade together. No protocol version bump is involved. Two consequences worth stating outright:

- **An already-oversized substate becomes unusable until it is shrunk.** Creation was always checked and growth never was, so the on-chain set may hold a substate over 1 MiB — that is the gap being closed. After this, any transaction that *touches* one is rejected, and `get_for_mut` moves a substate into `new_substates` on any `&mut` access, so a read through a mutable handle is enough. The remedy is one transaction that brings it back under the limit; an oversized NFT vault does not block fee payment, since fees come from the TARI vault. If a fee-intent substate were the oversized one the transaction is rejected with no fee collected. I do not expect esmeralda to have any, but that should be confirmed rather than assumed.
- **`resource.create`'s `image_url` is now indirectly capped at ~2 KiB.** It is the only builtin event payload field carrying an unbounded user-controlled value (the token symbol is already capped at 10 bytes), so a resource created with a larger `image_url` fails with `EventSizeExceeded`. That is the intended shape of the limit, but it is a behaviour change a template author can hit without emitting an event of their own.

## Tests

- `max_event_size_limit` — an event at the size limit is emitted; one byte more is rejected with the exact measured size. The boundary is solved for rather than computed, because a CBOR length prefix widens at 24/256/65536 bytes.
- `max_substate_size_limit_applies_to_mutations` — seeds an account's NFT vault near the limit and deposits faucet tokens into it. A deposit with room to spare commits; one that takes the vault over is rejected, in a transaction that creates no substate of its own.
- Full `-p tari_engine` suite: 425 passed.

### DIFF
--- a/crates/engine/src/runtime/working_state.rs
+++ b/crates/engine/src/runtime/working_state.rs
@@ -25,7 +25,7 @@ use tari_engine_types::{
     id_provider::{IdProvider, ObjectIds},
     indexed_value::{IndexedValue, IndexedWellKnownTypes},
     limits,
-    lock::LockFlag,
+    lock::{LockFlag, LockId},
     logs::LogEntry,
     non_fungible::NonFungibleContainer,
     proof::{ContainerRef, LockedResource, Proof},
@@ -234,14 +234,14 @@ impl<TStore: StateReader> WorkingState<TStore> {
         self.store.exists(address)
     }
 
-    fn enforce_substate_size_limit(&self, value: &SubstateValue) -> Result<(), RuntimeError> {
+    fn enforce_substate_size_limit(id: &SubstateId, value: &SubstateValue) -> Result<(), RuntimeError> {
         // Published template has its own size restriction
         if value.published_template().is_some() {
             return Ok(());
         }
         let size = encoded_len(value);
         if size > limits::ENGINE_LIMITS.max_substate_size {
-            return Err(LimitError::SubstateSizeExceeded { size }.into());
+            return Err(LimitError::SubstateSizeExceeded { id: id.clone(), size }.into());
         }
         Ok(())
     }
@@ -254,15 +254,24 @@ impl<TStore: StateReader> WorkingState<TStore> {
         let address = address.into();
         self.check_write_allowed(&address)?;
         let value = value.into();
-        self.enforce_substate_size_limit(&value)?;
+        Self::enforce_substate_size_limit(&address, &value)?;
         self.current_call_scope_mut()?.add_substate_to_scope(address.clone())?;
         self.store.insert(address, value)?;
         Ok(())
     }
 
     fn lock_substate(&mut self, addr: SubstateId, lock_flag: LockFlag) -> Result<LockedSubstate, RuntimeError> {
-        let lock_id = self.store.try_lock(addr.clone(), lock_flag)?;
+        let lock_id = self.try_lock(addr.clone(), lock_flag)?;
         Ok(LockedSubstate::new(addr, lock_id, lock_flag))
+    }
+
+    /// Every lock this state takes goes through here, so a write lock cannot be acquired without the frame write
+    /// mode permitting it. Callers that need the raw [`LockId`] use this directly rather than the store.
+    fn try_lock(&mut self, addr: SubstateId, lock_flag: LockFlag) -> Result<LockId, RuntimeError> {
+        if lock_flag.is_write() {
+            self.check_write_allowed(&addr)?;
+        }
+        self.store.try_lock(addr, lock_flag)
     }
 
     pub fn read_lock_substate(&mut self, addr: SubstateId) -> Result<LockedSubstate, RuntimeError> {
@@ -270,7 +279,6 @@ impl<TStore: StateReader> WorkingState<TStore> {
     }
 
     pub fn write_lock_substate(&mut self, addr: SubstateId) -> Result<LockedSubstate, RuntimeError> {
-        self.check_write_allowed(&addr)?;
         self.lock_substate(addr, LockFlag::Write)
     }
 
@@ -387,7 +395,7 @@ impl<TStore: StateReader> WorkingState<TStore> {
 
         for input in &stmt.inputs_statement.inputs {
             let address = UtxoAddress::new(resource_address, input.commitment.into());
-            let lock_id = self.store.try_lock(address.clone().into(), LockFlag::Write)?;
+            let lock_id = self.try_lock(address.clone().into(), LockFlag::Write)?;
             let utxo = self.store.down_utxo(lock_id)?;
             self.store.try_unlock(lock_id)?;
             if utxo.is_frozen() {
@@ -541,6 +549,12 @@ impl<TStore: StateReader> WorkingState<TStore> {
     }
 
     pub(super) fn validate_finalized(&self) -> Result<(), RuntimeError> {
+        // A substate can be grown through any of the `&mut SubstateValue` handles this state hands out, so the
+        // size limit binds on what is actually persisted rather than on what was created.
+        for (id, value) in self.store.mutated_substates() {
+            Self::enforce_substate_size_limit(id, value)?;
+        }
+
         if self.buckets.iter().any(|(_, b)| !b.is_empty()) {
             return Err(TransactionCommitError::DanglingBuckets {
                 count: self.buckets.len(),
@@ -656,7 +670,7 @@ impl<TStore: StateReader> WorkingState<TStore> {
             .into_iter()
             .map(|commitment| {
                 let address = ConfidentialOutputAddress::new(resource_address, commitment);
-                let lock_id = self.store.try_lock(address.clone().into(), LockFlag::Write)?;
+                let lock_id = self.try_lock(address.clone().into(), LockFlag::Write)?;
                 let output = self.store.down_confidential_output(lock_id)?;
                 self.store.try_unlock(lock_id)?;
                 if output.is_frozen() {
@@ -1899,6 +1913,11 @@ impl<TStore: StateReader> WorkingState<TStore> {
     }
 
     pub fn push_event(&mut self, event: Event) -> Result<(), RuntimeError> {
+        let size = encoded_len(&event);
+        if size > limits::ENGINE_LIMITS.max_event_size_bytes {
+            return Err(LimitError::EventSizeExceeded { size }.into());
+        }
+
         if self.events.len() >= limits::ENGINE_LIMITS.max_events {
             return Err(LimitError::MaxEventsExceeded.into());
         }
--- a/crates/engine/src/runtime/working_state.rs
+++ b/crates/engine/src/runtime/working_state.rs
@@ -1914,13 +1914,15 @@ impl<TStore: StateReader> WorkingState<TStore> {
     }
 
     pub fn push_event(&mut self, event: Event) -> Result<(), RuntimeError> {
+        // The count cap is tested first because measuring the event walks its payload, unlike `push_log`, whose
+        // size check is a `String::len`.
+        if self.events.len() >= limits::ENGINE_LIMITS.max_events {
+            return Err(LimitError::MaxEventsExceeded.into());
+        }
+
         let size = encoded_len(&event);
         if size > limits::ENGINE_LIMITS.max_event_size_bytes {
             return Err(LimitError::EventSizeExceeded { size }.into());
-        }
-
-        if self.events.len() >= limits::ENGINE_LIMITS.max_events {
-            return Err(LimitError::MaxEventsExceeded.into());
         }
         self.events.push(event);
         Ok(())
--- a/crates/engine/src/runtime/working_state.rs
+++ b/crates/engine/src/runtime/working_state.rs
@@ -550,7 +550,8 @@ impl<TStore: StateReader> WorkingState<TStore> {
 
     pub(super) fn validate_finalized(&self) -> Result<(), RuntimeError> {
         // A substate can be grown through any of the `&mut SubstateValue` handles this state hands out, so the
-        // size limit binds on what is actually persisted rather than on what was created.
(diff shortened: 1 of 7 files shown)


sdbondi (commented):
_Posted by Claude._

Reviewed at `5ec1585`. Nothing blocking — all three changes hold up under tracing. Notes below are non-blocking.

## What I verified

- **Write chokepoint.** `try_lock` is genuinely the only route to `store.try_lock` now (`working_state.rs:270`) — `lock_substate`, the two confidential/stealth spend paths and `write_lock_substate` all funnel through it, and nothing else in the crate calls `self.store.try_lock` directly. The five newly-covered sites cannot regress a legitimate flow: `drop_proof` and `burn_bucket` are reachable only via `proof_invoke`/`bucket_invoke`, both in `FORBIDDEN_IN_OWN_COMPONENT` and `FORBIDDEN_IN_READ_ONLY` (`impl.rs:274-275`), and `withdraw_fees_from_pool_up_to` runs at instruction level where `current_frame_write_mode()` defaults to `Full` with no frames (`working_state.rs:1465-1470`). The hook component's own write lock is still taken by the *parent* frame — `lock_component` at `processor.rs:933` precedes `push_call_frame`, which is where the restriction is applied (`impl.rs:3908-3916`) — so `OwnComponent` frames are unaffected.
- **Mutation sweep.** `WorkingStateStore::get_for_mut` (`state_store.rs:132-145`) moves a substate into `new_substates` on *any* `&mut` handle, so `mutated_substates()` really is the full set of `&mut SubstateValue` handles the working state hands out; the `modify_component_with`/`get_vault_mut`/`get_resource_mut`/`get_non_fungible_mut` gap the description names is closed by one sweep. A failure here lands in `process_instructions`' `validate_finalized` (`processor.rs:351`), so it rejects the main intent and still commits the fee intent — no free execution.
- **Receipt budget.** The 256 × 2 KiB arithmetic checks out against what `TransactionReceipt` actually carries: `diff_summary` is `UpSubstate` (id + version + 32-byte hash) capped at `max_substate_outputs` = 1000, plus `fee_withdrawals` and `FeeReceipt`. Logs are *not* in the receipt (they would have blown the budget at 256 × 32 KiB), so ~600 KiB worst case against the 1 MiB limit.
- `push_event` mirrors `push_log`'s size-then-count ordering exactly (`working_state.rs:1895-1920`), and the custom topic was already capped at 255 bytes by `Event::validate_custom_topic`, so the new check adds no ordering or field surprise.

## Non-blocking

1. **`impl.rs:255` still names the old chokepoint.** The doc for `enforce_frame_restrictions` says "layer (a) in `WorkingState::write_lock_substate` / `new_substate`". After this PR that is `try_lock` / `new_substate` — the whole point being that `write_lock_substate` is no longer where the check lives. The sibling comment in `scope.rs:407-413` was updated; this one is the pair to it.

2. **"what is actually persisted" overstates the sweep's reach** (`working_state.rs:550-551`). Two persisted things bypass it, both after `validate_finalized` has run: the refunds in `finalize_fees_and_refunds` mutate vaults in `substates_to_persist` directly (`working_state.rs:2002-2018`), and the transaction receipt is upped in `tracker.rs:540-544`. Neither is exploitable — a refund deposits into an existing container of the same resource, and the receipt is exactly what the new event limit bounds — but the comment reads as a stronger guarantee than the code gives. "every substate the transaction's instructions persist" would be accurate.

3. **Already-oversized substates become unusable, and the compatibility note doesn't say so.** Creation was always checked, growth never was, so the on-chain set may already contain a substate over 1 MiB — that is precisely the gap being closed. After this change any transaction that *touches* one is rejected, and because `get_for_mut` moves a substate into `new_substates` on any `&mut` access even when nothing changes, a read-through-`&mut` is enough. The remedy (one transaction that brings it back under the limit) exists, and an oversized NFT vault doesn't block fee payment since fees come from the TARI vault — but if a fee-intent substate were the oversized one, `checkpoint_fee_intent` (`impl.rs:3664-3678`) fails outright and the transaction is rejected with no fee collected at all. Worth a line in the Compatibility section even if the answer is "esmeralda has none".

4. **`resource.create`'s `image_url` is now indirectly capped at ~2 KiB** (`impl.rs:1596-1604`). It is the only builtin event payload field with an unbounded user-controlled value — the token symbol is already capped at 10 bytes — so a resource created with a larger `image_url` now fails with `EventSizeExceeded` rather than the template's own error. Probably the desired outcome, but it is a user-visible behaviour change beyond "which transactions succeed" in the sense the description means it.

## Nits

- `new_substate`'s `enforce_substate_size_limit` call is now redundant with the sweep. Keeping it is defensible — it fails at the creating instruction rather than at the end of the transaction, which is the better error — just noting it is deliberate duplication rather than an oversight.
- `largest_fitting_nft_count` (`tests/limits.rs`) builds a 200k-id `BTreeSet` for the bracketing assert and ~18 more during the search. Correct and terminating, just the slowest part of the test; `encoded_len` is `minicbor::len`, so it is pure arithmetic over the tree rather than an encode, which keeps it tolerable.
- `seed_nft_vault` picks the first vault holding `NFT_FAUCET_RESOURCE_ADDRESS`. That is unambiguous today only because the faucet template holds no vault of its own (it mints on demand from a serial counter). If the faucet ever gains one, this silently seeds the wrong vault.

crates/engine/src/runtime/working_state.rs
```diff
@@ -1899,6 +1914,11 @@ impl<TStore: StateReader> WorkingState<TStore> {
     }
 
     pub fn push_event(&mut self, event: Event) -> Result<(), RuntimeError> {
+        let size = encoded_len(&event);
+        if size > limits::ENGINE_LIMITS.max_event_size_bytes {
+            return Err(LimitError::EventSizeExceeded { size }.into());
+        }
+
```
sdbondi: This check can go after the next check (cheaper check first)

Done — the count cap now runs first. Noted at the call site why this one differs from `push_log`: there the size check is a `String::len`, so ordering is free, whereas measuring an event walks its payload.

One consequence worth naming: an oversized event pushed when the transaction is already at 256 now reports `MaxEventsExceeded` rather than `EventSizeExceeded`. Reject reasons are not consensus-visible (`AbortReason` collapses to `ExecutionFailure`), so this is a local diagnostic change only.

crates/engine/src/runtime/working_state.rs
```diff
@@ -1914,14 +1914,16 @@ impl<TStore: StateReader> WorkingState<TStore> {
     }
 
     pub fn push_event(&mut self, event: Event) -> Result<(), RuntimeError> {
+        // The count cap is tested first because measuring the event walks its payload, unlike `push_log`, whose
+        // size check is a `String::len`.
```
sdbondi: delete this (narrating a change)

Deleted in `2e65a3e`. You're right — it justified the ordering by contrast with `push_log`, which is both narration and defining the code by what it is not. The ordering is a micro-optimisation that stands on its own.

crates/engine/src/runtime/working_state.rs
```diff
@@ -541,6 +549,13 @@ impl<TStore: StateReader> WorkingState<TStore> {
     }
 
     pub(super) fn validate_finalized(&self) -> Result<(), RuntimeError> {
+        // A substate can be grown through any of the `&mut SubstateValue` handles this state hands out, so the
+        // size limit binds on every substate the transaction's instructions persist rather than on what was
+        // created.
+        for (id, value) in self.store.mutated_substates() {
+            Self::enforce_substate_size_limit(id, value)?;
```
sdbondi, line 557: cant we validate the substate size on ingress from the engine call (set state etc) rather than on finalise, to terminate the transaction as early as possible?

sdbondi: _Posted by Claude._

Yes for the writes that can cross it in one step — added in `377af6f8`. But it cannot replace the sweep, and it turns out it cannot fire today either. Three parts.

**1. Where an ingress check works.** `SetState`, `UpdateMetadata` and `UpdateNonFungibleData` each *replace* a substate's contents with a value that arrived as an engine argument, which `max_internal_call_size` (1 MiB) permits to be as large as `max_substate_size` (1 MiB). Those three now enforce the limit at the write, so the instruction responsible fails.

**2. Why the sweep has to stay.** The other mutation paths do not take a whole new value — they accumulate. A vault gains ~10 bytes of non-fungible id per deposit, so no single write can see it cross the limit; only the total can. And checking there is not free: `encoded_len` walks the whole substate, so measuring on every deposit makes a transaction that deposits into one vault N times cost O(N × vault size) instead of the O(vault size) the sweep costs once.

**3. The uncomfortable part — the ingress checks cannot currently fire.** I tried to write the test and could not. A component's state has to be built and encoded inside a 2 MiB WASM linear memory (`max_memory_pages: 32`), and the value is copied several times on the way out, so it traps on allocation well before the limit. Measured on the real engine, `grow_data(n)` on a component:

| n | outcome |
|---|---|
| 200 KiB | commits |
| 250 KiB | commits |
| 300 KiB | `Wasm RuntimeError: unreachable` |
| 400 KiB – 1 MiB | `Wasm RuntimeError: unreachable` |

So a component tops out around 250–300 KiB, roughly a quarter of `max_substate_size`, and `UpdateMetadata` and `UpdateNonFungibleData` travel the same encode-into-WASM-memory path. **The only reachable route past `max_substate_size` today is exactly the accumulating one — vault non-fungible ids — which is what the sweep catches.**

That leaves a judgement call I'd rather put to you than make silently. The three ingress checks are correct and land the error in the right place, but they cost one `encoded_len` walk per `SetState` on a path that cannot fail, and they only become load-bearing if `max_memory_pages` rises or `max_internal_call_size` and `max_substate_size` diverge. I have kept them because a size limit that holds only because an unrelated limit happens to bind first is not a limit. Say the word and I will drop them and leave the sweep alone.

Full `-p tari_engine` suite: 425 passed.

sdbondi: _Posted by Claude._

Settled on sweep-only — `377af6f8` reverted in `4f2a4eb1`. `validate_finalized` is the single check; `new_substate` keeps its own so creation still fails at the creating instruction.

Two things decided it.

**The sweep is the cheaper half, not the expensive one.** It measures each substate once per transaction. An ingress check measures it once per *write*, and the `#[template]` macro emits a `SetState` at the end of every `&mut self` method — so a transaction calling five mutating methods pays five walks against the sweep's one. Removing the sweep for a type removes the 1, not the 5.

**Skipping a type in the sweep turns coverage into a standing claim.** The sweep covers every path not because anyone enumerated them but because `mutated_substates()` *is* the set about to be persisted. A `continue` for components would instead assert that every component write routes through `modify_component_with` — true today, and `get_component_mut` and `get_locked_substate_mut` are both public and hand out unchecked `&mut`s. A future caller of either would silently exempt components, with no compile error and no test failure, and the sweep is also what rejects a transaction *touching* an already-oversized substate, so it would stay unbounded rather than be repairable.

For the record, from enumerating the paths while working this out:

- **Components** were fully covered by the ingress check — `SetState` and `SetAccessRules` both route through `modify_component_with`, and `update_component_template` only swaps a 32-byte address.
- **Resources** were not. `update_access_rule` takes an `AccessRule`, and `RestrictedAccessRule::AnyOf(Box<[RestrictedAccessRule]>)` is recursive and bounded only by the call size; `set_auth_hook` is a third arg-driven path. Metadata was not the only route, so the ingress set was incomplete for resources anyway.
- None of it could fire in any case: WASM's 2 MiB linear memory caps a component's state near 250 KiB, a quarter of `max_substate_size`.

sdbondi (commented):
reviewed

PR #2592, opened 2026-09-11
fix(engine): make the workspace deterministic and enforce the lock scope

sdbondi:
Engine audit wave 4, second of four. Two hygiene items in the call-frame and workspace machinery. Independent of #2591; both touch `working_state.rs`, so whichever lands second needs a trivial rebase.

## Deterministic workspace collections

`Workspace.items` was a `HashMap` and `Workspace.proofs` a `HashSet`, and both get iterated on paths that reach consensus:

- `drain_all_proofs` drives the order `drop_proof` runs in (`tracker.rs` fee checkpoint, `WorkspaceAction::DropAllProofs`, `WorkspaceAction::DropAll`). Dropping a vault-backed proof can append that vault to `new_substates`, and that order is hashed into the `TransactionReceipt` diff.
- `all_ids_iter` lands in `ItemNotOnWorkspace` reject strings. These are local/UX only — a block's `TransactionAtom` carries `Decision`, `AbortReason` collapses every execution failure to `ExecutionFailure`, and the receipt carries an outcome with no reason — so this half is cosmetic. The `drain_all_proofs` half stands on its own.

Neither diverges today — a vault-backed proof has already mutated its vault by the time it is dropped, so the append is a no-op on ordering — but the safety is incidental, not structural. `CallScope` already carries a `NOTE: HashSet is not appropriate due to non-determinism`; `Workspace` now carries the same reasoning and the same collection types.

`WorkingState::address_allocations` has the same shape and is included: `get_allocated_address_by_address` scans it, so a hashed map would let iteration order decide which allocation a lookup finds.

## Lock scope

`CallScope::add_lock_to_scope`, `remove_lock_from_scope` and `is_lock_in_scope` had no callers. `lock_scope` was therefore always empty, and the `DanglingSubstateLocks` branch in `pop_frame` was unreachable — a callee could leak a lock and grief the rest of the caller's transaction with a lock nothing would ever release.

`lock_substate` now records into the current frame and `unlock_substate` releases. The release searches the frame stack rather than assuming the current frame, because the frame releasing a lock need not be the one that took it: a component lock is taken by the caller and released when the frame it was pushed into is popped, by which point that frame is off the stack. A lock taken before the first frame exists (fee settlement, transaction setup) belongs to no frame and stays untracked.

`remove_lock_from_scope` now reports whether the frame held the lock instead of returning an error, which is what a stack search needs; `is_lock_in_scope` is deleted rather than left dead.

I checked the check is genuinely live rather than newly-inert in a different way: disabling the release side turns every nested component call into `2 substate locks were still active after call`, and inverting the emptiness test fails every frame. Both directions behave, and the suite passes with the real code.

## Fee pool lock leak

`withdraw_fees_from_pool_up_to` took the fee pool's write lock and returned the container without ever releasing it. Because `ClaimValidatorFees` runs at instruction level there is no frame to record the lock against, so the new check cannot see it either — the one place the engine actually leaks a lock was exempt from the motivation for this PR. Fixed here. The live consequence was that a second `ClaimValidatorFees` against the same pool in one transaction failed with `MultipleWriteLockRequested`; it now succeeds.

## Tests

- `test_two_claims_against_one_pool_in_a_transaction` — two claims against one pool in one transaction. Fails on the pre-fix code with `Multiple write locks requested for substate vnfp_5c7f0fe…`.
- `it_lists_existing_workspace_ids_in_insertion_order` — three workspace outputs then a bad reference, asserting the reject string lists `[0, 1, 2]`. The existing sibling test only ever had one id on the workspace, so it could not see the order.
- Full `-p tari_engine` suite: 425 passed.

### DIFF
--- a/crates/engine/src/runtime/working_state.rs
+++ b/crates/engine/src/runtime/working_state.rs
@@ -145,8 +145,10 @@ pub(super) struct WorkingState<TStore> {
     events: Vec<Event>,
     logs: Vec<LogEntry>,
     buckets: HashMap<BucketId, Bucket>,
-    address_allocations: HashMap<AddressAllocationId, AllocatedAddress>,
-    used_address_allocations: HashMap<AddressAllocationId, SubstateId>,
+    /// Insertion-ordered: `get_allocated_address_by_address` scans these, so a hashed map would let the iteration
+    /// order decide which allocation a lookup finds.
+    address_allocations: IndexMap<AddressAllocationId, AllocatedAddress>,
+    used_address_allocations: IndexMap<AddressAllocationId, SubstateId>,
     address_allocation_id: u32,
     proofs: HashMap<ProofId, Proof>,
     object_ids: ObjectIds,
@@ -197,8 +199,8 @@ impl<TStore: StateReader> WorkingState<TStore> {
             buckets: HashMap::new(),
             proofs: HashMap::new(),
             address_allocation_id: 0,
-            address_allocations: HashMap::new(),
-            used_address_allocations: HashMap::new(),
+            address_allocations: IndexMap::new(),
+            used_address_allocations: IndexMap::new(),
 
             store: WorkingStateStore::new(state_store),
 
@@ -262,6 +264,12 @@ impl<TStore: StateReader> WorkingState<TStore> {
 
     fn lock_substate(&mut self, addr: SubstateId, lock_flag: LockFlag) -> Result<LockedSubstate, RuntimeError> {
         let lock_id = self.try_lock(addr.clone(), lock_flag)?;
+        // Every lock a frame takes must be released before that frame is popped, which `pop_frame` enforces. A lock
+        // taken before the first frame is pushed — fee settlement, transaction setup — belongs to no frame and is
+        // left untracked.
+        if let Some(frame) = self.call_frames.last_mut() {
+            frame.scope_mut().add_lock_to_scope(lock_id);
+        }
         Ok(LockedSubstate::new(addr, lock_id, lock_flag))
     }
 
@@ -295,6 +303,13 @@ impl<TStore: StateReader> WorkingState<TStore> {
 
     pub fn unlock_substate(&mut self, lock: LockedSubstate) -> Result<(), RuntimeError> {
         self.store.try_unlock(lock.lock_id())?;
+        // The frame releasing a lock need not be the one that took it: a component lock is taken by the caller and
+        // released when the frame it was pushed into is popped, by which point that frame is off the stack.
+        for frame in self.call_frames.iter_mut().rev() {
+            if frame.scope_mut().remove_lock_from_scope(lock.lock_id()) {
+                break;
+            }
+        }
         Ok(())
     }
 
@@ -1245,7 +1260,7 @@ impl<TStore: StateReader> WorkingState<TStore> {
         }
         let alloc_addr = self
             .address_allocations
-            .remove(&id)
+            .shift_remove(&id)
             .ok_or(RuntimeError::AddressAllocationNotFound { id })?;
         self.current_call_scope_mut()?.remove_address_allocation_from_scope(id);
         self.used_address_allocations
--- a/crates/engine/src/runtime/working_state.rs
+++ b/crates/engine/src/runtime/working_state.rs
@@ -148,6 +148,7 @@ pub(super) struct WorkingState<TStore> {
     /// Insertion-ordered: `get_allocated_address_by_address` scans these, so a hashed map would let the iteration
     /// order decide which allocation a lookup finds.
     address_allocations: IndexMap<AddressAllocationId, AllocatedAddress>,
+    /// Only ever inserted into and looked up by id. Ordered for symmetry with `address_allocations`.
     used_address_allocations: IndexMap<AddressAllocationId, SubstateId>,
     address_allocation_id: u32,
     proofs: HashMap<ProofId, Proof>,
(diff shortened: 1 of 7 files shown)


sdbondi (commented):
_Posted by Claude._

Reviewed at `0a02a93`. Nothing blocking — the mechanism is correct. Two notes, one of which is a factual correction to the rationale rather than to the code.

## What I verified

- **The stack search is unambiguous.** `LockedSubstates::next_id` is a monotonic counter with no reuse (`locking.rs:117-128`), so a `LockId` names exactly one lock for the life of the transaction and `unlock_substate`'s reverse walk cannot hit the wrong frame.
- **The component-lock case works because of the ordering in `pop_frame`.** The frame is popped at `working_state.rs:1588` *before* `unlock_substate(component_lock)` runs, so the search sees the caller's frame — which is where `lock_component` (`processor.rs:933`, before `push_call_frame`) recorded it — and the emptiness test that follows applies to the popped frame's own scope. Reversing those two lines would break it, which is worth knowing given how quietly it reads.
- **No lock escapes the accounting.** Every `store.try_lock` outside `lock_substate` (`working_state.rs:405`, `434`, `674`) pairs with a direct `store.try_unlock`, so those stay untracked on both sides and cannot leave a stale id behind to trip `DanglingSubstateLocks`.
- **Workspace ordering.** `insert` rejects duplicate ids, so `IndexMap` order is insertion order with no re-insert repositioning; `address_allocations` uses `shift_remove`, which keeps the map ordered by the sequential allocation id rather than swapping the tail in.

## Non-blocking

1. **The `all_ids_iter` half of the determinism rationale does not hold — validators do not compare reject strings.** A block's `TransactionAtom` carries `decision`, `evidence`, `transaction_fee`, `leader_fee` and nothing else (`command.rs:32-44`), and `Decision::Abort` carries `AbortReason` (`commit_result.rs:517-531`), which collapses every execution failure to `ExecutionFailure`. The hashed side is no better a witness: `TransactionReceipt` carries `outcome: FinalizeOutcome` — `Commit` or `FeeIntentCommit` — and no reason at all. So `ItemNotOnWorkspace`'s id list is a local/UX string.

   The `drain_all_proofs` half *is* real, and on its own justifies the change: the order of `new_substates` becomes the order of `substates_to_persist`, which `DiffSummary::from_diff` walks into `upped`, which is encoded into the receipt substate and hashed. Worth correcting in the body and in the `Workspace` doc comment (`workspace.rs:22-25`), which repeats the "validators compare" claim — a future reader who checks it will find it false and may discount the real reason next to it.

2. **The check cannot see the one lock the engine actually leaks.** `withdraw_fees_from_pool_up_to` (`working_state.rs:1285-1316`) takes the fee pool's write lock and returns the container without ever unlocking it. `ClaimValidatorFees` runs at instruction level, so `lock_substate` finds no frame, leaves it untracked, and `DanglingSubstateLocks` stays quiet. The live consequence today: a second `ClaimValidatorFees` against the same pool in one transaction fails with `MultipleWriteLockRequested`, since the first never released. It is one line to fix and this PR is the natural home for it — otherwise the "a callee could leak a lock and grief the rest of the transaction" motivation is served everywhere except the place it already happens.

## Nit

- `used_address_allocations` is only ever `insert`ed and `get`-ed (`working_state.rs:1250`, `1268`) — never iterated — so its `IndexMap` change is symmetry with its sibling rather than load-bearing. Fine to keep; just not the same claim as `address_allocations`, whose `.values()` scan in `get_allocated_address_by_address` is the one that matters.

crates/engine/src/runtime/working_state.rs
```diff
@@ -1229,7 +1245,7 @@ impl<TStore: StateReader> WorkingState<TStore> {
         }
         let alloc_addr = self
             .address_allocations
-            .remove(&id)
+            .shift_remove(&id)
```
sdbondi: Would `swap_remove` break anything? I don't see how insert order is important for address_allocations.

crates/engine/src/runtime/workspace.rs
```diff
@@ -57,7 +58,7 @@ impl Workspace {
         Ok(())
     }
 
-    pub fn drain_all_proofs(&mut self) -> HashSet<ProofId> {
+    pub fn drain_all_proofs(&mut self) -> IndexSet<ProofId> {
```
sdbondi:
```suggestion
    pub fn take_all_proofs(&mut self) -> IndexSet<ProofId> {
```

crates/engine/src/runtime/working_state.rs

You're right — `swap_remove` in `dbdce0b`. I conflated two things: what this map needs is an iteration order that follows from the transaction's own operations rather than from a hash seed, and `swap_remove` gives that just as well as `shift_remove` does. Insertion order specifically is not load-bearing, and `swap_remove` is O(1).

Corrected the field comment too. It claimed insertion order mattered; what actually matters is that two allocations *can* name the same address — `derive_new_component_address(template, public_key)` is a pure function of its inputs — so the scan in `get_allocated_address_by_address` has to resolve the same way on every validator. Any deterministic order does that; a hash order does not.

crates/engine/src/runtime/workspace.rs
sdbondi: _Posted by Claude._

Applied in `dbdce0b`, renamed at all five call sites. Agreed — `drain` reads as a `Vec::drain`-style iterator, and this returns the whole set out of a `mem::take`.

sdbondi (commented):
Reviewed

PR #2593, opened 2026-09-11
fix(engine)!: resource and proof lifecycle corrections

sdbondi:
Engine audit wave 4, third of four. Six items in the resource and proof lifecycle. Independent of #2591 and #2592; all three touch `working_state.rs`, so whichever lands later needs a small rebase.

## Consuming a bucket with locked funds

A bucket whose funds a proof has locked may only be held, never consumed — every operation that empties one works on the unlocked funds alone, so consuming a locked bucket destroys the locked portion while the proof still names it.

Deposit and burn checked for this. Two paths did not:

- `Bucket::join` moved the source bucket in with no check. The locked funds went away and the proof failed later with nothing pointing at the cause.
- `PayFee::FromBucket` likewise. This one was caught eventually by `validate_finalized`, but only after the fee had been metered against a bucket that could not pay it.
- `execute_stealth_transfer`'s revealed-funds bucket, which is compared on `unlocked_amount()` and then dropped, so a bucket whose unlocked amount matches the statement passes with its locked portion destroyed.

All five now reject up front. The error carries the operation, so `InvalidOpDepositLockedBucket` becomes `InvalidOpLockedBucket` and stops claiming every failure is a deposit.

`BucketAction::DropEmpty` needs no such check: it gates on `ResourceContainer::is_empty()`, whose confidential arm counts `commitments.len() + locked_commitments.len()`, so a bucket holding locked commitments is not empty and does not drop.

## `GetOwnerProof` on a component created in the same transaction

It read through `get_unmodified_substate`, which sees the backing store and the loaded set but not what this transaction has written. Asking for the owner proof of a component created earlier in the same transaction therefore failed with `SubstateNotFound` — the create-then-use shape that address allocations exist to support.

It now reads the value the transaction has, through a new `WorkingStateStore::get_latest_substate`. The owner rule is immutable, which is why this read needs no lock; that reasoning is unchanged and now stated where it applies.

## Auth hooks and the resource write lock

Mint, UpdateMetadata and Burn take the resource's write lock, then invoke the resource's auth hook while still holding it. A hook that reads the resource it guards — `total_supply()`, `resource_type()`, anything — is refused:

```
Resource Auth Hook Denied Access for action native.resource.Mint: Runtime error: Lock error:
Requested Read lock on substate resource_ed5e... but it is already locked with Write
```

The lock is released across the hook call and retaken after it. Nothing can slip in between: the hook frame runs in `FrameWriteMode::OwnComponent` and cannot write, so what it reads is what the operation goes on to alter. The eight other hook sites hold a read lock, which stacks, and are untouched.

## `validate_finalized`

- Vaults were tested with `locked_balance()`, which reports zero for a confidential vault whose locked value is a set of commitments with hidden amounts. `has_locked_funds()` exists for exactly this distinction and is now used; `Vault` grows the forwarding accessor `ResourceContainer` already had.
- The dangling-bucket count included buckets that had already been emptied. An emptied bucket carries nothing, so only buckets that still hold funds are counted.
- `TransactionCommitError::WorkspaceNotEmpty` was never constructed. Deleted rather than left as a claim the engine does not make.

## `withdraw_confidential` change guard

```rust
if *revealed_amount < validated_proof.change_revealed_amount { ...InsufficientBalance... }
*revealed_amount += validated_proof.change_revealed_amount;
```

The guard compares the remaining revealed amount against the change and then *adds* the change to it, so it can never fire. The balance proof is what constrains the change; the only thing that can go wrong on that line is the sum. It is now a checked add reporting `BalanceOverflow`.

## Id counters

`next_proof_id` incremented the bucket counter. That is now deliberate rather than incidental: one counter, named `bucket_or_proof_id`, shared by both. Buckets and proofs are transient handles in the same runtime scope, and a shared counter guarantees an id is never both.

The counters also lose their atomics and their `Arc`. Execution is single threaded, and `Runtime` holds a `NonNull` to one boxed `RuntimeInterface`, so `runtime.clone()` shares that single instance — no copy ever needed a counter of its own.

**Breaking.** `IdProvider::new` takes `&mut ObjectIds`; the `IdProvider`, `ObjectIds` and `EntityIdProvider` allocators take `&mut self`; `EntityIdProvider` is no longer `Clone`; `RuntimeInterface::next_entity_id` takes `&mut self`. `tari_engine_types` and `tari_engine` are both published, hence the `!` on the title.

## Tests

- `it_rejects_joining_a_bucket_with_locked_funds` — a template locks part of a bucket with a proof and joins it into another.
- `it_allows_a_hook_to_read_the_resource_it_guards` — a mint from outside the hook's own component, with a hook that reads the resource. Verified it fails on the pre-fix code with the `InvalidLockRequest` above.
- `transfer_rejects_a_revealed_funds_bucket_with_locked_funds` — a stealth transfer handed a bucket a proof has locked. A bucket proof locks the whole bucket, so the statement reveals no input: that is the shape whose unlocked amount matches and reaches the drop. The proof is dropped in the same transaction so the dangling-proof check does not stand in for the guard. Verified it fails on the pre-fix code with `Bucket not found with id BucketId(1)` at the drop.
- `test_get_owner_proof_for_a_component_created_in_the_same_transaction`.
- Full `-p tari_engine -p tari_engine_types` suite: 534 passed.

### DIFF
--- a/crates/engine/src/runtime/working_state.rs
+++ b/crates/engine/src/runtime/working_state.rs
@@ -1516,15 +1516,17 @@ impl<TStore: StateReader> WorkingState<TStore> {
         Ok(frame.current_template_name())
     }
 
-    pub fn id_provider(&self) -> Result<IdProvider<'_>, RuntimeError> {
-        self.call_frames
+    pub fn id_provider(&mut self) -> Result<IdProvider<'_>, RuntimeError> {
+        let entity_id = self
+            .call_frames
             .last()
-            .map(|frame| IdProvider::new(frame.entity_id(), self.transaction_hash, &self.object_ids))
-            .ok_or(RuntimeError::NoActiveCallFrame)
+            .map(|frame| frame.entity_id())
+            .ok_or(RuntimeError::NoActiveCallFrame)?;
+        Ok(IdProvider::new(entity_id, self.transaction_hash, &mut self.object_ids))
     }
 
-    pub fn id_provider_for_entity(&self, entity_id: EntityId) -> IdProvider<'_> {
-        IdProvider::new(entity_id, self.transaction_hash, &self.object_ids)
+    pub fn id_provider_for_entity(&mut self, entity_id: EntityId) -> IdProvider<'_> {
+        IdProvider::new(entity_id, self.transaction_hash, &mut self.object_ids)
     }
 
     pub fn new_bucket_id(&mut self) -> BucketId {
--- a/crates/engine/src/runtime/working_state.rs
+++ b/crates/engine/src/runtime/working_state.rs
@@ -2172,6 +2172,17 @@ impl<TStore: StateReader> WorkingState<TStore> {
         }
 
         let revealed_funds_bucket = revealed_funds_bucket_id.map(|id| self.take_bucket(id)).transpose()?;
+        // The bucket is consumed whole by the transfer, so funds a proof has locked in it would be destroyed while
+        // the proof still names it.
+        if let (Some(bucket_id), Some(bucket)) = (revealed_funds_bucket_id, revealed_funds_bucket.as_ref()) &&
+            bucket.has_locked_funds()
+        {
+            return Err(RuntimeError::InvalidOpLockedBucket {
+                op: "stealth transfer from",
+                bucket_id,
+                locked_amount: bucket.locked_amount(),
+            });
+        }
         if let Some(ref bucket) = revealed_funds_bucket &&
             *bucket.resource_address() != resource_address
         {
--- a/crates/engine/src/runtime/working_state.rs
+++ b/crates/engine/src/runtime/working_state.rs
@@ -557,9 +557,11 @@ impl<TStore: StateReader> WorkingState<TStore> {
             Self::enforce_substate_size_limit(id, value)?;
         }
 
-        if self.buckets.iter().any(|(_, b)| !b.is_empty()) {
+        // An emptied bucket carries nothing and is tolerated, so the count is of those that are not empty.
+        let dangling_buckets = self.buckets.iter().filter(|(_, bucket)| !bucket.is_empty()).count();
+        if dangling_buckets > 0 {
             return Err(TransactionCommitError::DanglingBuckets {
-                count: self.buckets.len(),
+                count: dangling_buckets,
             }
             .into());
         }
@@ -579,7 +581,9 @@ impl<TStore: StateReader> WorkingState<TStore> {
         }
 
         for (vault_id, vault) in self.store.new_vaults() {
-            if !vault.locked_balance().is_zero() {
+            // A confidential vault's locked value is a set of commitments whose amounts are hidden, so the locked
+            // balance alone reports zero for it.
+            if vault.has_locked_funds() {
                 return Err(TransactionCommitError::DanglingLockedValueInVault {
                     vault_id,
                     locked_amount: vault.locked_balance(),
@@ -792,7 +796,8 @@ impl<TStore: StateReader> WorkingState<TStore> {
         // it. Callers reject this earlier to avoid charging for a burn that cannot succeed; the check lives here so
         // that it holds for every caller.
         if bucket.has_locked_funds() {
-            return Err(RuntimeError::InvalidOpDepositLockedBucket {
+            return Err(RuntimeError::InvalidOpLockedBucket {
+                op: "burn",
                 bucket_id,
                 locked_amount: bucket.locked_amount(),
             });
--- a/crates/engine/tests/shenanigans.rs
+++ b/crates/engine/tests/shenanigans.rs
@@ -15,11 +15,13 @@ use tari_template_lib::{
 use tari_template_test_tooling::{TemplateTest, support::assert_error::assert_reject_reason};
 
 const CRATE_PATH: &str = env!("CARGO_MANIFEST_DIR");
+const TEMPLATE_PATHS: &[&str] = &["tests/templates/shenanigans"];
+const RESOURCE_TEMPLATE_PATHS: &[&str] = &["tests/templates/resource_shenanigans"];
 const TEMPLATE_NAME: &str = "Shenanigans";
 
 #[test]
 fn it_rejects_dangling_vaults_in_constructor() {
-    let mut test = TemplateTest::new(CRATE_PATH, ["tests/templates/shenanigans"]);
+    let mut test = TemplateTest::new(CRATE_PATH, TEMPLATE_PATHS);
     let template_addr = test.get_template_address(TEMPLATE_NAME);
 
     let reason = test.execute_expect_failure(
@@ -38,7 +40,7 @@ fn it_rejects_dangling_vaults_in_constructor() {
 
 #[test]
 fn it_rejects_dangling_vault_that_has_been_returned() {
-    let mut test = TemplateTest::new(CRATE_PATH, ["tests/templates/shenanigans"]);
+    let mut test = TemplateTest::new(CRATE_PATH, TEMPLATE_PATHS);
     let template_addr = test.get_template_address(TEMPLATE_NAME);
 
     let reason = test.execute_expect_failure(
@@ -53,7 +55,7 @@ fn it_rejects_dangling_vault_that_has_been_returned() {
 
 #[test]
 fn it_rejects_dangling_vaults_in_component() {
-    let mut test = TemplateTest::new(CRATE_PATH, ["tests/templates/shenanigans"]);
+    let mut test = TemplateTest::new(CRATE_PATH, TEMPLATE_PATHS);
     let template_addr = test.get_template_address(TEMPLATE_NAME);
 
     //  Create with vault
@@ -84,7 +86,7 @@ fn it_rejects_dangling_vaults_in_component() {
 
 #[test]
 fn it_rejects_dangling_resources() {
-    let mut test = TemplateTest::new(CRATE_PATH, ["tests/templates/shenanigans"]);
+    let mut test = TemplateTest::new(CRATE_PATH, TEMPLATE_PATHS);
     let template_addr = test.get_template_address(TEMPLATE_NAME);
 
     let reason = test.execute_expect_failure(
@@ -99,7 +101,7 @@ fn it_rejects_dangling_resources() {
 
 #[test]
 fn it_rejects_unknown_substate_ids() {
-    let mut test = TemplateTest::new(CRATE_PATH, ["tests/templates/shenanigans"]);
+    let mut test = TemplateTest::new(CRATE_PATH, TEMPLATE_PATHS);
     let template_addr = test.get_template_address(TEMPLATE_NAME);
 
     let reason = test.execute_expect_failure(
@@ -121,7 +123,7 @@ fn it_rejects_unknown_substate_ids() {
 
 #[test]
 fn it_rejects_references_to_buckets_that_arent_in_scope() {
-    let mut test = TemplateTest::new(CRATE_PATH, ["tests/templates/shenanigans"]);
+    let mut test = TemplateTest::new(CRATE_PATH, TEMPLATE_PATHS);
     let template_addr = test.get_template_address(TEMPLATE_NAME);
     let (account, owner_token, owner_key) = test.create_funded_account();
 
@@ -151,7 +153,7 @@ fn it_rejects_references_to_buckets_that_arent_in_scope() {
 
 #[test]
 fn it_rejects_double_ownership_of_vault() {
-    let mut test = TemplateTest::new(CRATE_PATH, ["tests/templates/shenanigans"]);
+    let mut test = TemplateTest::new(CRATE_PATH, TEMPLATE_PATHS);
     let template_addr = test.get_template_address(TEMPLATE_NAME);
 
     let reason = test.execute_expect_failure(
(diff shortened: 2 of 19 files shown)


sdbondi (commented):
_Posted by Claude._

Reviewed at `fe9f310`. Nothing blocking. The six items each hold up; the substantive note is that the locked-bucket sweep misses two more consumption paths of the same class.

## What I verified

- **`join` really does destroy the locked portion.** `Bucket::join` is `resource_container.deposit(other.resource_container)` (`bucket.rs:122-124`), and `deposit` destructures `other` down to `amount` / `token_ids` / `commitments` + `revealed_amount` (`resource_container.rs:276-324`) — every `locked_*` field on the source is dropped on the floor. Same for `PayFee::FromBucket` via `take_all`.
- **The hook unlock/relock is safe.** The claim rests on an `OwnComponent` frame having no write path at all, and it does: `proof_invoke` and `bucket_invoke` are refused outright (`impl.rs:274-275`), every vault and resource mutation goes through `write_lock_substate` → `check_write_allowed`, and the hook's own component is written through the lock taken at push rather than a new one. So nothing can alter the resource between the unlock and the relock, and the fields captured under the first lock (`auth_caller`, `has_view_key`, `tracks_supply`) stay valid. The auth decision made before the hook is equally safe: the hook cannot reach the access rules, and rules evaluate against the auth scope rather than component state.
- **`get_latest_substate`'s lock-free read is justified.** There is no `SetOwnerRule` in `ComponentAction` (`template_lib/src/args/types.rs:87-100`) and no setter on `ComponentHeader`, so the owner rule is immutable in fact and not just by intent. The precedence — written, then loaded, then store — is the right order for "what this transaction has".
- **`withdraw_confidential`.** `validate_confidential_statement` rejects a negative `change_revealed_amount` (`confidential/validation.rs:48`) and the balance proof binds its magnitude, so the sum is genuinely the only thing that can fail on that line. The removed guard compared the post-subtraction balance against a value about to be *added* to it.
- **Proof ids.** `ProofId` and `BucketId` index separate maps and are distinguished by CBOR tag in `IndexedValue`, so sharing a counter was cosmetic — no collision was possible and none is created by splitting it.
- The old `DanglingBuckets { count: self.buckets.len() }` counted *every* bucket, empty ones included, while the condition that fired was "any non-empty". The new filter fixes a real miscount, not just the wording.

## Non-blocking

1. **Two more paths consume a bucket without the new check.** Both are the class the PR opens with — the failure surfaces later with nothing pointing at the cause:

   - **`BucketAction::DropEmpty`** (`impl.rs:3281-3292`) gates on `bucket.is_empty()`, which is `locked_amount().is_zero() && unlocked_amount().is_zero() && number_of_confidential_commitments() == 0` (`resource_container.rs:946-950`). `number_of_confidential_commitments()` counts only the unlocked `commitments` set, so a confidential bucket whose entire value sits in `locked_commitments` reports empty, gets dropped, and the locked commitments vanish while the proof still names the bucket. This is exactly the `locked_balance()` vs `has_locked_funds()` distinction this PR draws for vaults two hunks earlier. `validate_finalized`'s dangling-bucket filter uses the same `is_empty()`, so it does not catch such a bucket either — the transaction dies later at `drop_proof` with `BucketNotFound`, or at `DanglingProofs`.
   - **`execute_stealth_transfer`'s revealed-funds bucket** (`working_state.rs:2134`): `take_bucket`, then a comparison against `bucket.unlocked_amount()`, then the bucket is simply dropped. A bucket with locked funds passes the comparison and the locked portion is destroyed.

   Either fold both into `check_bucket_is_unlocked`, or teach `ResourceContainer::is_empty` to consult `has_locked_funds()` — the latter covers `DropEmpty` and the dangling-bucket check at once.

2. **`DanglingBuckets { bucket_ids }` is collected from a `HashMap`** (`working_state.rs:543-549`, `buckets: HashMap<BucketId, Bucket>`), so the ids come out in a different order run to run. I checked whether that reaches consensus and it does not — blocks carry only `Decision`/`AbortReason` and the receipt carries no reason string — so this is a local/UX wart rather than a divergence. Still worth sorting: the sibling PR in this same wave exists to remove precisely this shape, and the `count` it replaces was deterministic.

## Nit

- `relock_resource_for_write` runs unconditionally, so a resource with no auth hook now does unlock-then-relock for nothing. Harmless, and the lock is needed on both paths, so the only waste is the release in the no-hook case — but the closure could hand the lock back when `auth_hook` is `None` if you want the common path unchanged.

crates/engine/src/runtime/working_state.rs
```diff
@@ -541,9 +541,17 @@ impl<TStore: StateReader> WorkingState<TStore> {
...
+            .buckets
+            .iter()
+            .filter(|(_, bucket)| !bucket.is_empty())
+            .map(|(id, _)| *id)
+            .collect::<Vec<_>>();
+        dangling_buckets.sort_unstable();
+        if !dangling_buckets.is_empty() {
             return Err(TransactionCommitError::DanglingBuckets {
-                count: self.buckets.len(),
+                bucket_ids: dangling_buckets,
```
sdbondi: We dont need the bucket ids in the error, not worth the allocation

crates/engine/tests/shenanigans.rs
```diff
@@ -492,6 +492,33 @@ fn it_rejects_a_bucket_that_is_neither_consumed_nor_returned_by_a_call() {
     assert_reject_reason(reason, "were neither consumed nor returned by the call");
 }
 
+/// `Bucket::join` moves only the unlocked funds, so joining a bucket a proof has locked would destroy the locked
+/// portion while the proof still names it.
+#[test]
+fn it_rejects_joining_a_bucket_with_locked_funds() {
+    let mut test = TemplateTest::new(CRATE_PATH, ["tests/templates/shenanigans"]);
```
sdbondi: Move ["tests/templates/shenanigans"] into a constant to keep it consistent with other tests

crates/engine_types/src/id_provider.rs
```diff
@@ -131,6 +131,7 @@ pub struct ObjectIds {
     max_ids: usize,
     current_id: AtomicU32,
     bucket_id: AtomicU32,
+    proof_id: AtomicU32,
```
sdbondi: We shouldnt need another counter for proof id, maybe rename bucket_id to bucket_or_proof_id (or better name) for clarity and only use one.

crates/engine/src/runtime/working_state.rs

Reverted — `DanglingBuckets` is back to `{ count: usize }`, no `Vec<BucketId>` and no allocation. The semantic part of the change stays: the count is of buckets that still hold funds, so an emptied bucket left in the working state no longer aborts the transaction.

crates/engine/tests/shenanigans.rs
sdbondi: _Posted by Claude._

Done — added `TEMPLATE_PATHS`, `RESOURCE_TEMPLATE_PATHS`, `TEMPLATE_NAME` and `CRATE_PATH` at the top of the file and applied them to all 25 call sites, not just the new one.

sdbondi (commented):
_Posted by Claude._

Re-reviewed the force-push, `fe9f310` → `cd34fd4` (rebased onto development, base commit amended, four new commits). CI still running at the time of writing. Both items from my last review are resolved; the notes below are about the two new refactor commits and the description, not about the lifecycle fixes.

## Resolved

- **Stealth revealed-funds bucket** (`4247dfa`). The check sits before the resource-address and amount comparisons so it is the error that fires, uses `has_locked_funds()` rather than `locked_amount()` so it covers confidential locked commitments, and takes-then-checks like the sibling paths. Correct.
- **`DanglingBuckets` ordering**, by reverting to a count. That resolves it — and the non-empty filter survives the revert, so the real miscount (`self.buckets.len()` counting empty buckets too) is still fixed.

## Still open from the last review

**`BucketAction::DropEmpty` is a sixth consumption path.** `impl.rs:3275-3292` gates on `bucket.is_empty()`, which is `locked_amount().is_zero() && unlocked_amount().is_zero() && number_of_confidential_commitments() == 0` (`resource_container.rs:946-950`). For a confidential bucket whose value sits entirely in `locked_commitments`, all three are true while `has_locked_funds()` is false — so it drops, and the commitments go with it. `validate_finalized`'s dangling-bucket filter uses the same `is_empty()` and misses it for the same reason. The body now says "All five now reject up front", which reads as a closed set. Teaching `ResourceContainer::is_empty` to consult `has_locked_funds()` covers both sites at once, and is the same distinction the vault hunk in this PR already draws.

## New

1. **The description contradicts the code in two places.**
   - "Dangling buckets are reported by id instead of by count… The ids come out of a hashed map, so they are sorted." The amended base commit reverted this; `DanglingBuckets { count: usize }` is what ships, with no ids and no sort. `4247dfa`'s commit message carries the same stale paragraph, describing a change that is in neither that commit nor the branch.
   - "## Proof ids — `next_proof_id` incremented the bucket counter. Proofs get their own." `1e424c24` reverses that: the counter is now deliberately shared and named `bucket_or_proof_id`. Net behaviour against development is unchanged — which is fine, and the shared-counter rationale is better than the split — but the section claims the opposite of what the branch does.

2. **The atomics refactor (`cd34fd4`) is the widest change on the branch and is not in the description at all.** Seven files, two crates, and public API: `IdProvider::new` now takes `&mut ObjectIds`, most `IdProvider`/`ObjectIds` methods take `&mut self`, `EntityIdProvider` loses `Clone`, and `RuntimeInterface::next_entity_id` takes `&mut self`. `tari_engine_types` and `tari_engine` are both published tier-3 crates (`scripts/publish_crates.py:44,53`), so that is a breaking public change in a PR titled `fix(engine):` without the `!` its siblings in this wave carry. Either describe it and mark it, or split it out — it has no behavioural motivation shared with the lifecycle fixes.

   The refactor itself is sound, and I checked the parts that could have bitten:
   - **Dropping `EntityIdProvider`'s `Arc<AtomicU32>` is safe because `Runtime` is a `NonNull<Box<dyn RuntimeInterface>>`** (`runtime/mod.rs:307-309`) — `runtime.clone()` shares one interface instance, so no second copy ever needed a shared counter. "Execution is single threaded" is not by itself the reason the `Arc` could go; a by-value clone would have handed out duplicate entity ids regardless of threading. Worth saying in the commit message, because that is the fact a future reader would have to re-derive.
   - `ObjectIds`'s derived `Clone` matches the hand-written one it replaces — snapshot, not share — so the fee-checkpoint clone of `WorkingState` is unchanged.
   - `next_id` and `EntityIdProvider::next` no longer increment on the error path. Observationally identical: successful ids are still `0..max_ids-1` and `MaxIdsExceeded` aborts the transaction.
   - `read_with` → `write_with` in `generate_uuid`/`get_pseudorandom_bytes` is behaviour-neutral; both are plain accessors (`tracker.rs:626-638`).

3. **The new stealth test does not exercise the case the fix exists for.** `Bucket::create_proof` is the only lock action a template has (`BucketAction` has no partial-lock variant) and it calls `lock_all` (`impl.rs:3222`), so a bucket proof always locks the whole bucket and `unlocked_amount()` is zero. The test pairs that with a statement whose revealed input is 100, so pre-fix it fails at "Revealed funds bucket amount (0) does not match…" rather than destroying anything — it pins the new error, but nothing vanishes in it. The reachable fund-destroying shape is a statement with revealed input **0** plus a fully locked bucket: `bucket.unlocked_amount() != 0` passes, the bucket is dropped, and every locked commitment in it is gone. That is the case worth asserting.

## Nit

- `bucket_or_proof_id += 1` and `uuid += 1` are unchecked and, unlike `current_id`, have no `max_ids` gate. Unreachable — metering dies orders of magnitude before 2^32 — but #2594, merged two commits below this branch's base, added `checked_add` to three counters on exactly the "not reachable today; the point is the failure mode if it ever is" reasoning, and moving off atomics turns a silent wrap into a debug-build panic. Either gate them the same way or note why these two differ.
- `53185cb` ("hoist the shenanigans template paths into constants") also changes an error string, `{count} dangling bucket(s)` → `{count} dangling buckets`. It is tidying its own earlier hunk, but it is invisible from that subject.

crates/engine/tests/shenanigans.rs
```diff
@@ -492,9 +494,36 @@ fn it_rejects_a_bucket_that_is_neither_consumed_nor_returned_by_a_call() {
...
+        .unwrap();
+
+    let reason = test.execute_expect_failure(
+        test.transaction()
+            .call_method(component, "join_locked_bucket", args![])
+            .build_and_seal(test.secret_key()),
+        vec![],
+    );
+
+    assert_reject_reason(reason, "Cannot join bucket");
```
sdbondi: I want to avoid using execution_results (potentially to remove it in future for perf/storage/wire savings). This transaction can be done in a single transaction, which also makes the test faster.

```suggestion
    let reason = test.execute_expect_failure(
        test.transaction()
            .call_function(template_addr, "with_fungible_vault", args![])
            .put_workspace("component")
            .call_method("component", "join_locked_bucket", args![])
            .build_and_seal(test.secret_key()),
        vec![],
    );

    assert_reject_reason(reason, "Cannot join bucket");
```

crates/engine/tests/stealth.rs
```diff
@@ -375,6 +375,70 @@ fn transfer_revealed_between_accounts() {
...
+            mask: mint.output_masks[0].clone(),
+            value: 100,
+        }],
+        0u64,
+        [40, 60],
+        0u64,
+    );
+
+    let reason = test.execute_expect_failure(
+        Transaction::builder_localnet(Epoch(1))
```
sdbondi:
```suggestion
        test.transaction()
```

crates/engine/tests/shenanigans.rs
sdbondi: _Posted by Claude._

Applied in `843b820`. The component address now comes off the workspace via `put_workspace("component")`, so the test is one transaction with no `execution_results` read.

crates/engine/tests/stealth.rs
sdbondi: _Posted by Claude._

Applied in `843b820`. The rest of the chain is unchanged — `test.transaction()` returns the same `TransactionBuilder`, so `.finish()`, the three `add_signer` calls and `.seal(&alice_sk)` still apply; it just picks up the harness epoch and nonce instead of a hardcoded `Epoch(1)`.

sdbondi (commented):
Reviewed

PR #2594, opened 2026-09-11
chore(engine): small hardening

sdbondi:
Engine audit wave 4, last of four. Defence-in-depth items with no behaviour change a template can see. Independent of #2591, #2592 and #2593; small overlaps in `working_state.rs` and `error.rs`, so whichever lands later needs a rebase.

## MSM pricing decoded its argument twice

`price(RISTRETTO_MSM)` did `args.get::<Vec<RistrettoPublicKeyBytes>>(0)?.len()` — a full CBOR decode of the point vector to read one number — and `dispatch` then decoded the same vector again. Pricing must not do work proportional to the input it is pricing: that first decode runs *before* anything is charged, so a caller gets it for free, and `max_call_size` allows a few thousand points.

The term count now comes from the encoded length of argument 0 divided by the smallest a point can encode to (32 bytes plus a CBOR byte-string header). That bounds the count from above, so the charge is never short, and a new unit test pins the per-point minimum so a future encoding change cannot silently make the bound wrong. This matches how `per_byte` already prices the hash and Schnorr intrinsics.

## Unchecked arithmetic

Three sites get what `LockedSubstates::next_id` already had:

- `address_allocation_id += 1` — u32 counter in `new_address_allocation`
- `existing_state.version() + 1` — u32 version bump in `generate_substate_diff`
- `*count_mut -= 1` — usize read-lock refcount in `try_unlock`

None is reachable today. The point is the failure mode if one ever becomes reachable: the release profile is `panic = 'abort'`, so an overflow takes the validator down rather than failing the transaction.

## `IsNotFoundError for RuntimeError` was too wide

It classified eight variants as not-found: `SubstateNotFound`, `ComponentNotFound`, `VaultNotFound`, `BucketNotFound`, `ResourceNotFound`, `NonFungibleNotFound`, `ProofNotFound`, `VirtualSubstateNotFound`.

Most of those are not "the ledger does not have this". A missing bucket or proof means a call is wrong about what is in its scope — which is exactly the kind of error that must surface, not become `None`. No current `.optional()` caller reaches one, but the trait is an open invitation: the next `.optional()` added anywhere in the engine would swallow them silently.

Narrowed to `SubstateNotFound` and a not-found `StateStoreError`, which is what all six existing callers mean. `ComponentNotFound` is never constructed anywhere in the workspace, so it goes too.

## Documentation

Passing a `Proof` as an argument authorizes it in the callee, with no `authorize()` call — `include_refs_in_scope` routes argument proofs through `add_proof_to_scope`, which adds to the auth scope. `add_proof_to_scope` said so; the crossing point did not, which is where a reader is when the question arises. Now noted there, including the consequence: a caller that wants a callee to merely inspect a proof has no way to say so.

## Already done elsewhere

Two items from the audit list were fixed by wave 1 and needed nothing here: the `expect`s in `finalize_fees_and_refunds` (#2579) and the stale comment in `check_component_scope` (#2575).

## Tests

- `encoded_point_len_is_at_least_the_minimum` — pins the constant the MSM price divides by.
- Full `-p tari_engine -p tari_engine_types` suite: 512 passed.

### DIFF
--- a/crates/engine/src/intrinsics/mod.rs
+++ b/crates/engine/src/intrinsics/mod.rs
@@ -43,6 +43,10 @@ use tari_template_lib::{
 
 use crate::runtime::{EngineArgs, RuntimeError};
 
+/// The fewest bytes a [`RistrettoPublicKeyBytes`] can occupy inside an encoded list: 32 bytes of key and the
+/// CBOR byte-string header in front of them. `encoded_point_len_is_at_least_the_minimum` pins it.
+const MIN_ENCODED_POINT_LEN: u64 = 33;
+
 /// The metering points an intrinsic call costs, derived from its declared arguments alone.
 ///
 /// Charged before [`dispatch`] runs, so the price may not depend on anything the work itself
@@ -59,12 +63,14 @@ pub fn price(intrinsic: IntrinsicId, args: &EngineArgs) -> Result<u64, RuntimeEr
         I::RISTRETTO_MUL => NativeExecutionPoints::PER_RISTRETTO_MUL,
         I::RISTRETTO_MUL_BASE => NativeExecutionPoints::PER_RISTRETTO_MUL_BASE,
         I::RISTRETTO_MSM => {
-            // Priced off the declared point count, which is read before any point is decompressed.
-            // A mismatch against the scalar count is rejected in `dispatch`, after this charge —
-            // a caller cannot get free work out of an argument this never looks at.
-            let points_len = args.get::<Vec<RistrettoPublicKeyBytes>>(0)?.len() as u64;
+            // Priced off the encoded length of the point list rather than a decoded one: pricing must not do
+            // work proportional to the input it is pricing, and that decode is unmetered. Dividing by the
+            // smallest a point can encode to bounds the term count from above, so the charge is never short. A
+            // mismatch against the scalar count is rejected in `dispatch`, after this charge — a caller cannot
+            // get free work out of an argument this never looks at.
+            let terms = args.arg_encoded_len(0) / MIN_ENCODED_POINT_LEN;
             NativeExecutionPoints::PER_RISTRETTO_MSM
-                .saturating_add(NativeExecutionPoints::PER_RISTRETTO_MSM_TERM.saturating_mul(points_len))
+                .saturating_add(NativeExecutionPoints::PER_RISTRETTO_MSM_TERM.saturating_mul(terms))
         },
         I::SCALAR_ADD |
         I::SCALAR_SUB |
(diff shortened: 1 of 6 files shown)


sdbondi (commented):
_Posted by Claude._

Reviewed at `7d0cf8c`. Nothing blocking, and the one item with real behaviour behind it — the `IsNotFoundError` narrowing — checks out against every existing caller.

## What I verified

- **The MSM bound is sound and tight.** A `RistrettoPublicKeyBytes` encodes as a 32-byte CBOR byte string, whose length needs the one-byte-length form (`0x58 0x20`), so 34 bytes per point. Dividing by 33 over-counts by ~3% and can never under-count, and the list header only adds to the numerator. `max_call_size` caps the result near 4k terms and the multiply saturates, so there is no way to turn the divisor into an overflow. The `len = 1000` case in the new test is the one that would actually catch a future 32-byte encoding — at small `len` the list header would mask it.
- **Pricing can no longer fail, and that is the right trade.** A malformed argument 0 is now charged MSM terms before `dispatch` rejects it, rather than erroring for free in `price`. That is the direction the pricing discipline wants, and a caller only pays it on a transaction that was going to fail anyway.
- **All six `.optional()` callers still mean what they meant.** `state_store.rs:195`, `236`, `268`, `306` and `working_state.rs:2035` bottom out in `get_unmodified_substate`/`get_state`, i.e. `SubstateNotFound` or a not-found `StateStoreError`. The one non-obvious caller is `processor.rs:716` — `CreateAccount`'s idempotent existence check — and `load_component` reaches `load_and_cache` (`state_store.rs:182-203`), which produces `SubstateNotFound` for a missing component. Its other two outcomes, `SubstateAlreadySpent` and `InvariantError`, were never in the old list and should not be swallowed.
- `ComponentNotFound` is constructed nowhere in the workspace — the only two occurrences are its own definition and the line being deleted — so removing it is dead-code removal, not a behaviour change.
- The three checked-arithmetic sites all return through existing error types with no new variants, and `generate_substate_diff` still has `id` available for the message at the point it is used.

## Nit

- The closing sentence of the new `include_refs_in_scope` doc (`scope.rs:315-320`) — "a caller that only wants the callee to inspect a proof has no way to say so" — records an audit observation about a capability the API lacks rather than an invariant the code holds. The repo's comment convention is present-tense statements of intent, and the first two sentences already carry the whole value ("a proof among them also *authorizes* here from the moment it arrives, without the frame calling `authorize()`"), which is exactly what a reader at that line needs. The missing-capability observation reads better in the PR body, where it is already stated.

sdbondi (commented):
Reviewed

PR #2595, opened 2026-09-11
feat(transaction): add put_workspace alias on the transaction builder

sdbondi:
`put_last_instruction_output_on_workspace` is the most-used call in builder chains. This adds a short `put_workspace` alias that forwards to it; the original name is unchanged, so nothing else needs to move.

https://claude.ai/code/session_01A357R7azUjRU9qwrRwVi4i


sdbondi (commented):
👍

PR #2596, opened 2026-09-11
feat(substate)!: widen the substate version to u64

sdbondi:
## Description

Widens the substate version from `u32` to `u64` everywhere it is carried: the engine, consensus, storage, the wire formats, and the wallet/indexer SQLite stores.

### Why

A version counts one write per transaction that touches the substate, so its ceiling is a **throughput** ceiling on a single contended substate, not a calendar one. Extrapolating from testnet traffic is misleading — the faucet vault reached v33,691 in about a month, which is ~10,600 years to exhaust `u32`. The rate that matters is the one a hot vault or AMM pool sees at target TPS:

| sustained writes/s on **one** substate | `u32` lifetime |
| --- | --- |
| 10 | 13.6 years |
| 100 | 1.4 years |
| 1,000 | **50 days** |

`u64` at 1,000/s is ~585 million years.

Overflow today is also unsafe rather than a clean reject. Every increment was a plain `+ 1` (`working_state.rs`, `versioned_substate_id.rs`, `pending_store.rs`), and the version is the low bytes of `SubstateAddress`, so a wrapped version collides with the substate's v0 address and reuses a JMT key for a substate that is already down — silent corruption instead of an aborted transaction.

### Consensus-breaking

Networks must **reset**. Upgrading every node at once is not sufficient, because the break is in the on-disk state, not in what nodes say to each other:

- The RocksDB substates column family is keyed by `SubstateAddress`, which grows 36 -> 40 bytes. Existing rows are not merely stale, they are at keys the new binary never looks up.
- `BlockDiffKey`'s codec reads a 4-byte version where it now writes 8, so existing block diff rows decode to garbage.
- JMT leaf keys are `jmt_node_hash(&VersionedSubstateId)`, which is Borsh and therefore fixed-width, so every leaf key moves; leaf value hashes come from `hash_substate`, likewise Borsh, so every value hash changes. The same logical state produces a different merkle root.
- Committed block headers ratify the old roots through their QCs, so an upgraded node recomputes roots that disagree with its own history.

A migration could re-key the substate rows deterministically (the new address is the old object key, four zero bytes, then the old four version bytes) and rebuild the tree from the live set, but the rebuilt root matches nothing in the committed history. That needs a fork point past which root verification starts fresh, and a sync path that declines to re-derive pre-fork roots - more work than the change itself, and not worth it for a testnet.

- `hash_substate` chains the version through **Borsh**, which is fixed-width, so every substate value hash changes. There is no "small values encode identically" escape as there would be with CBOR.
- `SubstateAddress` grows 36 → 40 bytes, so every JMT key changes with it.
- The version inside a transaction's input requirements is part of the signed preimage, so transaction hashing changes too.

Shard routing is untouched: the version is a pure suffix of the address and `to_shard` reads only its first byte, so the shard/shard-group/range maths is unaffected.

The pinned-preimage tests in `substate_hasher.rs` are re-pinned to the new hashes, and the `hash_at` helper now mirrors `hash_substate`'s u64 chaining. Their doc comments previously claimed a specific old binary's hash must be reproducible; that continuity is what this change deliberately breaks, so they now state what they pin rather than where the value came from.

### Notable details

- **TS bindings are unchanged.** `ts-rs` maps a bare `u64` to `bigint`, which would break every JS/TS consumer for no reason — a version stays far below 2^53. Each exported version field is pinned to `ts(type = "number")`, matching the existing precedent on `NodeHeight`. A static check over the tree confirms no exported version field is left unpinned.
- **`prost` wire types** move from `uint32` to `uint64` in `common.proto` (`OptionalVersion`), `consensus.proto` (`Substate`), and `rpc.proto` (`GetSubstateResponse`, `SubstateData`, `SubstateDestroy`), plus the hand-written prost types in `tari_indexer_client`. `IdentitySignature.version` and `ShardStateVersions` are unrelated and untouched.
- **SQLite migrations** widen the version columns in both the wallet and indexer stores. The indexer schema has no foreign keys, so those tables are rebuilt normally. The wallet's `vaults` **is** a foreign key target, and a rebuild is not available: inside the transaction a diesel migration runs in, `PRAGMA legacy_alter_table` does not take effect, so renaming `vaults` repoints every referencing table at the temporary name and the referencing queries then fail on `vaults_old`. Both wallet columns are therefore replaced in place with `ADD COLUMN` / `UPDATE` / `DROP COLUMN` / `RENAME COLUMN`, which leaves foreign key clauses untouched; they move to the end of their tables, and the generated schema and models were reordered to match.
- **Memory guardrail raised.** `ProposedBlockChangeSet`'s worst case moves from 13.4MB to 14.3MB — the ~960KB is the wider version fields in `SubstateChange` and `SubstateLock`. The assertion ceiling goes to 15MB and the "last checked" figure is updated.
- **Fee estimation** shifts marginally: `TransactionReceipt::encoded_size_upper_bound` now uses `u64::MAX` as the worst-case version, which is 9 CBOR bytes instead of 5 per upped substate.

## Motivation and Context

`u32` is not wide enough for a substate that is written at the rates Ootle is designed to sustain, and the failure mode at the ceiling is JMT key reuse rather than a rejected transaction. Doing this while networks are still being reset is far cheaper than doing it after mainnet.

## How Has This Been Tested?

`cargo lints clippy --all-targets` is clean, and `cargo +nightly-2025-12-05 fmt --all` has been applied.

`cargo nextest run` passes for: `tari_ootle_common_types`, `tari_engine_types`, `tari_ootle_storage`, `tari_state_tree`, `tari_state_store_rocksdb`, `tari_ootle_p2p`, `tari_consensus`, `consensus_tests`, `tari_indexer_lib`, `tari_indexer`, `tari_ootle_wallet_storage_sqlite`, `tari_ootle_wallet_sdk`, `tari_ootle_walletd`, `tari_engine`, `tari_ootle_transaction`, `tari_transaction_manifest`, `ootle_sdk_core` and `tari_validator_node_rpc`.

The wallet migration is exercised by the `tari_ootle_wallet_storage_sqlite` suite, which runs migrations against a fresh database on every test — the foreign-key breakage above was caught there and the current form passes.

Not run locally: the `integration_tests` cucumber suite and the bindings regeneration (`bindings/build.sh`), both left to CI.

- Confirm the hash re-pinning is intended rather than accidental: `cargo nextest run -p tari_engine_types substate_hasher`.
- Check that the generated bindings really are unchanged by running `bindings/build.sh` and confirming an empty diff.
- Sanity-check the wallet migration against a populated pre-change wallet database: run the daemon against it and confirm vault balances and lock sweeping still work (the foreign keys on `vault_locks` are the thing to watch).

### Breaking Changes

- [ ] None
- [ ] Requires hard fork
- [x] Other - Please specify

Every substate value hash and every JMT key changes, so state roots committed by a pre-change binary cannot be re-derived, and the RocksDB substate and block-diff keys change shape underneath the existing store. Networks must be reset; a coordinated upgrade of every node does not avoid it. The wire formats (`SubstateRequirement`, substate RPC responses) and the signed transaction preimage change with them, so old and new nodes cannot interoperate.

### DIFF
--- a/crates/engine_types/src/substate.rs
+++ b/crates/engine_types/src/substate.rs
@@ -72,11 +72,12 @@ pub struct Substate {
     #[n(0)]
     substate: SubstateValue,
     #[n(1)]
-    version: u32,
+    #[cfg_attr(feature = "ts", ts(type = "number"))]
+    version: u64,
 }
 
 impl Substate {
-    pub fn new<T: Into<SubstateValue>>(version: u32, substate: T) -> Self {
+    pub fn new<T: Into<SubstateValue>>(version: u64, substate: T) -> Self {
         Self {
             substate: substate.into(),
             version,
@@ -95,7 +96,7 @@ impl Substate {
         self.substate
     }
 
-    pub fn version(&self) -> u32 {
+    pub fn version(&self) -> u64 {
         self.version
     }
 
@@ -111,7 +112,7 @@ impl Substate {
         hash_substate(network, self.substate_value(), self.version, epoch)
     }
 
-    pub fn previous_version(&self) -> Option<u32> {
+    pub fn previous_version(&self) -> Option<u64> {
         self.version.checked_sub(1)
     }
 }
@@ -119,7 +120,7 @@ impl Substate {
 /// Hashes a substate into its canonical value hash. The `epoch` argument binds the schema version
 /// (derived from `network` and epoch via `ProtocolVersion::at`) into the hash preimage, so substates
 /// produced under different schema versions can never collide in the JMT.
-pub fn hash_substate(network: Network, substate: &SubstateValue, version: u32, epoch: Epoch) -> Hash32 {
+pub fn hash_substate(network: Network, substate: &SubstateValue, version: u64, epoch: Epoch) -> Hash32 {
     let proto_version = ProtocolVersion::at(network, epoch);
     substate_value_hasher32()
         .chain(&substate.as_hash_message(proto_version))
@@ -981,7 +982,7 @@ pub struct SubstateDiff {
     #[n(0)]
     up_substates: Vec<(SubstateId, Substate)>,
     #[n(1)]
-    down_substates: Vec<(SubstateId, u32)>,
+    down_substates: Vec<(SubstateId, u64)>,
     #[n(2)]
     fee_withdrawals: Vec<ValidatorFeeWithdrawal>,
 }
@@ -1014,11 +1015,11 @@ impl SubstateDiff {
         self
     }
 
-    pub fn down(&mut self, id: SubstateId, version: u32) {
+    pub fn down(&mut self, id: SubstateId, version: u64) {
         self.down_substates.push((id, version));
     }
 
-    pub fn extend_down(&mut self, iter: impl Iterator<Item = (SubstateId, u32)>) -> &mut Self {
+    pub fn extend_down(&mut self, iter: impl Iterator<Item = (SubstateId, u64)>) -> &mut Self {
         self.down_substates.extend(iter);
         self
     }
@@ -1031,7 +1032,7 @@ impl SubstateDiff {
         self.up_substates.into_iter()
     }
 
-    pub fn down_iter(&self) -> impl Iterator<Item = &(SubstateId, u32)> + '_ {
+    pub fn down_iter(&self) -> impl Iterator<Item = &(SubstateId, u64)> + '_ {
         self.down_substates.iter()
     }
 
--- a/applications/tari_indexer/src/rest_api/handlers/substates.rs
+++ b/applications/tari_indexer/src/rest_api/handlers/substates.rs
@@ -48,7 +48,7 @@ fn substate_lookup_error(e: SubstateManagerError) -> ErrorResponse {
     params(
         ("substate_id" = String, Path, description = "The substate ID to fetch"),
         ("local_search_only" = bool, Query, description = "If true, only search local storage for the substate"),
-        ("version" = Option<u32>, Query, description = "Minimum version of the substate to fetch"),
+        ("version" = Option<u64>, Query, description = "Minimum version of the substate to fetch"),
     ),
     responses(
         (status = 200, description = "Substate details", body = GetSubstateResponse),
--- a/applications/tari_indexer/src/storage_sqlite/migrations/2026-09-11-000000_substate_version_u64/down.sql
+++ b/applications/tari_indexer/src/storage_sqlite/migrations/2026-09-11-000000_substate_version_u64/down.sql
@@ -1,0 +1,125 @@
+alter table substates rename to substates_old;
+
+create table substates
+(
+    id               integer   not NULL primary key AUTOINCREMENT,
+    address          text      not NULL,
+    version          int       not NULL,
+    data             text      not NULL,
+    template_address text      NULL,
+    module_name      text      NULL,
+    updated_at       timestamp not null default current_timestamp,
+    created_at       timestamp not null default current_timestamp
+);
+
+insert into substates (id, address, version, data, template_address, module_name, updated_at, created_at)
+select id, address, version, data, template_address, module_name, updated_at, created_at
+from substates_old;
+
+drop table substates_old;
+
+create unique index uniq_substates_address on substates (address);
+
+alter table substate_transitions rename to substate_transitions_old;
+
+create table substate_transitions
+(
+    id            integer   not NULL primary key AUTOINCREMENT,
+    shard         int       not NULL,
+    state_version bigint    not NULL,
+    epoch         bigint    not NULL,
+    substate_id   text      not NULL,
+    version       int       not NULL,
+    substate_type text      not NULL,
+    is_up         bool      not NULL,
+    value_hash    text      NULL,
+    created_at    timestamp not null default current_timestamp
+);
+
+insert into substate_transitions (id, shard, state_version, epoch, substate_id, version, substate_type, is_up,
+                                 value_hash, created_at)
+select id,
+       shard,
+       state_version,
+       epoch,
+       substate_id,
+       version,
+       substate_type,
+       is_up,
+       value_hash,
+       created_at
+from substate_transitions_old;
+
+drop table substate_transitions_old;
+
+create unique index substate_transitions_substate_id_version_uniq on substate_transitions (substate_id, version, is_up);
+create index substate_transitions_shard_state_version_idx on substate_transitions (shard, state_version);
+
+alter table utxos rename to utxos_old;
+
+create table utxos
+(
+    id               integer   not NULL primary key AUTOINCREMENT,
+    commitment       text      not NULL,
+    public_nonce     text      not NULL,
+    version          int       not NULL,
+    resource_address text      not NULL,
+    shard            int       not NULL,
+    state_version    bigint    not NULL,
+    output           blob      NULL,
+    utxo_tag         int       not NULL,
+    epoch            bigint    not NULL,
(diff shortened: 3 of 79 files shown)


crates/engine_types/src/substate.rs
```diff
@@ -981,7 +982,7 @@ pub struct SubstateDiff {
     #[n(0)]
     up_substates: Vec<(SubstateId, Substate)>,
     #[n(1)]
-    down_substates: Vec<(SubstateId, u32)>,
+    down_substates: Vec<(SubstateId, u64)>,
```
sdbondi, line 986: _Posted by Claude._

`SubstateDiff` is `ts(export)`ed and this field is not pinned, so the "TS bindings are unchanged" claim does not hold for it. Every other exported version field in the diff got `ts(type = "number")`; here the version is the second element of a tuple rather than a named `version:` field, which is presumably why the static check missed it.

ts-rs maps a bare `u64` to `bigint` in this repo (cf. `LeaderFee.ts`, `FeeBreakdown.ts`). The checked-in `bindings/src/types/SubstateDiff.ts` currently reads:

```ts
export type SubstateDiff = {
  up_substates: Array<[SubstateId, Substate]>;
  down_substates: Array<[SubstateId, number]>;
  fee_withdrawals: Array<ValidatorFeeWithdrawal>;
};
```

After `bindings/build.sh` it regenerates as `Array<[SubstateId, bigint]>`. That breaks JS/TS consumers of `SubstateDiff`, and it is wrong at runtime too: the value arrives over JSON as a plain number, so `JSON.parse` yields a `number` while the declared type says `bigint` — `v === 0n` / `v + 1n` against it fails. Since no CI job regenerates and diffs the bindings, the checked-in file will quietly disagree with the Rust type until someone runs the generator.

Note that `#[ts(type = "...")]` drops the field's ts-rs dependencies, so naming `SubstateId` inside the override would emit an unresolved reference (no `import type { SubstateId }`). `SubstateId` is `export type SubstateId = string`, so spelling it `string` is type-compatible — the same shape `ShardGroupEvidence.outputs` already uses with `Record<string, number>`:

```suggestion
    #[cfg_attr(feature = "ts", ts(type = "Array<[string, number]>"))]
    down_substates: Vec<(SubstateId, u64)>,
```

sdbondi: _Posted by Claude._

Confirmed and fixed in 55db3c4701, using the suggested form.

The static check I ran matched named `version: u64` fields, so a version carried as a tuple element was invisible to it. I re-ran the sweep over every `ts(export)`ed struct field mentioning `u64` regardless of name, diffed against the lines this branch introduces, and `SubstateDiff.down_substates` was the only one. The rest of the unpinned `u64` fields in the tree (`LeaderFee.fee`, `FeeBreakdown`, the wallet client fee fields) predate this branch and already render as `bigint` in the checked-in bindings.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 1 non-blocking, 1 nit · 1 resolved since `3f28c75` (1 non-blocking) · head `092845b5`

Re-review of the three commits since the last pass. `6a4bad7b` is a pure rebase of the commit reviewed before — diffed against `3f28c753`, only blob hashes and hunk offsets move, so that review still stands.

### Resolved

- **`SubstateDiff.down_substates` ts pin** (`4afb1e73`) — verified: `#[cfg_attr(feature = "ts", ts(type = "Array<[string, number]>"))]` on `crates/engine_types/src/substate.rs:986`. The field no longer renders as `bigint`.

### Verified clean

- **`a6b578f5`** — `version_of` in `crates/engine/tests/access_rules.rs:421` is the only u32 version the rebase pulled in from #2677; a fresh grep over `crates/`, `applications/`, `clients/` finds no other residual narrowing on the widened path. The wallet migration redate to `2026-09-23-000000` puts it last: the newest migration on `development` is `2026-07-15-100000_transaction_requests`.
- **`092845b5` (indexer squash)** — applied the full 17-migration chain and the new single `2023-02-16-145719_initial` to two fresh SQLite databases and compared normalised schemas (`pragma table_info` / `foreign_key_list` / `index_list` / `index_info` per table, plus views and triggers). Identical except column *ordering* in two tables: `events.resource_address` and `template_catalogue.metadata_hash` now sit where they are declared instead of where `ALTER TABLE ADD COLUMN` appended them. That is safe here — diesel names columns explicitly, and all four raw-SQL sites in the indexer (`store_factory.rs:641,1002`, `writer.rs:716`) are column-explicit too, so nothing binds positionally.
- `down.sql` drops all 13 tables `up.sql` creates; up-then-down on a fresh database leaves `sqlite_master` empty.
- The 72 removed lines in `store_factory.rs` are only the two tests for the deleted backfills. The `resource_address`-is-resource-only rule they covered is still enforced at the type level by `EventFilter::event_resource_address` → `SubstateId::as_resource_address` (`network_state_sync/event_filter.rs:81`), so no runtime invariant loses its last guard.
- Bindings: none of the four PR commits touch `bindings/`.

One correction to the description's operator note: an indexer database that is *fully* migrated does not have to be deleted. `run_pending_migrations` (`store_factory.rs:44`) ignores recorded versions it no longer knows about, and the resulting schema is the one verified identical above. Deletion is genuinely required for a *stale* database — one that stopped partway down the old chain — because `2023-02-16-145719` is already recorded and the new initial will never run to add what it is missing. Worth narrowing the note so operators know which case they are in.

### Non-blocking

- `utilities/validator_rollback/src/audit_writer.rs:162` — `version_to_u32` saturates the audit's `pre_rollback_version` / `post_rollback_version` (`audit.rs:94-96`) at `u32::MAX`. Pre-existing on the base and unreachable in practice, but it is now the one place in the tree that still narrows a substate version, and the PR is framed as an end-to-end widening. Widening the two `AuditRecord` fields and dropping the helper would finish the sweep. (I said last pass that no residual narrowing remained — that grep matched casts, not this `try_into().unwrap_or`.)

### Nit

- `bindings/src/types/SubstateDiff.ts:8` still reads `Array<[SubstateId, number]>` while the new pin emits `Array<[string, number]>`, so "TS bindings are unchanged" is now off by that one token and the checked-in file drifts on the next regeneration. `SubstateId` is a `string` alias, so nothing breaks either way. Pinning to `Array<[SubstateId, number]>` instead would make the generated file byte-identical and keep the nominal type — the import resolves because `up_substates` still carries the `SubstateId` dependency. The tradeoff is that the pin then depends on `up_substates` keeping that field type, which is why the original suggestion said `string`; either is fine, but the checked-in file should match whichever is chosen.

CI is still pending on `clippy`, `test (1-3/3)`, `integration test` and the four `fuzz` legs at this head.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 1 non-blocking, 1 nit · 0 resolved since `092845b5` · head `3e30b496`

CI has settled at this head. `integration test` concluded success; the one red check is `Test Results (CI)`, whose only failure annotation is `stall_recovery.feature:7:3` — the known RocksDB LOCK flake, unrelated to this change. Everything else is green, including clippy, all three test shards, all four fuzz legs, fmt, prettier, cargo_deny, ledger app and the web UI build.

`3e30b496` itself is a merge of `development` carrying only #2684's `real_time_ticker.rs` test-timing fix; no conflict resolution and no PR-side file changed, so the review of `092845b5` carries over unchanged.

Nothing blocking stands. The `validator_rollback` u32 saturation and the `SubstateDiff.ts` drift from the previous comment are both still open and both are fine to merge over or fold into a follow-up. I cannot click approve on this PR, so this stays a comment — it needs another maintainer's approval.

sdbondi (commented):
Tested and Approved

PR #2597, opened 2026-09-11
feat(engine)!: charge bulk memory and table operators by length

sdbondi:
## Problem

`memory.copy`, `memory.fill` and `memory.init` are priced at a flat 2–4 points regardless of the length operand; `table.copy`/`fill`/`grow` at 2. The meter therefore bounds a template's *instruction count*, not its work. Against a 2 MiB linear memory and the 250M per-transaction point budget, a template can move on the order of 70 TB of bytes through `memory.copy` inside budget.

This is audit finding H1.

## Fix

A `BulkMetering` middleware, pushed **after** `Metering`, emits an inline length-proportional charge against the same `wasmer_metering_remaining_points` global before each bulk operator runs:

```
global.set $scratch_len      ;; pop the length operand
...  len * RATE -> $scratch_cost
if remaining < cost { points_exhausted = 1; unreachable }
remaining -= cost
global.get $scratch_len      ;; push it back
<original operator>
```

Push order matters twice: stages run in push order with each consuming the previous stage's output, so pushing second means (a) the metering global indexes are already recorded and (b) the operators emitted here are not themselves statically charged.

A middleware cannot add function locals (`locals_info` is read-only), so the length operand is parked in two middleware-private module globals. Threads are off in the engine's feature set, so those are single-writer.

Reading `Metering`'s global indexes needs them to be public, which upstream `wasmer-middlewares` does not expose. This swaps the dependency to `tari-wasmer-middlewares` 7.4.0 — a three-commit fork of v7.4.0 on `tari-project/wasmer` (branch `tari/middlewares`) whose only delta is making `MeteringGlobalIndexes` and `Metering::global_indexes()` public. An upstream PR is open separately.

## Rates

| operator | unit | points |
| --- | --- | --- |
| `memory.copy` / `fill` / `init` | byte | 1 |
| `table.copy` / `fill` / `init` / `grow` | element | 16 |
| `memory.grow` | 64 KiB page | 65536 |

Points are calibrated at ~8.4M points/ms of validator CPU; a host memcpy sustains ~10 GB/s, i.e. ~10M bytes/ms, so cost-neutral is ~0.84 points/byte. Rounded up to 1 — a bulk copy is never cheaper per byte than the meter believes.

## Consensus impact

`ENGINE_FINGERPRINT` moves to **v6** (the middleware list is baked into the cached artifact).

Every template rustc compiles emits `memory.copy` for slice moves, so point totals rise across the board — a six-call transaction in `fees.rs` moved from 1000 to 1350 µT. This changes execution results and must ship with the coordinated protocol version bump alongside the other wave 3 changes.

## Tests

`crates/engine/tests/wasm_loader.rs`, using the hand-written WAT harness:

- `memory_copy_points_scale_with_the_bytes_copied` — a 64 KiB copy costs exactly 65536 points more than an empty one.
- `a_small_memory_copy_stays_cheap` — a 32-byte copy stays under 1000 points, so the charge does not price ordinary code out.
- `an_unbounded_memory_copy_traps_on_the_meter` — `len = 0xFFFFFFFF` trips the meter's `unreachable` rather than wasmer's bounds check, i.e. the charge lands before the copy is attempted.

`cargo nextest run -p tari_engine`: 445 passed. `cargo lints clippy --all-targets -p tari_engine`: clean.


crates/engine/src/wasm/bulk_metering.rs
```diff
@@ -0,0 +1,216 @@
...
+    fn feed(&mut self, operator: Operator<'a>, state: &mut MiddlewareReaderState<'a>) -> Result<(), MiddlewareError> {
+        let Some(rate) = rate_for(&operator) else {
+            state.push_operator(operator);
+            return Ok(());
+        };
+
+        // Every operator priced here leaves its count — bytes, elements or pages — on top of the
+        // stack, so the charge is computed from the value popped here and the value is pushed back
+        // unchanged before the operator runs.
+        state.extend(&[
```
sdbondi: _Posted by Claude._

**The emitted preamble is itself unmetered, so every bulk operator now runs ~13 points of work for free.**

`BulkMetering` is pushed after `Metering` precisely so its own operators escape static costing — but that means the ~19 operators emitted here (4 global loads/stores, an `i64.extend`, an `i64.mul`, an `i64.lt_u`, a conditional block, an `i64.sub`) are charged nothing at all. Priced by the engine's own table in `metering.rs` the executed portion is worth ~13 points (`GlobalSet` 1 ×3, `I64ExtendI32U` 1, `I64Mul` 3, `I64LtU` 1, `If` 1, `End` 1, `I64Sub` 1), against the 2 points `memory.copy` is statically charged.

Concrete misbehaviour: a straight-line unrolled body of `(memory.copy (i32.const 0) (i32.const 0) (i32.const 0))` is charged 5 points per copy (2 for the copy + 3 for the operands) while executing the work of ~18, so 250M points buys roughly 3× the CPU the meter believes it sold. These are real loads/stores to the VMContext — cranelift does not promote wasm globals to SSA, so the `global.set $len` / `global.get $len` round trip is not folded away even when the length is a constant 0. The same gap applies to every honest template, in proportion to how many `memory.copy`s rustc emitted.

The cheap fix is to fold a constant into the charge the middleware already computes, e.g. `len * rate + PREAMBLE_POINTS` (one extra `i64.add`), or to raise the static costs of the matched operators in `metering.rs` so the preamble is paid for there.

crates/engine/src/wasm/bulk_metering.rs
```diff
@@ -0,0 +1,216 @@
...
+        Operator::MemoryCopy { .. } | Operator::MemoryFill { .. } | Operator::MemoryInit { .. } => {
+            Some(POINTS_PER_MEMORY_BYTE)
+        },
+        // `[dst, src, len]` / `[dst, value, len]` / `[value, delta]`: table elements.
+        Operator::TableCopy { .. } |
+        Operator::TableFill { .. } |
+        Operator::TableInit { .. } |
+        Operator::TableGrow { .. } => Some(POINTS_PER_TABLE_ELEMENT),
+        // `[delta]`: 64 KiB pages.
+        Operator::MemoryGrow { .. } => Some(POINTS_PER_MEMORY_PAGE),
```
sdbondi: _Posted by Claude._

**`memory.grow` / `table.grow` are charged on the *requested* delta, so a grow that the tunables will refuse is billed as if it succeeded — and a large enough request now traps instead of returning `-1`.**

Unlike `memory.copy`, `grow` has a defined failure path: it returns `-1` and does no work. `LimitingTunables` caps linear memory at 32 pages and tables at 16 384 elements, so any delta past that does *zero* bytes of work — but the charge is emitted before the operator and uses the operand, not the outcome.

Two concrete regressions:

1. `Vec::with_capacity(10_000_000)` in a template → dlmalloc calls `memory.grow(153)` → 153 × 65 536 ≈ **10M points burned**, the grow then returns `-1`, dlmalloc returns null and Rust aborts. Before this change the same path cost 4 points.
2. `Vec::with_capacity` of anything over ~250 MB (delta > 3 814 pages) now exceeds the whole per-transaction budget, so the meter sets `points_exhausted` and traps. `WasmProcess::invoke` charges `points_before` — the **entire pre-call budget** — for an out-of-gas trap, where previously the template got a clean `-1`, aborted through `on_panic`, and was charged for what it actually consumed. Failure mode and fee both change for input-sized allocations.

The in-repo shape that relies on the `-1` return is `crates/engine/tests/wasm_loader.rs::a_table_grows_no_further_than_the_limit` — `table.grow 0 (ref.null func) (i32.const 16385)`. It only survives because 16 385 × 16 = 262 160 points happens to fit the budget; the same test written against `memory.grow` past the page cap would now trap on the meter rather than observe `-1`.

Suggest clamping the charged delta to the headroom the tunables can actually grant (`min(delta, max_pages)` / `min(delta, max_table_elements)`), which keeps the charge an upper bound on real work while preserving the spec'd `-1` result.

crates/engine/src/wasm/bulk_metering.rs
```diff
@@ -0,0 +1,216 @@
...
+        _ => None,
+    }
+}
+
+impl<'a> FunctionMiddleware<'a> for FunctionBulkMetering {
+    fn feed(&mut self, operator: Operator<'a>, state: &mut MiddlewareReaderState<'a>) -> Result<(), MiddlewareError> {
```
sdbondi, line 184: _Posted by Claude._

**Per-occurrence code expansion is an amplifier on the registration compile cost, which is priced on source bytes.**

Every bulk operator in a published module expands to 20 operators including a control-flow block, and rustc emits `memory.copy` for every slice move. A `memory.copy` with three `i32.const` operands is ~7 source bytes, so an adversarial 100 KB module can hold on the order of 14 000 of them — this middleware turns that into ~280 000 extra operators and 14 000 extra cranelift blocks for the same 100 KB of source.

Template registration charges the compile against `code_size` (the WASM source byte count), which does not see the expansion, so the cost per registered byte moves by a constant that isn't in the fee. Worth checking against #2599, which prices the publish compile — the calibration there should be taken after this middleware lands, not before.

sdbondi (commented):
_Posted by Claude._

Re-reviewed at `a206a30c5` (`fix(engine): pay for the charge sequence; clamp a grow to what it can get`). Both medium findings are genuinely resolved; I re-derived the numbers rather than taking the commit message for it.

**Charge sequence now paid for — correct, and the constant is exactly right.** Summing the emitted operators against the cost table at this head: `GlobalSet` 1 + `GlobalGet` 0 + `I64ExtendI32U` 1 + `I64Const` 0 + `I64Mul` 3 + `GlobalSet` 1 + 2×`GlobalGet` 0 + `I64LtU` 1 + `If` 1 + `End` 1 + 2×`GlobalGet` 0 + `I64Sub` 1 + `GlobalSet` 1 + `GlobalGet` 0 = **12** on the non-trapping path. The clamp arm adds `GlobalGet` 0 + `I32Const` 1 + `GlobalGet` 0 + `I32Const` 1 + `I32LtU` 1 + `Select` 1 = **4**, so 16 for a `grow`. Worst-case dispatch is `MemoryGrow` at 4 → 20 exactly, and `TableGrow` (clamped, dispatch 2) is 18. `BULK_OPERATOR_COST = 20` covers every instrumented operator with nothing to spare — worth noting for anyone who later lengthens the sequence or adds an operator with a dispatch cost above 4. `memory.copy` pays 20 against a real 14, which is the safe direction.

Routing `cost_function` through `bulk_metering::charge_for` is the right call — the two halves can't drift now. The eight entries still in the match arm (`MemoryGrow => 4`, `MemoryCopy | MemoryFill => 2`, the table group `=> 2`) are unreachable after the early return, but the `BULK_OPERATOR_COST` doc comment names them as the dispatch reference, so they read as intentional.

**Grow clamp — correct, and the trap is gone.** Emitted `min` verified by hand: stack `[len]` → `GlobalSet $len` → `[len, grantable, len, grantable]` → `I32LtU` → `[len, grantable, len<grantable]` → `Select` picks `val1` when the condition is true, i.e. `len`, else `grantable`. Unsigned compare, so a `0xFFFF_FFFF` delta clamps rather than reading as negative. The push-back at the end is `GlobalGet $len` — the original, unclamped operand — so the operator's own `-1` result is untouched. With `max_memory_pages = 32` the worst grow charge is 32 × 65 536 = 2 097 152 points and `max_table_elements = 16 384` gives 262 144, both far under the 250M budget, so the meter can no longer trap on a refusal. For `table.grow` the clamp operates entirely above the `funcref` operand, which is left alone.

Both new tests hold up. `an_impossible_memory_grow_is_charged_only_for_what_could_be_granted` works because both deltas clamp to the same 32 pages from a 3-page module; `a_zero_length_copy_pays_for_the_charge_sequence` gives a marginal of 20 against its floor of 12.

**Remaining, non-blocking:**

1. A refused `grow` is still charged the full clamped delta — 2M points for pages that were never allocated. Bounding it by real headroom would need `memory.size` at runtime; at 0.8% of the budget for an operator templates rarely reach for, the static cap seems the right trade. Flagging only so it's a decision rather than an oversight.
2. The earlier low finding stands: the clamp arm adds 6 more operators per `grow`, so cranelift input amplification for bulk-dense modules is slightly higher than at the last head. Still a calibration input for #2599's publish-compile pricing, not a problem here.

`ENGINE_FINGERPRINT` is `v6` at this head, which covers the changed cost function — and since #2599 leaves the constant at `v5`, there's no collision to resolve.

No blocking items from me. CI is still pending (clippy, the three `test` shards, fuzz, integration) — worth waiting on those before merge. I can't approve on your own PR; another maintainer will need to.

sdbondi (commented):
_Posted by Claude._

Re-reviewed `6672cc5e5` (`fix(engine): make the bulk metering scratch indexes write-once`).

**The change is correct.** `ScratchGlobals` is `Copy + Debug`, so `*self.scratch.get().expect(…)` and `set(…).expect(…)` both compile and read cleanly; `OnceLock` gives the happens-before that the cranelift workers need, and `set` returning `Err` carries the one-instance-per-module invariant the removed assert had. I checked the lifetime that invariant depends on: `create_engine` builds a fresh `BulkMetering` per call (`module.rs:174`) and every caller — `load_template_from_code` (`module.rs:79`) and `load_template_from_serialized` (`module.rs:110`) — makes a new engine per module, so `transform_module_info` runs exactly once per instance and the `expect` is unreachable. The deserialize path never runs middleware at all, so an unset cell there is fine. Lock-free reads across the rayon pool are a genuine improvement over taking a mutex per function.

**The commit message describes a defect this branch never had.** It says the previous code was "a `RefCell` behind an `unsafe impl Sync`" and that "`take()` leaves the cell empty after the first function reaches it, so every other function panics". Every published head of this branch used `Mutex<Option<ScratchGlobals>>` with `.lock()` and no `take()` — `2da29e716`, `a206a30c5` and `4abc6472c` all the same. That's `Sync` without any unsafe promise and has no race and no empty-cell panic; the worst of it was mutex contention across the cranelift workers.

So this commit is a lock-free refactor, not a race fix, and the message as written will mislead anyone reading `git log` or the squash-merge body — particularly the claim that a shipped path could panic nondeterministically. Worth rewriting to what it actually is: `OnceLock` fits a value written once during `transform_module_info` and read from every worker afterwards, so the per-function lock goes away and the one-module invariant moves into the type.

No blocking findings; the earlier two mediums remain verified fixed.

PR #2603, opened 2026-09-15
perf(engine)!: drop redundant fields from builtin event payloads

sdbondi:
## Motivation

Builtin `std.*` events are persisted in the transaction receipt and priced by their encoded length (`TransactionReceipt::encoded_size_upper_bound`), so every payload entry is paid for in storage fees by whoever runs the instruction that emits it.

A number of those entries duplicate state the same transaction already ups. A client holding the event has its `substate_id`; the resource it names is a lookup away and almost certainly cached, since resource type and symbol are immutable. Charging for it in the event too means paying twice for the same bytes.

## Changes

Removed:

- `resource_type` from `std.resource.{create,mint,recall,update_nonfungible_data,update_access_rule,update_metadata}` and `std.vault.{deposit,withdraw}`. On `update_nonfungible_data` it was the literal constant `"NonFungible"` — it carried no information at all.
- `token_symbol` and `image_url` from `std.resource.create`, and `token_symbol` from `std.resource.update_metadata` — a straight echo of the resource metadata the same instruction writes.

`std.resource.{create,update_nonfungible_data,update_metadata}` now carry an empty payload; the `substate_id` names the resource.

Kept:

- `resource_address` on `std.vault.{deposit,withdraw}`. The indexer's `EventFilter::event_resource_address` matches on it while streaming; removing it would force a vault substate fetch per event.
- `amount`, `action`, `vault_id`, `recall_desc`, `flags` — the "what changed" discriminators, not recoverable without the prior substate version.

The rule is documented on `emit_std_event` so it doesn't drift back.

## Consensus impact

Receipt bytes and therefore storage fees change, so this is a breaking change requiring a coordinated upgrade. It is not a WASM artifact change — `ENGINE_FINGERPRINT` stays at `v5`.

## Testing

`cargo nextest run -p tari_engine` — 364 passed. `cargo lints clippy -p tari_engine --all-targets` clean. The fee suite (`storage_does_not_follow_the_max_fee_digit_count`, `required_fees_covers_a_real_run_at_any_max_fee`, and the rest) passes unchanged.

## Follow-ups found while auditing, not addressed here

1. **`Metadata` is `BTreeMap<String, String>`**, which is where the remaining cost is. A `resource_address` encodes as a 73-char hex string (~92 bytes including the key) where the raw address is 33 bytes; `amount` is decimal text, ~20 bytes for 9. Roughly 3x on every payload entry, engine and template alike. Fixing it needs a typed value in `tari_template_lib_types` — a template ABI break, far larger than this PR.
2. **`std.resource.update_nonfungible_data` does not say which NFT changed.** Its payload was all-constant before and is empty now. The `nf_id` is the one thing a reader cannot derive from the topic, and the only non-redundant thing that could go in that payload.
3. **`account::withdraw_approved` emits the wrong value** — `("spender_badge", badge_resource.to_string())` is the badge *resource* address, where `approve`/`revoke_approval` both emit the full `NonFungibleAddress` (`crates/template_builtin/templates/account/src/lib.rs:337`). Looks like a bug rather than waste.
4. `liquidity_pool`'s `contribute`/`redeem` echo `resource_a`/`resource_b`, which are fixed pool config readable from the component. Minor, and template payloads are the author's call.


crates/engine/src/runtime/impl.rs
```diff
@@ -397,6 +397,14 @@ impl<TStore: StateReader + Clone + 'static, TTemplateProvider: TemplateProvider<
...
     }
 
+    /// Emits one of the engine's builtin `std.*` events.
+    ///
+    /// Events are persisted in the transaction receipt and priced by their encoded length, so a
+    /// `payload` entry is paid for by every caller of the instruction that emits it. A payload must
+    /// therefore carry only what a reader cannot recover from the substates the same transaction
+    /// ups: the amount moved, which rule changed, which vault was reached. Anything the reader can
+    /// read off `substate_id` or off a substate it names — a resource's type, symbol or metadata —
+    /// is charged for twice and must stay out.
```
sdbondi, line 407: _Posted by Claude._

The rule as written is stricter than what the PR actually keeps, and following it literally breaks the indexer.

`std.vault.{deposit,withdraw}` still carry `resource_address`, which *is* readable off the vault substate the same transaction ups — exactly the case this paragraph says "must stay out". The PR description records the exception (`EventFilter::event_resource_address` matches on that payload entry while streaming, `applications/tari_indexer/src/network_state_sync/event_filter.rs:88`), but the doc comment does not. A later contributor trimming payloads against this rule would drop `resource_address` from the two vault events and silently break every `resource_address`-filtered event subscription, since `event_resource_address` falls back to the payload only for vault events (their `substate_id` is the vault, not the resource).

Suggest adding the carve-out here: an entry may stay when a consumer must filter on it during streaming, before it can fetch the substate.

crates/engine/src/runtime/impl.rs
```diff
@@ -1976,13 +1968,7 @@ where
...
                 self.tracker.write_with(|state_mut| {
                     let resource_mut = state_mut.get_resource_mut(&resource_lock)?;
-                    let payload = Metadata::from_iter([
-                        ("resource_type", resource_mut.resource_type().to_string()),
-                        (
-                            "auth_hook",
-                            auth_hook.as_ref().map_or_else(|| "none".to_string(), |h| h.to_string()),
-                        ),
-                    ]);
+                    let payload = Metadata::from_iter([("auth_hook", auth_hook.display().to_string())]);
```
sdbondi: _Posted by Claude._

Two things on this line, both outside what the PR description covers.

1. The sentinel for "no hook" changes from `none` to `None`. `Displayable`'s `Option` impl writes `None` (`crates/common_types/src/displayable.rs:51`), where the old `map_or_else(|| "none".to_string(), ..)` wrote lowercase. The payload value is persisted in the receipt, so any consumer that matched the old `"none"` string to detect hook removal stops matching. Nothing in-tree reads it, and the length (and therefore the fee) is unchanged, but the PR body lists this commit as removals only and no test pins the value — worth either keeping `none` or calling the rename out.

2. By the rule this same commit documents on `emit_std_event`, `auth_hook` should not be here at all: the new hook is exactly what `set_auth_hook` writes to the resource substate the transaction ups on the next line, so a reader holding the event can read it back off `substate_id`. That is the same argument used to drop `resource_type` and `token_symbol` elsewhere in this diff. Either drop it (the topic plus `substate_id` already say "the hook on this resource changed") or note why it is exempt, otherwise the new doc comment reads as violated by the call site right below it.

crates/engine/src/runtime/impl.rs
sdbondi: _Posted by Claude._

Agreed — the rule as written condemned an entry this PR deliberately keeps. Added the carve-out to the doc comment:

> An entry earns its place despite that only when a consumer has to match on it while streaming, with no substate in hand yet. `std.vault.{deposit,withdraw}` carry `resource_address` for that reason: their `substate_id` is the vault, so the indexer's `EventFilter` has nothing else to filter a subscription on.

crates/engine/src/runtime/impl.rs
sdbondi: _Posted by Claude._

Both correct. Resolved together by dropping the sentinel rather than renaming it — when the hook is removed the `auth_hook` key is now absent, so `get_payload("auth_hook")` returns `None` instead of the string `"None"`:

```rust
let mut payload = Metadata::new();
if let Some(hook) = auth_hook.as_ref() {
    payload.insert("auth_hook", hook.to_string());
}
```

That also removes the `Displayable` call the `none` -> `None` rename came from, so point 1 no longer applies. The commit message now records the change.

On point 2: `auth_hook` stays, at the author's direction. The rule does bite — the new hook is what `set_auth_hook` writes on the next line. What is not derivable from the substate alone is the removal, and the key's absence now carries that for a reader who never fetches the resource. The value being present alongside it is the part that duplicates state; it is kept because the removal signal is only legible against it.

sdbondi (commented):
reviewed

PR #2604, opened 2026-09-15
fix(account): emit the spending badge, not its resource, on withdraw_approved

sdbondi:
## Problem

`Account::approve` and the `revoke_approval` branch both record the full `NonFungibleAddress` of the spender under `spender_badge`. `withdraw_approved` recorded `spender_proof.resource_address()` under the same key — the badge's *resource* address, not the badge.

```rust
let badge = NonFungibleAddress::new(badge_resource, badge);
// ...
emit_event("withdraw_approved", [
    ("spender_badge", badge_resource.to_string()),  // wrong value
```

A reader correlating `approve` / `revoke_approval` / `withdraw_approved` cannot tell which badge spent an approval, and one badge resource commonly covers many spenders — every account's owner badge shares `ACCOUNT_NFT` style resources, so the field is close to a constant in practice.

## Fix

Emit `badge`. The event now moves above the zero-approval cleanup (which consumes `badge`), with the exhaustion check read into a local first; nothing else about the ordering is observable.

Added `the_withdraw_approved_event_names_the_spending_badge`, which asserts the payload against the approving badge and fails on the old value.

## Impact

Builtin template change, so the account template's compiled bytes change — a coordinated upgrade like any other builtin edit. Event payload only; no behavioural change to approvals or withdrawals.

## Testing

`cargo nextest run -p tari_template_builtin --test account` — 4 passed.

Found while auditing builtin event payloads for #2603; independent of it.


sdbondi (commented):
LGTM

PR #2605, opened 2026-09-15
fix: guard arithmetic that untrusted input can overflow

sdbondi:
_Posted by Claude._

## Summary

Findings from a workspace-wide run of `clippy::arithmetic_side_effects` (962 hits, every one read in context). Release builds set `overflow-checks = true`, so each item below was a panic rather than a silent wrap. The engine hot path needed no changes: all 108 engine/engine_types hits are guarded, limit-bounded, or curve arithmetic.

## Reachable panics fixed

| Site | Trigger | Fix |
|---|---|---|
| `NumPreshards::all_shard_groups_iter` | `num_committees > num_preshards` (~10k VNs on mainnet constants, fewer on devnets) underflowed `shards_per_committee - 1`; reached from evidence insertion for global substates | clamp to `1..=num_shards`, as `to_shard_group` already does |
| `BaseLayerConsensusConstants::height_to_epoch` | base node returning `epoch_length == 0` (proto3 default) divided by zero in the epoch oracle | reject at fetch; accessors use `max(1)` / `saturating_mul` |
| `ConfidentialTransferApi` `ConfidentialOnly` path | transfer larger than the confidential balance: `lock_outputs_by_amount` returns short, then `Amount - Amount` panicked | check the locked total and `checked_sub` → `InsufficientFunds` |
| `SortedPrefixFileLookup` | table declaring `min = 0, max = u64::MAX`, or corrupt record offsets | checked span at load; clamped offset on read |
| `ManifestInstructionGenerator` | `u16` workspace-id counter was uncapped, and only call depth was limited so 16 functions each calling the next 4× expands to 4^16 instructions | `checked_add` → `WorkspaceIdOverflow`; total leaf-instruction cap → `TooManyInstructions` |
| db_inspector handlers | `page * limit` / `height + 1` from query params | saturating ops |

## Defensive hardening

- Vote-path `total_leader_fee` now uses `checked_add` → `InvariantError`, matching the propose path.
- `check_proposed_by_leader` uses `checked_sub` instead of relying on `check_height` having run first.
- `Committee::total_power` saturates; `VotePower::saturating_add` and `Epoch::saturating_add` added.
- `VersionedSubstateId::to_next_version`, `PendingSubstateStore` version bump, `Shard`/`ShardGroup::to_substate_address_range` bounds, `LeaderFee` exhaust burn sum, request-id counter (`wrapping_add`), round-robin leader with empty committee.
- Wallet: vault lock-amount folds from SQLite, key index increment, `Instant`/`SystemTime` additions from caller durations, page offset cast.
- `CallInfo::next_arg` uses `checked_add` so a malformed length yields `None` on wasm32 rather than overflowing before the bounds check.
- CLI tools: `transaction_submitter`, `validator_rollback` epoch arithmetic.

## Tests

Regression tests for the committee clamp, the full-range lookup header, and the manifest fan-out cap. `cargo lints clippy --all-targets` is clean on all touched crates; unit tests for the touched crates pass (418 + new).


crates/transaction_manifest/src/generator.rs
```diff
@@ -282,24 +287,37 @@ impl ManifestInstructionGenerator {
                 self.call_depth += 1;
                 let mut instructions = Vec::with_capacity(body.len());
                 for intent in body {
-                    instructions.extend(self.translate_intent(intent)?);
+                    // Nested calls account for their own leaf instructions, so only count the ones produced here.
+                    let is_call = matches!(intent, ManifestIntent::CallLocalFunction(_));
```
sdbondi: _Posted by Claude._

The fan-out cap only counts instructions a body actually emits, so a manifest whose leaves emit nothing still expands exponentially and `generated_instructions` never leaves 0.

`fn leaf() {}` parses to an empty body (`parse_block` maps over zero statements), and `ManifestIntent::AssignInput` / `AssignBlob` also return `Ok(vec![])`. So:

```
fn leaf() { }
fn f0()  { leaf(); leaf(); leaf(); leaf(); leaf(); leaf(); leaf(); leaf(); }
...
fn f14() { f13(); f13(); ... }
fn main(){ f14(); }
```

Every intent in every body is a `CallLocalFunction`, so `is_call` is true at every level and the counter is never incremented. Depth stays under `MAX_CALL_DEPTH`, and `translate_intent` is entered N^15 times, each with a `HashMap` lookup plus a `Vec` clone. With N=8 that is ~3.5e13 calls: `transactions.submit_manifest` pins a core forever, which is exactly the DoS this cap is meant to close.

Counting visits rather than emitted instructions closes it — charge the budget once per `translate_intent` (or once per `CallLocalFunction`) instead of by `produced.len()`.

applications/tari_validator_node/src/consensus/leader_selection.rs
```diff
@@ -14,6 +14,6 @@ impl RoundRobinLeaderStrategy {
 
 impl<TAddr: NodeAddressable> LeaderStrategy<TAddr> for RoundRobinLeaderStrategy {
     fn calculate_leader(&self, committee: &Committee<TAddr>, height: NodeHeight) -> u32 {
-        (height.as_u64() % committee.len() as u64) as u32
+        (height.as_u64() % (committee.len() as u64).max(1)) as u32
```
sdbondi: _Posted by Claude._

`.max(1)` moves the panic rather than removing it. `LeaderStrategy::get_leader` does `committee.get(index as usize).unwrap()` (`crates/consensus/src/traits/leader_strategy.rs:53`), so with an empty committee `calculate_leader` now returns 0 and the caller panics on `Option::unwrap` instead of on the modulo — with a less informative message. If an empty committee is reachable, the strategy needs to return `Option`/`Result`; if it isn't, this only hides the invariant.

crates/wallet/ootle-rs/src/provider/tx_watcher.rs
```diff
@@ -116,7 +116,11 @@ impl TransactionWatcher {
     async fn handle_request(&mut self, request: TxWatchRequest) {
         let _enter = self.span.enter();
         tracing::debug!("Received watch request for tx_id: {}", request.tx_id);
-        self.reap_times.insert(Instant::now() + request.timeout, request.tx_id);
+        // A timeout too large to represent as an `Instant` is treated as never reaping.
+        let reap_at = Instant::now()
+            .checked_add(request.timeout)
```
sdbondi: _Posted by Claude._

The fallback re-introduces the unchecked addition it is guarding: `Instant::now() + Duration::from_secs(u32::MAX as u64)` panics on overflow precisely in the case where `Instant::now()` is close enough to the representable maximum that `checked_add` failed. Chaining a second `checked_add` (falling back to `Instant::now()`, or skipping the reap entry altogether) keeps the guard total.

crates/consensus/src/hotstuff/substate_store/pending_store.rs
```diff
@@ -148,7 +148,11 @@ impl<'a, TTx: StateStoreReadTransaction> PendingSubstateStore<'a, TTx> {
                     shard,
                 });
                 let version = substate.version();
-                let next_version = version + 1;
+                let next_version = version
+                    .checked_add(1)
```
sdbondi, line 148: _Posted by Claude._

This arm now errors on version overflow, but the `SubstateChange::Down` arm ~20 lines below still calls `id.into_next_version()`, which this PR changed to `version.saturating_add(1)` (`crates/common_types/src/versioned_substate_id.rs`). At `u32::MAX` that silently yields the same version, so the store emits an `Up` at the version it just `Down`ed rather than the `InvariantError` raised here. Worth making both paths agree — ideally both reject.

Minor: the `SubstateChange::Down` insert a few lines above happens before this check, so the early return leaves a half-applied change in the pending store.

crates/wallet/sdk/src/apis/confidential_transfer.rs
```diff
@@ -102,9 +102,12 @@ where TSpec: WalletSdkSpec
 
         match &input_selection {
             UtxoInputSelection::ConfidentialOnly => {
-                let (confidential_inputs, _) =
+                let (confidential_inputs, locked_amount) =
                     self.confidential_outputs_api
                         .lock_outputs_by_amount(lock_id, &src_vault.id, spend_amount)?;
+                if locked_amount < spend_amount {
```
sdbondi: _Posted by Claude._

This check is unreachable. `ConfidentialOutputsApi::lock_outputs_by_amount` already returns `ConfidentialOutputsApiError::InsufficientFunds` when `total_output_amount < amount` (`crates/wallet/sdk/src/apis/confidential_outputs.rs:70`), and `?` converts it here. `Amount` is a `u128` newtype, so the subtraction panic the PR body attributes to this path could not occur either — the effective fix is the `checked_sub` further down in `resolve_inputs`. Harmless, but it reads as a guard that does something.

crates/template_test_tooling/src/template_test.rs
```diff
@@ -710,7 +710,7 @@ impl TemplateTest {
 
     fn next_key_seed(&mut self) -> u8 {
         let seed = self.key_seed;
-        self.key_seed += 1;
+        self.key_seed = self.key_seed.wrapping_add(1);
```
sdbondi: _Posted by Claude._

`key_seed` is a `u8`, so `wrapping_add` makes the 257th key silently reuse seed 0 — a long template test then derives two identical keys/accounts and fails somewhere far from the cause, where the previous overflow panic pointed straight at it. `checked_add(1).expect("test key seeds exhausted")`, or widening the seed, keeps the failure loud.

crates/wallet/storage_sqlite/src/reader.rs
```diff
@@ -1659,7 +1676,7 @@ impl WalletStoreReader for ReadTransaction<'_> {
         }
         let templates = query
             .limit(page_size as i64)
```
sdbondi: _Posted by Claude._

The `offset` on the next line is now guarded, but this `page_size as i64` still wraps: a `page_size` above `i64::MAX` from the JSON-RPC caller becomes a negative limit, and SQLite treats a negative `LIMIT` as no limit, so the handler loads the whole `authored_templates` table. `i64::try_from(page_size).unwrap_or(i64::MAX)` would match the treatment `offset` just received.

crates/transaction_manifest/src/generator.rs
sdbondi: _Posted by Claude._

Confirmed: empty bodies and `AssignInput`/`AssignBlob` emit nothing. Fixed in 197f095c0852551e056db4f37c915fefe05379c2 by charging the budget once per visited intent, before translating it, and added a test with 8-way fan-out of an empty leaf.

applications/tari_validator_node/src/consensus/leader_selection.rs
sdbondi: _Posted by Claude._

Agreed, the caller unwraps the index. Reverted in 197f095c0852551e056db4f37c915fefe05379c2; the invariant stays where it was.

crates/wallet/ootle-rs/src/provider/tx_watcher.rs
sdbondi: _Posted by Claude._

Fixed in 197f095c0852551e056db4f37c915fefe05379c2: an unrepresentable deadline now logs and skips the reap entry rather than adding again.

crates/consensus/src/hotstuff/substate_store/pending_store.rs
sdbondi: _Posted by Claude._

Fixed in 197f095c0852551e056db4f37c915fefe05379c2: the version is checked before the `Down` is recorded, and the `Down` arm now rejects the same way instead of going through the saturating `into_next_version`.

crates/wallet/sdk/src/apis/confidential_transfer.rs
sdbondi: _Posted by Claude._

Confirmed, `lock_outputs_by_amount` already errors on a shortfall. Removed the redundant check in 197f095c0852551e056db4f37c915fefe05379c2; the `checked_sub` in `resolve_inputs` stays.

crates/template_test_tooling/src/template_test.rs
sdbondi: _Posted by Claude._

Changed to `checked_add(1).expect("test key seeds exhausted")` in 197f095c0852551e056db4f37c915fefe05379c2.

crates/wallet/storage_sqlite/src/reader.rs
sdbondi: _Posted by Claude._

Fixed in 197f095c0852551e056db4f37c915fefe05379c2: `limit` now goes through `i64::try_from(..).unwrap_or(i64::MAX)` like `offset`.

sdbondi (commented):
Reviewed

PR #2606, opened 2026-09-15
perf(engine)!: drop resource_address from the vault events

sdbondi:
## Motivation

Follow-up to #2603, which trimmed every redundant entry from the builtin `std.*` event payloads except one: `resource_address` on `std.vault.deposit` / `std.vault.withdraw`.

It survived because the indexer's `EventFilter` matched on it, and the indexer is not in a position to resolve the vault itself — it keeps no substate store, only a demand-populated cache (`network_state_sync/worker.rs` calls `upsert_substate` for validator fee pools and nothing else), so a lookup during ingest means a validator RPC inside the write transaction. The SSE live filter has no store handle at all.

That makes the field a workaround for an indexer limitation, paid for by every user who moves tokens: ~92 bytes of receipt storage on every deposit and withdraw, forever, for a value that never changes. A vault's resource is fixed for the vault's life, so any consumer can resolve it once and cache it — which is what a wallet or dapp does anyway.

## Changes

**Engine** — `std.vault.{deposit,withdraw}` now carry only `amount`. The `emit_std_event` doc comment loses the carve-out that existed solely to excuse this entry.

**Indexer** — `EventFilter::event_resource_address` no longer falls back to the payload. It derives the resource from `substate_id` alone, so `resource_address` filters match the `std.resource.*` family and nothing else. A `resource_address` entry in a *template's* payload is now just template data, and a new test pins that it cannot make an event match.

A subscriber that wants one resource's transfers filters on the vault's `substate_id` instead. That is the more precise subscription regardless: a resource filter matches every account on the network holding it, handing the subscriber a firehose it has to re-filter client-side.

**Storage** — the `events.resource_address` column stays and is still populated for `std.resource.*` events. It is NULL for vault events going forward; rows written before this change keep the values they have.

**Docs/bindings** — the OpenAPI parameter description, the `tari_indexer_client` doc comments and the two generated `.ts` bindings all described the payload fallback; updated by hand rather than regenerating (the OpenAPI generator needs a release build of the indexer). `openapi.json` re-validated as JSON.

## Consensus impact

Receipt bytes and therefore storage fees change, so this needs a coordinated upgrade. Not a WASM artifact change — `ENGINE_FINGERPRINT` stays at `v5`.

## Testing

- `cargo nextest run -p tari_engine` — 453 passed
- `cargo nextest run -p tari_indexer` — 128 passed
- `cargo lints clippy -p tari_engine -p tari_indexer -p tari_indexer_client --all-targets` clean

`builtin_vault_events` now asserts what the event should say — the `substate_id` is a vault, the payload carries the amount and no resource address.


applications/tari_indexer/src/storage_sqlite/migrations/2026-04-20-000000_add_events_resource_address/up.sql
```diff
@@ -1,8 +1,8 @@
 -- Store the resource address (when present) an event refers to, so SSE
 -- consumers can filter a stream down to a single token's activity server-side.
 --
--- Populated from event.substate_id for `std.resource.*` events (the substate_id
--- is the resource address), or from the `resource_address` payload entry for
--- `std.vault.deposit` / `std.vault.withdraw`.
+-- Populated from event.substate_id for `std.resource.*` events, whose substate_id
+-- is the resource address. Other events, vault events included, leave it NULL:
```
sdbondi: _Posted by Claude._

"Other events, vault events included, leave it NULL" only holds for rows written *after* this upgrade. Rows already in an operator's DB keep whatever the old ingest wrote, and nothing backfills them — so the two halves of a single SSE subscription now disagree.

`run_replay_then_live` replays through `store.get_events_after_id(..., filter.resource_address, ...)`, which filters on this column, then switches to a live phase that filters with `EventFilter::matches`. A client reconnecting with `?resource_address=<TARI>&after_id=N` gets the pre-upgrade `std.vault.deposit` / `std.vault.withdraw` rows replayed from the DB and then never sees another one — the stream silently changes meaning at the replay/live boundary, with no error. `GET /transactions/events?resource_address=...` has the same split: transfer history ends abruptly at the upgrade point.

A follow-up migration that NULLs the stale values (e.g. `UPDATE events SET resource_address = NULL WHERE resource_address IS NOT NULL AND substate_id NOT LIKE 'resource_%'`) would make the column mean one thing across the whole table and make this comment true.

docs/developer-docs/public/indexer/indexer-api.html
```diff
@@ -635,7 +635,7 @@
...
.../events/stream",{"_index":217,"title":{},"description":{"38":{}}}],["transactions/rec",{"_index":179,"title":{},"description":{"34":{}}}],["transactions/{transaction_id",{"_index":205,"title":{},"description":{"35":{}}}],["transactions/{transaction_id}/result",{"_index":196,"title":{},"description":{"35":{},"36":{}}}],["two",{"_index":33,"title":{},"description":{"6":{},"35":{}}}],["up",{"_index":152,"title":{},"description":{"34":{}}}],["updat",{"_index":226,"title":{},"description":{"42":{}}}],["utxo",{"_index":220,"title":{},"description":{"40":{},"41":{},"42":{}}}],["utxos/fetch",{"_index":224,"title":{},"description":{"41":{}}}],["utxos/stream",{"_index":227,"title":{},"description":{"42":{}}}],["valid",{"_index":120,"title":{},"description":{"32":{},"44":{}}}],["via",{"_index":100,"title":{},"description":{"25":{},"38":{}}}],["volum",{"_index":66,"title":{},"description":{"13":{}}}],["wait-until-readi",{"_index":47,"title":{},"description":{"7":{}}}],["wait_until_readi",{"_index":37,"title":{"7":{}},"description":{}}],["whose",{"_index":168,"title":{},"description":{"34":{}}}],["window",{"_index":159,"title":{},"description":{"34":{},"35":{}}}]],"pipeline":[]}},"options":{}};
```
sdbondi, line 638: _Posted by Claude._

Only the embedded OpenAPI spec blob on this line got the new wording. The pre-rendered Redoc DOM earlier in the same file still carries the old description twice (around lines 598-599 — once inside the `html="&lt;p&gt;..."` attribute and once as the rendered `<p>`):

> Filter by resource address (derived from substate_id for std.resource.* events, or from the `resource_address` payload entry for std.vault.deposit / std.vault.withdraw)

`grep -c 'derived from substate_id' docs/developer-docs/public/indexer/indexer-api.html` is still 2 after this commit (it was 3 before). Anyone reading the checked-in HTML, or viewing the page before Redoc hydrates from the spec blob, is told the payload fallback still exists.

clients/tari_indexer_client/src/types.rs
```diff
@@ -307,9 +307,9 @@ pub struct QueryTransactionEventsRequest {
     pub topic: Option<String>,
     #[cfg_attr(feature = "utoipa", schema(value_type = Option<String>))]
     pub substate_id: Option<SubstateId>,
-    /// Filter by resource address. Matches when either the event's `substate_id` is the given
-    /// resource (std.resource.* events) or the event payload contains a `resource_address` entry
-    /// equal to the given address (std.vault.deposit / std.vault.withdraw).
+    /// Filter by resource address. Matches the events whose `substate_id` is that resource, which
+    /// is the `std.resource.*` family. Vault events name no resource — filter those by
+    /// `substate_id`, the vault's ID.
```
sdbondi, line 312: _Posted by Claude._

One more copy of this doc was missed — the Rust dapp SDK's own filter type, which is what most callers of `StreamTransactionEventsRequest` actually construct:

`crates/wallet/ootle-rs/src/provider/event_watcher.rs:24`
```rust
/// Filter by resource address. Useful for tracking all activity (deposits, withdrawals,
/// mints, burns, freezes, etc.) for a specific token regardless of which template
/// orchestrated the call.
pub resource_address: Option<ResourceAddress>,
```

After this change deposits and withdrawals no longer match a `resource_address` filter, so a dapp author following this doc sets `TransactionEventFilter { resource_address: Some(token), .. }` to watch a token's transfers and gets a stream that never yields one — silently, since an empty stream and a wrong filter look identical.

applications/tari_indexer/src/storage_sqlite/migrations/2026-04-20-000000_add_events_resource_address/up.sql
sdbondi: _Posted by Claude._

Confirmed and fixed in `fd012eb`. Verified the split: `run_replay_then_live` pages phase 1 through `store.get_events_after_id(.., filter.resource_address, ..)`, which filters on this column, then phase 2 filters with `EventFilter::matches` — so a mixed table does change the stream's meaning silently at the boundary, and the lagged-catch-up path inside phase 2 reads the column again.

Added `2026-09-15-000000_events_resource_address_resource_only` with your `UPDATE`, plus two things it needed:

- `NOT LIKE 'resource\_%' ESCAPE '\'` — `_` is a LIKE single-character wildcard, so the unescaped pattern also matches a substate_id like `resourceXcc` and would leave it set.
- `substate_id IS NULL OR ...`, since an event with no substate_id could also have been given a payload-derived value.

`the_backfill_clears_every_non_resource_resource_address` runs the shipped `up.sql` via `include_str!` against a seeded table and pins both, including the `resourceXcc` case.

I also reverted my edit to the original migration's comment. It correctly describes what that column meant when it was added; the new migration is the right place to record the change.

docs/developer-docs/public/indexer/indexer-api.html
sdbondi: _Posted by Claude._

Right — I had only replaced the plain-text copy. Both remaining copies are fixed in `fd012eb`: the HTML-escaped one inside the `html="..."` attribute (`&lt;code&gt;resource_address&lt;/code&gt;`) and the rendered `<p>`. `grep -c 'derived from substate_id' docs/developer-docs/public/indexer/indexer-api.html` is now 0.

These files are generated by `scripts/indexer-gen-api-docs.sh`, which needs a `--release` build of the indexer plus a `redocly` run, so I hand-edited rather than regenerating. `openapi.json` re-validated as JSON.

clients/tari_indexer_client/src/types.rs
sdbondi: _Posted by Claude._

Confirmed — that doc was the most actively wrong of the three, since it named deposits and withdrawals as the reason to use the field. Rewritten in `fd012eb`:

```rust
/// Filter by resource address. Matches the events that name a resource as their substate —
/// mints, burns, recalls, freezes, metadata and access-rule changes.
///
/// Vault deposits and withdrawals do not match: a vault event names the vault, not the
/// resource. To watch a token's transfers, resolve the vaults holding it — a vault's resource
/// is fixed for its life, so one lookup holds — and set `substate_id` to the vault instead.
```

The silent-empty-stream failure mode is the reason it says what to do instead rather than only what stops working.

sdbondi (commented):
Reviewed

PR #2608, opened 2026-09-15
feat(engine)!: check proofs at the call boundary

sdbondi:
Engine audit wave 3, PR 11 — the last of the auth-model findings. Replaces the "first-frame proof inheritance" risk accepted in wave 1.

## Problem

A top-level instruction copied the transaction's base auth scope into the callee's frame — signer virtual badges **and** any `Proof` the submitter left on the workspace. The callee held those proofs for its whole frame, so it could perform any resource action the badge guarded and forward the proof into nested calls, without the submitter ever naming it as an argument. Only the nested id-guessing case was fixed, by #2577.

## Model

- A method access rule is evaluated at the **call boundary**. The base scope's proofs are in scope for that check and revoked immediately after.
- A frame then acts with the badges it is stamped with (`set_caller`, plus signer badges on a first frame) and the `Proof` arguments it was handed. Nothing else.
- Resource-level checks inside a frame use that frame's scope, unchanged.

## Changes

`CallScope::boundary_proofs` records what `set_auth_scope` seeds from the base scope, excluding proofs that arrived as arguments. `revoke_boundary_proofs` drops that set from `proof_scope`, `inherited_proofs` and `auth_scope` together, so proof accounting in `pop_frame` is unaffected and a callee cannot return a boundary proof into its caller's scope.

Revocation lives in `invoke_template` — the sole path from the engine into template code (four callers, each immediately after a push; the only `WasmProcess::init` site). The migration path revokes at its own ownership check as well, since a migration with no `migrate` function never reaches `invoke_template` and `update_component_template` runs in between.

`set_auth_scope` now merges argument proofs into the base scope rather than replacing them. It previously discarded them from `auth_scope`, harmless only because base proofs were a superset of argument proofs — load-bearing once base proofs stop being seeded.

Also removes `update_from_parent`, an empty function whose body was commented out.

### Wallet side

Flows that relied on a workspace proof reaching an account method now pass it as an argument, using the `_with_auth` methods added in #2588:

- `walletd` `accounts.transfer` with `proof_from_badge_resource` → `withdraw_with_auth`
- SDK confidential transfer with `proof_from_resource` → `withdraw_confidential_with_auth`
- SDK stealth transfer, both builders, all `BadgeUsage` variants → `withdraw_with_auth`

The native `stealth_transfer` / `pay_fee` instructions execute at top level against the base scope and are unaffected.

## Breaking change

A template method that acts on a resource guarded by a badge must take the badge as a `Proof` parameter. Documented in the authorization guide under a new "Proofs reach a frame only as arguments" section.

## Tests

`tari_engine` 462 run, 461 pass. The one failure, `fees::fail_partial_paid_fees`, fails identically on `development` — pre-existing, unrelated. `tari_template_builtin` 15/15, `tari_ootle_wallet_sdk` 42/42.

New:
- `a_workspace_proof_reaches_a_callee_only_as_an_argument` — a component called with no proof argument cannot withdraw a badge-guarded resource while the proof sits on the workspace; the same badge handed to the method works; a frame handed the proof may forward it into a nested call. Verified to fail without the engine change.
- `a_signer_gated_account_method_needs_no_proof_argument` — signer badges still flow into the first frame.

Four existing `access_rules` tests needed `deposit_with_auth`; all four broke on the badge-gated **deposit** rule rather than withdraw.

Not covered by tests, verified by reading: the `UpdateComponentTemplate` migration frame and the `CreateAccount`-with-existing-account deposit path.

`ENGINE_FINGERPRINT` stays at v5 — it keys the compiled-WASM artifact cache, and this changes no artifact, metering value or header field.

## Known gap

`utilities/traffic-sim` creates an admin-badge proof and then calls `withdraw` on an externally deployed faucet template. That template is not in this repo and exposes no proof-taking method, so the flow needs the deployed faucet updated before it works again.

## Out of scope — pre-existing, filed separately

Two findings surfaced while reviewing this change. Both exist unchanged on `development` and neither is touched by this PR:

1. `validate_component_state` does not reject `ProofId` / `BucketId` in component state. A template that stores a `Proof` persists a transaction-scoped id to the ledger, and `invoke_resource_access_hook` passes component state into the hook frame as an argument, where `include_refs_in_scope` treats a proof id as an authorizing proof. Bounded by the `OwnComponent` sandbox to information disclosure (the hook reads the acting caller's badge and its NFT ids). Fix: reject transient ids alongside the existing vault checks.
2. `AuthorizationScope::contains_badge_of_resource` compares resource addresses across `virtual_proofs.chain(caller_badges)`, and every caller-component badge shares one fixed resource address. `rule!(resource(CALLER_COMPONENT_RESOURCE_ADDRESS))` is therefore satisfied by any caller component. `is_system_reserved()` covers both caller-badge resources but is enforced only at resource creation, never at rule construction.


applications/tari_walletd/src/handlers/accounts.rs
```diff
@@ -1020,10 +1020,22 @@ pub async fn handle_transfer(
...
+                ])
+            } else {
+                builder.call_method(source_account_address, "withdraw", args![
+                    req.resource_address,
+                    req.amount
+                ])
+            }
+        })
         .put_last_instruction_output_on_workspace("bucket")
         .call_method(destination_account_address, "deposit", args![Workspace("bucket")])
```
sdbondi, line 1040: _Posted by Claude._

The withdraw side is now proof-passing, but the destination `deposit` still calls the plain method, so this flow regresses for any resource whose **deposit** rule names the badge.

Before this PR the workspace proof reached the destination account's first frame and authorized `Vault::deposit`'s `ResourceAuthAction::Deposit` check. After the boundary revoke, the destination frame holds only signer badges, and `deposit` takes no `Proof`, so `check_resource_access_rules(Deposit, ...)` fails with `AccessDenied`.

This is exactly the case the PR's own test changes ran into — "all four broke on the badge-gated **deposit** rule rather than withdraw" — and a permissioned/whitelisted token gating both withdraw and deposit on one badge is the motivating use of `proof_from_badge_resource`. `accounts.transfer` with such a resource now fails at this instruction.

Suggest routing through `deposit_with_auth` when `proof_from_badge_resource.is_some()`, mirroring the withdraw branch above.

crates/wallet/sdk/src/apis/confidential_transfer.rs
```diff
@@ -447,10 +447,22 @@ where TSpec: WalletSdkSpec
...
+                    )
+                } else {
+                    builder.call_method(*from_account.component_address(), "withdraw_confidential", args![
+                        params.resource_address,
+                        proof
+                    ])
+                }
+            })
             .put_last_instruction_output_on_workspace("bucket")
             .call_method(to_account.address, "deposit", args![Workspace("bucket")])
```
sdbondi, line 467: _Posted by Claude._

Same gap as in `walletd`'s `accounts.transfer`: `withdraw_confidential` was converted to `_with_auth`, but the destination `deposit` was not.

The badge proof is on the workspace, yet after the boundary revoke the destination account's frame no longer holds it, and plain `deposit` takes no `Proof`. A confidential resource whose deposit rule names the badge (the usual shape for a permissioned token — `proof_from_resource` exists precisely for those) now fails here with `AccessDenied { Deposit }`, where it succeeded before this PR.

`deposit_with_auth(bucket, proof)` when `params.proof_from_resource.is_some()` would close it.

crates/engine/src/transaction/processor.rs
```diff
@@ -945,12 +948,18 @@ where
...
+    /// into template code, so it is where a frame stops holding the proofs that were in scope for its call
+    /// boundary: whatever the frame's access rule was evaluated against, the code itself acts with its own badges
+    /// and its `Proof` arguments.
     fn invoke_template(
         module: LoadedTemplate,
         mut runtime: Runtime,
         function_def: &FunctionDef,
         args: &[tari_bor::Value],
     ) -> Result<InstructionResult, TransactionErrorKind> {
+        runtime.interface_mut().revoke_boundary_proofs()?;
```
sdbondi, line 961: _Posted by Claude._

This revoke also lands on the `CreateAccount` deposit paths, which have no way to pass a proof — so a badge-gated **deposit** resource becomes unsendable to a new (or bucket-funded) account, with no workaround at the instruction level.

Both `CreateAccount` branches reach template code through here:
- existing account: `invoke_component(..., ACCOUNT_DEPOSIT_METHOD, [Workspace(bucket)])` — plain `deposit`, no `Proof` parameter;
- new account: the `Account::create` constructor deposits the initial bucket inside its own frame.

In both the frame is left with signer badges only, so `check_resource_access_rules(Deposit, ...)` fails for a resource whose deposit rule names a badge. Previously the submitter's workspace proof was in that first frame and authorized it.

This is reachable from the SDK today: `stealth_transfer`'s `create_account_with_bucket(...)` for revealed outputs (api.rs:1064/1067 and 1177/1179) is a `CreateAccount` with a bucket, and it is used with every non-`None` `BadgeUsage`. Either `Instruction::CreateAccount` needs an optional proof workspace id threaded into the deposit call, or the limitation belongs in the "Known gap" section alongside traffic-sim.

docs/developer-docs/src/content/docs/guides/authorization-and-access.mdx
```diff
@@ -149,6 +149,38 @@ and update its own component state (when declared `&mut self`), but cannot write
...
+.call_method(component, "take_tokens", args![Workspace("proof"), 100])
+```
+
+The built-in account follows the same rule: `withdraw`, `deposit` and the `create_proof_*` methods each have a
+`_with_auth` variant taking the proof, and a resource whose withdraw or deposit rule names a badge is reachable
+only through those.
+
+A proof handed in remains the caller's — the callee may hold and forward it but never consumes it — and it stops
+authorizing when the caller drops it. Signer badges are not proofs: they are stamped into every frame a
+top-level instruction pushes, so a rule naming the transaction's signer needs no `Proof` argument.
```
sdbondi, line 182: _Posted by Claude._

"every frame a top-level instruction pushes" overstates where signer badges reach, and a template author acting on it will write a method that cannot be called from another component.

`WorkingState::push_frame` copies the base scope's `virtual_proofs` only when `call_frames` is empty — the *first* frame of a top-level instruction. A nested frame starts from `CallScope::new()` and gets `set_caller` badges only, so a rule naming the signer is **not** satisfied once a component calls the method on the submitter's behalf.

Suggest "stamped into the first frame a top-level instruction pushes", with a note that nested frames see only caller badges and `Proof` arguments — which is also what makes the `_with_auth` variants necessary for composability.

sdbondi:
_Posted by Claude._

Happy with this. No blockers. Traced the engine change end to end; notes below are for the record and one follow-up.

### Verified

- **Ordering is right on every path.** `call_method`: push → `check_component_access_rules` → `invoke_template` revokes before instantiation. `call_function` and the `CreateAccount` create path push `Static` frames with no method rule, and revoke before any code runs. The migration path revokes after its ownership check, and the second revoke inside `invoke_template` is a no-op via `mem::take`. Resource auth hooks reach template code through `invoke_component_method` → `call_method` → `invoke_template`, so hook frames are covered too.
- **Argument proofs survive when they coincide with base proofs.** `set_auth_scope` excludes them from `boundary_proofs` and merges them back into the installed scope. Previously the base scope replaced them, which only worked because base was a superset; the merge is the load-bearing fix now.
- **Pop accounting holds.** Revocation removes from `proof_scope`, `inherited_proofs` and `auth_scope` together, so `proofs_owed` doesn't see a phantom, and a callee that names a revoked proof in its return value fails `validate_return_value` rather than leaking it upward.
- **Nested frames** never call `set_auth_scope`, so `boundary_proofs` is empty and nothing changes for them. Signer badges still flow into the first frame, pinned by `a_signer_gated_account_method_needs_no_proof_argument`.
- `ENGINE_FINGERPRINT` unchanged is correct: nothing here touches a compiled artifact.

### Note: nested calls are stricter than the boundary model states

The body says a method rule is "evaluated at the call boundary against the caller". That's true for top-level instructions (base scope). For a nested call the check sees only the callee's stamped badges plus `Proof` arguments; the calling frame's own held proofs are not consulted. So a component that was handed a proof and calls another component whose method rule names that badge must forward the proof explicitly. That was already the behaviour before this PR and it matches "proofs reach a frame only as arguments", so nothing to change here. Worth one sentence in the authorization guide so template authors don't expect the base-scope behaviour one level down. If we ever want the symmetric model, seeding `boundary_proofs` from the calling frame's `proof_scope` is the one-line hook and the revoke path already handles it.

### CI

`test (3/3)` fails here and on development's latest `CI` run at 82152315b, consistent with the pre-existing `fees::fail_partial_paid_fees` failure the body cites. Not this PR's problem, but it means this merges without a green run; fixing it on development first would be cleaner.

### Out-of-scope findings

Both should become issues so they don't get lost. The first (`ProofId`/`BucketId` persisted in component state, then treated as an authorizing proof when state is passed into a hook frame) is a real gap in `validate_component_state` and belongs with the wave 4 hygiene set. The second (`rule!(resource(CALLER_COMPONENT_RESOURCE_ADDRESS))` satisfied by any caller component) is a footgun for template authors; rejecting system-reserved resources at rule construction closes it.

PR #2609, opened 2026-09-15
fix(engine): checksum the wasm cache header, fsync before publish

sdbondi:
## Description

The 40-byte header of an on-disk WASM cache entry (`code_size` plus the four `ModuleShape` counts)
sits **outside** the wasmer artifact, so corruption confined to it leaves `deserialize_unchecked`
happy. The four shape counts are consensus inputs: `instantiation_points` prices them as
`native_execution_points` into the committed fee receipt. A damaged header therefore yields a
divergent receipt with no error raised anywhere. `store` was `fs::write` + `rename` with no sync, so
a crash could publish a file whose header landed without its tail.

This adds a CRC32 over the header fields, verified on load, and makes the write durable before the
rename.

- CRC32 over the 40 header bytes, stored in a trailing 8-byte LE field (`HEADER_BYTES` 40 → 48).
  A mismatch in `try_load` takes the existing corrupt-file branch: unlink, recompile from source.
- The CRC field is padded to 8 bytes so the header length stays a multiple of 8. The artifact starts
  at `HEADER_BYTES` into a page-aligned mmap and `wasmer_types::MetadataHeader::parse` rejects a
  start address that is not 8-byte aligned — a 4-byte field breaks every cache hit. A `const` assert
  guards the invariant.
- `store` builds the header as a `[u64; HEADER_FIELD_COUNT]` array, so adding a field without
  widening the header is a compile error rather than a silently misaligned header.
- The tempfile is `fsync`'d before the rename, and removed if the write fails.

No `ENGINE_FINGERPRINT` bump. A file written under the old 40-byte header reads its CRC from the
artifact's magic bytes (`b"wasmer-u"` as a LE u64, far above `u32::MAX`), so it can never match and
always self-heals.

The CRC covers only the header; whole-file coverage would force a full read on every load and defeat
the mmap.

## Motivation and Context

First of four PRs on the compiled-artifact cache (eviction, single-flight, ingress validation and
prewarm follow). This one is the correctness fix that the rest build on: a cache entry is a memo of
consensus-priced values read back from an unauthenticated local file, so the reader needs a verdict
on the bytes no other check reaches.

## How Has This Been Tested?

`cargo nextest run -p tari_engine --features wasm-cache` — new cases cover a flipped shape byte, a
flipped CRC byte, a truncated header, and a round-trip hit that checks the stored CRC against the
file's own bytes. `cargo clippy -p tari_engine --features wasm-cache --all-targets` is clean apart
from a pre-existing `too_many_arguments` warning.

`fees::fail_partial_paid_fees` fails on this branch and on `development` alike; it does not touch
the disk cache.

Run a node with the disk cache enabled, flip a byte anywhere in the first 40 bytes of a
`{template_address}_v5.bin` file, and restart: the entry is removed and the template recompiles from
source, where before the change the node would have served the damaged shape counts into a fee
receipt.

## Breaking Changes

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


crates/engine/src/wasm/cache.rs
```diff
@@ -224,17 +260,30 @@ impl WasmModuleCache {
...
+        }
+        let crc = u64::from(crc32fast::hash(&header[..HEADER_FIELD_BYTES]));
+        header[HEADER_FIELD_BYTES..].copy_from_slice(&crc.to_le_bytes());
+
+        let mut bytes = Vec::with_capacity(HEADER_BYTES + serialized.len());
+        bytes.extend_from_slice(&header);
         bytes.extend_from_slice(&serialized);
 
-        if let Err(e) = fs::write(&tmp, &bytes) {
+        if let Err(e) = write_durable(&tmp, &bytes) {
```
sdbondi, line 297: _Posted by Claude._

**The tempfile name is not unique per writer, so the new durability guarantee can still publish a torn artifact.**

`tmp` is `{addr}_{ENGINE_FINGERPRINT}.bin.tmp.{pid}` — keyed by address and *process*, not by the caller. In the indexer, two independent writers in the same process share `wasm_cache_dir` (`applications/tari_indexer/src/bootstrap.rs:334` hands the same path to both `TemplateManager`'s `WasmModuleCache` and `build_dry_run_template_provider`'s `DiskCachedWasmTemplateProvider`), and only the latter is serialised by `MemoryCacheTemplateProvider`'s per-address mutex. So `TemplateManager::load_template_with_cache` and a dry-run miss on the same template address can call `store` concurrently with the same `tmp` path.

Both then run `write_durable`, whose `File::create` is `O_CREAT|O_TRUNC`: the second writer truncates the first writer's partially written file to zero while the first holds an independent write offset. Both then write from their own offsets into the same inode. If writer A renames first, writer B keeps writing into the *published* file (same inode); if A renames while B's truncation has left a hole in the middle of the artifact, the published file has a valid header (the header bytes are identical for both writers, so the new CRC passes) and a torn artifact body — which is precisely the `deserialize_unchecked`-on-garbage case this PR is trying to close.

The failure paths make it worse: after a failed `write_durable` or a failed `rename`, `fs::remove_file(&tmp)` can delete the *other* writer's in-flight tempfile, turning its `rename` into a spurious `ENOENT` warning.

Suggest making the tempfile unique per call (an `AtomicU64` counter or random nonce appended to the name), so `store` never shares a tempfile with a concurrent `store`.

crates/engine/src/wasm/cache.rs
```diff
@@ -72,12 +74,27 @@ pub const ENGINE_FINGERPRINT: &str = "v5";
...
+/// The fields sit outside the wasmer artifact, so `deserialize_unchecked` has no view of damage
+/// confined to them, while the four shape counts price `instantiation_points` into a committed fee
+/// receipt. The CRC is the only check that reaches those bytes.
+///
+/// The CRC occupies a full 8 bytes so that the total stays a multiple of 8: the artifact starts at
+/// `HEADER_BYTES` into a page-aligned mmap, and `MetadataHeader::parse` rejects a start address
+/// that is not 8-byte aligned.
+const HEADER_BYTES: usize = HEADER_FIELD_BYTES + 8;
+
+const _: () = assert!(HEADER_BYTES.is_multiple_of(8), "the artifact must start 8-byte aligned");
```
sdbondi: _Posted by Claude._

**This const assert is tautological and cannot guard the invariant it names.**

`HEADER_BYTES` is defined one line above as `HEADER_FIELD_COUNT * 8 + 8`. Both terms are hard-coded multiples of 8, so the sum is a multiple of 8 for every possible `HEADER_FIELD_COUNT` — the assert can never fire, whatever a future change does to the field count. The hazard the comment describes (a narrower CRC/pad field silently breaking every cache hit) is not expressible while the pad width is a literal `8`.

Also worth noting that 8 is the weaker of the two constraints. `MetadataHeader::parse` checks 8-byte alignment, but the slice handed to `SerializableModule::archive_from_slice` starts at `HEADER_BYTES + 32` and reaches `rkyv::access_unchecked`, and wasmer's own layout aligns metadata to `MetadataHeader::ALIGN == 16`. `HEADER_BYTES == 48` satisfies both by luck; a future 40 or 56 would pass this assert while leaving the rkyv root access misaligned. If the assert is meant to be load-bearing, asserting a multiple of 16 is the safer bound.

crates/engine/src/wasm/cache.rs
```diff
@@ -57,8 +57,10 @@ const LOG_TARGET: &str = "tari::engine::wasm::cache";
 ///   so a node serving a stale artifact meters against the old cap and diverges from one that compiled fresh.
 /// - The `wasmer` crate version (the serialized artifact format is internal to wasmer and not part of any stable wire
 ///   spec).
-/// - The layout of the file's own header (`HEADER_BYTES` and the fields it carries). A reader that disagrees with the
-///   writer about the header slices the artifact at the wrong offset and reads the fields from the wrong bytes.
+/// - The order or meaning of the header's fields at an unchanged `HEADER_BYTES`. A reader that agrees with the writer
+///   on the header's length but not on its field order recomputes a matching CRC over values it then reads from the
+///   wrong bytes. A change to the length needs no bump: the reader takes the CRC from the wrong offset and rejects the
```
sdbondi: _Posted by Claude._

**The stated reason a length change needs no fingerprint bump does not hold for a header that shrinks.**

For a *growing* header the claim is right: the reader takes its CRC from the artifact's first 8 bytes (`b"wasmer-u"` as LE u64, far above `u32::MAX`), which can never match.

For a *shrinking* header it isn't. Drop a shape field (5 fields -> 4, `HEADER_BYTES` 48 -> 40) and the new reader reads `stored_crc` from `bytes[32..40]`, which in the old layout is a real field value (`declared_table_slots`) — a small `u64` that is in the CRC's range. Acceptance is then a ~2^-32 coin flip rather than impossible, and on acceptance the reader has sliced the shape counts out of the wrong offsets and would price `instantiation_points` off them.

What actually saves that case is the artifact magic: the body would be sliced at the wrong offset, so `ArtifactBuild::is_deserializable` fails and the entry is removed. Since this doc is the thing a future maintainer will consult before changing a consensus-priced field, it's worth stating the real reason rather than one that only covers the grow direction.

crates/engine/src/wasm/cache.rs
```diff
@@ -474,6 +537,70 @@ mod tests {
...
+        let mut bytes = fs::read(&path).unwrap();
+        bytes[8] ^= 0x01;
+        fs::write(&path, &bytes).unwrap();
+
+        assert!(cache.try_load(&addr).is_none(), "a bad header CRC is a miss");
+        assert!(!path.exists(), "the file with the bad CRC should be removed");
+    }
+
+    #[test]
+    fn flipped_crc_byte_falls_back_to_recompile() {
```
sdbondi, line 572: _Posted by Claude._

**No test covers the backwards-compatibility claim the PR rests on.**

The description argues that no `ENGINE_FINGERPRINT` bump is needed because a file written under the old 40-byte header reads its CRC out of the artifact magic and therefore always self-heals. Every node upgrading onto this commit hits that path on its first load of each cached template, but none of the four new tests exercise it — `flipped_crc_byte` and `flipped_header_byte` both start from a file this build wrote.

A test that writes the legacy layout (40-byte header, no CRC, artifact appended) and asserts `try_load` returns `None` and unlinks would pin the claim down; as it stands the only evidence is the reasoning in the PR body.

crates/engine/src/wasm/cache.rs
sdbondi: _Posted by Claude._

Confirmed and fixed in 59594145d. `applications/tari_indexer/src/bootstrap.rs:334` hands `wasm_cache_dir` to both the `TemplateManager`'s `WasmModuleCache` (which calls `store` from `load_template_with_cache`, `manager.rs:133`) and `build_dry_run_template_provider`, so two unserialised writers in one process do share the pid-keyed tempfile.

`store` now appends a process-unique counter to the tempfile name, so no two calls can name the same file.

crates/engine/src/wasm/cache.rs
sdbondi: _Posted by Claude._

Right on both counts, fixed in 59594145d. `HEADER_FIELD_COUNT * 8 + 8` is a multiple of 8 unconditionally, so the assert could never fire.

It now asserts a multiple of 16 and the doc names the binding constraint: the metadata slice starts at `HEADER_BYTES + 32` (`MAGIC_HEADER` 16 + `MetadataHeader::LEN` 16) and reaches `rkyv::access_unchecked` via `archive_from_slice`, with `serialize_module` asserting the archived root's alignment is at most `MetadataHeader::ALIGN` (16). `MetadataHeader::parse`'s 8-byte check is the weaker bound. A future `HEADER_FIELD_COUNT` of 4 or 6 now fails to compile.

crates/engine/src/wasm/cache.rs
sdbondi: _Posted by Claude._

Agreed, fixed in 59594145d. The CRC argument only covers a header that grows; for a shrinking header the CRC lands on a real field value and acceptance is a coin flip.

The bullet now gives the reason that holds in both directions — a length change moves the artifact's start, so the body no longer begins with `ArtifactBuild::MAGIC_HEADER` and `Artifact::deserialize_unchecked` rejects it at `is_deserializable` before anything is read out of it — and notes the CRC as the earlier catch in the grow direction only.

crates/engine/src/wasm/cache.rs
sdbondi: _Posted by Claude._

Added in ee65097e8: `legacy_header_without_crc_falls_back_to_recompile` rewrites a freshly stored entry into the pre-CRC layout (the five fields, artifact directly after them) and asserts `try_load` returns `None` and unlinks the file.

Correcting the scope I gave earlier: this is not a path every upgrading node takes. No release has written a v5 file — v0.40.2 is still on `ENGINE_FINGERPRINT = "v4"` with an 8-byte header, and those files are orphans under a different filename suffix that `try_load` never opens. The pre-CRC layout only exists in a data directory written by an unreleased `development` build. The test is worth keeping for those, and as cover for the next header change.

sdbondi: _Posted by Claude._

Dropping this one instead. The test I added exercised no branch that `flipped_crc_byte_falls_back_to_recompile` does not already cover — a header whose CRC does not verify is unlinked, whatever the bytes in the CRC slot happen to be — and the scenario is narrower than I first said: no release has written a v5 file (v0.40.2 is still on `ENGINE_FINGERPRINT = "v4"`, whose files are orphans under a different suffix that `try_load` never opens), so the pre-CRC layout only exists in a data directory written by an unreleased `development` build, and it survives exactly one load there.

The reason a bump is unnecessary is a fact about constants rather than about this code: `b"wasmer-u"` as a LE u64 is `0x752d_7265_6d73_6177`, which no CRC32 can equal. That belongs in the fingerprint doc, where it now is, not in a test. Removed in c52aa075e.

sdbondi (commented):
Reviewed

crates/engine/src/wasm/cache.rs
```diff
@@ -217,24 +261,42 @@ impl WasmModuleCache {
...
+            shape.data_segment_count,
+            shape.element_segment_entries,
+            shape.declared_table_slots,
+        ];
+        let mut header = [0u8; HEADER_BYTES];
+        for (slot, value) in header.chunks_exact_mut(8).zip(fields) {
+            slot.copy_from_slice(&value.to_le_bytes());
+        }
+        let crc = u64::from(crc32fast::hash(&header[..HEADER_FIELD_BYTES]));
+        header[HEADER_FIELD_BYTES..].copy_from_slice(&crc.to_le_bytes());
```
sdbondi: _Posted by Claude._

**The "add a field and it's a compile error" invariant leads the next editor straight into a runtime panic.**

Both sides hardcode that the CRC slot is exactly 8 bytes wide: here `header[HEADER_FIELD_BYTES..]` must be 8 bytes for `copy_from_slice`, and in `try_load` `mmap[HEADER_FIELD_BYTES..HEADER_BYTES].try_into()` must produce `[u8; 8]` for the `.expect(...)` to hold. Nothing ties that to the constants.

Follow the path the code prescribes. A later PR in this series adds a sixth header field and bumps `HEADER_FIELD_COUNT` to 6. `HEADER_BYTES` becomes 56, the `const` assert fires — good, as designed. The assert's message says the total must be a multiple of 16, so the fix is to pad: `const HEADER_BYTES: usize = HEADER_FIELD_BYTES + 16;` -> 64, assert passes, it compiles. Now `header[48..64].copy_from_slice(&crc.to_le_bytes())` panics on a 16-vs-8 length mismatch on *every* store, and `try_load`'s `.expect("8 bytes of a slice at least HEADER_BYTES long")` panics on every load — in two functions documented as best-effort and non-fatal.

Name the offset and derive the field, e.g. `const CRC_OFFSET: usize = HEADER_BYTES - 8;`, index `header[CRC_OFFSET..CRC_OFFSET + 8]` on both sides, and assert `HEADER_BYTES >= HEADER_FIELD_BYTES + 8` alongside the alignment assert.

crates/engine/src/wasm/cache.rs
```diff
@@ -158,6 +183,25 @@ impl WasmModuleCache {
...
+        if stored_crc != computed_crc {
+            warn!(
+                target: LOG_TARGET,
+                "Cache file {} has a corrupt header (CRC {:#010x}, expected {:#010x}); removing.",
+                path.display(),
+                stored_crc,
+                computed_crc,
+            );
+            drop(mmap);
+            let _ignore = fs::remove_file(&path);
```
sdbondi, line 211: _Posted by Claude._

**The unlink can delete a good entry that a concurrent `store` just published.**

The `store` comment two hunks down establishes that two `store`s for one address run concurrently (the indexer opens the dir twice). The same holds for `try_load` racing `store`, and this removal is by path, not by the inode the CRC was actually computed over:

1. Thread A `try_load(addr)` mmaps the file and its header CRC mismatches (genuine corruption, a half-written entry from a previous crash, whatever).
2. Thread B misses, compiles, and `store` renames a fresh, valid file over the same path.
3. Thread A finishes its check, `drop(mmap)`, and `fs::remove_file(&path)` — unlinking B's good entry.

Cost is one extra compile, so it is recoverable, but it is the same window the new per-call tempfile was added to close, on the read side. The same shape already exists in the deserialize-failure branch below, so this is extending an existing pattern rather than introducing it — worth handling in the same pass, e.g. by only unlinking when the path still resolves to the inode that was mapped.

crates/engine/src/wasm/cache.rs
```diff
@@ -72,12 +78,31 @@ pub const ENGINE_FINGERPRINT: &str = "v5";
...
+const _: () = assert!(
+    HEADER_BYTES.is_multiple_of(16),
+    "the artifact's rkyv metadata must start 16-byte aligned",
+);
 
 /// Low-level on-disk cache for compiled wasmer modules.
 ///
 /// Files live at `{dir}/{template_address}_{ENGINE_FINGERPRINT}.bin`.
-/// The body is `[u64 LE: code_size][u64 LE x4: ModuleShape] || wasmer::Module::serialize(...)`.
+/// The body is `[u64 LE: code_size][u64 LE x4: ModuleShape][u64 LE: CRC32 of the preceding 40 bytes]
```
sdbondi: _Posted by Claude._

Nit: `40` is spelled literally here while everything else in the file derives it from `HEADER_FIELD_COUNT`. The rest of the PR's design is that adding a field forces the constants to move together; this line is the one place that won't, and it will quietly start describing a layout the code no longer writes. `CRC32 of the preceding HEADER_FIELD_BYTES` (or just "of the preceding fields") keeps it honest.

crates/engine/src/wasm/cache.rs
```diff
@@ -217,24 +261,42 @@ impl WasmModuleCache {
         };
 
         let path = self.path_for(addr);
+        // Every call gets its own tempfile. Two `store`s of one address can run concurrently — the
+        // indexer opens the cache directory twice, once for the template manager and once for the
+        // dry-run provider — and a shared tempfile would let them truncate and interleave into a
+        // single inode, publishing a torn artifact under a header whose CRC still verifies.
```
sdbondi: _Posted by Claude._

**Comment policy: this narrates the bug that is no longer in the code.**

`CLAUDE.md` is explicit that comments are present-tense statements of intent or invariant, never "without this guard X would happen", and never define the code by what it is not. "a shared tempfile *would* let them truncate and interleave" describes a shared tempfile — and there isn't one. A reader who never saw the old code has to reconstruct a hypothetical to parse the sentence, and the rationale for the fix belongs in the commit message (where it already is).

The invariant is one line: each `store` writes to a tempfile no other `store` can name, so concurrent writers for one address never share an inode. The concrete detail worth keeping is the one a future reader can't derive — that concurrent `store`s of a single address are real, because the indexer opens the cache directory twice.

Same pattern at the `fields` array below: "...is a compile error rather than a header the reader slices at the wrong offsets" — the second half describes the layout the code doesn't produce. "`HEADER_FIELD_COUNT` fields, in the order `try_load` reads them" says everything a reader needs.

crates/engine/src/wasm/cache.rs
```diff
@@ -474,6 +550,70 @@ mod tests {
...
+        let cache = WasmModuleCache::open(dir.path()).unwrap();
+        let (store, addr) = make_store();
+        let provider = DiskCachedWasmTemplateProvider::new(store, cache.clone());
+        let loaded = provider.get_template(&addr).unwrap().expect("loaded");
+
+        let bytes = fs::read(cache.path_for(&addr)).unwrap();
+        let stored_crc = u64::from_le_bytes(bytes[HEADER_FIELD_BYTES..HEADER_BYTES].try_into().unwrap());
+        assert_eq!(stored_crc, u64::from(crc32fast::hash(&bytes[..HEADER_FIELD_BYTES])));
+
+        let reloaded = cache.try_load(&addr).expect("hit");
```
sdbondi, line 621: _Posted by Claude._

**Nothing in this file asserts that the four `ModuleShape` counts round-trip, and this PR rewrote the code that lays them out.**

`store` went from four explicit `extend_from_slice` calls naming each field to an array + `chunks_exact_mut` loop. The reader still hardcodes slot -> field by index (`read_field(1)` -> `data_segment_bytes`, `read_field(2)` -> `data_segment_count`, ...). All four fields are `u64`, so swapping two entries in the new `fields` array — say `data_segment_count` and `element_segment_entries` — type-checks, and the CRC covers the bytes either way. Grepping this module for `shape` outside `store`/`try_load` turns up nothing: no test reads them back.

Walk the tests against that mutation:
- `round_trip_compile_then_deserialize` — checks name, `code_size`, function count. Passes.
- `flipped_header_byte` / `flipped_crc_byte` — assert only that the CRC rejects a flipped byte. Passes regardless of field order.
- `stored_header_crc_covers_the_fields` — recomputes the CRC over the same bytes the writer did, then asserts `code_size` (slot 0, the one field a transposition among slots 1-4 can't touch). Passes.

So a transposition ships green. `instantiation_points` prices each of the four with a *different* constant (`PER_TEMPLATE_DATA_SEGMENT_BYTE`, `PER_TEMPLATE_DATA_SEGMENT`, `PER_TEMPLATE_ELEMENT_ENTRY`, `PER_TEMPLATE_TABLE_SLOT`), so a cache hit charges a different fee than a fresh compile of the same template — the exact divergence class this PR's header CRC exists to rule out, reintroduced one layer up where the CRC cannot see it.

This test is the natural home for the assertion. `LoadedTemplate` only exposes `code_size()`, but the variant is reachable:

```suggestion
        let reloaded = cache.try_load(&addr).expect("hit");
        assert_eq!(reloaded.code_size(), loaded.code_size());

        // The four shape counts are consensus inputs — `instantiation_points` prices each with a
        // distinct constant, so a cache hit must serve them in the order a fresh compile produced.
        let LoadedTemplate::Wasm(reloaded) = &reloaded;
        let LoadedTemplate::Wasm(loaded) = &loaded;
        assert_eq!(reloaded.shape(), loaded.shape());
    }
```

`ModuleShape` already derives `Debug, PartialEq, Eq`, and `LoadedWasmTemplate::shape()` is `pub`, so that compiles as-is.

Separately, the second commit's per-call tempfile has no test at all. Two threads calling `store(addr, ..)` on one `WasmModuleCache`, joined, then `try_load(addr).is_some()` would fail against the old shared-tempfile name and pass now — cheap, deterministic, and it pins the property the commit exists for.

crates/engine/src/wasm/cache.rs
```diff
@@ -158,6 +183,25 @@ impl WasmModuleCache {
             return None;
         }
 
+        let stored_crc = u64::from_le_bytes(
```
sdbondi, line 196: _Posted by Claude._

**The `unsafe { Mmap::map }` SAFETY note just above (L156-162) now rests on a premise this PR refutes.**

It reads: *"The cache dir is owned by this process (single writer, atomic rename on update), so this is safe in the deployment model."* The new comment on the tempfile in `store` says the opposite in as many words — two `store`s for one address run concurrently, because the indexer opens the cache directory twice. "Single writer" is the load-bearing clause of a safety argument for a mapping that SIGBUSes if the mapped file is truncated underneath it, and it is now documented as false in the same file.

The mapping is still sound, but for the reason this PR supplies rather than the one written down: `store` never writes through the published path. It creates a per-call tempfile, fills it, and `rename`s — which swaps the directory entry to a new inode and leaves the mapped one untouched, so a concurrent writer cannot shrink the mapping. `write_durable`'s `File::create` does truncate, but only a tempfile no reader can name (which is exactly what the counter suffix guarantees).

Worth rewriting the SAFETY note to that argument while the concurrency model is being pinned down here — a reader who takes "single writer" at face value and later adds an in-place write to the published path would be reintroducing the SIGBUS with the comment's blessing.

crates/engine/src/wasm/cache.rs
```diff
@@ -158,6 +183,25 @@ impl WasmModuleCache {
             return None;
         }
```
sdbondi, line 194: _Posted by Claude._

**One damage mode still doesn't self-heal: a zero-length cache file.**

Every other branch in `try_load` now unlinks — short file, bad header CRC, failed deserialize. The mmap-failure branch above (L163-172) warns and returns `None` without removing. `memmap2` refuses a zero-length mapping outright (`"memory map must have a non-zero length"`), so a cache file truncated to 0 by an fsck, a full disk, or a botched restore never reaches the length check the new `truncated_header_treated_as_miss` test exercises — it is turned away one step earlier and left on disk permanently.

Inside `DiskCachedWasmTemplateProvider` the next `get_template` recompiles and `store` renames over it, so it heals there. It does not heal for the standalone `WasmModuleCache::try_load` users the module docs call out (the wallet daemon's template monitor), which load without ever storing: that entry is a permanent miss, and `has_template` keeps answering `true` for it because it only stats the path.

Removing the file on a map failure — it is unreadable by definition — would make the "any corrupt entry self-heals" property in the type's doc comment hold for real. The new truncation test extends naturally to cover it with a `0`-byte case alongside the `HEADER_BYTES - 1` one; note that it needs a different assertion path, since the two sizes take different branches today.

crates/engine/src/wasm/cache.rs
sdbondi: _Posted by Claude._

Confirmed — walked the same path, and both slot accesses do panic once `HEADER_BYTES - HEADER_FIELD_BYTES` stops being 8.

Fixed in 2914971d6e by naming the layout: `HEADER_PAD_BYTES` (0 today) sits between the fields and the CRC, `CRC_OFFSET = HEADER_FIELD_BYTES + HEADER_PAD_BYTES`, and `HEADER_BYTES = CRC_OFFSET + 8`. Both sides index `[CRC_OFFSET..HEADER_BYTES]`, which is 8 bytes by construction whatever the pad is, and the assert's message names the pad as the knob, so the fix it prescribes keeps the slot the right width. The CRC covers `[..CRC_OFFSET]` on both sides, so a pad change cannot desync writer and reader either.

crates/engine/src/wasm/cache.rs
sdbondi: _Posted by Claude._

Fixed in 2914971d6e — it reads `CRC32 of the preceding fields` now, with no width spelled out.

crates/engine/src/wasm/cache.rs
sdbondi: _Posted by Claude._

Agreed, and both were already rewritten in eebe84bc1f before this landed. The tempfile comment now states the invariant and keeps only the fact a reader cannot derive:

```rust
// Every call needs its own tempfile: two `store`s of one address can run concurrently,
// since the indexer opens the cache directory twice, once for the template manager and
// once for the dry-run provider.
```

The `fields` comment is now `// HEADER_FIELD_COUNT fields, in the order try_load reads them.`, taking your wording — it says more than the version it replaced, in fewer words. Grepping every added comment line in the diff turned up one more of the same shape in `write_durable`'s doc, fixed in the same commit.

crates/engine/src/wasm/cache.rs
sdbondi: _Posted by Claude._

Took the shape assertion in 2914971d6e, and checked it is load-bearing rather than decorative: transposing `data_segment_count` and `element_segment_entries` in the `fields` array fails `stored_header_crc_covers_the_fields` and nothing else, so the mutation you describe is caught and caught only there.

Not taking the concurrency test. Two threads calling `store` on one address and joining does not fail deterministically against the shared-tempfile name — both writers have to be inside `write_durable` at once for the interleave, and on a warm page cache the first commonly finishes before the second opens. A test that passes against the code it is meant to reject pins nothing, and a sleep or a barrier inside `store` to force the window is instrumentation this module does not otherwise need. The property is enforced by construction instead: the name carries a counter that `fetch_add` makes unique per call, so no two calls can name one file.

crates/engine/src/wasm/cache.rs
sdbondi: _Posted by Claude._

Good catch — the note's load-bearing clause was contradicted by the commit that made it false. Rewritten in 2914971d6e to argue from the invariant that actually holds:

```rust
// SAFETY: see the docs on `Mmap::map`. A mapping SIGBUSes if the file shrinks under it, so
// what keeps this sound is that `store` never writes through the published path: it fills a
// tempfile only that call can name and renames, pointing the directory entry at a new inode
// and leaving this mapping's inode whole. Writers under a different engine config target a
// different filename through the fingerprint suffix.
```

crates/engine/src/wasm/cache.rs
sdbondi: _Posted by Claude._

Confirmed: `memmap2` refuses a zero-length mapping, so an empty file is turned away before the length check and stays on disk, permanently for the standalone `WasmModuleCache::try_load` users.

Fixed in 2914971d6e, narrowed to the case that is unreadable by definition:

```rust
// An empty file cannot be mapped at all, so it never reaches the length check below.
// Other map failures are resource limits, under which the file may still be good.
if file.metadata().is_ok_and(|m| m.len() == 0) {
    let _ignore = fs::remove_file(&path);
}
```

Unlinking on every map failure would delete good artifacts under `ENOMEM` or a descriptor limit, which is a recompile storm exactly when the node is least able to afford one. No test: `File::set_len(0)` plus `try_load` asserts `memmap2`'s refusal to map an empty file more than it asserts anything of ours, and the branch is two lines with no state.

crates/engine/src/wasm/cache.rs
sdbondi: _Posted by Claude._

The race is real — I traced the same three steps — but I'm not taking the fix, and the thread is worth leaving on the record for the eviction PR rather than resolved silently.

Cost is one extra compile of one template, in a window that needs a corrupt or half-written entry *and* a concurrent `store` of that same address. The inode guard it would take is `MetadataExt::ino`, i.e. `cfg(unix)`-only code on the load path of a module whose entire contract is "any unreadable entry is a miss, recompile", to save a recompile that the miss already pays for. The trade reads the wrong way to me at this point in the series.

It changes character in PR 2: once eviction is choosing files to unlink by policy, deleting an entry another thread just published stops being a self-correcting miss and starts being a hole in the size accounting. That's where the inode identity check belongs, alongside the in-flight set that keeps eviction off entries being written.

sdbondi (commented):
rechecked 👍

PR #2599, opened 2026-09-12
feat(engine)!: price template instantiation, publish compile and no-argument calls

sdbondi:
## Problem

Three things a transaction can ask a validator to do, none of them priced. Audit findings H2 (pricing) and H3.

1. **Instantiation.** `invoke_template` builds a fresh `Store` and `Instance` per instruction — memory mapped, data segments copied, tables and imports wired — before the first metered operator runs. Unbilled.
2. **Publish compile.** `PublishTemplate` makes every validator Cranelift-compile up to 1.5 MiB. Measured at **52 ms for 151 KiB** and **142 ms for 530 KiB**. It paid a size-proportional *fee* but consumed no execution *points*, so `max_block_execution_points` never saw it. Worse, publish was permitted in **fee instructions**, where the compile ran before `checkpoint_fee_intent` enforced any payment at all.
3. **Weight floor.** `CallFunction`/`CallMethod` with no args weigh 0 (`calc_args_weight(&[])` = 0), so `max_transaction_weight` bounded a transaction's bytes but not its instruction count — ~40k minimal calls fit the size cap for nothing.

## Instantiation is priced per *data-segment* byte, not per binary byte

This is the part worth reading. Compiled code is laid down once at publish and costs nothing to instantiate. Measured:

| template | binary | data segments | instantiate |
| --- | --- | --- | --- |
| faucet | 151 KiB | 9.6 KiB | 0.0147 ms |
| liquidity_pool | 322 KiB | 15.7 KiB | 0.0153 ms |
| account | 530 KiB | 19.1 KiB | 0.0146 ms |

All three cost the same, 3.5x apart in binary size. Pricing off `code_size()` would have overcharged `account` by **18x** — every instruction of every transaction. So `validate_module_structure` now sums the data section, `LoadedWasmTemplate` carries it, and the module cache header gains a second `u64` to carry it across a cache hit.

Price: `100_000 + 5 per data byte` (built-ins land at ~1.8x measured, the largest data segment a publish admits at ~2.1x). New `examples/instantiation_points_calibrate.rs` derives it.

## Publish compile

Charged before the compile runs, at `140_000_000 + 2100 per binary byte` — the measured fit converted at the calibrated ~8.4M points/ms.

[...]

PR #2610, opened 2026-09-15
test(engine): re-tune fail_partial_paid_fees for current pricing

sdbondi:
`tari_engine::fees::fail_partial_paid_fees` fails on `development`. Bisected: it passes at a6c5a07c9 and fails at 82152315b, so #2599 merged with it broken. `development` HEAD is currently red with failing `test` jobs.

## Cause

The test funds a fixed 2000 and expects the main instructions to cost more than that, so the transaction rejects with `InsufficientFeesPaid` and only the fee intent commits. #2599 lowered what this transaction costs — it now reports a total fee of 1924, under the 2000 funded — so the transaction it expects to reject commits in full and `expect_failure()` panics with "Transaction succeeded but it was expected to fail".

## Fix

Fund 1700. The viable window measured against current pricing is roughly [1500, 1900]:

| funded | outcome |
| --- | --- |
| ≤ 1450 | rejects, but as an execution failure — the native verification allowance runs out part-way through an instruction |
| 1500 – 1900 | rejects as `InsufficientFeesPaid` — what the test asserts |
| ≥ 1950 | commits in full |

1700 sits in the middle, roughly 200 either side.

The lower bound is the part the old comment missed: fund too little and the transaction still fails, but with the wrong `RejectReason`, so the test is not simply "as low as possible". The comment now names all three bounds and says to re-tune against the fee the transaction reports when it succeeds, since engine pricing changes move the window.

## Tests

`cargo nextest run -p tari_engine -E 'test(/fees/)'` — 17/17 pass. Full `tari_engine` suite was 461/462 before this with this as the only failure.

Test-only change; no engine code touched.


sdbondi (commented):
test fix verified locally, pushing through

PR #2611, opened 2026-09-15
test(consensus): wait for the deferring validator's own EOE commit

sdbondi:
## Description

Fixes a flake in `consensus_tests epoch_change::epoch_change_deferred_eoe_resumes_without_scan_completion`.

The test waited on `wait_for_validators_at_epoch(.., excluded = "1", Epoch(2), ..)`, which only checks the three **non**-deferring validators, and then read validator `1`'s store once to assert it had committed the EOE. Validator `1` commits that same block independently, so the read can land before its commit.

Observed in [run 34963705574](https://github.com/tari-project/tari-ootle/actions/runs/34963705574/job/104363104053):

```
11:39:48.545  validators 2/3/4 COMMIT the EOE block (NodeHeight(2), Epoch(1))
11:39:48.547  .. PACEMAKER: reset View updated to Epoch(2)/NodeHeight(0)   <- wait returns
11:39:48.551  validator 1 COMMITs the EOE block                            <- 4ms after the read
```

```
thread '..' panicked at crates/consensus_tests/src/epoch_change.rs:349:5:
the deferring validator must have committed the EOE: it observed the boundary, so it ratifies and votes
```

## Motivation and Context

The assertion is correct — the deferring validator does observe the boundary and vote — it was just read too early. Polling its own chain, the way `wait_for_leaf_at_epoch` already does for the Epoch(2) leaf, removes the race without weakening what the test checks. The `leaf_at_2.is_none()` assertion below it is unaffected: the deferral is a stable state that holds until the oracle cap is cleared.

## How Has This Been Tested?

`cargo nextest run -p consensus_tests -E 'test(epoch_change)'` — 8/8 pass; the previously flaky test run 5 further times in isolation, all green.

Check out the parent commit and add a `tokio::time::sleep` (or a breakpoint) between the `wait_for_validators_at_epoch` call and the `committed_eoe` read inverted — i.e. read immediately under load — to reproduce. Easier: run the test under `--test-threads` pressure on a busy machine; the ordering that fails is the one in the log above.

## Breaking Changes

- [x] None
- [ ] Other (please specify)


sdbondi:
Flaky test fix, checked locally

PR #2612, opened 2026-09-15
fix(engine)!: report underfunded compute as a fee shortfall

sdbondi:
Noticed while re-tuning `fail_partial_paid_fees` in #2610: the same underpayment gets two different reject reasons depending on *where* the engine notices it.

## Problem

| how it is noticed | error | reject reason today |
| --- | --- | --- |
| transaction runs to the end, costs more than paid | `RuntimeError::InsufficientFeesPaid` | `InsufficientFeesPaid` |
| WASM metering exhausts the fee-funded allowance | `WasmExecutionError::InsufficientFeesForCompute` | `ExecutionFailure` |
| a native verification exhausts it | `RuntimeError::InsufficientFeesForNativeExecution` | `ExecutionFailure` |

The bottom two fall through the catch-all arms in `RuntimeError::to_reject_reason` and `TransactionError::to_reject_reason`. Both of their own messages already say to raise the fee — "Increase the transaction fee." — while the reason handed to the caller says the code failed.

That distinction is not something a payer can act on. `ootle_sdk_core` turns the reason into a stable code (`INSUFFICIENT_FEES_PAID` vs `EXECUTION_FAILURE`), so a wallet cannot reliably tell "resubmit with a bigger fee" from "this transaction will never work" in exactly the case where a bigger fee is the fix.

## Fix

Map both to `InsufficientFeesPaid`.

Deliberately unchanged, because a larger fee does **not** raise either ceiling:

- `FeeIntentComputeExceeded` — the fee intent's flat compute credit, which "does not rise with the fee paid"
- `MaxNativeExecutionPointsExceeded` — the per-transaction hard cap

Both keep rejecting as `ExecutionFailure`, and `assert_fee_intent_credit_exceeded` still pins that.

## Breaking

`RejectReason` converts to `AbortReason`, which rides in `Decision::Abort(..)`. An underfunded transaction now decides `Abort(InsufficientFeesPaid)` where it decided `Abort(ExecutionFailure)`.

`Decision::is_same_outcome` ignores the reason, but the prepare-vote path compares the whole `Decision` (`on_ready_to_vote_on_local_block.rs:607`) and abstains on a mismatch, so a split-version network would stall on these transactions. Needs a coordinated upgrade. No wire format change — the variants already exist and keep their CBOR indices.

## Tests

New `fees::underfunded_compute_rejects_as_a_fee_shortfall` covers both paths — funding 1000 stops inside WASM, 1300 reaches a native stealth verification it cannot fund. Verified to fail without the mapping change.

`compute_fee_budget`'s `assert_insufficient_fees` matched `ExecutionFailure` carrying the substring `"Insufficient fees"`. Its name says what it means to assert, so it now checks the variant instead of a substring — a tightening, not a re-tune.

`tari_engine` 461/461, `ootle_sdk_core` 283/283.


sdbondi (commented):
Makes sense

PR #2614, opened 2026-09-15
refactor(indexer): share one WasmModuleCache per directory

sdbondi:
## Description

The indexer opens its `wasm_cache` directory twice in one process:

- `bootstrap.rs` → `WasmModuleCache::open` for the `TemplateManager`
- `DryRunTransactionProcessor::new` → `build_dry_run_template_provider` → `DiskCachedWasmTemplateProvider::open`

`WasmModuleCache` is currently just a `PathBuf`, so the duplicate is harmless today. It stops being harmless as soon as the cache carries state about the directory's contents — the size index the bounded-cache work needs — because each instance would then account for only the files it wrote itself and evict against a total it under-counts.

This lands the sharing ahead of that change so the eviction diff reads clean.

- `build_dry_run_template_provider` takes a `WasmModuleCache` instead of `wasm_cache_dir: impl Into<PathBuf>` and uses `DiskCachedWasmTemplateProvider::new`. Nothing in it opens anything any more, so the `io::Result` is gone.
- `DryRunTransactionProcessor::new` takes `wasm_cache: WasmModuleCache` instead of `wasm_cache_dir: PathBuf` and forwards it verbatim. It still returns `io::Result` for `Handle::try_current`.
- `tari_indexer/src/bootstrap.rs` hoists the `WasmModuleCache::open` out of the `spawn_blocking` closure, clones it in for the `TemplateManager` and passes the original to `DryRunTransactionProcessor::new`.
- Corrects the justification on `WasmModuleCache::store`'s per-call tempfile, which named the indexer's double open as the source of concurrent stores. The tempfile is still required — one instance is shared across threads and other processes may point at the same directory — but the stated reason is what this PR removes.

The validator node is unaffected: it has a single instance already (`bootstrap.rs:327`), and `DiskCachedWasmTemplateProvider::open` stays for it.

## Motivation and Context

Pure refactor, no behaviour change. The only call removed is the redundant second `create_dir_all`. `WasmModuleCache` stays `Clone`; when the index moves behind an `Arc<Mutex<_>>`, clones will share one index, one byte total and one evictor, with no keying on directory path anywhere.

`WasmModuleCache::open` now runs on the async runtime thread rather than inside `spawn_blocking` — one `create_dir_all` on the startup path.

## How Has This Been Tested?

`cargo lints clippy --all-targets -p tari_indexer -p tari_validator_node -p tari_engine` — clean.
`cargo nextest run -p tari_indexer` — 129 passed.
`cargo nextest run -p tari_engine` — 463 passed.

Confirm `WasmModuleCache::open` has exactly one call site per process: `grep -rn "WasmModuleCache::open\|DiskCachedWasmTemplateProvider::open" applications/ crates/`. Start an indexer and check the dry-run path still populates and reads `<data_dir>/wasm_cache`.

## Breaking Changes

- [x] None
- [ ] Other


sdbondi (commented):
_Posted by Claude._

Reviewed at `32e5e40`. Nothing blocking — the refactor is mechanically correct and the description's claims hold up.

**Verified**

- `WasmModuleCache::open` / `DiskCachedWasmTemplateProvider::open` now have exactly one call site per process outside tests: `applications/tari_indexer/src/bootstrap.rs:336` and `applications/tari_validator_node/src/bootstrap.rs:327`. The VN holds no separate `WasmModuleCache` (`grep WasmModuleCache` returns nothing under `tari_validator_node`), so it was already single-instance as stated.
- `WasmModuleCache` is `#[derive(Clone)]` over a bare `PathBuf` (`crates/engine/src/wasm/cache.rs:117`), so the clone into `spawn_blocking` and the move into `DryRunTransactionProcessor::new` are behaviour-identical to the two `open`s they replace, minus one `create_dir_all`.
- Signature churn is complete: `PathBuf`/`io` imports dropped from both `dry_run` files with no leftover uses, and `DryRunTransactionProcessor::new`'s `io::Result` is still earned by `Handle::try_current` (`dry_run/processor.rs:73`).
- The intra-doc link `[`crate::template_manager::TemplateManager`]` (`dry_run/template_provider.rs:29`) resolves — `template_manager/mod.rs` re-exports `manager::*`. `docs build` is green.
- The rewritten `store` comment (`crates/engine/src/wasm/cache.rs:273`) is accurate on its own terms: with one shared instance, two `store`s of one address can still race across the template-manager thread and a dry-run execution thread, so the per-call tempfile is still required. Removing the now-false indexer-double-open justification was the right call.

**Non-blocking**

1. `crates/engine/src/wasm/cache.rs:125` — `open` is the footgun this PR exists to defuse, but nothing on it says so. Once the bounded-cache work puts a size index behind the handle, a second `open` on a directory that already has a live instance is a silent correctness bug (two indices, each under-counting the total), and `DiskCachedWasmTemplateProvider::open` (line 365) constructs one implicitly, so the mistake is one line away in any new service. A sentence on `open` stating the invariant — at most one instance per directory per process, clone to share — costs nothing now and is the thing a future caller needs to read. Same for `DiskCachedWasmTemplateProvider::open`, whose remaining VN call site is only correct because the VN has no second consumer.

2. `crates/engine/src/wasm/cache.rs:20-21` — pre-existing, but adjacent to what you corrected: the module doc offers "the wallet daemon's template monitor" as the example non-`TemplateProvider` consumer of `WasmModuleCache`. `tari_walletd` has a `TemplateMonitor` (`services/mod.rs:53`) but it never touches `WasmModuleCache`. The real example is the indexer's `TemplateManager` (`template_manager/manager.rs:129`), which is also the one that makes this PR's sharing necessary. Worth fixing while you're in the file.

**Nit**

3. `applications/tari_indexer/src/bootstrap.rs:335` — "the template manager and the dry-run provider" enumerates today's two consumers and will quietly go stale on the third. The invariant is the first clause; the list adds nothing a reader can't get from the ten lines below it.

Only startup-path behaviour change is `create_dir_all` moving from `spawn_blocking` onto the runtime thread, which the description already calls out — one mkdir during bootstrap, not worth the closure.

sdbondi (commented):
:+1:

sdbondi:
_Posted by Claude._

Addressed in `adc9ee5`.

1. **Documented the rule on both `open`s.** `WasmModuleCache::open` now states that a directory takes one instance per process, cloned to its consumers; `DiskCachedWasmTemplateProvider::open` says it opens a cache for that provider's exclusive use and points a caller with a second consumer at `WasmModuleCache::open` + `::new`. Phrased as the usage rule and its purpose rather than as a correctness invariant — with the handle still a bare `PathBuf`, two instances behave identically today, and a comment claiming otherwise would be describing the eviction PR rather than this file.

2. **Fixed.** Confirmed: `grep -rn "WasmModuleCache\|wasm_cache" applications/tari_walletd/src/` returns nothing, so the wallet daemon's `TemplateMonitor` was never a consumer. The module doc now names the indexer's `TemplateManager`.

3. **Fixed** — the enumeration is gone, the comment keeps only the invariant.

PR #2615, opened 2026-09-16
ci: fix and tidy the nightly binary build

sdbondi:
Nightly binary build fixes. No trigger changes: merges to `development` never ran this workflow, and the cron stays because it is the only cross-platform build signal.

### Dead configuration

- **`S3_DEST_OVERRIDE=daily/`** — the "Scheduled Destination Folder Override" step was the only writer and nothing read the variable. The S3 sync lives in `create-release`, which is gated on `refs/tags/v` and so never runs on a schedule.
- **`customTag` dispatch input** — required, defaulted to `development-tag`, referenced nowhere.
- **`cancel-in-progress`** — `!startsWith(ref, 'refs/tags/v') || ref != 'refs/heads/development' || ...` is true for every ref, including tags, so it never protected release builds. The workflow has no `pull_request` trigger either, so the group key's `pull_request.head.label || head_ref` fallbacks were dead. Same shape as #2613.

### Nightly schedule

`05 00 * * *` → `37 20 * * *`. GitHub defers scheduled runs queued at popular slots: recent nightlies were created 04:25–04:46 UTC against a 00:05 UTC cron, and the lag has grown from ~1.7 h in August to ~4.5 h now. An off-peak, off-the-hour slot lands the run overnight even with a multi-hour deferral.

### Intel macOS dropped

`macos-x86_64` is removed from the matrix, and with it the `macOS-universal-assemble` job — a universal archive with one architecture in it is just the arm64 archive. `macos-arm64` takes over as the macOS leg and loses `best_effort`, so a macOS regression now fails the run instead of passing quietly.

Releases lose `tari_ootle-*-macos-x86_64.zip` and `tari_ootle-*-macos-universal.zip`. `ffi_libs.yml` still builds `ootle_sdk_ffi_c` for `x86_64-apple-darwin` and is untouched here — that leg is green and has its own consumers, so say if it should go too.

### pnpm

`packageManager` moves from `pnpm@11.6.0` to `pnpm@12.4.2`. The runners ship pnpm 11.25.0 and delegate to the pin, which needs an `@pnpm/exe.<platform>` build for it — npm has no 11.x release of those packages at all, versions jump from `11.0.0-rc.3` to `12.x`. Linux and darwin-arm64 got away with it; the Intel macOS runner, whose preinstalled pnpm is the standalone native build, failed with `ERR_PNPM_PNPM_ENGINE_IDENTITY_UNVERIFIABLE`. The 12.x lockfile records `packageManagerDependencies` and every `@pnpm/exe.*` platform entry, so the delegation is verifiable everywhere.

Verified locally with pnpm 12.4.2: the lockfile stays at `lockfileVersion: '9.0'`, and `pnpm install --frozen-lockfile` leaves the tree clean (no `allowBuilds:` scaffold appended, which is what ci.yml asserts).

### Still broken after this PR

- **windows-x64**: `wasmer-compiler-cranelift` 7.4 is `compile_error!` on Windows (since #2553), and `crates/engine` enables `wasmer` with `features = ["cranelift"]` unconditionally, so no Windows binary compiles. Fix in progress: feature-gate the wasmer-backed parts of `tari_engine` so the wallet daemon builds without them. The validator node and indexer stay Linux/macOS only.
- **windows-x64 signing**, masked by the above: Azure Trusted Signing fails with `AADSTS7000222` — the `AZURE_CLIENT_SECRET` app secret has expired and needs replacing in the Azure portal.
- **windows-arm64** (`best_effort`): rustc links the `tari_crypto` cdylib with `C:\Program Files\Git\usr\bin\link.exe` instead of MSVC's linker.
- **linux-riscv64** (`best_effort`): `riscv64-linux-gnu-gcc` fails linking `tari_crypto`.

Worth noting separately: `builds` carries `continue-on-error: ${{ startsWith(github.ref, 'refs/tags/') || matrix.builds.best_effort }}`, so a failing leg on a release tag is reported as success. That is why v0.40.2 shipped with no `tari_ootle-*-windows-x64.zip` and nobody noticed.


sdbondi (commented):
Reviewed, macos x86 is no longer supported. Some other cleanup done. Next is fixing the windows wallet build

PR #2616, opened 2026-09-16
fix(engine)!: reject transient values in component state

sdbondi:
Closes the one real finding from #2608's "Out of scope" section. The second finding listed there turned out not to be a bug — see below.

## Problem

A bucket, proof or address allocation is a capability the transaction holds, named by a counter that restarts at zero every transaction. Nothing stopped a template writing one into its own state, so the id reached the ledger, where it can only ever alias an unrelated object of a later transaction.

The existing checks do not cover it:

- `pop_frame`'s dangling-bucket and dangling-proof checks read `buckets_owed` / `proofs_owed`, which filter out `inherited_*`. A frame does not owe what its caller lent it, so a `Proof` or `Bucket` taken as an **argument** and stored in state passes both.
- An address allocation is not accounted for at pop at all.
- `validate_finalized` rejects a bucket, proof or allocation left **live** at the end of the transaction. What it cannot see is an id whose object is gone or empty by then.

Measured against unfixed code, only two shapes escape everything else:

| shape | without this fix |
| --- | --- |
| proof stored, workspace copy dropped | **reaches the ledger** |
| bucket emptied into another, emptied one stored | **reaches the ledger** |
| bucket stored and left unconsumed | already rejected — `DanglingBuckets` |
| component / resource allocation stored, unused | already rejected — `DanglingAddressAllocations` |

A proof the frame creates itself was also already caught, by the dangling-proof check at pop. The argument case is the gap, and it is the one that matters — that is how a proof gets into a frame in the first place.

## Why it is more than hygiene

`invoke_resource_access_hook` hands the acting component's state to the hook inside an `AuthHookCaller` (`component_state: Option<tari_bor::Value>`, raw CBOR). That is a call **argument**, and argument indexing treats a tagged `ProofId` as a proof the callee was lent — `include_refs_in_scope` puts it in the frame's `proof_scope` *and* its `auth_scope`. So a stored proof lends authority to the hook: code the acting component never chose, bound by whoever created the resource, reachable by anyone who can get that resource deposited.

The `OwnComponent` sandbox bounds the damage to what the proof accessors report, and those report already-public data — `Proof::amount()` reads the revealed amount, which is zero for the confidential portion. So this is not an exposure anyone can exploit today, and no template in this repo stores a `Proof`. But the id had no business being in state, and the check is two lines.

## Fix

Reject `bucket_ids`, `proof_ids`, `component_address_allocations` and `resource_address_allocations` in `validate_component_state`, alongside the existing vault checks. Inspects `next_state` rather than the diff, so a component that already has one stored cannot be written to either.

New `RuntimeError::TransientValueInComponentState { kind, id }`.

## Breaking

A component whose state holds one of these ids can no longer be written to. No template in this repo stores one, and on testnet the cost of bricking any that do is accepted.

## Tests

`shenanigans::it_rejects_transient_values_in_component_state` covers all four branches. The proof and the emptied bucket are the two cases with no other net; the plain bucket and the two allocations are there for branch coverage and the test says so. Verified against unfixed code: the proof and emptied-bucket cases *succeed* and the id reaches the ledger.

`tari_engine` 464/464, `tari_template_builtin` 15/15.

## Known follow-up: the same id class in non-fungible data

`impl.rs:1878` — `set_mutable_data` writes caller-supplied CBOR into an NFT substate with no `IndexedWellKnownTypes` pass, and the mint path is the same. A `ProofId` or `BucketId` in `data` / `mutable_data` reaches the ledger and aliases exactly as described above. It is not the auth-hook vector — a hook receives component state, not NFT data — so it does not block this PR, but the class is not closed by it. Deliberately left for a follow-up.

## Not fixed: the second finding, which was not a bug

#2608's description also listed `AuthorizationScope::contains_badge_of_resource` matching any caller component for `rule!(resource(CALLER_COMPONENT_RESOURCE_ADDRESS))`, suggesting rule validation should reject it.

That was wrong, and no change is warranted. `rule!(any_caller_component)` **desugars to exactly** `RuleRequirement::Resource(CALLER_COMPONENT_RESOURCE_ADDRESS)` (`access_rules.rs:863-868`); `any_caller_template` likewise. Matching any caller component is the intended implementation of that rule, documented at `access_rules.rs:724-726` and pinned by a test at `:1172`. Rejecting the requirement would have broken both macros.


sdbondi (commented):
_Posted by Claude._

Reviewed at `be3e230`. The fix itself is sound and I could not find a hole in it:

- `validate_component_state` is the only gate on persisted component state — `tracker.rs:298` for creation and `working_state.rs:356-370` for mutation are the two sites that index the state, and every write path (`ComponentAction::SetState` at `impl.rs:1377`, `SetAccessRules` at `impl.rs:1429`) goes through `modify_component_with`. `update_component_template` (`impl.rs:3886`) touches only the header. So there is no third door.
- Inspecting `next_state` rather than the diff is the right call, and there is no depth escape: `walk_all_depth` (`crates/tari_bor/src/walker.rs:29`) *errors* past `MAX_VISITOR_DEPTH` rather than silently stopping, so a buried id cannot slip past the index.
- The `AuthHookCaller` argument story checks out — `impl.rs:456` folds the caller component's state into the hook argument, and `include_refs_in_scope` (`scope.rs:342-345`) does put an arg `ProofId` into both `proof_scope` and `inherited_proofs`.
- I grepped every template in `crates/template_builtin/templates` and `crates/engine/tests/templates` for `Bucket`/`Proof`/`*AddressAllocation` in a state struct: every hit is a function parameter, none is a persisted field. Nothing in-tree breaks.
- The "not a bug" section is correct: `rule!(any_caller_component)` does desugar to `RuleRequirement::Resource(CALLER_COMPONENT_RESOURCE_ADDRESS)`.

Nothing blocking. Three things below.

## Non-blocking

### 1. The description misstates what `validate_finalized` already catches, and two of the three test cases were already rejected before this change

`crates/engine/src/runtime/working_state.rs:577-598` — `validate_finalized` checks dangling **buckets**, dangling **proofs** *and* dangling **address allocations**, not just "orphans and dangling vault locks" as the PR body says. That matters for two of the three cases:

- `keep_bucket_in_state` (`shenanigans.rs:790`): the minted bucket is never consumed, so it is still non-empty in `self.buckets` at finalize → `DanglingBuckets` would have rejected this transaction before the fix.
- `keep_allocation_in_state` (`shenanigans.rs:805`): the allocation is never used, so it is still in `self.address_allocations` at finalize → `DanglingAddressAllocations` would have rejected it before the fix.

Only the proof case genuinely escaped, and only because `drop_all_proofs_in_workspace()` empties `self.proofs` while the id stays in state. The doc comment at `shenanigans.rs:762-766` claiming the bucket is in "the case nothing else covers" is not right for the bucket.

The tests still pass and still guard the new code (they assert on the new message, which fires first), but they exercise a variant that was already unreachable. The variant the new check is actually load-bearing for is **store the id and consume the object in the same transaction**:

```rust
pub fn keep_bucket_after_deposit(&mut self, bucket: Bucket) {
    self.bucket = Some(bucket.clone());   // `Bucket` is `Clone` (models/bucket.rs:68)
    self.vault.as_mut().unwrap().deposit(bucket);
}
```

Buckets are emptied but never removed from `self.buckets` (`working_state.rs:578` filters on `!bucket.is_empty()`), so post-deposit this reaches the ledger with no dangling error. Same shape for allocations, which are also `Clone`: stash a clone, then let `with_address_allocation` consume the original, and `self.address_allocations` is empty at finalize. Worth adding at least the bucket variant — that is the case that would regress silently if the check were ever removed.

### 2. The `resource_address_allocations` branch has no test

`working_state.rs:1387-1392` is untested. The test covers bucket, proof and component allocation only, though the PR body says "covers all three". The code path is identical to the component case, so this is cheap to add via a `ResourceAddressAllocation` field on `Shenanigans`.

### 3. Same id class can still be persisted in non-fungible data

`impl.rs:1878` — `contents.set_mutable_data(arg.data)` writes caller-supplied CBOR straight into the NFT substate with no `IndexedWellKnownTypes` pass, and the same holds for the mint path. A `ProofId` or `BucketId` in `data`/`mutable_data` reaches the ledger and aliases exactly as described in the PR body. It is not the auth-hook vector (the hook only receives component state), so I agree it does not block this PR — but the "an id that can only ever alias an unrelated object of a later transaction" argument applies unchanged, and the PR body's framing ("component state") reads as if the class is closed. A sentence in the description marking NFT data as a known, deliberate follow-up would keep the next reader from assuming it was covered.

## Nit

`working_state.rs:1370-1392` — four near-identical blocks. A small helper taking `(kind, Option<String>)` or a `(&'static str, Option<String>)` array would read better, though at four branches it is marginal either way.

sdbondi:
_Posted by Claude._

All three verified and addressed in `e125703`. Point 1 was right and I had it wrong in both the description and the test comment.

**1. `validate_finalized` coverage — correct, and I confirmed it empirically rather than by reading.** I ran each shape against unfixed code with a throwaway probe. Results:

| shape | without the fix |
| --- | --- |
| proof stored, workspace copy dropped | reaches the ledger |
| bucket emptied into another, emptied one stored | reaches the ledger |
| bucket stored, unconsumed | rejected — `DanglingBuckets` |
| component / resource allocation stored, unused | rejected — `DanglingAddressAllocations` |

So two of the original three cases did have a second net, as you said.

One correction to the suggested repro: `keep_bucket_after_deposit` would not have reached the ledger. `deposit` takes the bucket via `working_state.rs:670-681`, which calls `remove_bucket_from_scope`, so the stored clone fails `check_all_substates_in_scope` on the new state with `BucketNotInScope` — a different rejection, but a rejection. The shape that does escape is emptying the bucket rather than consuming it: `bucket.take(bucket.amount())` leaves the original empty and *in scope*, and `validate_finalized` filters on `!bucket.is_empty()`. That is what `keep_emptied_bucket_in_state` now does, and it is confirmed to succeed without the fix.

The test comment now says which cases carry their own weight and which are branch coverage.

**2. `resource_address_allocations` untested — correct.** Added `keep_resource_allocation_in_state`, so all four branches are covered. Noting for the record that this one is in the already-rejected column above; it is there for branch coverage, not because it escapes.

**3. NFT data — correct, and the description now says so.** Added a "Known follow-up" section naming `impl.rs:1878` and the mint path, and flagging that this PR closes the component-state door only, not the class.

**Nit** — applied. The four blocks are now one `[(kind, Option<String>); 4]` table; `.first().map(..)` allocates nothing on the happy path.

crates/engine/src/runtime/working_state.rs
```diff
@@ -1362,6 +1362,38 @@ impl<TStore: StateReader> WorkingState<TStore> {
...
+                    .map(|a| a.id().to_string()),
+            ),
+            (
+                "resource address allocation",
+                next_state
+                    .resource_address_allocations()
+                    .first()
+                    .map(|a| a.id().to_string()),
+            ),
+        ];
```
sdbondi: This generates and allocates strings for all occurences, but then the loop only uses one. Rather only allocate when needed

Fixed in `d8b424c`.

You are right about the waste, with one scoping note: on the common path all four `.first()` calls return `None`, so `.map(ToString::to_string)` never runs and nothing is allocated. The waste appears only when state carries more than one kind — two present means two strings formatted and one used — and that is already the rejection path. So it was cheap, not free.

Changed anyway, because the explicit form is strictly better and costs nothing: each check now returns where it fires, and `RuntimeError::transient_in_component_state` takes the id as `impl Display`, so exactly one id is ever formatted and only at the point it is reported.

Worth flagging that this reverses the nit on the previous review, which asked for the four blocks to become a `(&'static str, Option<String>)` array — that array is what introduced the eager formatting. The explicit branches satisfy both concerns: no wasted work, and the repetition that motivated the nit now lives in the shared constructor rather than in four copies of the struct literal.

464/464 engine tests, clippy clean.

sdbondi (commented):
_Posted by Claude._

Re-reviewed at `d8b424c`. All three items from my earlier comment are resolved, and CI is fully green — clippy, fmt, all three test shards, the four fuzz targets, docs, ledger app and **integration test** all pass on this head.

## What I verified in the two follow-up commits

**`e125703` — test coverage**

- `keep_emptied_bucket_in_state` (`shenanigans/src/lib.rs:71`) is the shape that mattered: `bucket.take(bucket.amount())` moves the contents into a vault and keeps the emptied bucket in state. `validate_finalized` filters on `!bucket.is_empty()` (`working_state.rs:578`) and buckets are never removed from `self.buckets`, only emptied — so this one really does reach the ledger with nothing else objecting. `Bucket::take` takes `&mut self` and the two-phase borrow on `bucket.take(bucket.amount())` is fine.
- The original unconsumed-bucket case is kept for branch coverage, and the doc comment at `shenanigans.rs:765-768` now says which cases have a second net rather than claiming they all escape.
- `keep_resource_allocation_in_state` closes the untested branch. `CallerContext::allocate_resource_address()` takes no argument (`caller_context.rs:77`) and `ResourceAddressAllocation` is exported from the prelude (`prelude.rs:123`), so the template compiles as written — and CI confirms.

**`d8b424c` — refactor**

Behaviourally identical to `be3e230`. `transient_in_component_state(kind, id)` taking `impl Display` works for all four call sites: `BucketId` and `ProofId` both have a `Display` impl and pick up the blanket `&T` one, and `AddressAllocationId` is a `u32`. Still allocates only on the error path.

**PR body** now carries the NFT `mutable_data` gap as an explicit, deliberate follow-up. That was the ask.

## Two comment nits, neither worth another round

`error.rs:388` — "Takes the id as `impl Display` so that only the id actually being reported is formatted." No version in this PR ever formatted more than one: the array form it replaced used `.first().map(ToString::to_string)`, which is `None` on an empty slice and so allocates nothing either. The comment describes a hazard that was never in the code, which is the shape `CLAUDE.md` asks comments to avoid. The `impl Display` signature is a fine ergonomic choice and needs no justification — dropping the second half of the sentence would leave it accurate.

`shenanigans.rs:766-768` — "The two that escape everything else are the proof ... and the emptied bucket." There is a third: a **consumed** allocation. `use_allocated_address` (`working_state.rs:1276`) does `swap_remove`, so after `with_address_allocation` consumes it `self.address_allocations` is empty at finalize and `DanglingAddressAllocations` has nothing to say — while a clone kept in state (`ComponentAddressAllocation` derives `Clone`) carries the id to the ledger. The check covers it regardless; only the count in the comment is off.

## Decision

The fix is correct and the coverage now matches what it claims. `gh` will not let me approve a PR authored by the same account, so this stays a comment — it needs a second maintainer's approve to land.

sdbondi (commented):
Reviewed

PR #2617, opened 2026-09-16
refactor(engine)!: fix windows wallet builds

sdbondi:
Restores a Windows wallet daemon build. `wasmer-compiler-cranelift` 7.4 is a `compile_error!` on Windows, and `crates/engine` enables `wasmer`'s `cranelift` feature, so every binary that links the engine fails to compile there. The wallet daemon linked it for two things that never compile or run a template: a fee table and a WASM custom-section read. Both move down to `tari_engine_types`.

### The moves

**`FeeTable` + `WasmMeteringRate` → `tari_engine_types::fees`**, beside `FeeRates`. No new dependencies. `LITERAL_BYTE_DIVISOR` moves with them from `tari_ootle_transaction` (which depends on `engine_types`, so it could not be read from there), and `tari_engine::fees` re-exports the table so engine-internal and validator-node imports are unchanged.

This also settles a note left on `FeeRates`: it said the table "cannot move here, because its `fee_estimate_allowance` reads `LITERAL_BYTE_DIVISOR` from `tari_ootle_transaction`, a crate downstream of this one". It can now, so the doc says what the two types are for instead.

**`extract_template_def` → `tari_engine_types::static_template_def`**, behind a new `extract-template-def` feature. It parses WASM with `wasmparser` — the only dependency this adds anywhere — pinned to the 0.257.1 that wasmer 7.4 itself uses, so a workspace that both parses and executes builds one copy. The feature is optional because ~35 crates depend on `engine_types`, including `crates/ootle_wasm/core` and the client crates, and none of them read a template binary.

### What is left gated

`tari_ootle_app_utilities` keeps `transaction_executor` and `claim_burn_proof_verifier`, which do execute, behind a default `execution` feature — two `#[cfg]`s, in a crate that already gates four modules that way. `tari_engine` is now an optional dependency of it, pulled in only by that feature. The validator node and indexer ask for it; the wallet daemon and everything else (swarm daemon, consensus_tests, integration_tests) do not. `crates/engine` itself gains no cfgs and keeps linking wasmer unconditionally.

Three added `#[cfg(feature = ...)]` attributes in total.

### CI

`windows-x64` and `windows-arm64` build `tari_ootle_walletd` only; the validator node and indexer stay Linux/macOS. The archive step already copies whichever of `TS_FILES` exist, so a Windows archive simply contains the wallet daemon.

### Verified

- `cargo check` on `tari_engine_types` (plain, `--all-features --all-targets`), `tari_ootle_transaction`, `tari_engine`, `tari_ootle_walletd`, `tari_validator_node`, `tari_indexer`, and `tari_ootle_app_utilities` both with and without `execution`.
- `cargo nextest run -p tari_engine_types --all-features`: 95 passed, including the six moved fee-table and extraction tests. The extraction tests read the builtin templates, so `engine_types` takes a path-only dev-dependency on `tari_template_builtin`, the same cycle-breaking pattern `tari_engine` already uses.
- `cargo lints clippy --all-targets --all-features` clean across the touched crates; `cargo machete` finds no newly unused dependencies.
- `cargo tree --target x86_64-pc-windows-msvc -p tari_ootle_walletd -i wasmer` finds nothing; the validator node still resolves it.
- `crate_versioning.py impact tari_engine_types --breaking`: no bumps needed, the pending 0.41.0 covers it.

A real Windows compile needs CI — there is no MSVC or mingw toolchain here. Pushing this branch as `build-bins-*` would exercise the full matrix before merge.

### Not fixed here

Windows code signing still fails with `AADSTS7000222` — the Azure Trusted Signing client secret has expired. That predates the cranelift break and needs a new secret in the Azure portal, not a code change.


sdbondi (commented):
👍

PR #2618, opened 2026-09-16
feat(transaction)!: batch-verify signatures and raise the cap to the stealth input ceiling

sdbondi:
Raises the authorization signature cap from 16 to the stealth input ceiling (1024), and makes a cap that high affordable by verifying a transaction's whole signature set as one batch.

## Why

A key-path stealth spend authorizes each input with its own one-time key, whose badge must be in the transaction's authorization scope (`verify_input_authorizations`), so spending _n_ stealth inputs takes _n_ distinct signatures. At a cap of 16 that made the signature cap — not `STEALTH_LIMITS`, which admits 1024 inputs — the binding limit on a multi-input spend, expressed in the wrong units and in a crate a wallet author reading the stealth limits would never consult.

The user-visible effects were a wallet whose balance is split across more than 17 stealth UTXOs being unable to spend it in one transaction, and a coinjoin capped at 17 participants.

## Batch verification

Every signature in a set is valid exactly when `sᵢ·G - Rᵢ - eᵢ·Pᵢ` is the identity. Weighting those _n_ equations by scalars `zᵢ` and summing gives one equation over the whole set:

```
Σ zᵢ·Rᵢ + Σ (zᵢ·eᵢ)·Pᵢ - (Σ zᵢ·sᵢ)·G == 0
```

so one multiscalar multiplication over `2n + 1` terms stands in for _n_ double-base multiplications. Only the scalar arithmetic stays linear in _n_; the point arithmetic is shared, so the cost per signature falls as the set grows.

| n | individual | decode | batch | per sig | speedup |
|---|---|---|---|---|---|
| 1 | 43.9µs | 8.1µs | 41.4µs | 41.4µs | 1.06x |
| 2 | 84.4µs | 16.0µs | 57.9µs | 28.9µs | 1.46x |
| 4 | 170.0µs | 32.1µs | 89.0µs | 22.2µs | 1.91x |
| 16 | 678.4µs | 128.7µs | 280.2µs | 17.5µs | 2.42x |
| 256 | 10.82ms | 2.05ms | 3.85ms | 15.0µs | 2.81x |
| 1024 | 43.35ms | 8.20ms | 14.39ms | 14.1µs | 3.01x |

Criterion medians, 10s windows, pinned to one core. Absolutes are clock-state dependent on this box (582–5086 MHz, `powersave`); the ratios are measured within a single run.

**Decoding is the floor, and it is why the ratio tops out near 3x.** A signature arrives as bytes, so every term costs two Ristretto decompressions before any equation can run — a flat 8.0µs per signature at every batch size, 57% of the batched cost at a thousand, and unamortisable by construction. The `decode` column measures exactly that.

This is also the whole of the gap against [tari-crypto#249](https://github.com/tari-project/tari-crypto/pull/249), which upstreams this verifier and measures the same equation at 5.84x: that benchmark verifies signatures whose points are already decompressed, and its `verify_batch` takes a decoded `&RistrettoPublicKey`. Net of decoding the two agree — 6.05µs against 5.75µs per signature batched, 34.3µs against 33.6µs individually — so the arithmetic is identical and the difference is entirely where the points come from. Further headroom is in decompressing once across the ingress and block-validation paths, not in the equation.

Three decisions worth review attention:
Three decisions worth review attention:

**Weights are hash-derived, not sampled.** A validity predicate a committee votes on must not depend on local randomness, so every node reaches the same verdict on the same bytes. Soundness then rests on the same Fiat-Shamir argument as the challenge scalars: the weights are fixed by a hash of everything they weigh, so a set whose errors cancel under its own weights takes a 2^127 search to find. It is also the derivation a half-aggregated signature would carry, where the aggregator is untrusted and the weights must be public, so this verification path is reusable as-is if the format ever carries one.

**The multiplication is variable-time**, which is where most of the gain comes from — the same equation through the constant-time `RistrettoPublicKey::batch_mul` is ~1.5x behind it. Every term is public (keys, nonces, signature scalars, and messages that are digests of a body gossiped in full), so it has no secret for its timing to leak; a batch over a *confidential* message would not be safe this way. That is what the direct `curve25519-dalek` dependency buys, pinned to the major version `tari_crypto` builds against — and it goes away with tari-crypto#249, which moves both this and the challenge derivation into the library.

**The generator is folded into the multiplication** rather than multiplied through dalek's precomputed basepoint table. The table makes `k·G` fast in isolation but is still a whole extra scalar multiplication worth a flat ~13µs, so folding pays while that constant is large next to the multiplication it joins: upstream measured 16% better at one signature, ~7% at two to four, a wash from sixteen to 256, and 8% worse at a thousand. A transaction carries one seal and usually one authorization, so the small end is the case to optimise.

**The seal is in the same batch.** The seal and the authorizations sign different messages, which the equation is indifferent to — each term derives its challenge from its own. The two commitments cross without being circular: the seal commits to the authorization signatures, the authorizations commit to the seal signer's key, so both digests are derivable before anything is verified. Worth ~30µs per transaction, which is most of the cost of a one-authorization transfer.

## Identity key fix

`SchnorrSignature::verify_challenge_scalar` refuses a zero public key outright. The batch equation did not, and a batch that holds is the whole verdict, so the check was bypassed: under `P = 0` the equation reduces to `s·G == R`, which anyone satisfies by choosing `s` and setting `R = s·G`. A signature under the identity key was accepted where the individual verifier rejects it, as an authorization or as a seal. `verify_batch` now refuses a zero-byte key — exact rather than a shortcut, since Ristretto's canonical encoding is unique, so the identity is 32 zero bytes and nothing else. The nonce stays unchecked because the stock verifier leaves it unchecked; the goal is to accept exactly what it accepts.

This was caught in review of the first commit, not by the equivalence tests, which had not planted an identity key.

## What bounds the aggregate

Weight, not the count — which is what the weight budget is for. A spend at the cap pays weight three times per stealth UTXO (the declared input, the signature authorizing it, the statement input) and comes to **22,628 against the 24,000 a leader packs**, so weight is what stops one monopolising a block. `the_block_budget_admits_a_transaction_at_the_signature_cap` builds that transaction and measures it, since every term comes from a factor that can move on its own.

Raising the cap also inverts the attacker's incentive at ingress: because a large batch verifies more cheaply per signature than a small one, the cheapest CPU per gossiped byte is now a transaction with _few_ signatures, so the worst-case cost per byte is lower at the cap than below it.

`SIGNER_FACTOR` is deliberately unchanged. A signature's 5 weight already over-states its verification cost by more than an order of magnitude (the weight unit is ~154µs of execution at the block calibration, against 15.6µs measured), and raising it to the ~32 its 96 gossiped bytes would carry at the blob divisor would put a spend at the cap over the block budget — buying nothing but a smaller achievable coinjoin.

There is also no per-signature execution-points charge: the work happens at ingress rather than during execution, weight already accounts for it, and 1024 signatures would come to ~120M points against the 32M `FREE_COMPUTE_GRACE_POINTS`, so a large coinjoin could not source its own fee. Charging the payer in _fees_ rather than in block space would be a new fee dimension, not a knob.

## Rejection

A rejection is a verdict on the set and does not name the signature at fault. Locating it means a pass of per-term equations that only a rejected transaction would ever need — and a rejected transaction pays no fee, so that was work an attacker could ask for and nobody covers. Without it an invalid set costs exactly what a valid one costs: **2.6x cheaper to refuse** than the per-signature path it replaces.

## Tests

- Batch and individual verification agree on valid sets and on a planted foreign signature at every index.
- A pair whose error terms cancel unweighted is rejected — asserting first that it *does* balance unweighted, so the case cannot silently stop exercising the weights.
- Signatures verified under each other's keys are rejected; undecodable keys and scalars are rejected rather than panicking.
- The identity key is rejected, asserting first that the stock verifier refuses it.
- Weights are reproducible, never zero, and redrawn by any change to a signature, a key or a message — including a batch whose terms carry different messages, where each term must be bound to its own.
- The seal verifies in the same batch as its authorizations at 0, 1 and 4 of them, and errors cancelling across the seal boundary are still refused.

155 tests across `tari_ootle_transaction`, `tari_ootle_transaction_validation` and `tari_consensus`. `cargo lints clippy --all-targets` clean.

## Breaking change

Transactions carrying 17 to 1024 authorization signatures are now admitted where they were previously rejected at ingress and in block validation. Every node must run this before any node submits one, so it should land with the testnet reset.

Wallets cannot use the raised cap yet: selection still caps at `STEALTH_LIMITS.max_inputs` and counts UTXOs rather than signer groups. That follows in a stacked PR.


sdbondi (commented):
_Posted by Claude._

Reviewed the full diff at d4e46d2, all files at head, with the batch maths and tari_crypto's verifier traced against the actual source. Nothing blocking. Local: 155 tests across the three crates pass, `cargo lints clippy --all-targets` clean.

**Sound**

- Batch equation and weight derivation. Weights are bound to `len ‖ (m, P, sig)*` with fixed-width fields, so the commitment is injective and any change redraws every weight; the forced top bit makes a zero weight impossible. Deterministic weights are the right call for a consensus predicate.
- `challenge_scalar` reproduces `SchnorrSignature::verify` byte for byte: same `construct_domain_separated_challenge::<_, Blake2b<U64>>`, same wide reduction (`schnorr.rs:190-208`).
- Identity key refused on the bytes before decode. Exact, since Ristretto decompression rejects non-canonical encodings so 32 zero bytes is the only identity. The test asserts the stock verifier refuses it too, so it fails loudly if that ever changes.
- Dropping the per-term walk. Nothing consumed the index beyond a `debug!`, and it made a rejected set cost more than an accepted one with no fee behind it.
- Validator ordering: `TransactionWeightValidator` precedes `SignatureLimitValidator` and `TransactionSignatureValidator` in `bootstrap.rs:555-562`, so weight bounds a transaction before any verification runs.
- Ristretto is prime-order, so the Ed25519 cofactor batch pitfalls do not apply. Single `curve25519-dalek` major means drift fails at compile time on `.point()`.
- `the_block_budget_admits_a_transaction_at_the_signature_cap` passes at 22,628 against 24,000.

**Non-blocking**

- `crates/transaction/benches/signature_verification.rs:11` refers to `MIN_BATCH_SIZE`, which does not exist anywhere in the crate.
- `crates/transaction/benches/signature_verification.rs:79-81` says a failing batch is "walked afterwards". That path was removed in the second commit; the comment now describes behaviour the code does not have.
- Per-transaction cost judgment, worth stating in the PR rather than fixing: the per-byte framing is correct, but a 1024-signature transaction with no stealth inputs weighs 5,120, passes weight validation, and costs ~16ms unpaid, up from ~0.7ms at the old cap. Any layer that limits by transaction count rather than bytes (mempool slots, per-peer message rate, tx-id dedup) is ~23x weaker for it. Nothing in the touched code assumes count-based limiting, so this is a note.
- Headroom on the block-budget test is ~6%. Any bump to `SIGNER_FACTOR`, input weight or `STEALTH_LIMITS` trips it. That is the intent, but whoever touches those should know.

**Nits**

- `crates/transaction/src/v1/signature.rs:416-419` collects a `Vec<BatchTerm>` only to hand it to `verify_batch`. An iterator-taking `verify_batch` would drop the allocation. Trivial at these sizes.

crates/transaction/src/v1/signature.rs
```diff
@@ -389,6 +399,209 @@ impl TransactionSignature {
...
+    authorization_message: [u8; 64],
+) -> bool {
+    let mut terms = Vec::with_capacity(signatures.len() + 1);
+    terms.push(BatchTerm::new(seal.public_key(), seal.signature(), &seal_message));
+    terms.extend(
+        signatures
+            .iter()
+            .map(|sig| BatchTerm::new(sig.public_key(), sig.signature(), &authorization_message)),
+    );
+    verify_batch(&terms)
```
sdbondi, line 449: Can be an iterator without allocating

sdbondi: rejected

sdbondi (commented):
Approved

PR #2619, opened 2026-09-16
feat(engine)!: bound the compiled-artifact caches

sdbondi:
Second in the WASM artifact cache series, after #2609 (header CRC) and #2614 (one cache instance per directory).

## Why

A compiled artifact runs about ten times the size of its WASM source — measured across real templates: faucet 244,520 → 2,519,440 (10.3x), liquidity_pool 322,684 → 3,071,464 (9.5x), account 75,930 → 772,360 (10.2x). Nothing pruned the on-disk cache, so it only grew: 10,000 templates averaging 300 KiB of source projects to ~29 GiB of artifacts, and ~98 GiB at the 1 MiB binary cap.

Eviction is node-local policy with no consensus reach — an evicted artifact costs a recompile of bytes that remain on-chain.

## Disk tier

`WasmModuleCache` is now bounded by total bytes rather than entry count: artifacts span hundreds of KiB to tens of MiB, so a count says little about disk. Over the cap, least-recently-used artifacts are unlinked.

- A running tally of the directory lives on the cache, so `store` answers "over the cap, and what is coldest?" without a `readdir` and a `stat` per file on the execution path. The directory stays the authority on contents; a tally that drifts costs a recompile, never correctness.
- Recency is seeded at open from modification time, the only ordering the filesystem retains across a restart. Hits are not written back — a `utimes` per read buys a better post-restart order at the cost of a write on every hit.
- Opening reaps tempfiles older than 5 minutes, left by a run that died between write and rename. Nothing collected them before. A tempfile a live `store` may still be writing is left alone.
- Opening also reaps artifacts under a superseded `ENGINE_FINGERPRINT`. Without that the cap bounds one generation rather than the directory: a bump follows any wasmer upgrade or metering change, and the build that wrote the previous generation is the one being replaced, so nothing else would ever collect those files. A directory shared with an older binary costs that binary a recompile per template.
- Opening also applies the cap before the first lookup, so lowering it in config takes effect on restart.
- Eviction and the corrupt-file paths check the file's identity (inode, on unix) before unlinking. `try_load` deletes by path while `store` publishes by renaming a new inode over that same path, so a delete decided against the file one call read can land on the file another has just published. The check narrows that window to the gap between the stat and the unlink, on platforms that expose an identity; what is left costs a recompile. This is the thread left open on #2609.
- The last entry is never evicted: an artifact larger than the whole cap would otherwise be evicted by the very `store` that wrote it, and recompiled on every call.

## Memory tier

Builtins move out of the evictable moka cache into a map held for the life of the provider. They deliberately bypass the disk cache (their addresses are hardcoded constants, not content-addressed, so a disk entry would pin a stale module across a builtin recompile) — which means an evicted builtin has no artifact to deserialize and recompiles from source on the caller's thread. `account` is 529 KB of source and every account transaction calls it. They now sit beside `max_cache_size_bytes` rather than inside it, so both memory models — the VN's `memory_budget.rs` and `tari_vn_bench` — carry a line for them.

## Config

`TemplateConfig` gains `max_disk_cache_size_bytes` (default 10 GiB) alongside the existing in-memory bound, which is raised from 200 MiB to 1 GiB in the same sizing pass. That second move is not a consequence of the disk work — the moka cap and how it evicts are unchanged — but the two defaults are read together, so they were set together. Both constants carry the figures behind them: at a 300 KiB average source, 10 GiB is roughly 3,500 artifacts and 1 GiB is roughly 870 modules resident.

The node's derived memory budget rises with it, from ~1.5 GiB to ~2.3 GiB enforced and ~2.3 GiB to ~3.5 GiB required; `validator-node-requirements.mdx` carries the new table. and the tally plus the cap are logged at `info!` on open, as the indexer's retention defaults already are. `applications/tari_indexer/docs/configuration.md` documents the new section. The indexer now carries a `templates` section as the validator node already did, so its dry-run provider and template manager are configured rather than hardcoded to defaults.

## Testing

`cargo nextest run -p tari_engine --features wasm-cache -E 'test(cache)'` — 13 pass, including LRU eviction over the cap, a hit sparing an artifact from the next eviction, an oversized artifact being kept, the tally rebuilding on reopen, selective tempfile reaping, and a republished file surviving a stale unlink. `cargo nextest run -p tari_ootle_template_provider` covers builtins being served under a cache budget that holds nothing.

`cargo lints clippy --all-targets` clean for `tari_engine`, `tari_ootle_template_provider`, `tari_indexer`, `tari_validator_node`.

No fingerprint bump: nothing about the file format changes. No reset implication — this is node-local policy throughout.

## Next in the series

Prewarm on mempool admission and publish commit, which also moves `store`'s inline `fsync` off the execution path and fixes the double-load in `ConcurrentMapSemaphoreGuard::drop`.


sdbondi (commented):
_Posted by Claude._

Read at `8ab90c5`, whole files at the head rather than the hunks. The eviction logic itself I traced through by hand and believe is correct: `record` subtracts the previous size before adding the new one, `take_coldest` cannot loop forever (each iteration removes an entry, and the `len() <= 1` guard terminates), the just-stored entry always carries the highest tick so a `store` never evicts its own write, and an oversized artifact that survives one `store` is correctly dropped by the *next* one rather than pinned forever. `build_index` seeding ticks in mtime order gives the coldest file the lowest tick, which is what `evict_to_fit` wants. The inode check closes the specific hazard #2609 left open, and the tempfile reaper's age gate is the right shape.

No blocking findings. Four things worth addressing, in descending order.

### 1. The cap bounds one fingerprint's files, not the directory — `crates/engine/src/wasm/cache.rs:302`

```rust
// Files under another fingerprint are another build's, and are left for its own cleanup.
let Some(stem) = name.strip_suffix(&suffix) else {
    continue;
};
```

Files that don't end in `_v5.bin` are never tallied and never evicted, so they sit outside `max_disk_cache_size_bytes` permanently. `ENGINE_FINGERPRINT` has moved six times in this file's history (`v1`→…→`v6`→`v5`), so a bump is a routine consequence of a wasmer upgrade or a metering change, not a rare event. Each one silently strands a full generation — up to another 4 GiB at the default — and "its own cleanup" never runs, because the build that wrote those files is the one that just got replaced.

That's directly at odds with what the change says it does: `memory_cache.rs:41` reads "bounding the directory", and the PR body's framing is disk growth. As written the directory grows by one cap per fingerprint bump.

The reaper is already walking the directory, so this is a few lines next to the tempfile branch: a name that parses as `{64 hex}_v{n}.bin` with an `n` that isn't this build's is an orphan and can be unlinked on the spot. The one caveat is a directory genuinely shared with an older binary — but that's the same assumption `open`'s "one instance per process" doc already makes, and a downgrade costs recompiles either way. If you'd rather not reap, the comment and the `memory_cache.rs:41` doc should stop claiming the directory is bounded.

### 2. The disk cap isn't logged, against the convention the same file states — `crates/engine/src/wasm/cache.rs:328`

`build_index`'s tally lands at `debug!`. Meanwhile `applications/tari_indexer/src/bootstrap.rs:390` says, of the transaction-retention defaults:

> Both of these have defaults that decide how much disk this node uses and what it deletes, so they are logged unconditionally: an operator who upgraded without touching their config should be able to see what changed underneath them in their own logs.

A 4 GiB artifact cap that starts unlinking files is exactly that kind of default, and it arrives on upgrade for every existing node. An `info!` on open — directory, artifacts held, bytes, cap — would put it in the same place the operator already looks. The VN's `memory_budget.rs` does the equivalent for memory at `info!`.

### 3. Builtins are now resident memory outside every enforced cap — `crates/template_provider/src/memory_cache.rs:113`

Moving builtins out of the moka cache is the right call for the reason given (no disk tier to fall back to). But it moves ~4 compiled artifacts out of `max_cache_size_bytes` and into memory that nothing bounds, and two models now describe the process slightly wrong:

- `applications/tari_validator_node/src/memory_budget.rs:88` — "Compiled template module cache" is `templates.max_cache_size_bytes()`, and the module header claims "Every term here is a limit the node enforces". The builtins are now a term with no limit.
- `utilities/tari_vn_bench/src/memory.rs:170-176` — the `source` note says "the cache evicts rather than growing over a node's lifetime", which is now false of the builtins specifically.

It's single-digit MB against a 200 MiB line, so nothing needs resizing; it's the model that should stay honest. A `BudgetLine` for the resident builtins, or a clause on the existing `source`, covers it.

### 4. `remove_tracked_file` narrows the race, it doesn't close it — `crates/engine/src/wasm/cache.rs:206`

The doc says "a delete decided against the file one call read **can**not land on the file another call has just published", and the PR body repeats it. `fs::metadata` followed by `fs::remove_file` is still TOCTOU — a rename landing between the two goes through — and on non-unix `identity_of` returns `None`, so the check is skipped entirely and the unlink is unconditional.

The consequence is one recompile, so the code is fine as it stands. It's the absolute phrasing that overstates what the check buys. "Narrows the window to the gap between the stat and the unlink, on platforms that expose an identity" would match the code.

### Nits

- `evict_to_fit` (`cache.rs:346`) holds the index mutex across `fs::metadata` + `unlink` for every victim. Draining the victims under the lock and unlinking after dropping it would keep the lock to memory work; matters only when a cap drop sheds many entries at once.
- `total_bytes()` (`cache.rs:574`) is `pub` with no non-test caller. If (2) lands it gains one; otherwise it's test-only surface on a public type.
- `#[serde(default = "…")]` sits on `max_disk_cache_size_bytes` but not on `max_cache_size_bytes` (`memory_cache.rs:34`). Inert either way — `DefaultConfigLoader` serializes the `Default` in as a source before deserializing, so both fields are always present — but the asymmetry reads as if one is optional and the other required.
- The new `[indexer.templates]` section isn't in `applications/tari_indexer/docs/configuration.md`, which does document the comparable disk knobs (`transaction_retention_epochs` and friends).
- Worth noting the asymmetry in `build_index` is deliberate and right: tempfiles are reaped under *any* fingerprint while artifacts are not, because a tempfile is never a valid artifact for anyone. The comment at line 302 sits close enough to read as if it governed both.

CI is still running at the time of writing; I have not built locally.

sdbondi:
_Posted by Claude._

All four addressed in 9ffec70, plus the nits. Verified each finding against the code first.

**1. Cap bounds one fingerprint, not the directory** — fixed by reaping. `open` now unlinks artifacts whose filename parses as `{template address}_{fingerprint}.bin` with a fingerprint that isn't this build's, in the same walk that reaps tempfiles. `parse_orphan_name` requires the address to parse as 32 hex bytes, so a file this cache didn't write is left alone — covered by `opening_reaps_artifacts_under_a_superseded_fingerprint`, which asserts a `notes.txt` survives.

One correction to the reasoning, not the conclusion: `open`'s "one instance per process" doc is about instances within a process, not about two builds sharing a directory, so it doesn't already carry that assumption. The reaping does now, and the doc on `build_index` states it and its cost — a downgrade recompiles per template on its next run.

**2. Cap not logged** — the tally moved from `build_index` at `debug!` to `open` at `info!`, carrying directory, artifacts, bytes and cap. Wording follows `bootstrap.rs:390`'s stated reason. That also gives `total_bytes()` a non-test caller, and `len()` alongside it.

**3. Builtins outside every cap** — `builtin_resident_bytes()` in `memory_cache.rs` weighs them the same way the cache weighs an entry, and both models gained a "Resident builtin templates" line: `memory_budget.rs` and `tari_vn_bench`'s `memory.rs`, where the neighbouring `source` note now says builtins are held apart. Single-digit MB, as you said — the sizing didn't need it, the model did.

**4. Phrasing** — the `remove_tracked_file` doc and the PR body now say the check narrows the window to the gap between the stat and the unlink, on platforms that expose an identity, and that what is left costs a recompile.

Nits: `evict_to_fit` drains victims under the lock and unlinks after dropping it; `#[serde(default)]` now on both `TemplateConfig` fields; `[indexer.templates]` documented in `applications/tari_indexer/docs/configuration.md`; the line-302 comment split so the tempfile branch carries its own, which also states why the age gate is its only question.

14 cache tests pass under `--features wasm-cache`; `cargo lints clippy --all-targets` clean across `tari_engine`, `tari_ootle_template_provider`, `tari_validator_node`, `tari_indexer`, `tari_vn_bench`.

sdbondi (commented):
_Posted by Claude._

Re-checked each item against `9ffec70` rather than taking the commit message for it. All four are genuinely addressed.

**1. Orphan generations.** `parse_orphan_name` (`cache.rs:202`) is correctly conservative: it only fires when `parse_artifact_name` has already declined, and `TemplateAddress::from_hex` goes through `fixed_bytes_from_hex`, which is length-strict, so the reap needs exactly `{64 hex}_{fingerprint}.bin`. A `{64 hex}_v5.bin` whose stem somehow fails to parse still returns `None` on the fingerprint equality check before `from_hex` is reached, so this build's files can never fall through to the reaper by a parse accident. Ordering is right too — the tempfile branch `continue`s first, so a `.bin.tmp.` name never reaches it. `notes.txt` in the new test pins the "not ours" case.

**2. Logging.** `info!` on open with directory, count, bytes and cap, and the `debug!` in `build_index` correctly removed rather than duplicated. `len()`/`is_empty()` give `total_bytes()` a non-test caller and satisfy `len_without_is_empty`.

**3. Builtin accounting.** `builtin_resident_bytes` uses `t.binary.len() * CODE_SIZE_TO_RESIDENT_BYTES_FACTOR`, and I checked this is the same quantity the moka weigher computes: `WasmModule::compile` passes `code.len()` as `code_size` (`module.rs:99`), and the cache header preserves that across a deserialize, so `LoadedTemplate::code_size()` is the WASM source length for both paths. "Directly comparable" holds literally. Both models get the line, and `memory_budget.rs`'s delta-based tests are unaffected by the extra entry.

**4. Race wording.** `cache.rs:225` now says what the check buys, including the non-unix case.

The `evict_to_fit` refactor is the one that needed the most care and it's right: collecting victims under the lock and unlinking after means a `store` can republish an address between the two, but that is exactly the case `remove_tracked_file`'s identity check catches, and the entry is already out of the index so the tally stays consistent. Worth noting the window it widens is the non-unix one, where there is no identity to check and the unlink is unconditional — a concurrent republish there loses its file. Self-healing (the next `try_load` misses and recompiles) and already covered by the doc you rewrote, so I'd leave it; flagging only so it's on the record.

Two leftovers, both small:

- `memory_budget.rs:243` — `the_constant_caps_are_present` still lists three lines. Its docstring says it exists because "the caps that are constants rather than configuration cannot be varied, so they are checked by the value they contribute instead", and `builtin_resident_bytes()` is now exactly that: a constant cap with a line. Adding `("resident builtins", builtin_resident_bytes())` to the array keeps the test doing the job it claims.
- The title is `feat(engine):`. `WasmModuleCache::open` and `DiskCachedWasmTemplateProvider::open` both gain a parameter, and `tari_engine` 0.41.0 is on crates.io, so that is a breaking change to a published crate's public API — behind the non-default `wasm-cache` feature, but public. #2617 (`refactor(engine)!: fix windows wallet builds`) used `!` for a comparable move with no consensus reach. I checked `scripts/release_check.py` and the history around `workspace.package.version`, and bumps land in release prep rather than per-PR, so no version change is expected here — it's only the marker.

Nothing blocking. CI is still running (clippy, test 1-3, fuzz, integration pending); the approval has to come from another maintainer since GitHub will not take mine on your own PR.

sdbondi:
_Posted by Claude._

Both leftovers done in 7757f00.

- `the_constant_caps_are_present` gains `("resident builtin templates", builtin_resident_bytes())`. Agreed on the reasoning: the test checks constant-derived lines by the value they contribute because configuration cannot vary them, and that is now what the builtins are.
- Title is `feat(engine)!:`. `WasmModuleCache::open` and `DiskCachedWasmTemplateProvider::open` both take a new parameter, and both are `pub` on a crate that is published, so the marker belongs there whether or not the feature is default. Squash-merge takes the PR title, so the branch commits are left as they are.

On the `evict_to_fit` note: agreed, and worth keeping on the record as you put it. The window it widens exists only where `identity_of` returns `None`, and the cost there is a lost cache file that the next `try_load` misses and recompiles. Nothing on the affected targets reaches it — `wasm-cache` is enabled only by `tari_validator_node` and `tari_indexer`, and the Windows entries in `build_binaries.json` build `tari_ootle_walletd` alone.

sdbondi (commented):
_Posted by Claude._

`d09f5d9a` raises the defaults. The arithmetic all checks out: `builtin_resident_bytes()` over the four builtins is 1,154,631 × 4 = 4,618,524 bytes, and the new budget sums to 256 + 128 + 128 + 768 + 1024 + 4.4 + 64 + 20 = 2392.4 MiB → 2.3 GiB enforced, × `HEADROOM_FACTOR` = 3.5 GiB required. Both figures in the requirements table and the "~2.3 GiB" in the recommendation row match what the code will actually compute. No stale references to `200 MiB` / `4 GiB` anywhere, and the vn-bench line picks the new value up from `TemplateConfig::default()` on its own.

Two things.

### The builtins line cannot render as `4.4 MiB` — `validator-node-requirements.mdx:67`

The block is introduced as "the node logs its own version of that table at startup", so it reads as verbatim output. `format_bytes` (`memory_budget.rs:179`) only takes a decimal above the GiB threshold:

```rust
if mib >= 1024.0 {
    format!("{:.1} GiB", mib / 1024.0)
} else {
    format!("{mib:.0} MiB")
}
```

4,618,524 bytes is 4.404 MiB, which is below the threshold, so the node prints `4 MiB`. Every other line in the table is reproducible; this one isn't. Either write `4 MiB` or give `format_bytes` a decimal under the threshold too — the latter is arguably worth it anyway, since a sub-MiB line would currently print `0 MiB`.

### The new numbers carry no stated basis

`default_max_cache_size_bytes` (`memory_cache.rs:51`) has no doc comment at all, and `default_max_disk_cache_size_bytes`'s (`memory_cache.rs:55`) still says only that the default "is sized to keep the working set of an active network resident while bounding the directory" — written when it was 4 GiB, and equally true of any number. The commit message states the move without a reason for landing on 1 GiB and 10 GiB rather than, say, 512 MiB and 20 GiB.

That stands out here specifically because every other constant in this change set argues for itself — `HEADROOM_FACTOR`, `CODE_SIZE_TO_RESIDENT_BYTES_FACTOR`, `STALE_TEMPFILE_AGE`, `DEDUP_BYTES_PER_TRANSACTION` all carry the reasoning that produced them, and the whole point of the disk cap is that an operator can size it against their machine. A default they cannot reason about is one they will either leave alone without knowing what it buys or change without knowing what they are giving up.

You have the material for it already: against the PR's own ~10x expansion and 300 KiB average source, 10 GiB is roughly 3,400 templates' artifacts, and 1 GiB at the weigher's 4x is roughly 890 templates resident. If those are the figures behind the choice, they belong on the constants.

Two smaller observations, neither needing action:

- Raising the *in-memory* default is the one piece of this PR that isn't about bounding the artifact caches — the moka cap predates the change and nothing here alters how it evicts. It's defensible as "the same sizing pass", just worth naming in the PR body so it isn't read as a consequence of the disk work.
- `check_against_available_memory` now warns below 3.5 GiB free rather than 2.3 GiB, and the swarm daemon sets no `[validator_node.templates]` override, so every VN in a local swarm re-runs that check against a machine the other VNs are already sitting on. Startup gets noisier on a dev box; nothing is wrong.

Everything from rounds one and two remains resolved — I re-checked `the_constant_caps_are_present` at `7757f00` and `builtin_resident_bytes` is in scope there through the module-level import.

sdbondi:
_Posted by Claude._

Both fixed in 68126f9.

**Render.** You're right that the block is introduced as the node's own output, so the line has to be what `format_bytes` prints: 4,618,524 bytes is 4.404 MiB, below the 1024 MiB threshold, so `{mib:.0}` gives `4 MiB`. Changed the table rather than the formatter. Giving `format_bytes` a decimal under the threshold would turn every other line into `256.0 MiB` for the sake of one, and the `0 MiB` case it would fix is hypothetical — no line is sub-MiB, and a line that became one would be worth noticing rather than rounding into visibility.

**Basis.** Both constants now carry their figures. 10 GiB against the ~10x expansion and 300 KiB average source is roughly 3,500 artifacts; 1 GiB at the weigher's 4x is roughly 870 modules resident. I added what follows from that rather than only the arithmetic: past the disk cap the cost of being too low is CPU and the cost of being too high is disk the operator may want elsewhere, which is the trade they are actually sizing.

On the in-memory raise not belonging to the disk work — agreed, and now stated in the PR body: the moka cap and its eviction are untouched, the two defaults were simply set in one pass because they get read together.

The swarm startup noise is worth knowing about; leaving it, since the warning is accurate about the machine.

sdbondi (commented):
Approved

sdbondi (commented):
_Posted by Claude._

All 21 checks pass at `68126f9`, and every finding from the three rounds is resolved. Nothing blocking remains — recording that here rather than as an approval, since GitHub will not take one from me on your own PR.

What I verified rather than took on trust, across the four heads:

**Eviction.** `record` subtracts the previous size before adding the new one; `take_coldest` terminates (each pass removes an entry, `len() <= 1` guards the rest); the just-stored entry always holds the highest tick, so a `store` cannot evict its own write; an oversized artifact that survives one `store` is dropped by the next rather than pinned. `build_index` seeds ticks in mtime order, which gives the coldest file the lowest tick — the ordering `evict_to_fit` needs.

**Orphan reaping.** `parse_orphan_name` only runs after `parse_artifact_name` declines, checks fingerprint equality before `from_hex`, and `from_hex` goes through the length-strict `fixed_bytes_from_hex`. This build's files cannot reach the reaper through a parse accident, and the tempfile branch `continue`s ahead of it.

**The lock refactor.** Draining victims under the lock and unlinking after lets a `store` republish an address in between — which is exactly what `remove_tracked_file`'s identity check catches, with the entry already out of the index so the tally stays consistent.

**`builtin_resident_bytes`.** Measures the same quantity as the moka weigher: `WasmModule::compile` passes `code.len()` as `code_size`, and the cache header preserves it across a deserialize, so `code_size()` is the WASM source length on both paths. 1,154,631 × 4 = 4,618,524 bytes.

**The budget.** 256 + 128 + 128 + 768 + 1024 + 4.4 + 64 + 20 = 2392.4 MiB → 2.3 GiB enforced, 3.5 GiB required at `HEADROOM_FACTOR`. Both match the requirements table, and the builtins line now renders as `format_bytes` actually emits it.

**The defaults.** 1 GiB ÷ (300 KiB × 4) = 873.8 and 10 GiB ÷ (300 KiB × 10) = 3495, so "roughly 870" and "roughly 3,500" are right.

Two residual behaviours are documented rather than fixed, which I read as deliberate: the orphan reap will delete a co-resident older build's artifacts, and the non-unix unlink is unconditional because there is no identity to check. Both cost recompiles only. They stop being free if a second binary sharing a `wasm_cache` directory ever becomes a real deployment shape.

I did not build or run tests locally — the clippy and test legs are what stand behind the compile-level claims here.

PR #2620, opened 2026-09-16
fix(wallet): share one input budget between a transfer and its fee

sdbondi:
Follow-up to #2618, which raised the signature cap so a fragmented wallet could spend its balance. This fixes the wallet side, where selection was bounded by the wrong limit.

## The bug

Stealth input selection used `STEALTH_LIMITS.max_inputs` — the cap on a **single statement** (1000). What bounds a wallet is the **per-transaction** total, `max_total_inputs_per_transaction` (1024), and a transfer that cannot source its fee from its own revealed remainder carries a *second* statement whose inputs come out of that same total.

Both selections were capped at the per-statement limit independently, so a fragmented wallet could build a transaction of up to 2000 inputs against a ceiling of 1024. The engine refuses it as `ExceedsStealthTransactionLimit { limit: "inputs" }`, and ingress refuses it before that — the same class of failure as the original report that started #2618: the wallet builds what the network will not take.

## The budget

`MAX_INPUTS_PER_TRANSACTION` is now the per-transaction total, and the two selections share it:

| | budget | why |
|---|---|---|
| transfer statement | `MAX_TRANSFER_INPUTS` = 960 | the total, less the reserve |
| fee statement | `FEE_INTENT_INPUT_RESERVE` = 64 | the fee intent runs on `FREE_COMPUTE_GRACE_POINTS` of credit, and that credit is sized for a fee-sourcing transfer of up to 64 dust inputs |

64 is not a round number picked for headroom: a fee statement wider than that traps out of gas whatever else the transaction does, so the same figure is both what a transfer must leave behind and what the fee selection is itself capped at.

The reserve is held back **unconditionally**, including on the merged-statement path that may not need a fee statement at all. Whether the merge fits the fee-intent credit is only known once the realised input count is — which is after selection — and a selection that had already claimed the whole budget could not then fall back to a separate statement. That costs a merged transfer 64 of its 1024 inputs, which is the price of the fallback staying available.

A `const` assertion ties the budgets to the **per-statement** cap as well as to the total they are carved from, because the two are different numbers (1000 against 1024). A reserve much smaller than this one would leave a transfer free to select more inputs than any single statement may spend, and the statement it built would be refused while the transaction stayed inside its total. That failure would otherwise only show up as a rejected transaction.

## Running out of budget is no longer "insufficient funds"

A wallet holding the money in more UTXOs than one transaction can spend was reporting a balance it *has* as a balance it does *not* — which tells the user nothing about what to do. `InputLimitReached` names the limit, how much was covered, and that consolidating is the fix.

Branch-and-bound returns no solution both when the outputs cannot reach the amount and when no `max_inputs` of them can, so the two are told apart by whether the available total reaches the amount.

## Also

Drops a `.take(max_inputs)` over the branch-and-bound result. The search already bounds its own solutions, and the surrounding code locked every selected key regardless of the take — so a truncation there would have locked more inputs than it returned rather than bounding anything.

## Testing

42 tests in `tari_ootle_wallet_sdk` pass, `cargo lints clippy --all-targets` clean on the SDK and walletd.

The budget-vs-per-statement invariant is a `const` assertion, so it cannot be skipped. What is **not** covered is the end-to-end path — that a transfer plus its fee statement now stays inside the total, and that a fragmented wallet gets `InputLimitReached` rather than `InsufficientFunds`. That needs a `WalletSdkSpec` fixture (store plus key manager) and no test in `apis/` constructs one today, so it would be the first of its kind and is better as its own change than buried in this one. The selection mechanism itself — that branch-and-bound honours `max_inputs` — is already covered by `test_max_inputs_limit_respected` and `test_max_inputs_limit_prevents_solution`.

## Still open after this

Selection quality is unchanged: a fragmented wallet sending a large amount still gets whatever subset branch-and-bound finds, and a 960-input transfer weighs ~21,000 against the 24,000 a leader packs, so it fits a block only by being nearly alone in it. Preferring fewer, larger inputs for sends — and keeping smallest-first for deliberate consolidation — is a separate change.


sdbondi (commented):
_Posted by Claude._

Reviewed the full diff at 28a70fa, both files at head, with the selection paths traced through `branch_and_bound::select`, `lock_inputs_within_budget` and walletd's specific-input flow. One item I would block on; the rest is sound. Local: 42 SDK tests pass, `cargo lints clippy --all-targets` clean on the SDK and walletd.

**Blocking**

- `crates/wallet/sdk/src/apis/stealth_outputs.rs:233` raises the `lock_specific_outputs` bound from the per-statement `max_inputs` (1000) to `MAX_INPUTS_PER_TRANSACTION` (1024). Those inputs go into one `generate_transfer_statement` call (`applications/tari_walletd/src/handlers/accounts.rs:1686-1694`), so 1001 to 1024 named UTXOs now pass the wallet and are refused by the engine's per-statement cap. That is the failure class this PR exists to remove, and the `const` assertion three lines above it spells out exactly this hazard for the selection path. The specific path also shares no budget with the fee statement, so even 1000 plus a 64-input fee statement lands over the total. Bounding it by `MAX_TRANSFER_INPUTS` fixes both and keeps one number for what a transfer statement may carry.

**Sound**

- The budget split. `MAX_TRANSFER_INPUTS` 960 plus `FEE_INTENT_INPUT_RESERVE` 64 equals the 1024 total, both under the 1000 per-statement cap, and the `const` assertions pin that. The 64 matches the `FREE_COMPUTE_GRACE_POINTS` sizing note at `crates/engine_types/src/limits.rs:120-121`.
- Holding the reserve unconditionally on the merged path. `merged_statement_fits_credit` runs after selection (`api.rs:609`), so the fallback to a separate fee statement has to have budget left. The doc comment states this correctly.
- `InputLimitReached` diagnosis for branch-and-bound. `select` has no iteration cap, it is an exhaustive search with pruning, so `None` means exactly that no subset of at most `max_inputs` reaches the target. With that, "available total covers the amount" is a correct discriminator between fragmentation and a real shortfall.
- Dropping `.take(INPUT_LIMIT)`. The search bounds its own results and `stealth_outputs_lock_many` locked every selected key regardless, so the take could only have desynced returned inputs from locked ones.
- The `max_inputs` range check in `lock_outputs_internal` covers both new callers. `lock_outputs_until_partial_amount` discarding `reached_input_limit` is right, since its caller tops up from revealed funds by design.
- The `confidential_transfer.rs:206` caller is on `ConfidentialOutputsApi`, not this API, so the signature change does not reach it.

**Non-blocking**

- Branch-and-bound is exhaustive with no iteration budget. Pre-existing, but with wallets now expected to hold hundreds of stealth UTXOs, the worst case on an adversarially shaped set is exponential. Worth a follow-up ticket alongside the selection-quality work the description already lists.

sdbondi:
_Posted by Claude._

Blocking item fixed in 5b50e67. You were right, and it was a regression rather than a pre-existing gap: the original `STEALTH_LIMITS.max_inputs` bound on `lock_specific_outputs` was correct for that path, and I loosened it while generalising the constant. Confirmed the named inputs reach a single `generate_transfer_statement` call, so the per-statement cap governs them. They now take `MAX_TRANSFER_INPUTS`, which satisfies that cap and leaves the fee statement its reserve.

Your reasoning applied one layer down too, so I tightened `lock_outputs_internal`'s range check with it: it admitted a `max_inputs` up to the per-transaction total, so a future caller could have been handed more than one statement can spend. A selection's inputs all land in one statement, so the transfer budget is the ceiling for any single selection — the per-transaction total is the sum of the budgets, not a bound on either.

On the non-blocking note about branch-and-bound being exhaustive: agreed it wants a ticket, and it belongs with the selection-quality work rather than here. Worth recording that the shape which makes it expensive — many small inputs, no exact match — is exactly the fragmented wallet this change is aimed at, so it is closer than "pre-existing" suggests.

sdbondi (commented):
_Posted by Claude._

Re-reviewed 5b50e67 against 28a70fa.

**Resolved**

- `lock_specific_outputs` is now bounded by `MAX_TRANSFER_INPUTS` (`stealth_outputs.rs:235`), which satisfies the 1000 per-statement cap and leaves the fee statement its reserve. Tightening `lock_outputs_internal`'s range check to the same ceiling (`stealth_outputs.rs:395`) is correct for the same reason and closes the hazard for future callers.

**Blocking**

- `crates/wallet/ootle-rs/examples/stealth_transfer.rs` was swept into this commit and has nothing to do with the PR. It changes the example's recipient address and collapses the `use ootle_rs::{...}` imports onto shared lines. That layout is what stable rustfmt produces; the repo's `cargo +nightly-2025-12-05 fmt` reformats it back to one import per line (verified against `rustfmt.toml`, the base version is clean and the head version is not), so the `fmt` CI job will fail on this head. Drop the file from the commit.

sdbondi (commented):
Tested, Approved

sdbondi (commented):
_Posted by Claude._

Re-verified db484b9 against 5b50e67: the only change is the removal of `crates/wallet/ootle-rs/examples/stealth_transfer.rs` from the diff; the two SDK files are byte-identical to the previous head, where the `MAX_TRANSFER_INPUTS` bound on named inputs and the tightened `lock_outputs_internal` range check were already confirmed. Both blocking findings are resolved. `fmt` and the other 15 checks pass; only the integration test is still running. Nothing further from me; approval needs another maintainer since this is the author's own PR.

PR #2622, opened 2026-09-16
feat(faucet)!: add permissionless deposit

sdbondi:
## Description

Adds a `deposit` method to the builtin `XtrFaucet` so funds can be returned to, or added to, the faucet vault by any caller:

```rust
pub fn deposit(&self, bucket: Bucket) {
    emit_event("deposit", metadata!["amount" => bucket.amount().to_string()]);
    self.vault.deposit(bucket);
}
```

Permissionless by design — there is nothing to protect, since the faucet gives its funds away on request. A bucket of any resource other than the faucet's own is rejected by the vault.

## Motivation and Context

This shipped in the v0.40.2 hotfix line (b3279bc12) but was committed straight to `main`, so `development` never got it. That is the only real code divergence between the two branches, and it would otherwise be re-carried by every future release merge.

## How Has This Been Tested?

`cargo check -p tari_template_builtin --features templates` — the template compiles to wasm with the new method.

Call `deposit` on the faucet component with a TARI bucket and confirm the vault balance increases and a `deposit` event is emitted; call it with a bucket of another resource and confirm the vault rejects it.

## Breaking Changes

- [ ] None
- [x] Other

BREAKING: requires coordinated upgrade


sdbondi (commented):
Approved

PR #2623, opened 2026-09-16
chore(esme): set V1 activation epoch

sdbondi:
Description
---
chore(esme): set V1 activation epoch 11084

Breaking Changes
---

- [x] None
- [ ] Other - Please specify


sdbondi (commented):
Approved

PR #2625, opened 2026-09-16
docs: add changelog entry for 0.41.0

sdbondi:
Covers the 114 commits from `v0.40.2` to `v0.41.0`.

## The upgrade notes are the part worth reading

They lead with the esmeralda `ProtocolVersion::V1` activation at **epoch 11086**, because that is what makes this release a coordinated upgrade rather than a reset. `hash_substate` selects its schema from a substate's *creation* epoch, so without a scheduled activation the hashed-schema changes — a receipt's `exhaust_burn`, a resource's `auth_hook_updater`, the block header and vote shape — would land mid-epoch, and an epoch-granular schedule cannot describe a mid-epoch cutover (#2551 says exactly that: such a network needs a reset). With 11086 set, substates created under V0 keep hashing under V0.

Two conditions are stated alongside it:

1. Every validator is on the new binary **before** epoch 11086. `check_activation_schedule` refuses to start a node whose schedule disagrees with what it has already run past, so a late one fails loudly rather than diverging quietly.
2. Esmeralda must never have run a post-#2520 binary while still on V0. #2520 added `FeeReceipt::exhaust_burn`, which borsh writes unconditionally; a receipt committed that way under V0 is hashed with the field and no gate can describe it. `v0.40.2` predates #2520, so this holds as long as nothing from `development` was ever deployed there.

The stricter requirement is called out separately, because it is the one that actually forces a lockstep restart: the **ungated** engine changes — metering rates, the weight floor, event payloads, the proof scope model, the account-ownership gate — switch at *binary deployment*, not at an epoch boundary. Two validators on different binaries compute different receipts for the same transaction.

What would previously have been a "needs a reset" list is now framed as **what breaks on esmeralda without one**, all accepted as testnet costs: a deployed template calling the retired `EngineOp::SignatureInvoke` or whose binary lacks the `tari_tdef` custom section; a component whose state already holds a `BucketId`, `ProofId` or address allocation; and any substate already over 1 MiB, which is rejected on touch until it is shrunk. #2591 asks for that last one to be confirmed on esmeralda rather than assumed.

## Ordering

Wallet and Indexer come before the engine sections. The signature cap is told as a wallet change — a balance split across more than 17 stealth UTXOs was unspendable in one transaction, and a coinjoin was capped at 17 participants — rather than as a transaction-format one, and the indexer section carries the RPC deadline/keepalive work that made follow mode possible.

## Not addressed here

`0.40.1` and `0.40.2` still have no entries of their own. Worth a follow-up, or folding into this one if you'd rather the file be complete.


sdbondi (commented):
Approved

PR #2626, opened 2026-09-16
fix(tariswap-bench): stamp a distinct nonce per transaction

sdbondi:
## Problem

Running the tariswap bench against a local swarm dies immediately after the accounts and faucet are set up:

```
INFO ⏳️ Creating 1000 tariswap components...
Error: Store error: General database failure for operation transactions_insert: UNIQUE constraint failed: transactions.transaction_id
```

Since #2352 the transaction id is a pure function of the body — the seal signature (and its fresh Schnorr nonce) is deliberately excluded so that a re-seal of the same intent keeps its id. `UnsignedTransactionV1::nonce` is what distinguishes otherwise-identical bodies; walletd and the integration-test helper both stamp `rand::random()`, but the bench's `Runner::new_transaction_builder` left it at `0`.

`create_tariswaps` is called 40 times with the same account, faucet, fee and 25 identical `call_function` instructions over unversioned inputs, so every call produces byte-identical bytes and therefore the same id. The first insert wins, the second hits the unique index. The account and funding loops survive only because their per-index arguments differ; the swap loop (`do_tariswap_swaps`, run 10 times over the same accounts and pools) would have failed the same way next.

## Fix

Give `Runner` a counter and stamp it via `with_nonce` on every builder it hands out.

## Test plan

- `cargo check -p tariswap_bench`


sdbondi (commented):
Approved

PR #2627, opened 2026-09-16
fix(account)!: allow anyone to call deposit_with_auth

sdbondi:
## Problem

`deposit_with_auth` (#2588) exists so that someone other than the account's owner can deposit a badge-restricted resource, handing the account the badge its deposit rule names. It was left under the account's `deny_all` default, so only the owner could call it. The owner is the one caller who never needs it, and #2608's tests passed only because the depositor was always the owner.

## Change

`deposit_with_auth` gets the same `allow_all` method rule as `deposit`. The account's owner rule gates what leaves; the resource's own deposit rule, satisfied by the handed badge, gates what arrives. The other `_with_auth` methods (withdraw and create-proof variants) stay owner-only, correctly.

## Consensus impact

Changes the builtin account binary (address unchanged, no substate moves). Validators and indexers roll together, as with #2588.

## Test

`access_rules::anyone_may_deposit_with_auth`: a sender who holds the badge withdraws restricted tokens with it and deposits them into another user's account via `deposit_with_auth`, signed by the sender alone. Fails on `development` with access denied on the method call; passes with the rule.

`cargo nextest run -p tari_engine --test access_rules` for the new test and the two adjacent `_with_auth` tests: 3 passed.


sdbondi (commented):
Approved

PR #2628, opened 2026-09-16
fix(swarm): burn funds into the wallet daemon, not the console wallet

sdbondi:
The base layer page rendered the burn form inside each Minotari console
wallet card and sent that card's instance id as `wallet_instance_id`.
The process manager resolves that id against its wallet daemon
collection, so the lookup could never match and every burn failed with
"No wallet daemon instances ... found. Please start a wallet before
burning funds" — including on the default swarm config, where all six
instances (and a wallet daemon among them) were running.

The id identifies the destination: the daemon holding the account whose
owner key the burn pays to. The console wallet spent from is chosen by
the process manager. Send the running wallet daemon's instance id, and
rename the request field to `wallet_daemon_instance_id` (aliasing the
old name) so the two wallet kinds can't be confused again.

Also drop the key-creating run from wallet daemon lookups: it shares the
collection but serves no JSON-RPC, so it is never a usable target.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_0159XHGPiDy9UWYYZEd14a9E


applications/tari_swarm_daemon/src/webserver/rpc/minotari_wallets.rs
```diff
@@ -32,7 +32,9 @@ pub async fn create(
 #[derive(Debug, Clone, Deserialize)]
 pub struct MinotariWalletBurnFundsRequest {
     pub amount: u64,
-    pub wallet_instance_id: InstanceId,
+    /// The wallet daemon that holds the account the burn is claimed into.
+    #[serde(alias = "wallet_instance_id")]
+    pub wallet_daemon_instance_id: InstanceId,
```
sdbondi: Is this change necessary, old name is fine

sdbondi: _Posted by Claude._

Not necessary — reverted in dde8a0e. The rename was only to make the two wallet kinds harder to confuse; the doc comment does that job, so the field is `wallet_instance_id` again (and the serde alias is gone with it).

---

sdbondi (commented):
Approve

PR #2629, opened 2026-09-17
refactor(engine_types): give the template blob a fixed wire ceiling

sdbondi:
## Motivation

`TemplateBlob` was `MaxBytes<{ ENGINE_LIMITS.max_template_binary_size_bytes }>`, which welds the **policy** limit on what may be published into the **wire format** of the `PublishedTemplate` substate. That left no room to move the limit in either direction:

- **Lowering it is the dangerous case, and it fails silently.** Dropping `max_template_binary_size_bytes` makes every already-published template above the new value undecodable — committed state a node can no longer read, surfacing as a codec error rather than anything actionable.
- **Template substates are exempt from `max_substate_size`** (`working_state.rs`), so this constant is the only size bound the substate format carries. That makes it load-bearing as a format constant and a poor place to express policy.

Decoding was never the enforcement layer, so nothing is lost by loosening it:

- `PublishTemplateLimitValidator` applies the limit at **mempool ingress and block validation**, so a transaction over it is never gossiped or stored, and a block carrying one is rejected.
- The engine re-checks it in `TransactionProcessor::publish_template` before the compile is charged.
- The minicbor `MaxBytes::decode` — what `tari_bor::decode` uses, and so the path this widens — borrows from the input and length-checks before copying, so the tight cap bought no DoS protection either. The real bound on untrusted bytes is the gossip/RPC frame limit. (The serde impl materialises the buffer before checking, but it is bounded by `N` all the same and is not the gossip path.)

## Change

State the ceiling outright rather than deriving it:

```rust
pub const MAX_TEMPLATE_BLOB_WIRE_BYTES: usize = 2 * 1024 * 1024; // 2 MiB = max_substate_size + 1 MiB

const _: () = assert!(limits::ENGINE_LIMITS.max_template_binary_size_bytes <= MAX_TEMPLATE_BLOB_WIRE_BYTES, ...);

pub type TemplateBlob = MaxBytes<MAX_TEMPLATE_BLOB_WIRE_BYTES>;
```

This is the same split already made deliberately for `MAX_TRANSACTION_SIZE_BYTES` in `consensus_constants.rs` — stated outright rather than derived, with an assertion failing if the two drift apart, so a change to either is reviewed against the other rather than inherited. The doc states what bounds 2 MiB from above (a template substate has to fit one `RPC_MAX_FRAME_SIZE`) so a future raise is a decision against stated constraints.

Also: docs on `max_template_binary_size_bytes` naming the ceiling it lives under, and the `working_state.rs` / `processor.rs` comments pointed at the actual enforcement sites.

`publish_template_too_big_binary` wrapped its binary in `TemplateBlob::new_unchecked` — the one documented way to break the `MaxBytes` invariant. The builder takes `Into<Blob>` and the binary travels as a `Blob`, so the wrapper bought nothing even before the ceiling moved; it now passes the `Vec<u8>` directly like every other call site in the file.

## Compatibility

No protocol version gate needed. The publish limit stays at 1 MiB, so no substate above the old ceiling can exist and nothing observable on the wire changes today. `Instruction::PublishTemplate` carries a `BlobIndex` into the unbounded `Blob` store rather than a `TemplateBlob`, so the only decode site this widens is the `PublishedTemplate` substate. The change only makes a future move of `max_template_binary_size_bytes` a policy decision rather than a format break.

## Testing

- New in `published_template.rs`: a blob past the publish limit but within the ceiling round-trips through CBOR; a blob past the ceiling fails to decode. Both go through the minicbor path; the serde impl's own bound is already covered generically in `max_bytes.rs`.
- `crates/engine/tests/publish_template.rs::publish_template_too_big_binary` is unaffected — its binary always travelled as a `Blob` and its rejection always came from `processor.rs`, not from decode.
- `cargo nextest run -p tari_engine_types -p tari_ootle_transaction_validation` and `-p tari_engine --test publish_template --test limits`: all green.
- `cargo lints clippy --all-targets` on the affected crates: clean.


sdbondi (commented):
_Posted by Claude._

Nothing blocking. The split is the right one and the guardrail chain around it holds: a raise of `max_template_binary_size_bytes` past 2 MiB fails the new const assert, a raise below it still fails `the_transaction_size_cap_still_admits_a_max_size_publish` (`consensus_constants.rs:604`) and `the_largest_publishable_binary_leaves_room_to_source_a_fee` (`limits.rs:473`), so the policy limit cannot move without a deliberate review. I also confirmed the change is inert on the wire: `Instruction::PublishTemplate` carries a `BlobIndex` into the unbounded `Blob` store (`crates/transaction/src/v1/instruction.rs:104`, `crates/transaction/src/blobs.rs:43`), not a `TemplateBlob`, so the only decode site this widens is the `PublishedTemplate` substate — and 2 MiB sits well under `RPC_MAX_FRAME_SIZE` (6 MiB, `networking/rpc_framework/src/lib.rs:38`).

`PublishTemplateLimitValidator` is wired into both chains as the new docs claim — `create_structural_transaction_validator` (`crates/transaction_validation/src/builder.rs:50`) and `create_node_transaction_validator` (`applications/tari_validator_node/src/bootstrap.rs:559`), the latter backing `TariBlockTransactionValidator`.

## Non-blocking

**1. The change makes an `unsafe` block redundant and its SAFETY comment false — `crates/engine/tests/publish_template.rs:92`.**

```rust
// SAFETY: We are intentionally publishing an oversized binary to test size limits.
.publish_template(unsafe { TemplateBlob::new_unchecked(random_wasm_binary) })
```

With the ceiling at 2 MiB, a `max_template_binary_size_bytes + 1` binary no longer breaks the `MaxBytes` invariant, so `new_checked(...).unwrap()` succeeds and the `unsafe` is unnecessary. `new_unchecked` is documented as "the only way to break the invariant guarantees of `MaxBytes<N>`" (`crates/template_lib_types/src/max_bytes.rs:44-49`), and this call site no longer does that — the comment now describes a hazard that isn't in the code. Since the PR's own argument is that the format is deliberately looser than the policy limit, this is the one place that should visibly stop relying on the escape hatch. Worth folding into this PR.

**2. The ceiling's value isn't justified, only its existence.** `MAX_TEMPLATE_BLOB_WIRE_BYTES` (`published_template.rs:163`) explains at length why a fixed ceiling rather than a derived one, but nothing says why *2 MiB*. The sibling precedent this PR cites states its derivation inline (`MAX_TRANSACTION_SIZE_BYTES: usize = 1_310_720; // 1 MiB + 256 KiB`). One line on what bounds the choice from above — it has to stay under `RPC_MAX_FRAME_SIZE`, and a published template substate is the only one that may exceed `max_substate_size`, which this now permits by 1 MiB rather than by 0 — would make a future raise a decision someone can make against stated constraints instead of re-deriving them.

**3. Two claims in the PR description don't survive checking.** Neither changes the conclusion, but they're the kind of thing that gets quoted back later:

- "`publish_template_too_big_binary` still passes — the rejection now comes from the engine's size check rather than from decode, which is the point." That test never went through a `TemplateBlob` decode. It bypasses the type with `new_unchecked` and the binary travels as a `Blob`; the rejection came from `processor.rs:645` before this change too. The test is unaffected, not relocated.
- "`MaxBytes::decode` borrows from the input and length-checks before copying." True of the minicbor impl (`max_bytes.rs:160-170`), which is what `tari_bor::decode` uses and therefore what matters here. The serde impl does the opposite — `deserialize_byte_buf` materialises the buffer at `max_bytes.rs:137`, then checks length at `:138`. Still bounded by `N`, and not the untrusted-gossip path, but the blanket statement isn't true of both impls.

## Nits

**4. `crates/engine/src/runtime/working_state.rs:243`** — "and sized by what the binary's compile costs" doesn't parse; "sized by what compiling the binary costs".

**5. The new tests cover only the minicbor path.** `a_blob_past_the_ceiling_does_not_decode` (`published_template.rs:239`) goes through `tari_bor::decode`, i.e. `minicbor::Decode`. The serde `Deserialize` impl carries its own independent bound (`max_bytes.rs:127` for hex, `:138` for bytes) and is what the JSON-RPC surface uses. `max_bytes.rs` already tests both paths generically, so this is a note rather than a gap — just that "does not decode" is narrower than it reads.

Once (1) is addressed and CI is green (clippy and `test (1-3/3)` are still pending at `01fb538`) this is good to merge.

sdbondi:
_Posted by Claude._

All four actionable points addressed in 3da1998.

**1. The redundant `unsafe`** — fixed, and further than suggested. `new_checked(...).unwrap()` would still be ceremony: the builder takes `Into<Blob>` and the binary travels as a `Blob`, so the `TemplateBlob` wrapper bought nothing even before the ceiling moved. It now passes the `Vec<u8>` directly, like every other `publish_template` call site in the file, and the `TemplateBlob` import is gone.

**2. The ceiling's value** — the doc now states what bounds it from above: a template substate has to fit one `RPC_MAX_FRAME_SIZE` (6 MiB), and the value keeps the one substate that may exceed `max_substate_size` within a megabyte of every other. Trailing comment reads `// 2 MiB = max_substate_size + 1 MiB`.

**3. Both description claims** — corrected.

- You're right that `publish_template_too_big_binary` never decoded through `TemplateBlob`; it's unaffected, not relocated. The description now says so.
- The `MaxBytes::decode` claim is now scoped to the minicbor impl, which is what `tari_bor::decode` uses and therefore the path this widens, with the serde difference noted.

**4. Grammar** — "sized by what compiling the binary costs".

**5.** Noted, no change — agreed that `max_bytes.rs` already covers both impls generically; the test names describe the minicbor path they exercise.

sdbondi (commented):
_Posted by Claude._

All three items from the previous review are addressed at `3da1998`, re-verified rather than taken on the commit message:

- `crates/engine/tests/publish_template.rs:91` now passes the `Vec<u8>` straight through — `From<Vec<u8>> for Blob` (`crates/transaction/src/blobs.rs:90`) covers it, `wasm_binary_size` is captured before the move, the `TemplateBlob` import is gone with no stale references left in the file, and `limits` is still used by the two assertions. The test asserts the same `WasmBinaryTooBig` rejection from `processor.rs:645`, so what it covers is unchanged — it just no longer reaches for `new_unchecked` to get there.
- `MAX_TEMPLATE_BLOB_WIRE_BYTES` (`published_template.rs:161-165`) now states what bounds the value from both sides: `RPC_MAX_FRAME_SIZE` above, a doubling of the publish limit below.
- Grammar fix at `working_state.rs:243`.

CI is green on this head, including `clippy`, all three `test` legs, the four fuzz targets and `integration test`.

The remaining description nits are cosmetic and don't need a push.

No blocking findings remain — this is a sign-off from my side; GitHub will not let a review approve its author's own PR, so another maintainer still has to click approve.

sdbondi (commented):
Approved

PR #2630, opened 2026-09-17
docs: add release checklists and pre-tag / post-tag release tooling

sdbondi:
## Description

Replaces the 23-line `CHECKLISTS.md` — which stopped at "create a release on GitHub" — with procedures that can be run top to bottom by a human or an agent, plus two scripts that do the checking.

Everything that has actually gone wrong lives past where the old checklist stopped:

- **v0.41.0 published without Windows wallet binaries.** A `best_effort` build leg failing does not fail the tag run, and the developer docs render the wallet downloads of the newest *non-draft* release — so publishing the draft shipped Windows users an empty download table.
- **Its draft carried assets from two commits** (`43d6617` and the tag's `ac73128`), which ootle-go's vendor script had to grow a `--commit` flag to disambiguate.
- **The downstream SDKs are each pinned to a different, older core than the one that shipped**: ootle-go on 0.39.1, ootle.ts on bindings 1.50.1 / wasm 0.39.0, ootle-py on a local 0.31.0 wasm build.

### `checklists/`

| File | When |
|---|---|
| `release.md` | every release |
| `release-breaking.md` | consensus, engine, on-chain ABI, config struct or SDK surface changes |
| `hotfix.md` | a fix on top of an already-tagged version |
| `README.md` | shared reference: what a tag sets off, what is immutable, recovery paths, and what changes when the `main`/`development` split goes away |

`release.md` covers the parts that were only ever in someone's head: waiting for the wallet builds to attach before publishing the draft, verifying the testnet deploy came up on the new binary, publishing crates, and then updating ootle-go (FFI libs), ootle.ts (catalog pins) and ootle-py (vendored wasm blob). `release-breaking.md` covers the activation-epoch decision, the upgrade notes, and the SDK surfaces — including that a behaviour change with no ABI change still breaks hosts, which is what the ABI tag cannot catch.

### `scripts/release_check.py` — the pre-tag gate

Must print `READY TO TAG`. Checks a clean tree on a release branch and in sync, a free tag locally and on origin, a changelog entry, a version bump for every crate **and npm package** whose files changed since the last release, a topological publish order, pins that all resolve, and flags a touched protocol activation schedule.

It also answers the question `cargo publish --dry-run` cannot. A dry run resolves every dependency against the registry, so it fails on any dep whose new version this release has not published yet — a failure that says nothing about the real publish. The publish preflight resolves pins against crates.io **plus** what this release will publish. `--package` additionally runs `cargo package` for the crates that genuinely can be packaged now.

On the current `development` tip it correctly reports `tari_template_builtin` and `tari_engine` as changed since v0.41.0 and owing a bump.

### `scripts/release_status.py` — the post-tag dashboard

`--watch` until the draft is complete. Renders the draft's assets against the binary and FFI build matrices (required legs vs `best_effort`), the tag's workflow runs, npm, crates.io, and what each downstream SDK currently carries. Refuses to call a draft safe to publish while a required artifact is missing — run against v0.41.0 it flags the missing Windows binaries and the foreign-commit assets.

## Motivation and Context

Releases were run from memory. The parts that bite are the ones no workflow fails on: a silently skipped npm publish, a best-effort build leg, a downstream SDK left a version behind.

## How Has This Been Tested?

Both scripts run against the current tree and against the published v0.41.0 release; their findings are the ones described above. `satisfies()` (pin resolution) spot-checked across caret/bare/exact requirement shapes.

```sh
./scripts/release_check.py            # gate on this tree
./scripts/release_status.py v0.41.0   # dashboard for the last release
```

Both are read-only.

## Breaking Changes

- [x] None
- [ ] Requires hard fork
- [ ] Other - Please specify


sdbondi (commented):
LGTM

PR #2631, opened 2026-09-17
fix(ci): select packages when building release binaries

sdbondi:
The Windows legs of the release build matrix have failed since the engine started
pulling in wasmer, most recently on
[v0.41.0](https://github.com/tari-project/tari-ootle/actions/runs/35087510150/job/104767637154):

```
error: The Cranelift compiler backend is not supported on Windows. Use the V8 backend instead.
  --> wasmer-compiler-cranelift-7.4.0/src/lib.rs:22:1
error: could not compile `wasmer-compiler-cranelift` (lib) due to 1 previous error
```

Windows only builds `tari_ootle_walletd`, and the wallet daemon deliberately takes
`tari_ootle_app_utilities` without its `execution` feature precisely so it does not need
the engine. The build command was

```
cargo auditable build --locked --release --target x86_64-pc-windows-msvc \
  --features "default" --bin tari_ootle_walletd
```

With no `--package`, cargo selects every default workspace member and resolves the whole
workspace's dependency graph, so cranelift lands in the build plan regardless of which
binary was asked for. Confirmed with `--unit-graph` for `x86_64-pc-windows-msvc`: 1330 units
including 28 cranelift/wasmer units without `--package`, 720 units and zero with it.

Passing `--package` alongside `--bin` scopes the graph to the packages owning the requested
binaries. Each entry in `TS_FILES` is named after its package, so the flag can be derived
from the same list.

Side effect: features no longer leak into the released binaries through workspace-wide
unification. The biggest of these is axum's `http2`, which none of the four binaries enable
(their JSON-RPC and REST clients are HTTP/1.1); 59 shared units lose at least one feature
that some other workspace member was enabling.

Not addressed here: `windows-arm64` fails separately, with rustc resolving `link.exe` to Git
for Windows' coreutils `link` instead of the MSVC linker (`/usr/bin/link: missing operand`).
That leg is `best_effort`.


sdbondi (commented):
utACK

PR #2632, opened 2026-09-17
ci: drop riscv64, cache vcpkg openssl, fix the windows-arm64 linker

sdbondi:
Follow-up to #2631. Three changes to the release build matrix, aimed at a tag producing a
complete set of binaries on the first run.

### Drop linux-riscv64

`Cross.toml` pinned `ubuntu:20.04` for that target. Its binutils 2.34 cannot merge the ISA
attributes rustc 1.97 emits (`rv64i2p1_m2p0_..._zmmul1p0_zaamo1p0_zalrsc1p0...`), so the leg
ends in

```
ld: failed to merge target specific data of file ...
error: could not compile `tari_crypto` (lib) due to 1 previous error
```

Bumping the image to ubuntu:24.04 would likely fix it, but no one is waiting on a riscv64
validator binary, so the leg goes rather than gets maintained.

### Cache vcpkg's built packages on Windows

Building OpenSSL from source is the most expensive single step on a Windows runner — 5m45s of
the 38m job on v0.41.0. vcpkg hashes the ABI of every port it restores, so a stale cache costs
a rebuild rather than yielding a wrong binary. Tag runs read the cache the scheduled runs on
the default branch populate.

The step's comment blamed hickory for the OpenSSL dependency; it actually arrives via
`webauthn-rs` → `webauthn-attestation-ca` → `openssl`.

### Fix the windows-arm64 linker

That leg fails before it links anything:

```
error: linking with `link.exe` failed: exit code: 1
  = note: "C:\\Program Files\\Git\\usr\\bin\\link.exe" "/DEF:..."
  = note: /usr/bin/link: missing operand after '\377\376'
```

rustc has no linker discovery for `aarch64-pc-windows-msvc`, so it spawns a bare `link.exe`;
the build step runs under Git bash, where Git for Windows' coreutils `link` precedes the MSVC
one on PATH. A new step resolves `link.exe` in pwsh — restricted to a `\VC\Tools\MSVC\` path,
and failing loudly otherwise — and hands cargo the absolute path via
`CARGO_TARGET_AARCH64_PC_WINDOWS_MSVC_LINKER`.

The OpenSSL triplet and the `<TARGET>_OPENSSL_DIR` variable now follow the matrix target, so
the ARM64 leg links an ARM64 OpenSSL rather than the x64 one it was pointed at before.

Both Windows legs build only `tari_ootle_walletd`. Whether the ARM64 one now gets all the way
through is what the verification run answers: `aws-lc-sys` and `wasm-opt-sys` (binaryen, C++)
are in the wallet daemon's dependency graph and have to cross-compile.

Verification: https://github.com/sdbondi/tari-ootle/actions/runs/35190055178 (fork, `build-bins-win-arm64`)


sdbondi (commented):
ACK

PR #2633, opened 2026-09-17
ci: shorten and de-risk the tag build run

sdbondi:
Follow-up to #2632. Two independent changes, one commit each.

### Let tag builds restore the cargo dependency cache

`Cache cargo files and outputs` was gated on `! startsWith(github.ref, 'refs/tags/v')`, so a
release leg compiled its entire dependency graph from scratch — roughly 20 of the 38 minutes
the Windows job took on v0.41.0.

The step now runs on tags too, with `save-if` keeping the *writes* off them. The scheduled runs
on the default branch keep the cache warm and a tag run is allowed to read the default branch's
caches, so what a release build restores was produced before the tag existed. `rust-cache` drops
the workspace's own crates before saving, so every Ootle crate and the binary itself is still
compiled and linked from the tag's commit.

### Fail a tag run when a required leg fails

`continue-on-error` was set for *every* leg on a tag ref, because `create-release` needs
`builds` and would otherwise be skipped along with the failure. The effect was that v0.40.2 and
v0.41.0 both reported a green build run while shipping no Windows binaries; the gap surfaced
only when someone read the draft.

Tolerance is now scoped to the legs marked `best_effort`, and `create-release` runs on anything
short of a cancellation. A failed required leg turns the run red, and the draft is still
assembled from the legs that did build — the artifacts attach as before, the failure is just
visible without reading the matrix.

One consequence worth naming: `create-release` now also runs when `ledger-app` fails, where it
used to be skipped. The draft is a draft, `release_status.py` is still the gate on publishing,
and a red run is a better signal than a skipped job.

The checklist gains a line saying a red run does not mean an empty draft.


sdbondi (commented):
ACK

PR #2634, opened 2026-09-17
fix(engine)!: reject transient values in non-fungible data

sdbondi:
Closes the non-fungible-data half of the class #2616 closed for component state, named there as a deliberate follow-up.

## Problem

A bucket, proof or address allocation is a capability the transaction holds, named by a counter that restarts at zero every transaction. Both non-fungible write sites take a caller-supplied `tari_bor::Value` and store it with no `IndexedWellKnownTypes` pass:

- **mint** — `WorkingState::mint_resource`, `MintArg::NonFungible`, writes both the immutable `data` and `mutable_data`;
- **update** — `ResourceAction::UpdateNonFungibleData`, writes `mutable_data`.

`NonFungible` holds both as raw CBOR, so a tagged id survives the round-trip to the ledger.

Measured against unfixed code, with the object gone by finalize so `validate_finalized` has nothing live to object to — all three shapes commit, and the stored value is `Tag(135, Integer(0))`, a `ProofId`:

| shape | without this fix |
| --- | --- |
| proof id in minted `data` | **reaches the ledger** |
| proof id in minted `mutable_data` | **reaches the ledger** |
| proof id written by `UpdateNonFungibleData` | **reaches the ledger** |
| control: minted bucket dropped on the floor | rejected — `1 dangling buckets remain` |

## Why it is more than hygiene

Weaker than the component-state case, and worth saying so. Non-fungible data does **not** reach a resource auth hook: a hook is invoked with `(action, AuthHookCaller)`, and `AuthHookCaller` carries the *acting component's* state. `UpdateNonFungibleData` does fire a hook, but the hook sees the action and the caller, never the data being written.

The reachable amplification is indirect: a template can read `mutable_data()` and forward it as a call argument, at which point argument indexing (`include_refs_in_scope`) reads a tagged `ProofId` as a capability the callee was lent. The primary argument is #2616's — a transaction-scoped id has no business on the ledger, because it can only ever alias something else later.

## Fix

The four checks #2616 put inline in `validate_component_state` become `reject_transient_values(location, &IndexedWellKnownTypes)`, plus `reject_transient_values_in(location, &tari_bor::Value)` for a value that has not been indexed yet. `RuntimeError::TransientValueInComponentState` becomes `TransientValueInSubstate`, carrying the field name; with `location = "Component state"` the component-state message is unchanged.

The mint path checks `data` and `mutable_data` before creating the substate. `UpdateNonFungibleData` checks `arg.data` before the resource auth hook runs, so a doomed transaction does not execute hook code first.

## Scope: transient ids only

Deliberately **not** widened to substate references. Component state additionally scope-checks new references and turns them into ownership edges via `attach_node_to_component`; non-fungible data does neither. That is inert today — `GetData`/`GetMutableData` return the raw CBOR without touching scope, and every vault operation gates on `is_substate_in_scope` — and the only code that walks NFT data for references, `indexer_lib::substate_decoder::find_related_substates`, has no callers. The latent issue is that NFT data can carry a reference that is not an ownership edge and nothing validates it; if that indexer function is ever revived, a forged reference becomes a lie a consumer believes. Blast radius is indexer-side relatedness, not funds, so it is noted rather than fixed here.

## Breaking

A non-fungible whose `data` or `mutable_data` holds one of these ids can no longer be minted or updated. No template in this repo stores one, and on testnet the cost of bricking any that do is accepted — the same posture taken for component state in #2616.

Two rejections come along with routing the value through `IndexedWellKnownTypes::from_value`, and they are not transient ids:

- **Depth.** The indexer walks to `MAX_VISITOR_DEPTH` (50) while `MAX_DECODE_DEPTH` is 64, so non-fungible data nested 50–63 deep decodes and mints today and now fails with `Maximum depth exceeded`.
- **Malformed well-known tags.** A value carrying a Tari `BinaryTag` whose payload does not decode now fails where it previously round-tripped, because `try_from_tag_and_value` propagates the error.

Both are defensible — component state has lived under exactly these limits since it was indexed — but a mint failing with a depth error is a surprise if the note only mentions transient ids.

## Tests

`shenanigans::it_rejects_transient_values_in_non_fungible_data` covers the three write paths. Verified against unfixed code: all three *succeed* and the tagged `ProofId` lands in the committed substate, while a control transaction that drops a minted bucket is rejected.

`it_rejects_transient_values_in_component_state` is unchanged and still passes — generalising the variant kept its message.

`tari_engine` 460/460.


sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 3 non-blocking, 0 nits · head `c0f4f3c` (cannot approve my own PR; tests still pending)

**What I verified**

- Write-site coverage is complete. `NonFungibleContainer::new` has exactly one engine call site (`working_state.rs:981`) and `set_mutable_data` exactly one (`impl.rs:1878`); both are now guarded. `mint_resource` is the only route to `MintArg::NonFungible` (`impl.rs:1627`, `impl.rs:1722`, covering both create-with-initial-supply and mint), and `NonFungibleAction` exposes only `GetData`/`GetMutableData`, which read. No third path.
- The `UpdateNonFungibleData` check does run before `check_resource_access_rules` and before `invoke_resource_access_hook`, as the description says.
- Description claims check out: `find_related_substates` (`indexer_lib/src/substate_decoder.rs:34`) has no callers anywhere in the tree, and `GetData`/`GetMutableData` return the raw CBOR without a scope check.
- The component-state message is genuinely unchanged: `location = "Component state"` reproduces the old prefix verbatim, so `it_rejects_transient_values_in_component_state` keeps asserting the same string.
- Extracting the four inline checks into `reject_transient_values` is a faithful move — same order, same kinds, same `first()` semantics.

**Non-blocking**

1. `crates/engine/src/runtime/working_state.rs:2331` — the breaking surface is slightly wider than the "Breaking" section states. `reject_transient_values_in` routes the value through `IndexedWellKnownTypes::from_value`, which also enforces `MAX_VISITOR_DEPTH` (50) and propagates a `try_from_tag_and_value` failure. `MAX_DECODE_DEPTH` is 64, so non-fungible data nested 50–63 deep decodes today and mints today, and after this change it does not — nor does data carrying a well-known tag with a payload that fails to parse. Both rejections are defensible (component state has lived under them since it was indexed), but they are not "transient ids only", and a mint failing with `Maximum depth exceeded` is a surprise if the release note says otherwise. Worth one line in the PR body / commit message.

2. `crates/engine/src/runtime/impl.rs:1838` vs `working_state.rs:980` — the two checks sit on opposite sides of authorization: update rejects before the access rule is evaluated, mint rejects after (`mint_resource` runs under an already-authorized lock). Side effect on the update path: a caller with no `UpdateNonFungibleData` permission passing a transient id now gets the transient error rather than access-denied. Harmless — the id is the caller's own and tells them nothing — and the stated reason for the ordering (don't run hook code for a doomed transaction) is sound. Noting it only so the asymmetry is deliberate rather than incidental.

3. `working_state.rs:569` — this is the second write site of the class to be patched individually (#2616 was the first), and the scope section names a third latent one. `validate_finalized` already iterates `self.store.mutated_substates()` for the size limit; the same loop could run `reject_transient_values` over each mutated substate's value as a backstop, which closes the class rather than the site. Cost is one extra walk per mutated substate — component state and non-fungible data would each be walked twice. Not for this PR, but it is the shape that stops the next one.

**On the test**

`it_rejects_transient_values_in_non_fungible_data` is the right construction: banking the minted token in a vault so the run reaches finalize, and dropping the proof from the workspace so `validate_finalized` has nothing live to object to, is what makes it a test of this check rather than of the dangling-proof check. The two location strings are distinct enough that the `contains` assertions cannot cross-match (`"Non-fungible data may not contain a proof"` is not a substring of the mutable-data message).

CI: `clippy`, `fmt`, `docs build` green; `test (1-3/3)` and `integration test` still pending at the time of writing.

sdbondi:
_Posted by Claude._

Thanks — all three verified, one change made.

**1. Widened breaking surface — correct, PR body updated.**

Confirmed: `MAX_VISITOR_DEPTH = 50` (`engine_types/src/indexed_value.rs:32`) against `MAX_DECODE_DEPTH = 64` (`tari_bor/src/value.rs:257`), so non-fungible data nested 50–63 deep decodes and mints today and now fails with `Maximum depth exceeded`. The second case holds too — `WellKnownTariValue::try_from_tag_and_value` (`indexed_value.rs:448`) returns `Err` on a Tari `BinaryTag` whose payload does not decode, and the visitor propagates it, so such a value round-trips today and does not after. The "Breaking" section now names both.

Left out of the commit message deliberately: it does not claim "transient ids only", so there is nothing there to correct, and a force-push to reword it is churn on a reviewed branch.

**2. Ordering asymmetry — deliberate, and I am leaving it.**

The check is a pure function of caller-supplied bytes, so it runs before any lock or authorization work on that path, which is the normal order for argument validation. The mint side is not the same decision: `mint_resource` runs under an already-authorized lock, so there is no earlier point to move it to without hoisting it out of `MintArg::NonFungible` entirely, and doing that would separate the check from the write it guards.

Agreed the error-precedence flip is inert — the id is the caller's own transaction-scoped counter and carries no information about the resource or the token.

**3. `validate_finalized` backstop — agreed, and it is the right shape.**

Two write sites patched individually with a third latent one named is exactly the argument for a backstop over per-site checks. It is out of scope here: it changes the rejection point for the two sites this PR and #2616 cover, which makes both tests assert against a different check than the one they were written for, and the extra walk per mutated substate wants its own measurement rather than riding along on a consensus-affecting fix. Filing it as a follow-up.

sdbondi (commented):
Approved

PR #2635, opened 2026-09-17
fix(consensus)!: require a candidate block to extend its justify block

sdbondi:
## The defect

A local proposal was validated against its justify block only loosely:

- `on_receive_local_proposal.rs` checked that the parent exists **only** when the parent happened to be the justify block (`justifies_parent()`), and nothing required the parent to be the justify block or to be one height below the candidate.
- `Block::is_safe`'s liveness rule passed on `max_certificate_height() > locked.height()`, which includes the height of a timeout certificate. A TC certifies no block.
- The timeout-certificate branch returned `Ok(ValidBlock::with_dummy_blocks(..))` **before** the safeNode predicate ran at all.

Together these let one byzantine leader split the committee permanently. It proposes `B4` (justify `QC3`, parent `B3`) to one replica and `B'` (justify `QC3`, parent `B0`) to the rest. `B'` passes: its justify block exists, its parent is not the justify block so existence is never checked, and the liveness rule passes on the justify height alone. Both branches gather a quorum, commit recursion follows parent links, and the two branches commit different chains. The same hole after a timeout is wider still: because the TC's height counts toward `max_certificate_height`, the leader at `T+1` could justify from an arbitrarily old QC and skip the predicate entirely.

## The fix

A candidate now reaches its justify block in exactly one of two ways, checked by `check_extends_justify`:

- no view gap: the parent **is** the justify block;
- a view gap: a timeout certificate is present, the dummy chain is filled to its full length, and the parent is the last dummy.

The parent therefore always exists, and the truncated-dummy-chain case (the early break in `calculate_dummy_blocks`) is rejected rather than silently accepted with a hole in the chain.

The safeNode predicate is then read off the justify block, which the candidate provably extends:

- liveness: `justify_block.height() > locked.height()` — a timeout certificate contributes nothing, since it proves only that its views produced nothing;
- safety: the justify block is, or extends, the locked block.

It is evaluated on the justify block rather than the candidate because the dummy blocks between them are not stored at validation time, so walking back from the candidate cannot reach the locked block.

The second commit fixes the proposer side. It built on `highest_seen_block` whenever that was one height below the proposal. A locally accepted block is not certified, so an equivocating leader can leave replicas holding different blocks at that height; building on one the proposer's own high QC does not certify produces exactly the shape validators now reject. The parent, and the state anchor for every speculative read, is now the justify block or the dummy chain extending it. In the steady state the justify block *is* the highest seen block, so honest proposals are unchanged — the whole `consensus_tests` suite passes without modification.

This is a local rule: no wire format, hash preimage or stored record changes, so it needs no state reset. Binding the high QC into timeout votes and the TC (so replicas can verify the leader used the committee's highest QC) is a separate, stricter design that does change the timeout-vote signature preimage and the persisted `TimeoutCertificate`; it will follow as its own PR against the testnet reset. It is a liveness hardening, not a safety requirement — the lock rule is what closes the fork above.

## Testing

- `crates/consensus_tests/src/extends_justify.rs` (new): accepts a contiguous parent and a full dummy chain; rejects a parent that is not the justify block, a truncated dummy chain, a view gap with no timeout certificate, and a parent that is not the last dummy.
- `crates/consensus_tests/src/safe_node_predicate.rs`: updated for the justify-block predicate, plus a new case asserting that a timeout certificate above the locked block does not satisfy the liveness rule.
- `cargo nextest run -p consensus_tests`: 77 passed locally; CI is authoritative.
- `cargo lints clippy -p tari_consensus -p tari_ootle_storage --all-targets`: clean.


crates/consensus/src/hotstuff/common.rs
```diff
@@ -163,6 +164,83 @@ pub fn calculate_dummy_blocks_from_justify<TAddr: NodeAddressable, TLeaderStrate
...
+            justify_block_height: justify_block.height(),
+            candidate_block_height: candidate_block.height(),
+        });
+    }
+
+    let num_dummies = candidate_block
+        .height()
+        .as_u64()
+        .saturating_sub(justify_block.height().as_u64())
+        .saturating_sub(1);
```
sdbondi, line 198: Unsure but could this give an invalid value? isnt it an invariant that the justify block height < candidate block? no need to add clarifying code comments, only make code changes if you agree.

sdbondi: _Posted by Claude._

It is an invariant, and it is enforced three lines above: `candidate_block.height() <= justify_block.height()` returns `CandidateBlockNotHigherThanJustify`. So the subtraction can never underflow and `saturating_sub` returns the exact difference here, never a clamped one.

I kept it saturating rather than plain arithmetic because this runs on attacker-supplied blocks and the release profile sets `overflow-checks = true` with `panic = 'abort'` — a plain `-` turns any future reordering of that guard into a node abort on a crafted proposal, while the saturating form falls through to the shape checks and rejects the block.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 1 blocking, 2 non-blocking, 2 nits · head `649f6b07`

The core rule is right, and the two halves fit together the way the description says: `check_extends_justify` makes the candidate→justify link total (parent *is* the justify block, or the parent is the last of a full dummy chain proven by a TC), and that is exactly what licenses reading safeNode off the justify block instead of the candidate. `Block::is_safe`'s new `justify_block.id() == locked.block_id()` clause is load-bearing, not defensive — `extends_pending` returns `false` for `self.id() == ancestor` (block.rs:736), so without it a proposal justifying precisely the locked block would fail both rules and the chain would wedge on the first leader that holds exactly `QC(locked)`. Dropping TC height from the liveness rule matches the paper: `qc.viewNumber` is the justify block's view, and a TC certifies no block.

I also checked the surrounding invariants the new rule leans on and they hold: `check_height` (validations/common.rs:152) already forces `height == max_certificate_height + 1`, and `check_proposal_certificate`/`check_timeout_certificate` bound both certificate heights, and all of these run in `check_local_proposal` on both the live-proposal and catch-up paths (on_message_validate.rs:211, :240) before `validate_local_proposed_block` ever sees the block — so `num_dummies` is quorum-bounded, and the removed `ParentNotFound` check is genuinely subsumed (num_dummies == 0 ⇒ parent is the justify block, which was just loaded from the store). `rpc_state_sync/src/manager_old.rs:329` still calls the old one-arg `is_safe`, but that module is commented out of the crate root, so it does not break the build.

The proposer-side change is a strict improvement over what it replaces. With `check_height` in force, the non-timeout `next_height` is always `justify_block.height() + 1`, so `parent_block = justify_block.as_leaf()` (on_propose.rs:372) is the same block as `highest_seen_block` in every honest steady state — and where they *differ* (a sibling at the justify block's height that this node happened to store), the old fallback produced precisely the shape validators now reject.

## Blocking

**1. Two `consensus_tests` regressions on this head, and the description claims the opposite.** The PR says "the whole `consensus_tests` suite passes without modification" and "`cargo nextest run -p consensus_tests`: 75 passed". CI run 35196030746, leg `test (2/3)`, disagrees:

```
FAIL [104.736s] consensus_tests epoch_change::single_shard_epoch_change
FAIL [105.562s] consensus_tests consensus::node_requests_missing_transaction_from_local_leader
```

Both panic at `crates/consensus_tests/src/support/harness.rs:279: Timeout waiting for Hotstuff event`. The last three completed `development` runs are green, and this is the only run on this head, so there is no flake history to lean on either way.

I could not settle flake vs. regression from the log — the runner was badly starved (42 s of wall clock just to spawn the second validator, 63 tests still running when nextest cancelled), which is the classic shape of a timing flake in these harness tests, and no `does not extend justify` / `NotSafeBlock` warning appears in the captured output. But two *liveness* tests going red on a PR that tightens a liveness rule is the exact failure this change would produce if it were wrong, so it needs to be settled rather than assumed. A clean re-run of that leg is enough; if it reproduces, the thing to look at first is that `is_safe` now runs on the TC/dummy path, which previously returned `Ok(ValidBlock::with_dummy_blocks(..))` before the predicate.

## Non-blocking

**2. The post-timeout liveness cost lands now; the mitigation is deferred.** `on_receive_new_view` validates the NEWVIEW's `high_pc` and even syncs when the block is missing, but it never calls `update_highest` on it (on_receive_new_view.rs:88–115 — the only `ProposalCertificate::update_highest` call sites are the vote collector and `block_store.rs:55`). So a leader whose own high QC sits below some replicas' locked block proposes a block those replicas now reject, where before this PR the TC branch waved it through. Recovery does exist — the `last_vote` carried in NEWVIEW aggregates into a QC via `proposal_collector.rs:96-109` — but it is worth confirming that path covers the case the old TC bypass was masking, since the description defers the stricter high-QC-in-TC binding to a later PR. This is correct HotStuff either way; I am flagging the ordering, not disputing the rule.

**3. `on_propose.rs:384-395`: the comment's claim about the epoch-end gate no longer holds in the case this PR is about.** The comment says reading the epoch-boundary checks at `highest_seen_block` "only ever suppresses commands". That is true when `highest_seen_block` is on the candidate's own chain. Now that the state anchor is unconditionally `justify_block` (:397), `highest_seen_block` can be on a branch forked *below* the justify block — which takes an equivocating leader, i.e. exactly this PR's threat model. Then `is_end_of_epoch_in_chain` (:374) and `is_epoch_end()` (:381) can fail to fire while the justify chain does carry the `EndEpoch`, and the proposer emits commands after an epoch end and gets its own block rejected. Either read those two from `justify_block` as well, or narrow the comment to the honest case it actually covers. (`locked_epoch` at :493 and the header's `epoch_hash` at :676 are fine — both are epoch-constant.)

## Nits

**4. `ProposalValidationError::ParentNotFound` (error.rs:165) is now unreachable.** The only constructor was the check deleted from `on_receive_local_proposal.rs`. Rust will not warn on an unused public variant, so it will sit there. Remove it.

**5. `common.rs:186-192` duplicates `check_proposal_certificate`.** `candidate.height() <= justify.height()` is already rejected at validations/common.rs:236 before the block reaches here. Keeping it makes `check_extends_justify` total on its own inputs (and the new unit tests call it directly), which is worth having — a one-line note saying so would stop a future reader deleting it as dead.

## On the tests

`extends_justify.rs` covers the four rejection shapes and both accept shapes, and `rejects_truncated_dummy_chain` does exercise the new length check specifically: parenting `chain[len-2]` makes `calculate_dummy_blocks` break early on the id match, so the parent check alone would pass and only `dummy_blocks.len() != num_dummies` catches it. `unsafe_when_only_the_timeout_certificate_is_above_locked` pins the TC-height removal at height 9 vs. a justify at 1, which is the right shape.

sdbondi:
_Posted by Claude._

Thanks — addressed at `0742b72`. Point by point.

**1 (blocking) — the two `consensus_tests` failures.** I ran both locally on the reviewed head and on the new head; both pass, along with the whole suite (78 tests on the stacked branch, 77 here). The CI evidence points at runner starvation rather than the rule: the log shows ~42 s of wall clock just to spawn the second validator in `node_requests_missing_transaction_from_local_leader`, with 63 tests still running when nextest cancelled, and `harness.rs:279` is a wall-clock deadline. There is also no `does not extend justify` or `NotSafeBlock` warning anywhere in the captured output, which is what a real rejection would have printed before the node stalled. That said, your reasoning for not assuming it is right, so this is settled by CI rather than by me: the push above started a fresh run and I am watching it. If `test (2/3)` reproduces, I will bisect it rather than re-run again.

The description's "75 passed" was accurate for the local run it described; I have updated it to say the suite passes locally and to point at CI for the authoritative result.

**2 (non-blocking) — the deferred high-QC adoption.** Correct, and it is deliberate: the adoption belongs with the rule that makes it necessary, which is #2636 (stacked on this PR, draft until the reset). It does `high_pc.update_highest(tx)` in exactly the place you point at, and rejects a NEWVIEW whose claimed height does not match the certificate it carries. Until then the recovery path you identified — `last_vote` re-aggregating into a QC via `proposal_collector.rs` — is the one that covers it, and the cost of not adopting is a burned view, not a wedge: the next leader's own high certificate is independent of this one's.

**3 (non-blocking) — the epoch-end gate.** Real, and fixed in `1a4bf3e`. `is_end_of_epoch_in_chain` and `is_epoch_end()` now read `justify_block`, which is the chain the candidate actually extends; an EndEpoch on a branch the high certificate does not reach binds nothing. The comment no longer claims the old invariant — it now covers only `locked_epoch` and the header's `epoch_hash`, which are epoch-constant as you say.

**4 (nit) — `ParentNotFound`.** Removed in `0742b72`.

**5 (nit) — the duplicated height check.** Kept, with a one-line note saying the function checks the pair on its own rather than relying on call order, so it does not read as dead.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 1 blocking, 1 non-blocking, 0 nits · 3 resolved since `649f6b07` (0 blocking, 1 non-blocking, 2 nits) · head `0742b725`

## Resolved

- **Epoch-end gate now reads the candidate's own chain** (on_propose.rs:377, :384). `is_end_of_epoch_in_chain` and the `is_epoch_end()` fallback both moved from `highest_seen_block` to `justify_block`. I traced the two directions: a seen-but-unjustified `EndEpoch` above the justify block is on a branch the candidate abandons (dummy-fill path), so ignoring it is right; and in the non-timeout case `check_height` forces the proposer at `h+1` to hold `QC(h)`, so an `EndEpoch` at `h` *is* the justify block and still suppresses. `locked_epoch` (:496) and the header's `epoch_hash` (:679) correctly stay on `highest_seen_block` — both are epoch-constant and `HighestSeenBlock::get` is epoch-scoped, so the new comment's claim holds.
  - One correction to what I wrote last time: I said the old reads could make the proposer "emit commands after an epoch end and get its own block rejected". The rejection half was wrong — `should_not_propose_commands` is proposer-side only; there is no symmetric replica rule keying off the chain, just `EndOfEpochWithOtherCommands` on the `EndEpoch` block itself. The read was still on the wrong chain, so the fix stands on its own.
- **`ParentNotFound` removed** (error.rs). No references remain anywhere in the tree.
- **`common.rs:186` carries the note** that the height pair is checked standalone rather than leaning on the stateless certificate checks having run.

## Still blocking

**1. The two `consensus_tests` failures are not yet answered.** `epoch_change::single_shard_epoch_change` and `consensus::node_requests_missing_transaction_from_local_leader` failed on `649f6b07` with `Timeout waiting for Hotstuff event`. Run 35197591554 is in flight on this head and `test (2/3)` is still pending.

Worth noting it is not a clean re-run of the same code: 1a4bf3e28 changes when the proposer suppresses commands at an epoch boundary, which is squarely in `single_shard_epoch_change`'s path. A green leg leaves it ambiguous between "was a flake" and "the epoch-end read was the cause" — if you want that distinguished, the cheap way is one re-run of the leg at `649f6b07`. A green leg is enough for me to drop this as blocking either way; a red one points at the epoch-change path first.

## Still non-blocking

**2. `on_receive_new_view` validates the NEWVIEW's `high_pc` but never adopts it.** Unchanged at this head (on_receive_new_view.rs:88–115; no `update_highest` call). Recovery from a leader holding a stale high QC rests entirely on the `last_vote` aggregation at proposal_collector.rs:96-109 now that the TC branch no longer bypasses safeNode. Reasonable to leave for the high-QC-in-TC PR, but it is the one behaviour this change makes strictly stricter before its mitigation lands.

sdbondi:
_Posted by Claude._

**2 (non-blocking) — moved into this PR, `3400f788`.** You are right that the strictness and its mitigation should not be separated, and on reflection the adoption does not belong in the stacked PR at all: the split criterion for these two PRs is "does it need a state reset", and `high_pc.update_highest(tx)` needs no wire, preimage or stored-record change. `on_receive_new_view` now adopts the certificate it already validated, so a leader takes the committee's high certificate instead of proposing from behind it. #2636 keeps only the part that genuinely needs the reset — the claim-vs-certificate check, which cannot exist before the timeout vote carries a claimed height.

Worth noting the adopted certificate advances `HighPc` without running the lock/commit recursion in `block_store.rs:36`, so `HighPc` can sit above the locked block until a proposal carrying that chain arrives. That is not a new state: `proposal_collector.rs:109` already calls `update_highest` on a QC it forms from votes, with the same effect. It is permissive in the safe direction — our own lock lags, so `is_safe` rejects less, never more.

**1 (blocking) — status.** I tried your suggestion first: `gh run rerun 35196030746 --failed` at `649f6b07`. It cannot be had — the workflow's concurrency group cancels a run for a ref that has a newer run, so the re-run was cancelled within seconds, twice. Getting a CI leg at that exact commit would mean pushing it to a scratch branch, which I have not done.

What I do have at `649f6b07` is a local full-suite run: 75/75, both named tests included. That is the run the description's number came from, so it does distinguish the two hypotheses you raised — the epoch-end read in `1a4bf3e2` cannot be what fixed them, because they passed at the commit before it. It does not rule out a CI-environment-only regression, which is why the in-flight run on `3400f788` is still the thing that settles it.

Correction to my previous comment: I wrote "77 here" for this branch's local suite count. It is 75 — I quoted the stacked branch's count minus its new tests instead of measuring. 75/75 at `649f6b07`, `0742b725` and `3400f788`; 78/78 on #2636.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 1 blocking, 0 non-blocking, 1 nit · 1 resolved since `0742b725` (0 blocking, 1 non-blocking, 0 nits) · head `3400f788`

## Resolved

**The NEWVIEW high certificate is now adopted** (on_receive_new_view.rs:115). I checked the ordering and the guards rather than the commit message:

- `validate_qc` (epoch + quorum signatures) and the `Block::record_exists` check both run before `update_highest`, so the adopted certificate is quorum-signed and its block is present — which matters, because `update_highest` does `self.as_leaf_block().set(tx)` and that call is documented to fail when the block is missing.
- The zero-block signature exemption in `check_quorum_certificate_signatures` cannot be leveraged here: such a certificate is at height 0, so it either loses the `local_high_qc.block_height > high_pc.height()` guard or hits `update_highest`'s `>=` early return.
- `LeafBlock` only moves forward. The write path in `update_highest` runs only when the incoming height is strictly greater than the local one, and `LeafBlock` already tracks the high certificate in lockstep (block_store.rs:55), so this keeps the existing invariant rather than bending it.
- Adoption happens before the leader check and before `timeout_vote_collector.check_and_collect_vote`, so every NEWVIEW that contributes to the TC has already contributed its certificate. By the time the quorum-th message forms the TC, the leader holds the highest certificate any sender reported — which is what closes the gap the stricter safeNode opened.

Moving the block to `with_write_tx` is fine: the `Ok(false)` paths commit an empty transaction and the `FallenBehind` path rolls back with nothing written.

## Still blocking

**1. The two `consensus_tests` failures from `649f6b07` are still unanswered.** The run on `0742b725` was cancelled by this push before `test (2/3)` finished, so there is still no green (or red) consensus-test leg on any head of this branch. `epoch_change::single_shard_epoch_change` and `consensus::node_requests_missing_transaction_from_local_leader` both timed out waiting for a Hotstuff event on the first head, and `development` is green. This is the last thing standing between the change and an approval from me — everything else I raised is closed.

## Nit

**2. Every NEWVIEW now takes a write transaction, and most of them write.** When the incoming certificate is not higher, `update_highest` falls into its "not new" branch and calls `tx.last_sent_new_view_clear()` — so a leader collecting `n` NEWVIEWs per timeout now performs `n` clears, on the path that is already degraded. No behavioural effect today: the persisted `LastSentNewView` is written at on_next_sync_view.rs:115 and never read back (the resend guard reads the in-memory `self.last_sent_new_view`). Worth knowing, though, because if that persisted record is ever made authoritative — which is what it looks like it is for — clearing it on every inbound NEWVIEW would defeat it. Guarding the adoption on `high_pc.height() > local_high_qc.block_height` before opening the write transaction would avoid both.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 0 non-blocking, 1 nit · 1 resolved since `3400f788` (1 blocking, 0 non-blocking, 0 nits) · head `3400f788`

This would be an approval; GitHub does not accept one from the author's own account, so it stands as a comment.

## Resolved

**The consensus-test failures were runner starvation, not this change.** `test (2/3)` on this head is green — 781 tests run, 781 passed — and both tests I held this on ran and passed:

```
PASS [19.760s] consensus_tests consensus::node_requests_missing_transaction_from_local_leader
PASS [20.140s] consensus_tests epoch_change::single_shard_epoch_change
```

19.8 s and 20.1 s against the 104 s and 105 s they took before timing out on `649f6b07`, which matches what the red run's log showed (42 s of wall clock just to spawn a validator, 63 tests still in flight at cancellation). Every leg on this head is green, integration test included.

## What I verified across the three heads

- `check_extends_justify` makes the candidate→justify link total, and that is what licenses reading safeNode off the justify block. The `justify_block.id() == locked.block_id()` clause is load-bearing — `extends_pending` returns false for `self == ancestor`, so without it the chain wedges on the first leader holding exactly `QC(locked)`.
- The upstream invariants the rule leans on hold: `check_height` forces `height == max_certificate_height + 1`, both certificate checks bound their heights, and all of them run in `check_local_proposal` on the live-proposal *and* catch-up paths before `validate_local_proposed_block` sees the block — so `num_dummies` is quorum-bounded and the deleted `ParentNotFound` check is genuinely subsumed.
- The proposer-side anchor is a strict improvement: with `check_height` in force the non-timeout `next_height` is always `justify_block.height() + 1`, so the new parent is the same block as `highest_seen_block` in every honest case, and where they diverge the old fallback produced exactly the shape validators now reject.
- The epoch-end gate reads the candidate's own chain in both directions (abandoned-branch `EndEpoch` ignored on the dummy-fill path; a justified `EndEpoch` still suppresses).
- NEWVIEW certificate adoption is correctly ordered behind signature validation and the block-existence check, cannot be reached by the zero-block signature exemption, moves `LeafBlock` forward only, and lands before timeout-vote collection — so the leader holds the highest reported certificate by the time the TC forms.
- `extends_justify.rs` covers both accept shapes and four rejection shapes, and `rejects_truncated_dummy_chain` genuinely exercises the length check rather than the parent check.

## Nit carried forward (not worth holding for)

Every inbound NEWVIEW now opens a write transaction, and the common "not higher" case writes — `update_highest` clears `last_sent_new_view` on that branch. No effect today, since the persisted `LastSentNewView` is written at on_next_sync_view.rs:115 and never read back, but it would defeat that record if it is ever made authoritative. A height comparison before opening the transaction avoids both.

sdbondi:
_Posted by Claude._

The nit is already fixed at the current head, `1175d075` — it was pushed a few minutes before that review landed, so it reviewed `3400f788`.

`on_receive_new_view` is back on a read transaction: it returns our own high certificate height alongside the validity result, and the write transaction opens only when the reported certificate is strictly higher. The equal-height case — every sender of a view in the common case — now writes nothing, so `last_sent_new_view_clear()` is not called on the inbound path at all and the persisted `LastSentNewView` stays available if it is ever made authoritative.

`cargo nextest run -p consensus_tests`: 75/75 locally at that head. CI is re-running on it; the green run you cited was `3400f788`, which differs only in this guard.

sdbondi (commented):
Makes sense

PR #2636, opened 2026-09-17
feat(consensus)!: bind the committee's high certificate into the timeout certificate

sdbondi:
**Draft: needs a state reset. Do not merge before the testnet reset.** Rebased onto `development` now that #2635 has merged, so this is a single commit on top of it.

## Why

#2635 closes the fork: a candidate must extend its justify block, and the safeNode predicate is read off that block, so the lock rule alone keeps the committee on one chain. What it deliberately does not do is stop a leader from *rewinding*: after a timeout, the next leader may justify from any certificate above the locked block, orphaning certified blocks above it. Replicas cannot distinguish that from a leader that has simply fallen behind, because a `TimeoutCertificate` says only that a view failed — nothing about how far the committee had got.

This is the stricter design: make the timeout certificate carry that information, so replicas can hold the leader to it.

## What

- `TimeoutVote` signs `(epoch, height, high_pc_height)` — the height of the certificate the sender held when the view failed.
- `TimeoutCertificate` carries `Vec<SignedTimeout>` (the signature plus the height it attests to) instead of bare signatures, and exposes `max_high_pc_height()`. Signature verification reconstructs each signer's own message, so a quorum cannot be assembled that hides one of the reported heights.
- `check_justify_reaches_timeout_certificate`: a block carrying a timeout certificate must justify from a certificate at least as high as the highest one attested to.
- A claimed height is only worth the certificate behind it, and only the collector of a NEWVIEW sees that certificate. `on_receive_new_view` therefore rejects a vote whose claim does not match the `high_pc` the message carries — otherwise `f` validators could attest to a height nobody can reach and wedge the committee. (Adopting that certificate needs no reset, so it now lands in #2635 instead.)

Note what this does *not* buy: it is a liveness hardening, not a safety fix. Safety is closed by #2635's lock rule. This stops a byzantine leader from using a leader failure to discard certified work, and stops an honest leader from burning a view by proposing on a stale certificate.

## Why it needs a reset

- The timeout vote's signature preimage changes, so votes and certificates signed under the old shape do not verify.
- `TimeoutCertificate`'s CBOR encoding and therefore its id change. That breaks two stores, not one: `TimeoutCertificateCf` (referenced by `HighTc`), written under `VersionedTimeoutCertificate::V1`, and every stored block, since `Block` embeds `Option<TimeoutCertificate>` at CBOR field 3 — `BlockCf` is the larger half. Decoding pre-upgrade records would need a `V2` variant and an upgrade step for both, which is the backward compatibility we are explicitly not engineering.

This PR on its own leaves block ids unaffected: the header preimage commits the justify id and the parent, never the timeout certificate. That is itself the gap: `max_high_pc_height` becomes a validity input that the proposer's signature does not cover. #2642, stacked on this PR, closes it by committing the header to the timeout certificate id under V1, which does change block ids and rides the same reset.

## Testing

- `crates/consensus_tests/src/timeout_certificate.rs` (new): a justify at the attested height is accepted, one below is rejected, and a block with no timeout certificate is unaffected.
- `cargo nextest run -p consensus_tests`: 78 passed — the leader-failure and dummy-block suites drive the new timeout vote and certificate end to end.
- `cargo lints clippy -p tari_consensus -p tari_consensus_types -p tari_ootle_p2p -p tari_state_store_rocksdb --all-targets`: clean.
- TS bindings hand-updated: `SignedTimeout.ts`, `TimeoutCertificate.ts`, `index.ts`.


sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 4 non-blocking, 2 nits · head `ce382d1d`

The rule is sound and the invariant that keeps it live holds. I traced the two properties that matter:

**The bound is real.** A leader can only *lower* `max_high_pc_height` by excluding signers, never raise it — the height is inside each signer's own preimage (timeout_vote.rs:96, reconstructed per-signer at common.rs:364). With `quorum_threshold() = n - f` (committee.rs:66), if a quorum holds `QC_H` then at most `f` do not, so any quorum of `n - f` contains at least `n - 2f ≥ f + 1` who do. `max_high_pc_height ≥ H` survives whatever subset a byzantine leader picks. The property the PR claims is the one the check enforces.

**It cannot deadlock the proposer.** The rule needs `HighPc.height ≥ HighTc.max_high_pc_height` to hold at all times, and both TC-adoption paths establish it. Locally built TCs: `on_receive_new_view` adopts each reported certificate (on_receive_new_view.rs:140) *before* `timeout_vote_collector.check_and_collect_vote`, votes are keyed by `(epoch, height)` with `AggregationKey = ()` so every collected vote passed that adoption, and `HighPc` only rises within an epoch. TCs arriving inside a block: `tc.update_highest` and `update_nodes`' `justify().update_highest` run in the same write transaction (on_ready_to_vote_on_local_block.rs:128), so `HighPc ≥ block.justify().height() ≥ max`. There is no third path — I checked every `timeout_certificate()` call site.

The leader-side claim check (on_receive_new_view.rs:93) is what makes the whole thing work, and it is correctly placed: an honest sender's claim and carried certificate come from one read of `HighPc` in a single transaction (on_next_sync_view.rs:78-88), so it never drops an honest vote, and a forged claim needs a quorum-signed certificate at that height, which cannot be fabricated. A byzantine leader that skips the check only poisons a TC it then cannot propose on — and that TC reaches no honest store, because both adoption paths are gated on validation.

Also checked and sound: the signature-verification restructure preserves dup and vote-power accounting on every signature in both arms, and the zero-block exemption still cannot apply to a TC (`as_proposal_certificate()` is `None`). `check_timeout_certificate` verifies the quorum *before* trusting the attested height. Commit proofs are untouched — they read `ProposalCertificate::signatures()` only (commit_proofs.rs:218), so the base-layer verifier is unaffected. Proto, conversions and the hand-written bindings are consistent with the Rust shapes.

## Non-blocking

**1. Block validity now depends on data the block's signature does not cover.** The header preimage commits `justify_id`, height and parent but never the timeout certificate (block_header.rs:339-386), and the proposer signs only the block id (block_header.rs:546) — which is exactly what makes the PR's "block ids are unaffected" claim true. Before this change the TC only fed height and dummy-chain arithmetic, both cross-checked against the header's own height. Now `max_high_pc_height` gates acceptance, and nothing binds it to the block.

I do not think it is exploitable today: dropping signers only lowers the max and relaxes the rule, and raising it needs a valid quorum-signed TC at the same height with a higher max, which only that view's leader could assemble — and that leader is the proposer. But it is a new dependency of validity on unsigned data in a consensus rule, and committing the TC id into the header preimage is a reset-gated change that is free to make while a reset is already on the table and expensive afterwards. Worth a decision on the record rather than leaving it implicit.

**2. The reset note understates the storage break.** It names `TimeoutCertificateCf` and `HighTc`, but `Block` embeds `Option<TimeoutCertificate>` at CBOR field 3 (storage/src/consensus_models/block.rs:105), so every stored block carrying a TC also fails to decode. The conclusion is unchanged — reset required — but the PR body is the record of what the reset is for, and `BlockCf` is the bigger half of it.

**3. `check_justify_reaches_timeout_certificate` asserts a property it does not establish** (common.rs:271-277). Its doc says "a quorum signs the height … so the committee has provably reached that height", but the function never verifies a signature or a quorum — `check_timeout_certificate` does that first, at common.rs:265. It is now `pub` at the crate root, where that ordering is invisible; the new test calls it directly with a TC of three zero signatures, which is precisely the shape it cannot detect. Either keep it crate-private and drive the test through `check_timeout_certificate`, or state the precondition in the doc. Relatedly, the comment above the re-export (lib.rs:12) still reads "Re-export the QC signature check for recovery probes outside the consensus crate" while now covering two items, only one of which has a consumer outside the crate.

**4. A leader that is ahead can still fail to build a TC, and this PR is what makes that fixable.** `local_high_qc.block_height > high_pc.height()` drops the whole NEWVIEW (on_receive_new_view.rs:104-131), timeout vote included, so a leader ahead of more than `f` replicas never reaches a quorum of timeout votes. That guard predates this PR, but this PR is what makes those votes provably harmless: a sender reporting a *lower* certificate cannot raise `max_high_pc_height`, so including its vote costs nothing and it is exactly the vote the quorum needs. Collecting the timeout vote independently of whether the reported certificate is worth adopting would close a real liveness hole at no cost to the new invariant.

## Nits

**5. `on_next_sync_view.rs:90` signs one epoch and ships another.** The signed `TimeoutVoteMessage` takes `epoch: high_pc.epoch()`, the `TimeoutVote` carries `epoch`, and verification reconstructs from `tc.epoch()`. They are equal by construction — `HighPc::get(tx, epoch)` is epoch-scoped — so this works, but signing a different field from the one shipped is fragile, and this PR adds a third field to that same preimage. Cheap to align while touching it.

**6. proto field 3 is reused with a changed type** (consensus.proto:187): `repeated SignatureAndPublicKey signatures` → `repeated SignedTimeout timeouts`. Both are length-delimited, so an old decoder mis-parses rather than failing cleanly. Irrelevant under a coordinated reset, but `reserved 3;` plus a new tag costs nothing and makes the break loud if a stale binary ever meets the new wire.

## Tests

`timeout_certificate.rs` covers accept-at-the-attested-height, reject-below, and no-TC, and `block_justifying` mirrors `check_height` by setting `height = tc.height() + 1`. What is not covered is the leader-side claim check at on_receive_new_view.rs:93 — the one thing standing between a byzantine validator and a TC that no leader can satisfy. The PR body argues that case explicitly; a test for it would be worth more than the three that are here.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 4 non-blocking, 2 nits · 0 resolved since `ce382d1d` · head `1e016a77`

The force-push is the one missed TS consumer of the renamed field (BlockDetails.tsx:334, `timeout_certificate.signatures` → `.timeouts`), which is what failed the `web ui build` leg at the previous head. I swept the rest of the TypeScript for the same drift: the only other `.signatures` reads are `justify.signatures` (still a `ProposalCertificate`, unchanged) and a transaction body's, so that is the last one.

Nothing else changed — the Rust diff is byte-identical to `ce382d1d`. All four non-blocking items and both nits from the previous review stand as written; the reservation is item 1, block validity now turning on a timeout certificate that the header preimage and the proposer's signature do not cover.

sdbondi:
_Posted by Claude._

Head is now `1e46f773`. Also fixed the `web ui build` failure that was red when you reviewed: `BlockDetails.tsx:334` read `timeout_certificate.signatures.length`, now `timeouts.length` — the only TS reference to the renamed field in the tree.

**1 — validity depending on unsigned data.** Agreed on the premise, and the header preimage is not the place to fix it. `BlockHeader::calculate_hash` carries a comment that its field selection must stay identical to `tari_sidechain::SidechainBlockHeader::calculate_hash`, which the base layer uses to verify a commit proof against the block id a committee signed — so committing the TC id is a change in another repo's crate, not a local one, and not free even during a reset.

It is also not needed for this rule, for a reason stronger than "not exploitable today": the bound is invariant under any valid substitution. Whoever relays the block can only attach a certificate that passes `check_quorum_certificate_signatures`, i.e. one holding ≥ `n - f` signers, and your own intersection argument then puts ≥ `f + 1` holders of `QC_H` in it — so every valid TC for that view reports `max ≥ H`. Pruning lowers nothing below the true committee maximum; substituting another view's TC breaks `check_height`, which pins `height == max_certificate_height + 1`; attaching a TC at a height ≤ the justify height is a no-op, since a validator that timed out at view `h` cannot hold a certificate at or above `h`. Stripping the TC does not help either: with a view gap it fails `check_extends_justify`, and without one the block was already being judged on its justify alone.

**2 — the reset note.** Fixed. `Block` embeds `Option<TimeoutCertificate>` at CBOR field 3, so `BlockCf` breaks too, and it is the larger half. The body now says so.

**3 — the doc asserted what the function does not check.** Fixed: the doc now states the precondition and names `check_timeout_certificate` as what establishes it. Kept `pub` — the test drives it directly and that is the point of the export — and the stale `lib.rs` comment now covers both re-exports.

**4 — the NEWVIEW dropped over a lower certificate.** Real, and I agree it costs a view for no benefit. I am not fixing it here: it needs no wire, preimage or stored-record change, so by the same criterion that split this PR from #2635 it belongs on `development` rather than behind the reset. Separating the two decisions in `on_receive_new_view` — "is this certificate worth adopting" and "is this timeout vote worth collecting" — is a small change and I will raise it as its own PR.

**5 — signing one epoch and shipping another.** Fixed, both are `epoch` now.

**6 — proto tag reuse.** Fixed: `reserved 3;` with the old shape named in a comment, `timeouts` moved to tag 4.

**Tests.** Added one for the claim check, which meant giving it a name: `NewViewMessage::timeout_claim_matches_high_pc`. The handler calls it before destructuring the message, and the test pins both directions. Note what it does not prove — that the handler still calls it — because the test harness's `MessageFilter` returns `bool` and can only drop a message, never rewrite one, so a tampered NEWVIEW cannot be injected. Making that filter able to rewrite messages would unlock byzantine-proposal tests for this rule and for #2635's, and is worth doing on its own.

`cargo nextest run -p consensus_tests`: 79/79.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 2 non-blocking, 0 nits · 4 resolved since `1e016a77` (0 blocking, 2 non-blocking, 2 nits) · head `1e46f773`

## Resolved

- **The reset note now names `BlockCf`.** It calls out the `Option<TimeoutCertificate>` at CBOR field 3 and says which of the two stores is the larger half. That is the record I wanted it to be.
- **`check_justify_reaches_timeout_certificate` states its precondition** (common.rs:271-278). The doc now says the attested heights carry weight only after `check_timeout_certificate` has verified the signatures, and names the caller that does it. That closes the gap between what the doc asserted and what the function establishes; leaving it `pub` is fine with the precondition written down, and the lib.rs comment now accounts for both exports.
- **`on_next_sync_view.rs:90` signs the epoch it ships.** `TimeoutVoteMessage` now takes `epoch`, which is what the `TimeoutVote` carries and what verification reconstructs from `tc.epoch()` (the collector builds the certificate with `current_epoch`, and `validate_vote` rejects a vote whose epoch differs). Signed, shipped and verified are the same value on all three paths now.
- **proto field 3 is reserved** and `timeouts` moved to 4 (consensus.proto:187-188). `reserved` is legal anywhere in a proto3 message body, the Rust field name is unchanged so `conversions/consensus.rs` needs nothing, and prost generates at build time (crates/p2p/build.rs) with no checked-in output to regenerate.

Also good: the claim check moved into `NewViewMessage::timeout_claim_matches_high_pc()` (new_view.rs:32-38) and runs as the first thing in `handle`. It is a pure predicate on the message, so hoisting it above the epoch and view-height guards changes nothing about ordering, and it gives the rule a testable seam — which the new `a_timeout_claim_above_the_carried_certificate_is_not_matched` case uses. That was the test gap I flagged; it is covered.

## Still open

**1. Block validity turns on a timeout certificate that nothing binds to the block.** Unchanged. The header preimage commits `justify_id`, height and parent but not the timeout certificate (block_header.rs:339-386), and the proposer signs only the block id (block_header.rs:546) — which is exactly why "block ids are unaffected" holds. `max_high_pc_height` now gates acceptance on a value outside that preimage. Not exploitable as far as I can see, since lowering the max only relaxes the rule and raising it needs a quorum-signed certificate at the same height that only the view's leader could assemble. Still the one thing I would want decided explicitly rather than left implicit, and committing the TC id into the header is free while a reset is already on the table.

**2. A leader that is ahead can still fail to build a timeout certificate.** Unchanged. `local_high_qc.block_height > high_pc.height()` drops the whole NEWVIEW (on_receive_new_view.rs:99-126), timeout vote included, so a leader ahead of more than `f` replicas never reaches a quorum. The guard predates this PR, but this PR is what makes those votes provably harmless to include: a lower-reported certificate cannot raise `max_high_pc_height`, and that vote is exactly what the quorum is short of. Collecting the timeout vote independently of whether the reported certificate is worth adopting costs the new invariant nothing.

sdbondi:
_Posted by Claude._

Both remaining items were answered in [the previous comment](https://github.com/tari-project/tari-ootle/pull/2636#issuecomment-5712595541), which the re-review at `1e46f773` does not appear to have read. Restating the two conclusions so they are not lost:

**1 — committing the TC id into the header is not free.** `BlockHeader::calculate_hash` carries a comment that its field selection must stay identical to `tari_sidechain::SidechainBlockHeader::calculate_hash`, which the base layer uses to verify a commit proof against the block id a committee signed. Changing the preimage is a change in another repo's crate, coordinated with the base layer, not a local one a reset covers.

It is also not required for this rule, because the bound is invariant under any valid substitution: an attached certificate must pass `check_quorum_certificate_signatures`, so it holds ≥ `n - f` signers, so ≥ `f + 1` of them hold `QC_H` and report `max ≥ H`. Pruning cannot go below the true maximum; another view's certificate fails `check_height`; an older-view certificate is a no-op, since a validator that timed out at view `h` holds nothing at or above `h`; and stripping it either fails `check_extends_justify` or changes nothing.

**2 — raised as #2638**, against `development`, as said. It needs no wire, preimage or stored-record change, so it ships without waiting for the reset. #2636 will want a trivial rebase once it merges, since both touch `on_receive_new_view`.

sdbondi (commented):
Reviewed and approved

PR #2637, opened 2026-09-17
feat(engine)!: classify execution failures with a coarse reason code

sdbondi:
## What

`RejectReason::ExecutionFailure` carried a single free-text string. A template panic, an access
denial, an out-of-gas trap and a blown engine limit all reached a consumer as one undifferentiated
case, so an application could only string-match to decide whether to retry, raise the fee, re-quote,
fix its own call, or tell the user they were refused.

The variant now carries a coarse `ExecutionFailureCode` alongside the message:

```rust
pub enum RejectReason {
    ExecutionFailure { code: ExecutionFailureCode, message: String },
    // ...
}
```

Thirteen codes, each earning its place by changing what an application does: `TemplateError`,
`AccessDenied`, `AssertionFailed`, `InsufficientFunds`, `OutOfCompute`, `LimitExceeded`,
`InvalidArgument`, `NotFound`, `DanglingResources`, `ResourceRestricted`, `InvalidProof`,
`EngineInvariant`, `Unclassified`.

## How it stays honest

Every engine error type gains an exhaustive `failure_code()` with **no wildcard arm** —
`RuntimeError` (106 variants), `WasmExecutionError`, `WasmValidationError`, `TransactionErrorKind`,
`TemplateLoaderError`, `ResourceError`, `AssertError`, `LimitError`, `TransactionCommitError`. A new
error variant has to be classified before it compiles, which is the only thing that stops the
taxonomy decaying back into `Unclassified`.

`Unclassified` stays as a fallback for a variant that slips through, with a `warn!` behind it. Note
that with every `failure_code()` exhaustive it is currently unreachable from engine code — the only
producer is `execution_failure_unclassified`, which engine code must not call. It is defence, not a
telemetry source.

## Two gaps found while wiring it up

Both were cases where the cause was already being discarded before it could be classified, and both
are fixed here because leaving them would have shipped the wrong answer for the two most common
failures an application sees.

1. **Cross-template calls rendered the callee's error to a string.** An access denial raised inside
   a call — the ordinary path for an account withdraw through an auth hook — surfaced as a template
   fault. `CrossTemplateCall{Function,Method}Error` now carry the callee's error whole, the way
   `SpendScriptRejected` already did, and delegate classification to it.
2. **A hard-cap out-of-gas trap was indistinguishable from a template panic.** When a fixed cap
   (rather than a fee allowance) bound the call, the trap fell through as a bare
   `wasmer::RuntimeError: unreachable`. It now reports `MaxComputeExceeded` and classifies as
   `LimitExceeded`, not `OutOfCompute` — no fee raises a hard cap, so reporting it as an
   underpayment would send callers down a dead end. `compute_exceeded_error` is total as a result.

The same test applies to every compute failure, so `OutOfCompute` is reserved for the shortfalls a
larger fee actually clears (`InsufficientFeesForCompute`, `InsufficientFeesForNativeExecution`).
`MaxNativeExecutionPointsExceeded` and both `FeeIntentComputeExceeded` variants are flat ceilings by
their own error messages — *"does not rise with the fee paid"* — so they classify as `LimitExceeded`
alongside `MaxComputeExceeded`.

## Boundary

`ootle_sdk_core` exposes the code as a `failure_code` field next to `abort_code`, on the same
never-rename-only-add contract, with the full code set pinned by tests:

```json
{ "code": "EXECUTION_FAILURE", "failure_code": "ACCESS_DENIED", "message": "..." }
```

A new golden vector (`parse_finalized_result/reject_access_denied`) pins both the wire shape and the
boundary shape. TS bindings hand-updated (`ExecutionFailureCode.ts`, `RejectReason.ts`, `helpers.ts`).

## Breaking

The wire shape of `RejectReason` changes, as does the rendered message of a hard-cap out-of-gas
failure. Both are consensus-visible, so this rides the testnet reset — no compatibility shim.

## Known coarseness

An argument the template's generated ABI shim refuses to decode arrives as `TemplateError`, not
`InvalidArgument`: the decode failure is a template-side panic and the engine cannot tell it from any
other panic without a distinct ABI error. `fees::fail_pay_negative_fee` is the case in point. Worth a
follow-up if it bites; it needs a new `WasmExecutionError` variant and template-lib cooperation.

## Testing

- `cargo nextest run -p tari_engine` across `access_rules`, `cross_template`, `fees`,
  `metering_budget`, `zk_verify`, `compute_fee_budget`, `test`, `wasm_loader`, `publish_template`,
  `template_upgrade` — all green.
- `cargo nextest run -p ootle_sdk_core -p tari_engine_types` — 380/380, golden vectors regenerated
  and idempotent.
- `cargo lints clippy --all-targets` on the touched crates — clean.
- `cargo check --all-targets` on `tari_ootle_walletd`, `transaction_generator`,
  `tari_ootle_wallet_sdk_tests`, `consensus_tests`, `integration_tests`.

`assert_access_denied_for_action` and `assert_insufficient_funds_for_action` now assert the code as
well as the message, so the whole `access_rules` suite covers the classification for free. The TODO
in `assert_error.rs` asking for engine-side enumeration is resolved.


sdbondi (commented):
_Posted by Claude._

**Changes requested** — 3 blocking, 2 non-blocking, 1 nit · head `0b013e65`

The shape of the change is right: exhaustive `failure_code()` with no wildcard arms is the only thing that keeps a taxonomy like this honest, the cross-template delegation is a genuine fix (`call_method`/`call_function` return `TransactionErrorKind`, so boxing the callee's error whole type-checks and leaves the rendered message identical), `compute_exceeded_error` is correctly total now, and `failure_code` on the boundary is additive and provably disjoint from `abort_code`. Three things stand.

## Blocking

**1. `crates/engine/src/runtime/error.rs:551` — the crate does not compile.**

```
error[E0599]: no variant named `TransientValueInComponentState` found for enum `runtime::error::RuntimeError`
```

The variant is `TransientValueInSubstate` (declared at `crates/engine/src/runtime/error.rs:283`, constructed at :400). `failure_code` names `TransientValueInComponentState`, which does not exist, so `tari_engine` fails to build and `test (2/3)` and `integration test` are both red on it. This also means the Testing section's "all green" claim cannot describe this head — please re-run those suites after the fix rather than carrying the old results forward.

**2. `bindings/src/helpers/helpers.ts:113` — the web UI build is broken.**

```
src/helpers/helpers.ts(113,57): error TS2339: Property 'code' does not exist on type 'unknown'.
src/helpers/helpers.ts(113,91): error TS2339: Property 'message' does not exist on type 'unknown'.
```

`rejectReasonToString` has two `"ExecutionFailure" in reason` branches (lines 93 and 111). The second is unreachable — by then TS has narrowed `ExecutionFailure` out of the union, so `reason.ExecutionFailure` is `unknown` and the new property access fails to type. The old code got away with interpolating it; the new one cannot. Delete the second branch; it was dead before this PR too.

**3. `crates/engine/src/runtime/error.rs:549` — `MaxNativeExecutionPointsExceeded` is classified `OutOfCompute`, which is the exact dead end gap #2 exists to close.**

Its own message is *"Native verification points {consumed_points} exceed the per-transaction maximum {max_points}. Split the work across transactions."* (`runtime/error.rs:274-278`), and the comment at :429-432 says outright that it is a flat ceiling a larger fee does not raise. That is the same thing `WasmExecutionError::MaxComputeExceeded` is, and this PR deliberately gives that one `LimitExceeded` on the reasoning that reporting a hard cap as an underpayment sends the caller to raise a fee that cannot help. The native counterpart should be `LimitExceeded` for the identical reason. The code is consensus-visible, so this is cheap now and a wire break later.

(`RuntimeError::FeeIntentComputeExceeded` staying `OutOfCompute` is fine — a larger fee intent does raise that credit.)

## Non-blocking

**`crates/engine/src/wasm/error.rs:26-29` — the `MaxComputeExceeded` wording over-claims which cap bound the call.** `binding_allowance` is `None` whenever the allowance is not the binding constraint (`process.rs:218-220`), which includes the case where the *per-call* cap `MAX_WASM_POINTS_PER_CALL` binds rather than `MAX_WASM_POINTS_PER_TRANSACTION`. Both are 250_000_000 today so it is unobservable, but the message says "the maximum compute a transaction may use" unconditionally and would misattribute the moment those constants diverge. The classification stays right either way; only the prose is at risk.

**`crates/engine/src/runtime/error.rs` — `TransactionCommitError::IdProviderError(_) => LimitExceeded` collapses two unlike variants.** `IdProviderError::MaxIdsExceeded` is a limit; `IdProviderError::LockingError` is an engine bug and belongs in `EngineInvariant`. Delegating to an `IdProviderError::failure_code()` would keep the "no wildcard" property one level deeper.

## Nit

**The description's "the engine logs a warning whenever it produces one" does not hold.** With every `failure_code()` exhaustive and no wildcard arms, `Unclassified` is unreachable from engine code — the only source is `execution_failure_unclassified`, whose doc says engine code must not call it. Both `warn!` sites are dead as written. Worth keeping as defence, but the "prioritise the tail from real traffic" story should not be in the PR body, because there will be no such traffic.

**`bindings/src/index.ts:5`** inserts `ExecutionFailureCode` between `AbortReason` and `AbridgedTransactionKernel`, out of the file's alphabetical order.

sdbondi (commented):
_Posted by Claude._

**Changes requested** — 2 blocking, 2 non-blocking, 1 nit · 1 resolved since `0b013e65` (1 blocking) · head `0b973e8a`

The push is a one-line change to `crates/engine/src/runtime/error.rs` and nothing else.

## Resolved

- **Blocking 1 — compile error.** `TransientValueInComponentState` → `TransientValueInSubstate` at `runtime/error.rs:551`. Verified against the variant declaration at :283; the arm now names a real variant and nothing else in the match moved.

## Still standing

- **Blocking 2 — `bindings/src/helpers/helpers.ts:113`.** Untouched by this push; the duplicate unreachable `ExecutionFailure` branch is still there, so the `web ui build` leg will fail on the same two TS2339 errors.
- **Blocking 3 — `runtime/error.rs:549`.** `MaxNativeExecutionPointsExceeded` still classifies as `OutOfCompute`. Unchanged reasoning: it is a hard per-transaction ceiling that no fee raises, same as the `MaxComputeExceeded` case this PR fixes, and the code is consensus-visible.
- Both non-blocking items and the nit are unchanged.

sdbondi:
_Posted by Claude._

All six addressed in `56366c4e`. Blocker 1 was fixed in `0b973e8a`; the rest are here.

**Blocking 2 — `helpers.ts`.** Second branch deleted. Verified with `tsc --project tsconfig.prod.json --noEmit`, exit 0.

**Blocking 3 — `MaxNativeExecutionPointsExceeded` → `LimitExceeded`.** Agreed, and the same test applies one variant further than the review allows for.

The review's parenthetical says `FeeIntentComputeExceeded` can stay `OutOfCompute` because "a larger fee intent does raise that credit". Both variants' own messages say the opposite:

> The credit is a flat allowance for sourcing a fee and **does not rise with the fee paid** — move this work to the main instructions.

(`runtime/error.rs:256-258`, `wasm/error.rs:36-39`.) So by the review's own principle — does paying more fix it? — the credit belongs with the ceilings. It moves to `LimitExceeded` too, which leaves a clean line:

- `OutOfCompute` — `InsufficientFeesForCompute`, `InsufficientFeesForNativeExecution`, `InsufficientFeesPaid`. Raise the fee.
- `LimitExceeded` — `MaxComputeExceeded`, `MaxNativeExecutionPointsExceeded`, both `FeeIntentComputeExceeded`. Move the work; no fee helps.

Two test assertions moved with it (`zk_verify::verify_does_not_fit_the_fee_intent_credit`, `compute_fee_budget::assert_fee_intent_credit_exceeded`).

**Non-blocking — `MaxComputeExceeded` wording.** Correct, it cannot know which cap bound the call. Now *"Exceeded a fixed compute maximum … whichever cap bound the call, per call or per transaction"*, with the stale per-transaction claim dropped from the `test_errors_on_infinite_loop` comment too.

**Non-blocking — `IdProviderError`.** Now delegates to `IdProviderError::failure_code`: `MaxIdsExceeded` → `LimitExceeded`, `LockingError` → `EngineInvariant`.

**Nits.** Both taken. The `Unclassified` claim is removed from the description — you are right that it is unreachable while every `failure_code()` is exhaustive, so there is no tail to prioritise; it is kept as defence only. `index.ts` export moved next to `ExecuteResult`.

**Re-run at this head**, since the rebase invalidated the earlier results:

- `tari_engine`: `access_rules`, `cross_template`, `fees`, `metering_budget`, `zk_verify`, `compute_fee_budget`, `test`, `wasm_loader`, `publish_template`, `template_upgrade` — 168 passed, 0 failed.
- `ootle_sdk_core` + `tari_engine_types` — 380 passed.
- `cargo lints clippy --all-targets` on `tari_engine`, `tari_engine_types` — clean.
- `tsc --noEmit` on `bindings` — clean.

sdbondi (commented):
_Posted by Claude._

**Approved with reservations** — 0 blocking, 0 non-blocking, 1 nit · 5 resolved since `0b973e8a` (2 blocking, 2 non-blocking, 1 nit) · head `56366c4e`

Everything I raised is addressed, and the push goes further than asked in one place — noted below rather than as a finding.

## Resolved (re-verified at this head)

- **Blocking — `helpers.ts`.** The duplicate unreachable `ExecutionFailure` branch is gone; the one remaining branch at :93 is reached while `reason` is still the full union, so `.code`/`.message` type. `prettier` is already green.
- **Blocking — `MaxNativeExecutionPointsExceeded`.** Now `LimitExceeded` (`runtime/error.rs:549-551`), alongside `FeeIntentComputeExceeded`, with `InsufficientFeesPaid` / `InsufficientFeesForNativeExecution` left on `OutOfCompute`. The split now matches the rule the PR states: pay-more clears the first pair, only moving the work clears the second.
- **Non-blocking — `MaxComputeExceeded` wording.** Reworded to "Exceeded a fixed compute maximum … whichever cap bound the call, per call or per transaction", which no longer asserts a cap it cannot distinguish. `test.rs:441` and the comment above it were updated to match.
- **Non-blocking — `IdProviderError`.** Split into its own exhaustive `failure_code()` (`id_provider.rs:40-48`): `MaxIdsExceeded` → `LimitExceeded`, `LockingError` → `EngineInvariant`. The no-wildcard property now holds one level deeper.
- **Nit — export ordering.** `ExecutionFailureCode` moved to its alphabetical slot in `index.ts`.

## Beyond what I raised

`WasmExecutionError::FeeIntentComputeExceeded` also moved `OutOfCompute` → `LimitExceeded`, with `compute_fee_budget.rs:96` and `zk_verify.rs:307` updated. I had called `OutOfCompute` defensible there; `LimitExceeded` is defensible too, and consistency with the `RuntimeError` twin is worth more than either reading on its own. No objection.

## Nit

`crates/engine_types/src/commit_result.rs:576` — `OutOfCompute`'s doc still reads *"Execution ran past what the fee paid for, **or past a hard compute ceiling**"*. That second clause is now exactly what `LimitExceeded` covers, and it is the one place a reader would go to decide which of the two a new error belongs in. Worth trimming.

## Reservation

Not a finding against the code: CI on this head is still pending, and the three legs that were red (`test (2/3)`, `integration test`, `web ui build`) have not yet re-run against these fixes. The fixes read correctly and the compile error is unambiguously gone, but I have not seen a green build, and I have not run the suites locally. I would want those three green before merge.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 0 non-blocking, 0 nits · 1 resolved since `56366c4e` (1 nit) · head `d591ef07`

Nothing stands. GitHub will not take an approve from me on this PR, so this is the approval in comment form — another maintainer still has to click it.

## Resolved at this head

- **Nit — `OutOfCompute` doc.** `commit_result.rs:576-583` now states the dividing line in both directions: `OutOfCompute` is "execution ran past what the fee paid for. Paying more clears it", `LimitExceeded` is "a fixed engine ceiling … no fee raises any of them". That is the rule the classification actually follows, in the place a reader goes to apply it to a new error.

## Verified across the series

- Every `failure_code()` is exhaustive with no `_` arm — `RuntimeError`, `WasmExecutionError`, `WasmValidationError`, `TransactionErrorKind`, `TemplateLoaderError`, `ResourceError`, `AssertError`, `LimitError`, `TransactionCommitError`, `IdProviderError`. A new error variant cannot compile unclassified.
- Cross-template delegation is type-correct (`call_method`/`call_function` return `TransactionErrorKind`) and leaves the rendered message byte-identical, so only the code is new on that path.
- `compute_exceeded_error` is total, and its three outcomes line up with the pay-more/move-the-work split the enum docs now state.
- Boundary: `failure_code` is additive, disjoint from `abort_code` (pinned by `sub_codes_do_not_overlap`), and the full 13-code set is frozen by `failure_codes_are_stable` plus the `reject_access_denied` golden vector.

## CI

Green on `d591ef07`, including the three legs that were red earlier in the series: `test (1/3)`, `(2/3)`, `(3/3)`, `integration test`, `web ui build`, `clippy`, `fmt`, all four fuzz targets.

sdbondi:
Closing this, as it landed in #2686

PR #2638, opened 2026-09-17
fix(consensus): count a NEWVIEW's timeout vote even when its certificate is behind

sdbondi:
Raised out of the review on #2636.

## The defect

`on_receive_new_view` drops a NEWVIEW whole when the certificate it reports is lower than the local high certificate:

```rust
if local_high_pc.block_height > high_pc.height() {
    return Ok(None);   // message discarded, timeout vote and all
}
```

The timeout vote goes with it. A leader that is ahead of more than `f` replicas therefore never reaches a quorum of timeout votes and cannot build the timeout certificate that carries its view forward — while the votes it is short of are sitting in the messages it just discarded. The laggards do recover by syncing off a later proposal, so this is a burned view rather than a wedge, but the view is burned for no reason: the vote that would have ended it was in hand.

## The fix

Whether the reported certificate is worth *adopting* and whether its timeout vote is worth *counting* are separate questions, and the code now asks them separately. A certificate ahead of ours is still adopted, and still triggers a sync when we do not hold its block. One level with or behind ours is simply not adopted, and the message carries on to vote collection.

The certificate's signatures are still verified before either question — a timeout vote arrives beside a certificate, and an unverifiable certificate takes the vote with it. That ordering is what makes this compose with #2636, where the timeout vote signs the height of the certificate beside it: a claim is only worth the certificate that backs it, and a vote whose certificate does not verify must not enter a timeout certificate. Under that PR a lower reported height also provably cannot raise `max_high_pc_height`, so counting these votes cannot weaken the bound it introduces.

Cost: a NEWVIEW carrying a lower certificate now has its quorum signatures verified where it was previously discarded unverified. That is the same verification already done for every other NEWVIEW.

## Testing

- `cargo nextest run -p consensus_tests`: 75 passed.
- `cargo lints clippy -p tari_consensus --all-targets`: clean.

Not covered by a test: the harness's `MessageFilter` returns `bool`, so it can drop a NEWVIEW but not rewrite one, and the scenario needs a leader whose high certificate is ahead of a specific subset of senders. Making that filter able to rewrite messages would make this and the byzantine-proposal cases in #2635 testable, and is worth doing on its own.


sdbondi (commented):
Makes sense

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 3 non-blocking, 0 nits · head `777057bf`

The fix is right and minimally expressed. I checked the diagnosis against the history: the guard arrives in #1437 with the comment "Only accept a higher QC than the local one" already gating the whole message through an `is_qc_valid` bool — so the drop really does look like adoption logic that swallowed the vote by accident, not a deliberate filter. Separating the two questions is the correct shape.

The behaviour delta is exactly one case. Old: `local > reported` → drop; `local == reported` → continue, then `reported > local` false → no adopt. New: `local >= reported` → `Ok(false)` → no adopt, continue. Equality and strictly-ahead both behave as before; only strictly-behind changes, from dropped to counted. That is the whole fix.

**Counting those votes is safe on the paths they reach.** `timeout_vote_collector.check_and_collect_vote` runs `check_eligibility` against the local committee and `validate_vote` (epoch + signature) before the vote is stored, and votes are keyed by `(epoch, height)` with `AggregationKey = ()`, so a late-arriving laggard's vote lands in the same bucket as everyone else's at that view. The certificate's own signatures are still verified first (:89), which is the ordering #2636 needs — a vote whose certificate does not verify must not reach a timeout certificate.

**And it actually ends the view rather than deferring it.** A replica one certificate behind still holds the *block* at that height (it voted on it); it is only short of the QC. So when the leader proposes at `T+1` justifying from its own higher certificate, that replica resolves the justify block locally, `check_extends_justify` recomputes the dummy chain from it, and the safeNode predicate reads off it — no sync needed. The body's "laggards recover by syncing off a later proposal" is the conservative case (a replica that never received the block at all); the common one is cheaper than that.

Hoisting `validate_qc` out of the read transaction is a bonus: it never took `tx`, and the transaction is now not held across a batch signature verification.

## Non-blocking

**1. The `>=` quietly removes a second thing: the `FallenBehind` trigger at equal height.** Old code reached `Block::record_exists` whenever `local <= reported`; new code returns at :100 before it, so equality no longer triggers a sync (:104-114). I think this is an improvement — holding a certificate at height H means we are not behind at H, and `update_highest` sets `as_leaf_block()`, which fails on a missing block, so our own certificate always has its block. The only case that differs is a *different* certificate at the same height, i.e. equivocation, where syncing onto the other branch was never obviously right and #2635's lock rule now closes the fork anyway. But it is a second behaviour change and the description covers only the first; worth saying it was intended.

**2. The proposal vote rides along too, not just the timeout vote.** The title and body are about the timeout vote, but a NEWVIEW that previously died at :100 also carried its `last_vote` past `proposal_vote_collector.check_and_collect_vote` (:136-146). That is fine and mildly useful — the sender only attaches `last_vote` when it voted above its own certificate (on_next_sync_view.rs, the `filter(|vote| high_pc.height() < vote.block_height())`), the collector needs a quorum on the same block to form anything, and `update_highest` only raises. Still, "both collectors are now reached" belongs in the description of a consensus change.

**3. The "not testable without a rewriting filter" claim looks stronger than it needs to be.** `MessageFilter` is indeed `Fn(&TestAddress, &TestAddress, &HotstuffMessage) -> bool` (support/network.rs:35), so it cannot rewrite — that part checks out. But the scenario does not need a rewrite, only a leader whose certificate is ahead of some senders, and dropping is enough to produce that: let the leader collect votes for block `h` (it now holds `QC_h`, nobody else does), drop the next proposal so the view times out, and the NEWVIEWs from the laggards arrive at a leader that is ahead. `dummy_fill_anchor` already starves one validator of a proposal with exactly this kind of filter, and the schedule is fully deterministic — `RoundRobinLeaderStrategy` is `height % committee.len()` (support/leader_strategy.rs:16) — so which view lands back on that leader is arithmetic, not luck. I have not built it, so I may be missing something the harness makes awkward; but it looks constructible with what is there, and this is a liveness property on the main consensus path.

## Note

#2636 edits this same region and still reads the `local_high_pc_height` binding that this PR removes, so it will want a rebase over this one rather than the other way round.

sdbondi:
_Posted by Claude._

Post-merge corrections to the record, on all three points.

**1 — the `>=` also drops the `FallenBehind` trigger at equal height, and that was intended.** The description covered only the vote-counting delta. The second change: the old code reached `Block::record_exists` whenever `local <= reported`, the new code returns before it at equality. It is a deliberate narrowing, for the reason you give — a certificate at height `H` is one we hold, and `update_highest` sets `as_leaf_block()`, which fails on a missing block, so our own certificate always has its block present. We are not behind at `H`. The only case that changes is a *different* certificate at the same height, i.e. equivocation, where syncing onto the sibling branch was never the right recovery and #2635's lock rule now keeps the fork from committing either way. I will put that invariant next to the check in the code, since a merged PR description is not where a future reader looks.

**2 — `last_vote` rides along too.** Also true and also missing from the description: a NEWVIEW that previously died at the guard carried its proposal vote past `proposal_vote_collector.check_and_collect_vote` as well. Both collectors are now reached. It is beneficial for the same reason as the timeout vote and cannot do harm: the sender only attaches `last_vote` when it voted above its own certificate, the collector needs a quorum on one block to form anything, and `update_highest` only raises.

**3 — my "needs a rewriting filter" claim was wrong, but the test is still not worth building, for a different reason.** You are right that dropping alone constructs the state: the leader of view `v+1` forms `QC_v` from the votes before anyone else, so dropping its proposal leaves it uniquely ahead, and `height % len` says exactly when the view returns to it.

What is missing is not the setup but the assertion. Reaching a view where the ahead node is leader needs its proposal at `v+1` dropped *and* the views in between kept from succeeding — a behind leader that adopts the ahead node's certificate from its NEWVIEW syncs and recovers the committee, which is the same thing that makes the bug self-heal one view later. So the scenario needs three or four coordinated drops to hold the state open, and what it can then assert is "this node's store holds a `HighTc` at this view", which the bug turns into "does not". That is a real regression test, but it is a heavily constructed schedule for a one-view delay, and the construction itself is what the drops have to defeat.

Worth noting the severity that follows from the same analysis, since the description does not say it: recovery costs a sync either way. The laggards are short of `QC_v`, not of the block, but the ahead leader's own recovery proposal justifies from `B_v`, so they fetch it in both worlds. What the fix buys is that the view advances at the leader that holds the messages, instead of burning until leadership rotates to a node that will sync.

sdbondi:
_Posted by Claude._

Correcting the last paragraph of my previous comment: "recovery costs a sync either way" is wrong, and it understates what this fix buys.

A replica votes on a block only after storing it — `save_block` runs at on_receive_local_proposal.rs:263, inside the same write transaction that validated it, and `on_ready_to_vote_on_local_block::handle` decides the vote later at :363. So every voter on `B_v` holds `B_v`; what the laggards are short of is `QC_v`, and that arrives inside the `justify()` of the recovery proposal itself. `Block::get` resolves the justify block locally, `JustifyBlockNotFound` never fires, and no catch-up sync happens.

The same correction applies to the without-the-fix path I described: the behind leader that receives the ahead node's NEWVIEW passes `Block::record_exists` for the same reason, so it adopts the certificate and proposes without syncing either. A sync is needed only by a replica that missed `B_v` outright — one that was offline for view `v`, not one made to lag by dropping a proposal at `v+1`.

So the cost of the bug is burned views until leadership rotates to a node that will act on the certificate, and the fix's benefit is the view advancing at the leader that already holds the messages. No sync on either side of the comparison in the normal case.

PR #2639, opened 2026-09-17
fix(wallet): bound the branch-and-bound input search

sdbondi:
## Problem

`branch_and_bound::select` runs an unbounded exhaustive search. It terminates quickly when some subset of the inputs totals the target exactly — the incumbent is optimal and the search breaks. When no subset does, it has to prove minimality by enumerating every subset below the incumbent, which is exponential.

Equal-valued outputs are the everyday shape of "no exact match". Measured on the current algorithm (identical-valued inputs, target not exactly reachable), the cost grows ~4x per two extra inputs:

| identical inputs | states explored | time |
|---|---|---|
| 24 | 10.4M | 0.28s |
| 28 | 155M | 3.9s |
| 30 | >200M | did not finish |
| 40 | >200M | ~hours extrapolated |
| 80 | — | C(80,40) ≈ 1e23 subsets; never |

This is on the default send path: `lock_outputs_for_at_least_amount` selects with `InputSelectionAlgorithm::BranchAndBound`, and it does so inside `store.with_write_tx`, so the wallet store's write lock is held for the duration — one such send stalls the whole wallet daemon, not just that request. An account can be put into that state by anyone who pays it in equal amounts.

Randomly-valued inputs are unaffected (60 inputs: ~2k states, 100µs); it is specifically the near-uniform, no-exact-fit set.

## Fix

- **Cap the search at 100,000 states** (the number Bitcoin Core uses for its own branch-and-bound) and take the best selection found so far when the cap is hit. The result still reaches the target and still respects `max_inputs`; it may leave more change than the minimal selection would.
- **Explore the include branch first.** The states were pushed include-then-skip and popped LIFO, so the search descended skip-first and its first incumbent was built from the *smallest* inputs — the worst answer to settle for, and a weak bound to prune against. Descending largest-first reaches the target within `max_inputs` states, so there is always a usable incumbent long before the cap, and every later state has a real total to be pruned against.
- **Decide reachability up front.** A cap alone would break the caller: `lock_outputs_for_at_least_amount` reads a `None` return as proof that no `max_inputs` of the outputs reach the amount, which is what separates `InputLimitReached` (fragmentation, consolidating fixes it) from `InsufficientFunds` (a balance). Since no selection of at most `max_inputs` inputs can total more than the `max_inputs` largest, a prefix sum over the sorted inputs decides it exactly. `None` keeps meaning "out of reach" and now costs O(n log n) instead of an exhaustive search.

The fallback path (cap reached before any incumbent) is unreachable in practice given the ordering above, but it is written out rather than asserted, and `select_bounded` takes the budget as an argument so the tests can drive it.

Selection *quality* is unchanged in scope: the truncated answer is largest-first, which is closer to what a send wants than what the old first incumbent was, but preferring fewer, larger inputs by design is separate work.

## Tests

Four added, all in `branch_and_bound.rs`:

- `it_terminates_when_no_subset_totals_the_target` — 80 identical inputs, unmatchable target; this hangs indefinitely before the fix, and completes in ~30ms (debug) after.
- `it_returns_a_largest_first_selection_when_the_search_is_cut_short` — with a zero-state budget, the largest-first selection comes back.
- `a_cut_short_search_respects_max_inputs` — the cut-short path honours the limit, and reports `None` when the limit puts the target out of reach.
- `it_reports_an_unreachable_target_without_searching` — over balance, over `max_inputs`, and the exact boundary.

`cargo nextest run -p tari_ootle_wallet_sdk`: 46 passed. `cargo lints clippy -p tari_ootle_wallet_sdk --all-targets`: clean.


sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 2 non-blocking, 2 nits · head `4c85d13a7`

Read the whole file at head plus the caller in `stealth_outputs.rs`. The three-part fix hangs together and I could not break it:

- The up-front reachability check is exactly right for the invariant it has to preserve. `items` is sorted descending before it, so `take(max_inputs).sum()` really is the maximum achievable under the limit, and `None` keeps meaning "out of reach" — which is what `lock_outputs_internal` (`stealth_outputs.rs:462-471`) relies on to split `InputLimitReached` from `InsufficientFunds`. That split is still sound after the cap.
- Pushing `with` last so it pops first does give a first incumbent within `max_inputs` states, and the incumbent is only ever set under `target <= state.total`, so anything returned at the cap genuinely reaches the target. The `max_inputs` gate is on the push, so a truncated answer cannot exceed the limit either.
- `take_largest_first` is unreachable in the ways the description claims, but it is correct on its own terms: the early check guarantees `total_value >= target` by the time `max_inputs` items are consumed, so its `None` arm is defensive only.
- `Amount` is an unsigned 128-bit integer, so summing `u64` values cannot overflow `max_achievable`, and the one subtraction on it (`state.total - target`) is guarded by the `target <= state.total` branch it sits in, so there is no underflow either.

Edge cases I re-checked by hand: `target == 0` (returns an empty selection before the cap can bite), `max_inputs == 0` (`take(0).sum() == 0`, so any positive target short-circuits to `None`), and `max_inputs` exceeding `inputs.len()`.

## Non-blocking

1. `branch_and_bound.rs:160-164` — `remaining_sum` recomputes an O(n) suffix sum at every state. The cap now bounds the state count, but per-state cost is still O(n) for that sum plus O(k) for the two `selected` clones, so the worst case is ~`max_states * n` additions (8M for the 80-input case) rather than `max_states`. A suffix-sum array computed once next to the sort makes the bound O(1) per state and is a couple of lines. Not a correctness issue — just that the PR's goal is a predictable ceiling, and this is the term that still scales with `n`.

2. `branch_and_bound.rs:123-126` — the `state.selected.len() > max_inputs` prune is now provably dead: the only push site gates on `len() < max_inputs`, and the root state has zero selected. It predates this PR, but this PR is what makes it unreachable; worth deleting while the file is open rather than leaving a branch that implies the invariant can be violated.

## Nits

3. `branch_and_bound.rs` (`a_cut_short_search_respects_max_inputs`) — the first assertion (`target 800`, `max_inputs 2`) is decided by the up-front reachability check and never reaches the cut-short path, so despite the test's name only the second assertion exercises truncation under a limit. Fine as coverage of the `None` contract, just not of what the name says.

4. `branch_and_bound.rs` (`it_terminates_when_no_subset_totals_the_target`) — `assert!(result.selected_keys.len() <= 80)` is trivially true for 80 inputs. `== 41` would actually pin the answer (41 x 1000 is the unique minimum over 40_500), and would catch a regression that returns a needlessly large truncated selection.

One behavioural note for the record, not a finding: descending include-first also flips tie-breaking. When two selections have the same total, `_ => {}` keeps the first incumbent, which is now the largest-first one rather than the smallest-first one. That is the better answer for a send and the description covers it, but no existing test pins *which* keys come back on a tie (e.g. `test_duplicate_values` only checks count and total), so the change is invisible to the suite.

sdbondi:
_Posted by Claude._

All four applied in 3bedc0413.

1. **Suffix sums** — totalled once beside the sort, so the upper-bound prune is a lookup and the cost is proportional to states explored rather than states × inputs. Right that this was the term still scaling with `n`; the ceiling is the point of the PR.
2. **Dead prune** — removed. Confirmed unreachable: the sole push site gates on `len() < max_inputs` and the root holds none, so no state can arrive holding more.
3. **`a_cut_short_search_respects_max_inputs`** — the `800` case was indeed decided by the reachability check, not the cut-short path. Retargeted to `650`, which is reachable within the limit and so actually runs and is truncated; the over-the-limit `None` case moved to `it_reports_an_unreachable_target_without_searching`, with a zero-budget variant alongside it so the contract is pinned on both paths.
4. **`it_terminates_when_no_subset_totals_the_target`** — now asserts `41_000` / 41 keys. As you say, that pins the answer to the largest-first descent rather than to whatever the cap happened to leave.

On the tie-breaking note: agreed, and it stays invisible to the suite — item 3's case is not a tie either (no other two-input selection reaches 650), so nothing here pins which keys come back when two selections have the same total. Leaving that rather than growing the diff: it is a property of the ordering this PR chose, and worth its own test if selection quality work follows.

`cargo nextest run -p tari_ootle_wallet_sdk`: 46 passed. Clippy clean.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 0 non-blocking, 0 nits · 4 resolved since `4c85d13a7` (2 non-blocking, 2 nits) · head `3bedc0413`

Re-read the file at `3bedc0413` and diffed it against the previous head. All four items are genuinely fixed, and the refactor introduces nothing new.

Resolved, with what I checked:

1. **Suffix-sum lookup** (`branch_and_bound.rs:111-117`) — `remaining_sums` is `items.len() + 1` long with a zero sentinel at the end, and the reverse fold makes `remaining_sums[i]` equal `items[i..].sum()`, so the lookup at line 167 is exactly the value the old inline sum produced. `state.index` is bounded below `items.len()` by the `state.index >= items.len()` guard above it, and the extra slot makes it in-range even without that. The per-state cost is now O(1) plus the `selected` clones, so the `MAX_SEARCH_STATES` cap is a real ceiling on work rather than one scaled by `n`.

2. **Dead `selected.len() > max_inputs` prune removed** — re-confirmed it was unreachable: the single push site gates on `< max_inputs`, the root state has zero selected, and `take_largest_first` caps with `take(max_inputs)`, so nothing can construct a state past the limit.

3. **Termination test now pins the answer** — `41_000` / 41 keys is right and deterministic: every subset total is a multiple of 1000, 41_000 is the smallest at or above 40_500, and the `best <= state.total` prune rejects equal totals, so the largest-first incumbent is what survives whether or not the cap is hit. This now actually fails if a future change returns a sloppier truncated selection.

4. **`a_cut_short_search_respects_max_inputs` exercises the cut-short path** — at target 650 with `max_inputs` 2, `max_achievable` is 700, so the early check passes and the zero budget breaks on the first pop into `take_largest_first`, which returns B+C = 700. The `None`-under-`max_inputs` assertion moved to `it_reports_an_unreachable_target_without_searching`, where the added zero-budget variant checks the same contract holds when the budget cannot rescue it. Both tests now say what their names claim.

Nothing further from me. The earlier tie-breaking note stands as a note, not a finding.

CI on this head is green for the three checks that have reported (signed commits, GitGuardian, check-title); I have not run the build locally.

sdbondi (commented):
Approved

PR #2640, opened 2026-09-17
feat(transaction)!: declare read or write intent on each transaction input

sdbondi:
## Why

A transaction declares its inputs, and a shard group that does not hold one of them cannot execute to find out how it is used. `SubstateRequirement` has nowhere to say "I only read this", so that shard group must lock for writing to be safe. Two costs follow:

- **Liveness.** An untouched declared input in a foreign shard group takes an exclusive write lock, held from proposal to commit. Honest transactions naming that substate at a specific version hard-abort; unversioned ones are deferred. The attacker pays 15 weight per input.
- **Lost read parallelism.** Every cross-shard read of a shared resource, oracle or config component serialises — exactly the hot substates where contention hurts.

`prepare_multishard_no_execution` currently guesses with `is_read_only() || is_resource()` and carries a TODO saying the guess breaks access-rule updates and that the user should be able to specify this instead. This is that.

## What changed

`InputDeclaration` — substate id, optional version, `is_write` — becomes the type of a transaction's `inputs`. `SubstateRequirement` is unchanged and still used everywhere a substate is merely *named*: indexer queries, wallet substate lookups, RPC. Only the input list moved.

The declaration is an upper bound on access rather than a hint, which takes three pieces:

1. **Lock from the declaration.** `prepare_multishard_no_execution` uses the declared intent, deleting the heuristic and its TODO.
2. **Enforce it in the engine.** A write to a read-declared input aborts in `WorkingState::try_lock`, alongside the existing frame-write-mode check. Under-declaring is never a way out of a lock another transaction has already taken.
3. **Default to write.** A declaration that says nothing declares a write — over-declaring locks more than it needs and runs slowly, where under-declaring aborts part-way through execution. `declare_input` widens when a substate is declared twice, so declaration order cannot decide the intent.

Enforcement is what makes the default matter. An earlier draft inferred the intent for wallet-resolved inputs, reproducing the old heuristic at the declaration site. That was dropped: the heuristic was safe only because it applied to foreign inputs that the engine never checked, and once the engine aborts on a bad read guess, guessing on the caller's behalf turns a working access-rule update into an abort. One rule instead — if you did not say `read`, you declared a write.

## Format change

`InputDeclaration` is inside `TransactionSignatureFields`, so this changes the preimage, and with it every transaction id and signature. **This rides the next network reset.**

No `schema_version` bump. The bump only buys concurrent v1/v2 acceptance, which a reset makes unnecessary — no already-signed transaction survives it. The derived `BorshSerialize` carries the new field into the preimage on its own.

- CBOR (the consensus form) **requires** `is_write`, so a peer's intent is never guessed for it.
- JSON (wallet RPC, SDK) **defaults it to write**, so a client that has not been updated over-declares rather than under-declares.
- `--input` on `transaction_generator` and `-i` on the wallet CLI take `<id>[:<version>][:read|:write]`; `InputDeclaration` also parses `<id>:?`, so its own `Display` round-trips (`SubstateRequirement` still has that gap — pre-existing, untouched here).

## What this does not do

The consumer — `prepare_multishard_no_execution` — is unreachable with a single shard group, and the local swarm runs one (`calculate_num_committees(7 VNs, committee_size 40) = 1`), so the consensus half of this cannot be exercised locally. The engine half is reachable today and is covered by `crates/engine/tests/input_intent.rs`.

Because the old heuristic is gone and nothing declares reads yet, a foreign input that used to take a read lock (a resource, a template) now takes a write lock. That is a concurrency regression confined to the unreachable path, and the fix is wallets declaring reads rather than the node guessing.

## Tests

- `input_declaration` unit tests: parse/Display round-trip for every version-intent combination, absent intent is a write, `declare_input` widens in either order.
- Preimage sensitivity: the seal, authorization and intent-commitment digest tests now assert that flipping `is_write` changes the digest.
- `crates/engine/tests/input_intent.rs`: a read-declared input may be read; writing to one aborts with `WriteToReadDeclaredInput` and leaves the component unchanged; a write-declared input may be written; a substate declared both ways is a write.
- `ootle_sdk_core` golden fixtures regenerated — the only content delta is the added `is_write`.
- `cargo lints clippy --all-targets` clean; `tari_engine`, `ootle_sdk_core`, `tari_ootle_wallet_sdk`, `tari_template_test_tooling` (789), `consensus_tests`, `tari_ootle_transaction`, `tari_ootle_common_types`, `transaction_generator` all pass.


sdbondi (commented):
_Posted by Claude._

**Approved** — 0 blocking, 3 non-blocking, 2 nits · head `ca3d07b6`

Reviewed at `ca3d07b6` (one commit past the `bbbc66c3` the description was written against). CI is green across the board.

## What I verified as sound

- **The engine check is a genuine chokepoint.** Every mutating path in `WorkingStateStore` first demands a `LockFlag::Write` lock obtained through `try_lock` — `get_locked_substate_mut`, `mutate_locked_substate_with`, `down_utxo` and `down_confidential_output` all go through `locked_substates.get(lock_id, LockFlag::Write)`. So the new guard at `crates/engine/src/runtime/state_store.rs:70` cannot be walked around by destroying a read-declared input instead of mutating it. Holding the set in the store rather than the working state is the right home for it.
- **The read declaration on TARI is safe.** `XTR_REQUIREMENT` became `is_write: false` (`crates/transaction/src/v1/transaction.rs:70`), which puts `resource_TARI` in `read_declared_inputs` on every transaction. No write lock can be taken on it: the only supply mutations are `increase_total_supply` / `decrease_total_supply`, and the genesis TARI resource disables supply tracking and denies `Burn` — the invariant is already written down at `crates/engine/src/runtime/impl.rs:3607-3611`. The remaining resource write-lock sites (Mint, UpdateAccessRule, UpdateAuthHook, UpdateMetadata, Recall) are all denied on TARI by its access rules.
- **Both-ways declarations close at both layers.** `declare_input` widens in the set, and `processor.rs:190`'s `read_declared_inputs.retain(...)` repeats the rule for an `Executable` that chains several sources. The version-keeping on widen is documented and right.
- **`PartialEq`/`Hash` keyed on substate id alone exactly mirrors `SubstateRequirement`** (`versioned_substate_id.rs:163,179`), so the dedup semantics of every `IndexSet<InputDeclaration>`, of `indexset_codec::decode`, and of `push_unique` in `ootle_sdk_core/src/inputs.rs:857` are unchanged from before. No silent behaviour delta there.
- Preimage sensitivity to `is_write` is asserted in all three digests, and `Evidence::from_inputs_and_outputs` records inputs unpledged with no lock type, so the initial evidence cannot disagree with the intent-derived locks taken later.

## Non-blocking

**1. `detect_inputs` silently widens a caller's read declaration back to a write** — `applications/tari_walletd/src/handlers/transaction.rs:148-173`

Every id the transaction already declares is chained into `locate_dependent_substates`, and everything that comes back is mapped to `InputDeclaration::write(...)` / `InputDeclaration::from(input)` — both writes — then folded in with `with_inputs`, which widens. So a client that declares `X:read` and submits through `transactions.submit` with `detect_inputs` on (the default path) gets `X:write`.

It is safe — widening always is — but it means the wallet daemon, the feature's main consumer, has no way to express a read at all. Same shape in `handle_detect_inputs` (`:374`) and `submit_dry_run_inner` (`:422`). Worth carrying the caller's declared intent through: when a detected input matches one already declared, keep the declaration's `is_write` rather than overwriting with `true`.

**2. `is_read_only()` substates lose their read lock for nothing** — `crates/consensus/src/hotstuff/transaction_manager/manager.rs:556-562`

The deleted heuristic was `is_read_only() || is_resource()`. The `is_resource()` half was genuinely unsound — that is the access-rule-update case the TODO named. The `is_read_only()` half is not: a `Template`, `TransactionReceipt` or `ClaimedOutputTombstone` has no engine path that writes it at all, so coercing those to a read lock is correct regardless of what the transaction declared. Keeping that half would erase the concurrency regression the description accepts for templates, at no safety cost, instead of waiting on wallets to start declaring reads.

As it stands `SubstateId::is_read_only` (`crates/engine_types/src/substate.rs:370`) now has no callers left in the tree.

**3. A write action on TARI now reports the wrong reason** — `crates/engine/src/runtime/impl.rs:1691` (and `:1905`, `:1947`, `:2002`)

`ResourceAction::Mint` and friends take the resource write lock *before* checking access rules. With TARI read-declared, a mint against TARI now aborts with `WriteToReadDeclaredInput` rather than the access-rule denial that actually governs. The outcome is right and the operation was always going to fail, but the error now points the caller at their input declaration instead of at the rule they violated.

## Nits

- `applications/tari_wallet_cli/src/command/transaction.rs:112` — `-i` silently gained the `:read` / `:write` suffix, but its help text wasn't extended the way `transaction_generator`'s `--input` was in `utilities/transaction_generator/src/cli.rs:48`.
- The description's "`<id>:?` now parses, so `Display` round-trips" holds for `InputDeclaration` only. `SubstateRequirement::from_str` (`crates/common_types/src/versioned_substate_id.rs:126`) still rejects `?` while its `Display` still emits it, so that type's own round-trip gap is untouched. Pre-existing, not this PR's to fix — just don't let the sentence read as if it were.

sdbondi:
_Posted by Claude._

Thanks — two of the three non-blocking items are fixed in `b4263e31`, one is declined, and one correction.

**1. `detect_inputs` widening a read back to a write — fixed.** Confirmed exactly as described: detection is seeded with the ids the transaction already declares, everything comes back as a write, and `with_inputs` widens. Detection now defers to an existing declaration for the same substate and only declares a write where the transaction said nothing — `declare_detected_input`, applied in `transactions.submit`, `transactions.detect_inputs` and the dry-run path, with four unit tests on the helper (read preserved, write preserved, undeclared → write, unversioned detection drops the version).

**Nit 1 — fixed.** `-i` now carries the same help text as the generator's `--input`.

**2. Keeping the `is_read_only()` half — agreed on the substance, left to the repo owner.** The reasoning is right: a `Template`, `TransactionReceipt` or `ClaimedOutputTombstone` has no engine write path, so read-locking one is sound regardless of the declaration, and it would erase the concurrency regression for templates without waiting on wallets. I'd raised the same option before opening this PR and it wasn't taken up, so I'm not adding it unilaterally — it changes consensus lock behaviour and belongs in an explicit decision rather than a review round.

One correction to the supporting detail: `is_read_only()` does still have callers — `crates/engine/src/runtime/scope.rs:332` and `applications/tari_indexer/src/network_state_sync/worker.rs:1302`. So it isn't dead code, and "no callers left" isn't an argument for restoring the branch.

**3. Wrong reject reason for a write action on TARI — declined.** The observation is correct (`impl.rs:1691` write-locks the resource before `check_resource_access_rules`, and the same shape at `:1905`, `:1947`, `:2002`). But the lock is what loads the resource, and the access rules are read off the loaded value — the rule check cannot precede it without a read-lock-then-upgrade dance across four call sites. That is real risk in the engine's locking for a better error string on an operation that fails either way. Not worth it here.

**Nit 2 — fair.** `SubstateRequirement::from_str` does still reject `?` while its `Display` emits it; that gap is pre-existing and untouched. I've reworded the description so the round-trip claim is scoped to `InputDeclaration`.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 2 non-blocking, 1 nit · 2 resolved since `ca3d07b6` (1 non-blocking, 1 nit) · head `b4263e31`

CI still running on this head; posting as a comment since I can't approve your own PR either way.

## Resolved since `ca3d07b6`

- **Non-blocking 1 — read declarations surviving detection.** `declare_detected_input` (`handlers/transaction.rs:820`) now looks the detected id up in the transaction's own declaration set and defers to it. Checked the lookup works as intended: `InputDeclaration` implements `Borrow<SubstateId>`, so `declared.get(detected.substate_id())` hits, and `is_none_or(|d| d.is_write())` gives write for an undeclared substate and read only where the caller asked for one. Version handling is byte-for-byte what the three call sites did before (`None` under `use_unversioned`, else the detected version), so nothing else moved. The subsequent `with_inputs` fold then widens read+read to read rather than back to write — which is the whole point. All three sites converted, and the four unit tests pin each corner.
- **Nit 1 — CLI help.** `-i` now documents the `:read` / `:write` suffix (`tari_wallet_cli/src/command/transaction.rs:112-115`).

## Still standing (both non-blocking, unchanged from the last pass)

- **`is_read_only()` substates lose their read lock for nothing** — `transaction_manager/manager.rs:556-562`. The `is_resource()` half of the removed heuristic was unsound; the `is_read_only()` half was not, and reinstating just that half would erase the template-contention regression the description accepts, at no safety cost. `SubstateId::is_read_only` currently has no callers.
- **A write action on TARI reports `WriteToReadDeclaredInput` rather than the access-rule denial** — `runtime/impl.rs:1691` and siblings take the resource write lock before the auth check.

Nit: the description's "`<id>:?` now parses" is true of `InputDeclaration` only; `SubstateRequirement::from_str` still rejects it. Pre-existing.

`handle_submit_manifest` (`:543`) still maps its dependencies to unconditional writes, which is right — the manifest DSL has no way to express a read, so there is no caller intent to preserve there.

sdbondi:
_Posted by Claude._

Nothing new to change on this pass — both standing items were answered on the previous round, and two points in this review are stale.

- **`is_read_only()` "currently has no callers" is still wrong.** Two callers: `crates/engine/src/runtime/scope.rs:332` and `applications/tari_indexer/src/network_state_sync/worker.rs:1302`. It was wrong last pass too. The substantive suggestion — reinstating just the `is_read_only()` half of the removed heuristic — remains sound and remains a call for the repo owner, since it changes consensus lock behaviour.
- **The `<id>:?` nit is already fixed.** The description was reworded when the previous reply was posted; line 30 now scopes the round-trip claim to `InputDeclaration` and names the `SubstateRequirement` gap as pre-existing.
- **Write action on TARI reporting `WriteToReadDeclaredInput`** — declined last pass, reasoning unchanged: the write lock is what loads the resource the access rules are read from, so the rule check can't precede it without a read-then-upgrade across four sites.

Agreed on `handle_submit_manifest` (`:543`): `TransactionSubmitManifestRequest` carries no `inputs` field — the transaction is built from the manifest text — so there is no caller declaration to defer to and unconditional writes are correct there.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 3 non-blocking, 1 nit · 0 resolved since `b4263e31` (the delta is a rebase plus one new commit) · head `8afbfba2`

Re-reviewed after the rebase onto `d8cb3b37`. The PR's own engine and consensus diff is byte-identical to what I checked at `b4263e31` — the `try_lock` guard, the both-ways `retain` in `processor.rs`, and `declare_detected_input` all survived unchanged. Everything else in the rebase is a mechanical `u32` → `SubstateVersion` adaptation: `FromStr` still parses through `SubstateVersion: FromStr`, `or_zero_version` is now `SubstateVersion::ZERO`, and the `:?` / `:read` / `:write` grammar is untouched. No new findings from the rebase itself.

## New, from `8afbfba2`

**1. The commit is not what it says, and it widens a type outside this PR.** — `crates/engine_types/src/substate.rs:81`, `crates/engine_types/src/substate_version.rs:94-102`

"prettier and lints" also changes `Substate::new` from taking a `SubstateVersion` to `V: Into<SubstateVersion>`, and adds `From<u32>` and `From<u16>` for `SubstateVersion`. Those impls have no callers — `grep` finds zero `SubstateVersion::from` anywhere — and the only bare-integer `Substate::new` call site in the tree is the one test line this PR adds at `state_store.rs:385`. So the whole change exists to serve a single test literal, and it does not even do that: the generic made a bare `0` ambiguous, which is why the line had to become `Substate::new(0u64, ...)`.

The cost is not local. #2687 introduced the newtype days ago precisely so an untyped integer version is a compile error; `From<u32>` / `From<u16>` make it an implicit conversion again, everywhere, for anyone who writes `.into()`. `Substate::new(SubstateVersion::ZERO, ClaimedOutputTombstone { value: 1 })` in the test reads better than `0u64` and leaves the newtype intact.

Separately: a commit titled "prettier and lints" that changes a shared constructor's signature will not be found by anyone bisecting for it later.

**2. `read_versioned` and `write_versioned` now disagree on their version parameter.** — `crates/common_types/src/input_declaration.rs:64,73`

`read_versioned` gained `V: Into<SubstateVersion>` in the rebase; `write_versioned` still takes a `SubstateVersion`. With the `From<u32>` impl from finding 1 in place, `read_versioned(id, 3u32)` compiles and `write_versioned(id, 3u32)` does not. Whichever way you settle finding 1, these two should match — they are two halves of one new public API.

## Still standing

**3. `is_read_only()` substates lose their read lock for nothing** — `transaction_manager/manager.rs:557-563`. Unchanged by the rebase: the `is_resource()` half of the deleted heuristic was unsound, the `is_read_only()` half was not, and reinstating just that half erases the template-contention regression at no safety cost. Correcting myself from the last pass — I said `is_read_only` had no callers left; post-rebase it has two (`engine/src/runtime/scope.rs:332`, `tari_indexer/src/network_state_sync/worker.rs:1302`). That does not change the suggestion, only my reason for noticing it.

Nit: a write action on TARI still reports `WriteToReadDeclaredInput` rather than the access-rule denial that actually governs (`runtime/impl.rs`, the resource write lock precedes the auth check). Cosmetic, and the operation was always going to fail.

CI at this head: 14 green, 6 still running.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 3 non-blocking, 1 nit · 1 resolved since `8afbfba2` (1 nit) · head `5a7c4db6`

## Resolved

**The misleading abort reason on a read-declared resource.** `5a7c4db6` fixes it at the root rather than at the message: mint and burn now take a read lock for the access checks and upgrade to a write lock only when the resource tracks supply. Verified rather than assumed:

- **Mint under a read lock performs no mutation.** `mint_resource` re-reads `is_supply_tracking_enabled` from the same locked resource (`working_state.rs:944`) and its only `get_resource_mut` is gated on that flag (`:1059`). Burn is the same shape — the sole resource mutation is `decrease_total_supply`, gated at `:878`.
- **A disagreement between the two flag reads fails safe.** The outer `tracks_supply` is captured before the hook, and `mint_resource` computes its own. If they ever diverged, `get_resource_mut` on a read lock returns a `LockError`, so the transaction aborts cleanly — it cannot mutate through a read lock.
- **Consensus lock types are unaffected.** `ExecutionOutput::resolve_input_locks` (`tari_ootle_app_utilities/src/transaction_executor.rs:57-66`) derives the input lock flag from whether the substate appears in the accept diff's `down_iter`, not from the engine's internal lock flag. A resource that is never mutated was never downed, so it already got a read lock there. The narrowing is confined to the engine's own `WorkingStateStore`, which is exactly the scope the commit message claims.
- **The lock is released before the hook and relocked after**, in both paths (`impl.rs:1790`, `:3231`), so nothing is held across the call.

The test pair is the right one: `minting_and_burning_leave_a_resource_without_supply_tracking_as_a_read` proves the narrowing works, and `burning_from_a_read_declared_resource_that_tracks_supply_aborts` proves it did not become a hole. Without the second, the first would be indistinguishable from removing the check.

## Still standing

**1. `Substate::new` widened and `From<u32>`/`From<u16>` added to `SubstateVersion`** — `engine_types/src/substate.rs:81`, `substate_version.rs:94-102`. Still no callers for either impl, and the only bare-integer `Substate::new` site is the test literal at `state_store.rs:385`, which still needs its `u64` suffix. It gives back the implicit integer conversion #2687's newtype was introduced to remove. `SubstateVersion::ZERO` in the test costs nothing.

**2. `read_versioned` and `write_versioned` disagree** — `input_declaration.rs:64,73`. `read_versioned` takes `V: Into<SubstateVersion>`, `write_versioned` takes a `SubstateVersion`. Two halves of one new public API.

**3. `is_read_only()` substates lose their read lock for nothing** — `transaction_manager/manager.rs:557-563`. The `is_resource()` half of the deleted heuristic was unsound; the `is_read_only()` half was not.

Nit: `UpdateAccessRule` (`impl.rs:1993`) and `UpdateAuthHook` (`:2035`) still take the write lock before `check_resource_access_rules`, so on those two a read-declared resource reports `WriteToReadDeclaredInput` rather than the rule denial — the residue of what `5a7c4db6` just fixed for mint, burn and `UpdateMetadata`. Both always mutate, so the abort itself is correct either way; only the reason is off.

CI at this head: 4 green, 4 still running.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 3 non-blocking, 1 nit · 1 resolved since `5a7c4db6` (1 non-blocking) · head `50e05ea1`

## Resolved

**Read-only substates now take a read lock whatever was declared.** `50e05ea1` does the thing the earlier note asked for, and confines it correctly — resources in general are *not* coerced, only the five kinds in `is_read_only()`.

This is the one direction that can be unsafe, because it overrides the declaration rather than following it, so I checked each kind rather than taking the doc comment's word:

- **`STEALTH_TARI_RESOURCE_ADDRESS` / `PUBLIC_IDENTITY_RESOURCE_ADDRESS`** — both carry `SubstateOwnerRule::None`, which `check_ownership` resolves to `false` for every caller (`tracker_auth.rs:127`). `ResourceAccessRules::new()` denies mint, burn, recall, freeze and update_metadata, and every updater is `UpdateRule::Locked` except `metadata_updater` and `nft_data_updater`, which are `UpdateRule::Owner` — unsatisfiable with no owner. So `Mint`, `Burn`, `Recall`, `Freeze`, `UpdateAccessRule`, `UpdateAuthHook` and `UpdateMetadata` are all closed, permanently. The `AllowAll` defaults that remain (`withdraw`, `deposit`, `update_nft_data`) write vaults and `SubstateId::NonFungible` substates, neither of which is coerced.
- **`Template`** — only ever created, via `new_substate` at `impl.rs:4035`. There is no lock or mutation path for a `SubstateId::Template` anywhere in the runtime, and `is_migration` is about calling a migration function, not rewriting the template.
- **`TransactionReceipt`** — up'd at `SubstateVersion::ZERO` by the finalizer (`tracker.rs:555`) and never locked.
- **`ClaimedOutputTombstone`** — `new_substate` only (`impl.rs:3695`).

The coercion is sound. `Evidence::from_lock_intents` and `try_lock_all` consume the same iterator, so evidence and locks cannot disagree, and the other prepare paths derive locks from the accept diff, which already read-locks anything not downed — so the two paths stay consistent.

## New

**1. The coercion has no engine backstop.** — `transaction_manager/manager.rs:681`, `engine/src/transaction/processor.rs:183`

This is the one structural difference from the declaration-based read lock. When a transaction declares `template_x:write`, consensus now takes a read lock, but the engine puts that id in `write_declared_inputs` — the declaration did say write — so `read_declared_inputs` never contains it and `WriteToReadDeclaredInput` cannot fire. The two layers disagree by construction; they only happen to agree because nothing can write those substates today.

The new doc comment on `is_read_only` states the invariant ("an id added here must be impossible to write"), which is the right thing to write down, but it is enforced by review alone. Adding read-only ids to `read_declared_inputs` in `processor.rs` regardless of what was declared would make the engine abort rather than silently write under a read lock, and would cost one line. Worth doing while the invariant is still true — the failure mode if someone later adds a writable id is divergence, not an abort.

**2. `fmt` is red, and it is this branch's doing.** `cargo +nightly-2025-12-05 fmt --all -- --check` flags only `crates/engine/tests/input_intent.rs:122` and `:132` — the two `args![Amount::new(..)]` calls from `5a7c4db6`. Cosmetic, but it is the one failing check.

## Still standing

**3. `Substate::new` widened and `From<u32>`/`From<u16>` added to `SubstateVersion`** — `engine_types/src/substate.rs:81`, `substate_version.rs:94-102`. Still no callers; still serving one test literal that still needs its `u64` suffix.

Nit: `UpdateAccessRule` (`impl.rs:1993`) and `UpdateAuthHook` (`:2035`) still take the write lock before the access check, so they keep the misleading reason that mint, burn and `UpdateMetadata` lost in `5a7c4db6`. Also `read_versioned` (`input_declaration.rs:73`) still takes `V: Into<SubstateVersion>` while `write_versioned` (`:64`) takes a `SubstateVersion` — folded in here since it is one line next to finding 3.

CI at this head: 9 green, 1 red (`fmt`, above), 10 running.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 2 non-blocking, 1 nit · 2 resolved since `50e05ea1` (1 non-blocking, 1 nit) · head `27b09342`

## Resolved

- **`fmt`.** `cargo +nightly-2025-12-05 fmt --all -- --check` is clean at this head — `27b09342` applied exactly the two `args![Amount::new(..)]` reflows the check was asking for.
- **`read_versioned` / `write_versioned` asymmetry.** `f1bcbbab` makes `write_versioned` generic over `V: Into<SubstateVersion>` too, so the two halves match.

One coupling worth naming: that fix leans on the `From<u32>` impl from the finding below. If those impls are dropped, both `*_versioned` constructors go back to taking a `SubstateVersion` — resolve them together, in whichever direction.

## Still standing

**1. The read-only lock coercion has no engine backstop.** — `transaction_manager/manager.rs:681`, `engine/src/transaction/processor.rs:183`. A write-declared read-only substate gets a read lock in consensus, but the engine files that id under `write_declared_inputs`, so `WriteToReadDeclaredInput` cannot fire on it. The layers agree only because nothing can write those substates today; the new doc comment on `is_read_only` records that invariant but nothing enforces it. Adding read-only ids to `read_declared_inputs` regardless of declaration makes the engine abort instead of writing under a read lock — one line, while the invariant still holds.

**2. `Substate::new` widened and `From<u32>`/`From<u16>` added to `SubstateVersion`** — `engine_types/src/substate.rs:81`, `substate_version.rs:94-102`. Callers of the two `From` impls: still the `*_versioned` constructors above and nothing else. The bare-integer `Substate::new` site that motivated the widening still carries its `u64` suffix.

Nit: `UpdateAccessRule` (`impl.rs:1993`) and `UpdateAuthHook` (`:2035`) still take the write lock before `check_resource_access_rules`, keeping the misleading abort reason that mint, burn and `UpdateMetadata` lost in `5a7c4db6`.

CI at this head: 2 green, 2 running.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 0 non-blocking, 1 nit · 3 resolved since `27b09342` (2 non-blocking, 1 nit) · head `397cfaae`

Nothing substantive left open. Detail below.

## Resolved

**1. The read-only coercion now has its engine backstop.** `processor.rs:180` puts a read-only id into `read_declared_inputs` whatever the declaration says, so consensus's coercion and the engine's guard agree by construction rather than by coincidence. Checked the interaction with the widening rule: a read-only id can no longer enter `write_declared_inputs`, so the `retain` two lines down cannot strip it back out — the two rules compose in the right order. Also checked that this cannot break a legitimate transaction: creation goes through `WorkingStateStore::insert`, which does not consult `read_declared_inputs`, so publishing a template is unaffected, and the transaction receipt is `diff.up`'d in `tracker.rs` without touching the store at all.

**2. `From<u32>` / `From<u16>` for `SubstateVersion` are gone.** `209cffcc` removes them, so a bare `u32` no longer converts implicitly and #2687's newtype discipline is back. The generic `*_versioned` constructors still compile against `From<u64>` and the identity conversion.

**3. `UpdateAccessRule` and `UpdateAuthHook` check the updater before taking the write lock** (`impl.rs:1993`, `:2036`), which is the last of the misleading-reason cases. `an_unauthorized_access_rule_update_on_a_read_declared_resource_is_denied_access` is a real guard — it asserts `AccessDenied { UpdateResourceAccessRule(Mint) }` on a read-declared resource, which is exactly the assertion that would have failed before this commit.

With `5a7c4db6` and this one, all five mutating resource actions now evaluate authorization under a read lock and upgrade only when they are actually going to write. That is a better invariant than the one I originally asked for.

## Nit

`Substate::new` is still generic over `V: Into<SubstateVersion>` (`engine_types/src/substate.rs:81`) and `state_store.rs:385` still passes `0u64`. With the `u32`/`u16` impls gone this no longer costs anything — it accepts `u64` and `SubstateVersion` and nothing else — so it is down to taste. `SubstateVersion::ZERO` at that one call site would let the signature go back to being concrete. Not worth a push on its own.

CI at this head: 9 green, 11 running, nothing red.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 1 non-blocking, 2 nits · 0 resolved since `397cfaae` (no new commits; this is a second, adversarial pass) · head `397cfaae`

A second pass focused on soundness rather than the diff. The chokepoint, the consensus/engine agreement, the read-only coercion and the lock narrowing all came back clean under a harder look — details at the end. One thing is worth acting on.

## Non-blocking

**Every client write-declares resources, so the multishard path serialises hot resources — and the PR description understates this.**

`declared_lock_intent` (`manager.rs:682`) honours the declaration, and the deleted heuristic's `|| req.substate_id().is_resource()` → read arm is what used to keep shared resources concurrent. The read-only coercion does not cover user-created resources, only the two genesis ones. Meanwhile every resolver declares the resource a write: `confidential_transfer.rs:279,298`, `stealth_transfer/api.rs:489,682,691,794`, `ootle_sdk_core/src/inputs.rs:749,822`, and `handle_submit_manifest` (`transaction.rs:536`) for every detected dependency.

Scenario: token R lives in SG_R; Alice in SG_A and Bob in SG_B both transfer R. SG_R holds R but neither vault, so it cannot execute and lands in `prepare_multishard_no_execution` for both — taking a **write** lock on R each time, held from prepare to finalisation. Before this PR both took read locks and ran concurrently.

One correction to the sharper version of this claim: the *default* path degrades to a deferral, not an abort. `detect_inputs_use_unversioned` defaults to true (`clients/wallet_daemon_client/src/types.rs:320`), so declarations are unversioned, `requested_version()` is `None` (`lock_intent.rs:308`), and `try_lock_all` records a soft conflict (`pending_store.rs:437-442`). The hard abort is the opt-in case — a client that sets `detect_inputs_use_unversioned: false`, or declares a version explicitly, gets `FailedToLockInputs` instead of waiting.

So: a throughput regression on every cross-shard transfer of a non-TARI token, plus a cheap griefing lever against a hot resource, and an abort rather than a deferral for versioned clients.

The description calls this "a concurrency regression confined to the unreachable path". That reads as *rare*, but `prepare_multishard_no_execution` is unreachable only in the **local swarm**, which runs a single shard group — it is the normal path on any multi-committee network. Worth rewording, since it is the sentence a future reader will use to decide the regression was not worth measuring.

The engine already makes the fix safe: a transfer touches the resource only through `read_lock_substate` (`impl.rs:2463,2520`), and after `5a7c4db6` mint and burn on a non-supply-tracking resource are reads too. So the resolvers can declare `read` for the resource today. Given this rides a reset, landing that with the format change is cheaper than landing it after — otherwise the reset ships with every token transfer serialised on its resource.

## Nits

**`declare_input` drops the second declaration's version when it widens** — `input_declaration.rs:262-275`. `add_input(X:?:read)` then `add_input(X:5:write)` gives `X:?:write`. The doc comment says so, but a caller who pinned a version on the write silently gets an unversioned lock.

**Wire decode and the builder disagree on duplicates** — `tari_bor/src/adapters.rs:122` keeps the *first* duplicate, while `declare_input` widens. A JSON body of `[X:read, X:write]` is signed as `{X:read}` and then aborts with `WriteToReadDeclaredInput`. Self-inflicted only, and identity-by-substate-id is inherited from `SubstateRequirement` rather than introduced here — the preimage is taken over the deduped set, so no two encodings give one id and no id has two intents.

## Categories that came back clean

- **Engine chokepoint.** `new_substates` — the sole source of the up-diff — is written by exactly four paths in `state_store.rs`: `insert` (refuses an existing id), `get_for_mut` (private, reached only through `get_locked_substate_mut`), `mutate_locked_substate_with`, and the two down paths. All four demand a `LockFlag::Write` lock, and `LockedSubstates::try_lock` is called only from `WorkingStateStore::try_lock`, where the guard sits. `LockState` is `Read(n) | Write`, so a read lock id can never satisfy `has_access(Write)`. The one mutation outside a lock — the fee refund in `finalize_fees_and_refunds` — writes into the already-taken `new_substates`, where the vault is present only because `TakeFees` write-locked it.
- **Consensus/engine agreement.** The two predicates are now literally the same expression over the same iterator (`manager.rs:682`, `processor.rs:181`). `resolve_input_locks` derives Write only from `down_iter`, and a down needs a write lock, so a read-declared input cannot come back Write.
- **Read-only coercion.** Re-derived per kind. Both genesis resources: `SubstateOwnerRule::None` resolves to `false` for every caller, all of mint/burn/recall/freeze/update_metadata are `DenyAll`, every updater is `Locked` except metadata and nft_data which are `Owner` (unsatisfiable), and `Resource` exposes no setter for the supply-tracking flag. NFT-data updates write the NFT substate, not the resource.
- **Lock narrowing and the auth-hook boundary.** The hook frame is `OwnComponent`, `check_write_allowed` refuses any new write lock, and `call_invoke`/`bucket_invoke`/`proof_invoke` are forbidden inside it — so the hook cannot change the updater, the access rule or the supply flag that the decision was based on. No TOCTOU across the unlock/relock.
- **Parsing and encoding.** `is_write` has no `#[cbor(default)]`, so a peer cannot omit it; JSON's `serde(default)` is write; `Display`/`FromStr` round-trip and reject `:read:7`, `:?:?` and trailing parts.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 1 non-blocking, 3 nits · 1 resolved since `397cfaae` (1 non-blocking) · head `525dc8fe`

## Resolved

**Resources are declared reads.** This is the reservation from the last pass, and landing it with the format change rather than after it is the right call.

I verified every new read declaration against the engine rather than the comment. The complete set of paths that write a resource substate is: `Create` (`impl.rs:1707`, a new resource, never a declared input), `Mint` (`:1797`) and `Burn` (`:3240`) under supply tracking, `StealthUtxoBurn` under supply tracking (`:2442`), and `UpdateAccessRule` / `UpdateAuthHook` / `UpdateMetadata` (`:2011`, `:2054`, `:2122`). Nothing else. Transfers, deposits, withdrawals, proof creation and `execute_stealth_transfer` (`working_state.rs:2191`) all take a read lock on the resource, and an auth hook runs in an `OwnComponent` frame that cannot take a write lock at all — so none of the declarations changed here can abort a transaction that the engine would otherwise accept.

The split is right where it should be: `resolve_vault` narrows to a read, `resolve_all_vaults` keeps writes because the called method is unknown, and both carry the doc note pointing at `SpecificSubstate` as the escape hatch for a mint or a rule change. `push_unique` now widens instead of keeping the first, matching `declare_input`, with a test.

## Non-blocking

**`declare_written_inputs_only` can turn a working manifest into a state-dependent abort.** — `applications/tari_walletd/src/handlers/transaction.rs:807-833`

The heuristic is "whatever the dry run downed is a write, the rest are reads". The dry run and the real execution run against different state — the dry run happens at submit time, the real one after sequencing — so a manifest whose write set depends on chain state can be declared read for an input that the real execution then writes. The engine aborts it with `WriteToReadDeclaredInput`.

This is not hypothetical: the engine deliberately makes writes conditional. `mint_resource` only touches the resource `if is_total_supply_tracking_enabled` with a comment about avoiding needless mutation, and templates branch on their own state routinely. A manifest that mints only when a balance crosses a threshold dry-runs below the threshold, gets its resource declared read, and aborts when it executes above it. Nothing changed in the manifest and nothing is wrong with it.

The failure is safe — the engine catches it, no divergence — but it converts a deterministic success into a concurrency-dependent abort, and the reported reason points at an input declaration the user never wrote. The fallbacks cover a dry run that *fails* or is only partially accepted; they do not cover one that succeeds and is then stale.

Worth considering: narrow only the inputs whose read-ness is structural rather than observed (a resource the manifest never names in a mint/burn/update instruction), or keep the dry-run narrowing but re-widen on a `WriteToReadDeclaredInput` reject and resubmit once.

Also note every non-dry-run manifest submit now performs an extra full signed dry run before the real submission, where it previously did none. That is a round trip and an execution per submit.

## Nits

- `with_input_intents` (`unsigned_transaction.rs:187`) overwrites the intent of *every* declaration unconditionally. Harmless today because `TransactionSubmitManifestRequest` carries no caller-supplied inputs (`types.rs:352`), so there is no caller intent to clobber — but the method is public and the next caller may have some.
- `declare_input` drops the second declaration's version when it widens (`input_declaration.rs:262-275`); a caller who pinned a version on the write silently gets an unversioned lock.
- `tari_bor/src/adapters.rs:122` keeps the *first* duplicate on decode while the builder widens, so a hand-written `[X:read, X:write]` body signs as `{X:read}` and then aborts. Self-inflicted, and inherited from `SubstateRequirement` rather than introduced here.

CI at this head: 10 green, 10 running, nothing red.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 1 non-blocking, 0 nits · 3 resolved since `525dc8fe` (3 nits) · head `57f360e4`

## Resolved

- **Version kept when widening.** `merge` (`input_declaration.rs:107`) takes `self.version.or(other.version)`, so a pin on either declaration survives, preferring the existing one. The test pins both orders, including the two-pins case where taking the first is arbitrary but deterministic.
- **Duplicate declarations rejected on the wire.** Both layers: `indexset_codec::decode` errors on a repeated element (definite and indefinite arrays, both tested), and `deserialize_input_declarations` does the same for serde. Checked the blast radius of the codec change, since it is shared: `indexset_codec` has exactly two users in the tree, `UnsignedTransactionV1::inputs` and `PrunedUnsignedTransactionV1::inputs`, so nothing else can start rejecting messages it used to accept. And no honest encoder can emit a duplicate — the source is an `IndexSet` — so the stricter decode only rejects what a peer had to construct deliberately. Both struct fields carry the serde attribute; those are the only two `IndexSet<InputDeclaration>` fields, the rest are accessors.
- **`with_input_intents` → `narrow_to_reads`** (`unsigned_transaction.rs:186`), which now only turns a write into a read and leaves a read alone. A caller's read declaration can no longer be clobbered by the dry-run pass.

## Still standing

**The manifest dry run can still turn a working transaction into a state-dependent abort** — `handlers/transaction.rs:807`. `narrow_to_reads` makes the operation strictly one-directional, which removes the clobbering problem, but the direction it keeps is the one that can fail: a write narrowed to a read on the strength of a dry run that ran against earlier state. A manifest whose write set depends on chain state — and the engine makes writes conditional on purpose, `mint_resource` only touching the resource when supply tracking is on — dry-runs one way and executes the other, aborting with `WriteToReadDeclaredInput` on an input the user never declared.

Safe, and caught by the engine rather than diverging. Still worth either restricting the narrowing to inputs whose read-ness is structural rather than observed, or re-widening and resubmitting once on that specific reject.

CI at this head: 3 green, 13 running, nothing red.

sdbondi:
_Posted by Claude._

Re: the manifest dry-run narrowing ([review](https://github.com/tari-project/tari-dan/pull/2640#pullrequestreview-5302513765), `handlers/transaction.rs:807`) — this is accepted as is. A manifest whose write set changes between the dry run and execution aborts with `WriteToReadDeclaredInput`, the engine catches it without divergence, and the user retries. No re-widen/resubmit logic or structural-only narrowing will be added; the concurrency gain on the common case is worth the occasional retry.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 0 non-blocking, 0 nits · 1 accepted as a design decision · head `829b1500`

Closing out the one item I left open. The manifest dry-run narrowing (`handlers/transaction.rs:807`) is accepted as it stands: a manifest whose write set changes between the dry run and execution aborts with `WriteToReadDeclaredInput`, the engine catches it without divergence, and the caller retries. Re-widen-and-resubmit and structural-only narrowing were both considered and rejected — the concurrency gain on the common case is worth the occasional retry. Recording that here so the finding above does not read as outstanding.

`829b1500` itself is a pure rename (`with_intent` → `with_is_write`, five call sites); I checked for stale references and there are none.

Nothing else is open from my side. To summarise what this review verified across the series, since it ran over many heads:

- `WorkingStateStore::try_lock` is a genuine chokepoint — the four paths that write `new_substates` and the two down paths all demand a write lock obtained through it, and a read lock id can never satisfy `has_access(Write)`.
- Consensus and the engine derive the lock from the same expression over the same iterator, and the local-execution path derives Write only from the accept diff's `down_iter`, so a read-declared input cannot come back a write.
- The read-only coercion's premise holds per substate kind: templates, receipts and tombstones have creation-only paths, and both genesis resources deny every mutating action through rules that no caller can change.
- The read declarations the wallet resolvers now emit are correct against the complete set of resource-write paths (Create, Mint/Burn/StealthUtxoBurn under supply tracking, and the three Update actions) — none of which is a transfer, deposit, withdrawal or proof creation.
- An auth hook cannot take a write lock at all, so the read-lock-then-upgrade pattern has no TOCTOU across the hook boundary.
- `is_write` is in all three digests with tests, CBOR requires it, JSON defaults it to write, and duplicate declarations are now rejected on both decode paths.

I cannot approve this PR — GitHub does not accept an approval from the author — so it still needs another maintainer's click.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 0 non-blocking, 0 nits · head `829b1500` · CI red, see below

CI is red at this head on all three `test` shards, but the failure looks unrelated to this PR:

```
FAIL [37.908s] consensus_tests leader_fast_skip::a_returning_validator_is_restored
```

The surrounding log is leader-selection noise — `is not the leader for NodeHeight(13). Expect ... TestAddress(4)` — and then a cascade of missing-justified-block catch-up.

Why I read it as a flake rather than a regression:

- The test lives in `crates/consensus_tests/src/leader_fast_skip.rs`, which this PR does not touch. Its last two commits are #2652, which introduced it, and #2674, which was itself a timing fix for this exact test ("keep the returning validator in view while it is suspended").
- This PR's entire consensus footprint is `transaction_manager/manager.rs` and `foreign_proposal_processor.rs` — input resolution and lock intents. Nothing in leader selection or validator suspension.
- The head it failed on, `829b1500`, renames `with_intent` to `with_is_write` and changes nothing else. A rename cannot regress leader selection.

What I cannot show is an earlier green run of this test on this branch — the run for `57f360e4` was cancelled when the rename landed, and older runs have aged out — so this is "almost certainly a flake" rather than proven. A re-run of the failed jobs would settle it.

The verdict above is unchanged and independent of this: it tracks the findings, and CI being red does not add one.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 1 non-blocking, 1 nit · 0 resolved since `829b1500` (the delta is a rebase plus one new commit) · head `7aefd5b5`

Rebased onto `74c32c48`, plus `7aefd5b5` weighing inputs by declared intent.

## Non-blocking

**Weight is charged on the raw declaration, not the lock actually taken.** — `crates/transaction/src/v1/transaction.rs:207-213`

`calculate_transaction_weight` folds over `self.inputs()` and prices `input.is_write()` at 15. But the lock a shard group actually takes is decided by `decl.is_write() && !decl.substate_id().is_read_only()` — the expression `declared_lock_intent` (`manager.rs:682`) and `processor.rs:181` now share. A write-declared template, receipt, tombstone or genesis resource is read-locked and priced as a write.

That is not rare: write is the default, so any transaction that names a template pays 15 for a lock that excludes nobody. It over-charges, so nothing is exploitable, but it is the third place that needs the same predicate and the only one that does not use it. Same expression, same result, and the weight then means what the commit message says it means.

`self.inputs()` is also the unfiltered set, where `all_inputs_iter` drops TARI and substitutes the read `XTR_REQUIREMENT` — so a client that write-declares TARI is charged 15 for a declaration that is discarded. The same fix covers it, since TARI is `is_read_only()`.

## Nit

**A read declaration still blocks writers, so write-denial got cheaper.** The commit message justifies 5 by exclusion breadth — "a write lock excludes every other transaction naming the substate, a read lock only its writers". Correct, but worth following through: `pending_store.rs:530` (conflict 2) rejects a requested write against an existing read lock for a different transaction, so a read declaration denies a hot substate to every writer for the same prepare-to-finalisation window. The PR's own motivation prices that attack — "The attacker pays 15 weight per input". After this commit the same write-denial costs 5.

Honest readers are the ones who benefit, and an attacker can no longer block readers at all, so this is likely the right trade. It is just a 3× discount on the specific grief the PR set out to price, and the description still quotes the old figure.

## Re-verified after the rebase

The core files came through unchanged except `state_store.rs`, which #2692/#2693 reworked underneath this PR: the two typed down-sets collapsed into one `downed: IndexSet<SubstateId>`, and there is a new `mutate_unpersisted` for substates consensus applies itself. That is a fifth path that mutates state, so the chokepoint claim from my earlier review needs restating rather than repeating: `mutate_unpersisted` opens with `locked_substates.get(lock_id, LockFlag::Write)`, and the new private `down` does the same, so every path still demands a write lock obtained through `try_lock` where the read-declaration guard sits. The guard, `processor.rs`, `manager.rs` and `input_declaration.rs` are all byte-identical to what I checked before.

CI at this head: 11 green, 9 running. The `leader_fast_skip` failure from the previous head has not recurred so far.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 1 non-blocking, 1 nit · 0 resolved since `7aefd5b5` (rebase only) · head `dd9494a5`

CI status, closing two loops from earlier comments.

**The `leader_fast_skip` failure was a flake, now confirmed.** I called it "almost certainly a flake, not proven" at `829b1500` because the run that would have shown it green had been cancelled. All three `test` shards are green at this head with that test passing, on a head whose only difference in that area is a rebase. Nothing to chase.

**The one red check is the known `stall_recovery` flake.** `Test Results (CI)` reports 2661/2662 passing; the single failure is

```
Scenario: All committee members restart after consensus stalls before an epoch boundary
tests/features/stall_recovery.feature:7:3
```

Every actual job is green, including `integration test` — the aggregator reddens on the cucumber annotation alone.

So this head is green other than a pre-existing flake unrelated to input declarations.

The two open items are unchanged, both from `dd9494a5`'s weighting commit: weight charged on the raw declaration rather than the lock actually taken, and the 3× discount on write-denial that pricing a read at 5 implies.

PR #2641, opened 2026-09-17
docs(consensus): state why a report level with ours needs no sync

sdbondi:
Follow-up to #2638, raised from its review.

That PR narrowed the `FallenBehind` sync trigger as well as the vote-drop it was about: the old code reached `Block::record_exists` whenever the reported certificate was at or above ours, the new code returns at equality. That was intended, and the reasoning is worth having beside the check rather than in a merged PR description.

A certificate at height `H` is one we hold, and `update_highest` sets `as_leaf_block()`, which fails when the block is missing — so the certificate we hold at that height always has its block, and we are not behind at `H`. The case that differs is a *different* certificate at the same height, which is equivocation: the branch it names is one #2635's lock rule keeps from committing, not one to sync onto.

Comment only, no behaviour change.


sdbondi (commented):
🆗

PR #2642, opened 2026-09-17
feat(consensus)!: commit the block header to its timeout certificate

sdbondi:
**Draft: needs a state reset. Stacked on #2636 (its commit is the first of two here, since the base branch lives only on the fork); merge after it and after the testnet reset. Review only the top commit.**

## Why

#2636 makes a block's validity depend on a third property of its timeout certificate, `max_high_pc_height`, read by `check_justify_reaches_timeout_certificate`. The block id commits to the justify id, height, parent, roots and metadata hash, never the timeout certificate, and the proposer signs only the block id. Presence and height were pinned by the header's own signed height; the new bound is pinned by nothing. Any committee member can self-sign a `SignedTimeout` for any `(epoch, height, high_pc_height)` and splice it into a certificate it observes: the certificate still passes `check_quorum_certificate_signatures` (each signature verifies against its own reconstructed message), and the block id is unchanged, so the proposer's signature still verifies.

This is not a safety hole. Splicing can only raise the bound, so the mutated block is rejected and leaves no trace. It is a validity rule reading unauthenticated data, and the fix changes block ids, so it is free now and costs another reset later.

## What

- `BlockHeader` carries `timeout_certificate_id: Option<TcId>` (CBOR field 17). From `ProtocolVersion::V1`, `calculate_metadata_hash` commits to it via a new `MetadataHashFieldsV2`, which puts it in the header hash, the block id and the proposer's signature. V0 keeps `MetadataHashFieldsV1`, so V0 block ids stay reproducible, the same gating `calculate_hash` already applies to its own preimage.
- The metadata hash is the route because the base layer never recomputes it: `tari_sidechain::SidechainBlockHeader` takes `metadata_hash` as an opaque `FixedHash` in commit proofs. The header preimage the base layer verifies is unchanged, so no `tari_sidechain` release is needed. The alternative, a `BlockHeaderHashFieldsV3` upstream, is the more visible place for a consensus-relevant field but is a cross-repo change gated on a `tari` release; happy to switch if that is preferred.
- On decode the id is derived from the certificate the block carries, exactly as `justify_id` is, so a splice changes the derived id and fails the signature check. `check_timeout_certificate` additionally requires the header's id to equal the carried certificate's id (or both absent), via the new `check_block_commits_to_timeout_certificate`, so a locally built block cannot disagree with itself.
- Dummy blocks, genesis and the zero block carry `None`. `Block::create` derives the id from its `high_tc` argument, so its callers are unchanged.
- `BlockHeader.ts` binding hand-updated.

## Follow-up for #2636

Its description says "Block ids are unaffected". With this on top that is no longer true under V1; that line now points here.

## Testing

- `crates/storage/src/consensus_models/block_header.rs`: the id round-trips; under V1 a header with, without, and with a different certificate id hashes to three different block ids; under V0 the id is not in the block id; the no-protocol-version decode test now truncates at field 16 rather than dropping the trailing element.
- `crates/consensus_tests/src/timeout_certificate.rs`: a certificate spliced after signing keeps the block id but fails `check_block_commits_to_timeout_certificate`, while carrying the spliced certificate honestly changes the block id; a header and block must agree on whether there is a certificate at all.
- `cargo nextest run -p tari_ootle_storage -p tari_consensus -p tari_ootle_p2p -p consensus_tests`: 136 passed.
- `cargo lints clippy -p tari_ootle_storage -p tari_consensus -p tari_ootle_p2p -p consensus_tests -p tari_state_store_rocksdb --all-targets`: clean.


sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 3 non-blocking, 2 nits · head `4e4beecb` (draft; own PR; several CI legs still pending)

Reviewed the top commit only, as the description asks. The mechanism is sound and I verified each claim it rests on:

- `TimeoutCertificate::calculate_id` hashes the whole certificate through Borsh, `timeouts` vector included, so a spliced `SignedTimeout` does change the id (`crates/consensus_types/src/certificates/timeout_certificate.rs:72`).
- The only wire path that builds a `BlockHeader` from a proto message is `TryFrom<proto::consensus::Block>`, and it derives the id from the carried certificate, exactly as it already does for `justify_id` (`crates/p2p/src/conversions/consensus.rs:593`). There is no other call site of `try_convert_proto_block_header`, and `proto::consensus::BlockHeader` gains no field, so a splice or a drop of the certificate moves the block id and fails `check_block_signature`.
- `on_propose` passes the same `propose_high_tc` to `BlockHeader::create_unsigned` and to `Block::new`, so the header cannot disagree with the block it ships in (`crates/consensus/src/hotstuff/on_propose.rs:670`, `:688`).
- The metadata-hash route checks out: `convert_block_to_sidechain_block_header` hands `metadata_hash` to `tari_sidechain::SidechainBlockHeader` as an opaque `FixedHash` (`crates/consensus/src/hotstuff/commit_proofs.rs:198`), and `calculate_hash`'s `BlockHeaderHashFieldsV1`/`V2` field sets are untouched. No `tari_sidechain` release is implied.
- The V0 path keeps `MetadataHashFields::V1`, so its Borsh discriminant and preimage are unchanged and V0 block ids stay reproducible.

## Non-blocking

**1. V0 still reads the unauthenticated certificate.** `check_justify_reaches_timeout_certificate` is not version-gated (`crates/consensus/src/validations/common.rs:299`), while the binding is (`block_header.rs:346`). Esmeralda's schedule is `[(0, V0), (11086, V1)]` and LocalNet's is `[(0, V0), (5, V1)]`, so a reset that restarts Esmeralda at epoch 0 puts the network back under V0 with #2636's new rule live and nothing pinning `max_high_pc_height` — the exact window this commit exists to close, reopened for the whole pre-activation stretch. LocalNet has the same five-epoch window on every swarm. Two ways out: gate the rule on `ProtocolVersion::V1`, or move the Esmeralda and LocalNet genesis entries to V1 as part of the reset. The second is probably what the reset intends anyway; either way it is worth being deliberate rather than inheriting the schedule.

**2. The new check cannot fire on the wire path, and its doc reads as though it can.** Because the header's id is derived from the carried certificate on decode, `header_tc_id != tc_id` is unreachable for any block that arrived over p2p; it only guards a locally constructed block. That is a reasonable belt-and-braces check, but the doc comment at `common.rs:272-277` frames it as what makes a certificate-reading rule safe, and the actual enforcement is the signature check over the derived id. Saying that in the comment would keep the protection from being lost silently if the field is ever put on the wire.

**3. The sidechain hash-equivalence test only covers the `None` case.** `build_header` in `commit_proofs.rs:311` passes `None`, so `assert_hashes_identically_to_sidechain_header` never runs a V1 header that actually carries a certificate id. The preimage claim in the description is the one thing a base-layer mismatch would break; a `Some(..)` variant of that test would pin it.

## Nits

**1.** `timeout_certificate_id` has `#[n(17)]` but no `#[serde(default)]`, whereas `protocol_version` at `block_header.rs:113` has one. Serde-decoding a header emitted before this field now fails. It matches `signature` at `:81`, which also has none, so this may be deliberate — but the neighbouring field chose the other convention.

**2.** `crates/consensus_tests/src/timeout_certificate.rs:85` swaps `ProtocolVersion::at(NETWORK, TEST_EPOCH)` for a hardcoded `ProtocolVersion::V1`. Necessary for what the new tests assert, but those cases no longer track the version the network is actually configured for; the comment above it explains the swap, which helps.

The follow-up note about #2636's "Block ids are unaffected" line is right and worth carrying into that PR's description before it merges.

crates/p2p/proto/consensus.proto
```diff
@@ -183,7 +184,13 @@ message QuorumCertificate {
 message TimeoutCertificate {
   uint64 epoch = 1;
   uint64 block_height = 2;
-  repeated tari.ootle.common.SignatureAndPublicKey signatures = 3;
+  reserved 3; // was: repeated SignatureAndPublicKey signatures
+  repeated SignedTimeout timeouts = 4;
```
sdbondi: No need to reserve, we'll upgrade all vns together in testnet

sdbondi:
_Posted by Claude._

Addressed in `ffedc834`:

- **Non-blocking 2:** the doc on `check_block_commits_to_timeout_certificate` now says the enforcement is the block signature over the decode-derived id, and that this check guards locally built blocks and holds the derivation in place if the id is ever sent on the wire.
- **Non-blocking 3:** `it_hashes_the_header_identically_to_sidechain_header` now runs each protocol version with and without a timeout certificate id.

Not changed:

- **Non-blocking 1:** agreed that a reset restarting Esmeralda at epoch 0 under `[(0, V0), (11086, V1)]` reopens the window until activation. The fix belongs to the reset itself: the schedule comment already allows a reset network to collapse to a single genesis entry at the newest version, so Esmeralda should become `[(0, V1)]` there. Gating the rule on V1 would make #2636 inert on LocalNet before epoch 5, including in the consensus tests, so I have left the rule ungated.
- **Nit 1:** serde's derive already deserializes a missing `Option<T>` field as `None`, so no `#[serde(default)]` is needed.
- **Nit 2:** no change; the comment stays.

sdbondi (commented):
_Posted by Claude._

**Changes requested** — 1 blocking, 2 non-blocking, 2 nits · 0 resolved since `4e4beecb` (no push; re-derived on new information about the reset) · head `4e4beecb`

Follow-up to my earlier comment. I was told the testnet reset will likely reset everything into `V0`, i.e. what is `V1` today becomes `V0`. That turns non-blocking item 1 into a blocking one for the merge, and supersedes it.

## Blocking (at merge, not at this head)

**The binding is gated to a version the post-reset network will not run.** `calculate_metadata_hash` selects `MetadataHashFieldsV2` only under `ProtocolVersion::V1` and keeps the unbound `MetadataHashFieldsV1` for V0 (`crates/storage/src/consensus_models/block_header.rs:346-361`). If the reset renumbers so that current-V1 semantics become V0, every block on the live network takes the V0 arm and its id does not commit to the certificate id. `check_justify_reaches_timeout_certificate` is not version-gated (`crates/consensus/src/validations/common.rs:299`), so it goes on reading unauthenticated data and this commit ships inert — the splice it describes stays available.

The only thing the gate buys is reproducibility of pre-reset V0 block ids, and a reset discards that chain. Under the stated plan the branch should go away and the certificate id should be committed unconditionally, leaving one field set. The `calculate_hash` branch immediately below it (`:378-409`) is on the same footing, but that one is the schema `tari_sidechain` mirrors, so collapsing it is a separate, cross-repo call and not something I would fold into this PR.

If the reset instead keeps the enum as it is and just moves every schedule to `[(Epoch(0), V1)]`, nothing here needs to change and this item drops. Worth pinning down before this leaves draft, since the two plans want opposite code.

## Unchanged from the previous comment

Non-blocking 2 (the new check cannot fire on the wire path, and its doc reads as though it can) and non-blocking 3 (the sidechain hash-equivalence test only covers the `None` case) stand as written, as do both nits. Everything I verified as sound in that comment is unaffected by this.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 0 non-blocking, 2 nits · 3 resolved since `4e4beecb` (1 blocking, 2 non-blocking) · head `ffedc834` (draft; own PR; CI still pending)

## Resolved

**The V1 gate (blocking, withdrawn).** Raised on the understanding that the reset would renumber current-V1 semantics into V0, which would have left the binding gated to a version nothing runs. Confirmed out of band that V1 is the right home for now and a separate PR will move V1 into V0. Nothing to change here; the note below is the only residue.

**The doc on `check_block_commits_to_timeout_certificate`** (`crates/consensus/src/validations/common.rs:272-279`). Now says plainly that the block signature over the derived id is what stops the splice, that a block off the wire cannot fail this check, and that it guards a locally built block and pins the derivation. That is what the code does.

**Sidechain hash-equivalence coverage** (`crates/consensus/src/hotstuff/commit_proofs.rs:290-297`). `it_hashes_the_header_identically_to_sidechain_header` now runs the cross product of both protocol versions and `None`/`Some(..)`, so the V1-with-certificate case that actually exercises `MetadataHashFieldsV2` is pinned against `tari_sidechain`'s preimage. That was the one path where a divergence would have surfaced only on the base layer.

## Note for the renumbering PR

When V1 moves into V0, `MetadataHashFieldsV2` has to move with it, or the binding stops applying. `check_justify_reaches_timeout_certificate` is not version-gated (`common.rs:299`), so a mismatch there is silent: the rule keeps reading the certificate while the block id no longer commits to it. Worth a line in that PR's description.

## Nits (unchanged)

**1.** `timeout_certificate_id` has `#[n(17)]` but no `#[serde(default)]`, while `protocol_version` at `crates/storage/src/consensus_models/block_header.rs:113` has one. Matches `signature` at `:81`, so possibly deliberate.

**2.** `crates/consensus_tests/src/timeout_certificate.rs:85` hardcodes `ProtocolVersion::V1` in place of `ProtocolVersion::at`. Needed for what the new tests assert, and the comment explains it; it will want revisiting at the renumbering.

Not approving by click: this is a draft on your own PR and CI has not finished. On the code as it stands I have nothing blocking.

crates/p2p/proto/consensus.proto
sdbondi: _Posted by Claude._

Dropped the reservation: `timeouts` is field 3 now. Landed on #2636's branch as `8791d5d3` (the proto is that PR's change), and this PR is rebased on top of it.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 0 non-blocking, 2 nits · 0 resolved since `ffedc834` (rebase only) · head `a50e3048`

This head is a rebase onto #2636's new head, not a change to this PR. Both of this PR's own commits carry the same patch ids as the ones I reviewed at `ffedc834`, so every finding and every resolution from the previous comment stands unchanged.

The one new commit in the range, `8791d5d3` reusing proto field 3 for the timeout certificate's `timeouts`, belongs to #2636 and is out of scope here; it should be reviewed on that PR. Flagging one thing for whoever looks at it there: field 3 was explicitly `reserved` for the old `repeated SignatureAndPublicKey signatures`, and reusing a reserved tag with a different message type means a peer on the old format mis-parses rather than ignores. With the reset and a coordinated upgrade that is fine, but it reverses the marker that was put there to prevent exactly this, so it is worth being explicit about in that PR.

The two nits from `ffedc834` still stand. Nothing blocking on the code as it is.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 0 non-blocking, 2 nits · 0 resolved since `a50e3048` (no push) · head `a50e3048`

All 17 CI legs are green at this head, clippy, the three test shards and the integration test included. Combined with the review at `ffedc834`, whose findings carry over unchanged through the rebase, I have nothing blocking.

This stays comment-only rather than an approval for two reasons that have nothing to do with the code: GitHub does not accept an approval on your own PR, and it is still a draft pending the reset and #2636. It needs another maintainer's click.

crates/storage/src/consensus_models/block_header.rs
```diff
@@ -561,6 +591,7 @@ impl SignedMessage for BlockHeader {
 #[derive(Debug, BorshSerialize)]
 enum MetadataHashFields<'a> {
     V1(MetadataHashFieldsV1<'a>),
+    V2(MetadataHashFieldsV2<'a>),
```
sdbondi: No need for V2, testnet reset

crates/consensus/src/validations/common.rs
```diff
@@ -269,6 +270,52 @@ pub(super) fn check_timeout_certificate<TConsensusSpec: ConsensusSpec>(
...
+///
+/// What makes a rule that reads the certificate read signed data is the block signature: on decode the header's
+/// id is derived from the certificate the block carries (`try_convert_proto_block_header`, as for `justify_id`),
+/// so a certificate swapped in flight moves the block id out from under the proposer's signature. A block that
+/// arrived over the wire therefore cannot fail here; this guards a locally built block, and it holds the
+/// derivation in place should the id ever be sent on the wire instead.
+pub fn check_block_commits_to_timeout_certificate(block: &Block) -> Result<(), ProposalValidationError> {
+    let header_tc_id = block.header().timeout_certificate_id().copied();
+    let tc_id = block.timeout_certificate().map(|tc| tc.calculate_id());
+    if header_tc_id != tc_id {
```
sdbondi: nit: no need to copy the 32 bytes

```suggestion
    let header_tc_id = block.header().timeout_certificate_id();
    let tc_id = block.timeout_certificate().map(|tc| tc.calculate_id());
    if *header_tc_id != tc_id {
```

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 1 non-blocking, 2 nits · 2 resolved since `a50e3048` (2 nits) · head `ceeb5c06` (own PR; clippy, tests and integration still pending)

## Resolved since `a50e3048`

- **Hardcoded `ProtocolVersion::V1` in the test helper.** `block_justifying` now uses `ProtocolVersion::at(NETWORK, TEST_EPOCH)` (`crates/consensus_tests/src/timeout_certificate.rs:85`). It no longer needs V1 because the id is committed under every version.
- **Missing `#[serde(default)]` on `timeout_certificate_id`.** Withdrawn: serde's derive already treats a missing `Option` field as `None`, so the attribute would change nothing.

The earlier note about moving `MetadataHashFieldsV2` during the renumbering no longer applies. `MetadataHashFieldsV1` commits the id for every protocol version, so there is no gate left to lose.

## What I checked at this head

- **Coverage.** `TimeoutCertificate::calculate_id` hashes the whole certificate: epoch, height and every `SignedTimeout`, including `high_pc_height`. A spliced signature therefore always moves the id, and with it `max_high_pc_height`'s binding.
- **Construction sites.** `try_convert_proto_block_header` is the only proto→header path, and `Block::try_from` is its only caller. It derives the id from the carried certificate, as it does for `justify_id`. `rpc_state_sync` decodes through the same `Block::try_from`, so synced blocks are covered too. `on_propose` builds the header and the `Block` from the same `propose_high_tc`. Dummy blocks (`hotstuff/common.rs:321`), genesis and zero all pair `None` with `None`.
- **Persistence.** RocksDB stores the whole `Block` (certificate included) in `BlockCf`. A block served to a syncing peer therefore re-derives the same id.
- **Base layer.** `metadata_hash` is opaque in `SidechainBlockHeader`. `it_hashes_the_header_identically_to_sidechain_header` pins both versions × `None`/`Some`.
- The earlier inline nit (the `TcId` copy) is applied exactly as suggested (`validations/common.rs:286-288`).

## Non-blocking

**GitHub no longer marks this PR as a draft, but the description still says "do not merge before the testnet reset."** Every block id changes, genesis and V0 included, so merging early breaks every running network. If un-drafting was deliberate because the reset is imminent, ignore this. Otherwise, re-draft it.

## Nits

**1.** The `protocol_version` doc says blocks written before the field existed "decode and hash unchanged" (`crates/storage/src/consensus_models/block_header.rs:107`, mirrored in `bindings/src/types/BlockHeader.ts:84`). After this PR the hash part is false for every pre-existing header, because `MetadataHashFieldsV1` gains a field. Suggest limiting the claim to decoding, or to hashing under the V0 schema selection.

**2.** `a_header_and_block_must_agree_on_whether_there_is_a_timeout_certificate` (`crates/consensus_tests/src/timeout_certificate.rs:178`) documents both directions, but it only tests a header that commits to a certificate the block drops. Asserting a `None`-committing header with `Some(tc)` carried would cover the other direction.

On the code I have nothing blocking. This review is comment-only because it is your own PR and three CI legs are still running.

sdbondi (commented):
Reviewed, Tested and Approved

PR #2643, opened 2026-09-17
refactor(consensus): prove the tail of a commit proof with chain links

sdbondi:
## Summary

A commit proof needs certificates only for the 3-chain that commits `b` (`QC(b'')`, `QC(b')`, `QC(b)`); every block below `b` is committed by being its ancestor, so the descent to the committed block is a hash chain. The generator emitted a certificate for every ancestor that justified its parent, which the L1 verifier accepts but must verify signature-by-signature. It now emits the three 3-chain certificates and a single `ChainLinks` element for the rest.

- Fallback: when the committed block is `b`'s direct parent, `b`'s justify is emitted as a 4th QC — a link chain names only the blocks strictly between the last certificate and the header, so that case cannot be expressed with links.
- A block inside the 3-chain that does not justify its parent is now an invariant error rather than silently producing a proof the verifier rejects.
- The proof is now at most 4 elements regardless of how far below `b` the committed block sits. `on_commit_block_recurse` commits every ancestor down to the last executed block, so after a stall the old generator could exceed the verifier's 20-element cap.

Not consensus-breaking: the `tari_sidechain` verifier is unchanged and already accepts this shape, so proofs from either generator validate on either side (foreign proposals, epoch checkpoints, state-sync RPC). Tightening the L1 verifier to exactly 3 QCs is a separate Minotari change.

## Test plan

- [x] `cargo lints clippy -p tari_consensus --all-targets`
- [x] `cargo nextest run -p tari_consensus` — includes new store-backed tests asserting the emitted shape and running it through `check_proof_elements`
- [x] `cargo nextest run -p consensus_tests -E 'test(/multi_shard|foreign|epoch/)'` — foreign proposals round-trip generate → verify


sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 2 non-blocking, 1 nit · head `8c0c649a`

Traced the generator against the verifier in `tari_sidechain-5.7.0-pre.5` (`src/validations.rs::check_proof_elements`, `src/commit_proof.rs::ChainLink::calc_block_id`) and against the commit rule in `crates/consensus/src/traits/block_store.rs::update_nodes`. The change is sound.

### What I verified

- **Element chain matches the verifier's state machine.** The verifier threads `last_parent = qc.parent_id` into the next element and requires the next QC's *justified* block (or the next link's `calc_block_id`) to equal it. Walking the generator: `commit_qc` certifies `b''` and sets `last_parent = b'`; `b''.justify` certifies `b'`, sets `last_parent = b`; `b'.justify` certifies `b`, sets `last_parent = b.parent()`. The links branch then starts at `b.parent()` — exactly where `last_parent` points — and the last link's `parent_id` is the committed block, which is what the verifier compares to `header.calculate_block_id()`. Correct in both the links case and the 4-QC fallback.
- **`justifies_parent()` really is an invariant for the first two hops.** `update_nodes` only commits when `justified_node.parent() == new_locked.id() && *new_locked.parent() == commit_node`, so `b''` and `b'` justify their parents by construction. Turning a violation into an `InvariantError` (commit_proofs.rs:98) cannot fire on a well-formed 3-chain.
- **The old code really did emit an invalid proof in the direct-parent case.** With `!justifies_parent()` and `block.parent() == committed`, the old inner `while` exited immediately and the trailing `if block.id() != committed_block.id()` was false, so it pushed `ChainLinks(vec![])` — which the verifier rejects outright ("Dummy chain must contain at least one element"). The new code cannot emit an empty link vector: the links branch fetches the parent first and only loops while that parent is not the committed block, so there is always ≥1 link. Good fix, and the error is the right outcome for the unreachable variant.
- **`ChainLink` hashing.** `calc_block_id()` is `H(parent_id || header_hash)`, matching `Block::calculate_id` — already asserted by `assert_hashes_identically_to_sidechain_header`.
- **The PR's "not consensus-breaking" claim holds**: the verifier is unchanged and its QC-count gate is `>= 3`, so 3-QC and 4-QC proofs both pass, as do proofs from the old generator.

### Non-blocking

1. **This also fixes a latent `MAX_PROOF_ELEMS` overflow that the description doesn't mention.** `check_proof_elements_num_qcs` caps a proof at 20 elements. In the foreign-proposal path `on_commit_block_recurse` commits every ancestor down to `last_executed`, so after a stall `committed_block` can sit many blocks below `b`; the old generator emitted one QC per justifying ancestor and would blow the cap. The new one emits at most 4 elements regardless of depth. Worth a line in the description — it's a correctness fix, not just a signature-count optimisation.

2. **No test covers the emitted shape.** `commit_proofs.rs:252` only tests header hashing and vote signing; nothing asserts "3 QCs + 1 `ChainLinks`", the 4-QC direct-parent fallback, or that the links are non-empty. The empty-`ChainLinks` bug above survived precisely because this path has no unit coverage, and the integration tests in the test plan exercise only the shallow case. A store-backed test that builds `b''/b'/b` over a few dummy blocks and runs the result through `check_proof_elements` would pin all three shapes.

### Nit

- commit_proofs.rs:81 — "`QC(b'')`, `QC(b')`, `QC(b)`, each justifying the next one's parent" is slightly off: each certificate justifies the *next one's subject*, not its parent (`QC(b'')` justifies `b''`, and `b''` is the parent of nothing in the list — `b'` is). "each certifying the next one's block" reads correctly.

CI: the fast legs pass; `clippy`, `test (1-3/3)` and `integration test` are still pending. Comment-only because GitHub does not accept an approving review on one's own PR.

sdbondi:
_Posted by Claude._

Addressed in `HEAD`:

- **Tests** — three store-backed tests in `commit_proofs.rs` pin the emitted shape (3 QCs + one `ChainLinks` over dummy and real ancestors; 4 QCs for the direct-parent case; exactly 3 for the head) and run each proof through `check_proof_elements`.
- **Nit** — comment reworded to "where each certified block is the parent of the one before it".
- **`MAX_PROOF_ELEMS`** — added to the description.

On the empty-`ChainLinks` point: the old branch could only push an empty vector when a certified block did not justify its signed direct parent, which `check_extends_justify` rejects for a direct child, so it was unreachable rather than latent. The new code errors there instead of producing a proof.

sdbondi (commented):
Approved

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 0 non-blocking, 1 nit · 3 resolved since `8c0c649a` (2 non-blocking, 1 nit) · head `da937f51`

Re-reviewed `69e3542a2` and `da937f518`. Everything I raised is addressed; CI is fully green.

### Resolved

- **Test coverage for the emitted shape** (was non-blocking 2). `69e3542a2` adds three store-backed tests that pin all three shapes. The first one is the right test: `b` is built with `justify = qc_for(&r1)` while its parent is the dummy `d2`, so `b` does not justify its parent and the links branch is genuinely forced — asserted links `[d2, d1, r1]` with `r1.parent_id == committed`. The second pins the 4-QC direct-parent fallback, the third the exact-3-QC case. `assert_verifies` runs each proof through the real `tari_sidechain::check_proof_elements`; I checked `validate_qc` and confirmed that `quorum_threshold: 0` with empty signature vectors passes every gate, so the structural checks (QC count >= 3, `last_parent` chaining, non-empty links, last element == header) are actually exercised rather than short-circuited.
- **Doc comment wording** (was the nit). commit_proofs.rs:82 now reads "where each certified block is the parent of the one before it", which is accurate.
- **`MAX_PROOF_ELEMS`** (was non-blocking 1) — noted in the description.

Also checked the new dev-dependencies: `tari_state_store_rocksdb` depends on `tari_consensus_types`, not `tari_consensus`, so there is no cycle; both are workspace-defined, and `machete` passes. `da937f518` is a pure `Cargo.lock` update adding exactly those two entries to the `tari_consensus` node, nothing else.

### Nit

- `qc_element_ids` also returns ids for the `ChainLinks` arm, so the name undersells what it collects — `proof_element_ids` would read truer. Not worth a push on its own.

CI: all 17 checks pass, including `clippy`, `test (1-3/3)` and the 52-minute `integration test`.

No approving review because GitHub does not accept one on your own PR — this needs a click from another maintainer.

PR #2644, opened 2026-09-17
feat(validator): record diagnostic events in the state store

sdbondi:
## Motivation

Everything a validator knows about its own abnormal moments — leader failures, no-votes, consensus errors, state sync entry and exit, crashes, epoch oracle failures — only ever reaches a log line. Logs are per-process, rotated, often missing on a remote node, and not queryable. Answering "what went wrong, and when" currently starts by asking the operator for files.

This adds a small, always-on, bounded event log kept in the state store, queryable over the validator's JSON-RPC and visible in the web UI.

It is purely local observability: nothing here participates in consensus, is gossiped, or affects state.

## What it records

| Topic | Level | Source |
| --- | --- | --- |
| `consensus.leader_failure` | warn | `on_leader_timeout` |
| `consensus.no_vote` | warn | a no-vote on an otherwise valid local proposal |
| `consensus.block_validation_failed` | warn | a rejected proposal consensus carries on from |
| `consensus.error` | error | an error that stalls this node until someone intervenes |
| `consensus.needs_sync` | warn | an error state sync resolves on its own |
| `consensus.state_transition` | info | every state machine transition |
| `consensus.crashed` | error | the failure transition into `Sleeping` |
| `sync.started` / `sync.completed` | info | entering and leaving `Syncing` |
| `node.started` / `node.shutdown` | info | process lifecycle |
| `node.panic` | error | the process panic hook |
| `node.task_failed` / `node.service_exited` | error/warn | a service task ending |
| `epoch.changed` / `epoch.hash_corrected` | info/warn | epoch manager |
| `epoch.event_error` / `epoch.oracle_stopped` | error | epoch oracle |
| `diagnostics.cleared` / `diagnostics.dropped` | info/warn | the log about itself |

## Design

**Storage.** A new `KeyPrefix::DiagnosticEvents` inside the existing `diagnostics` column family, alongside the no-vote breadcrumbs — no new CF and no migration. The key is a monotonic big-endian sequence, so iteration order is insertion order and the sequence doubles as the event's id and the pagination cursor.

**Schema.** `topic` is a free-form dotted string and `fields` a flat `BTreeMap<String, String>`. Adding an event never touches storage or bindings, and rows written by an earlier binary always decode.

**Write path.** Emitters push onto a bounded channel via a cheap cloneable `DiagnosticsHandle`; a background task batches them into its own write transaction every ~200ms or 64 events. Consensus is never blocked, and a consensus write transaction that rolls back still leaves the record of the failure that caused it. A full channel drops rather than blocks, and the drop count is itself recorded as `diagnostics.dropped`, so silent loss is visible.

**Retention.** Bounded by count (10,000) and age (7 days), both configurable. Pruning is O(deleted): ids only increase, so keeping ids above `newest - max_events` leaves at most `max_events` rows without a counting scan.

**Emission.** `ConsensusHooks` gains `on_state_transition` and `on_no_vote` (both defaulted, so `NoopHooks` and the test harness are unaffected); the validator composes its `DiagnosticHooks` with the existing metrics hook via `CompositeHook`. `crates/consensus` stays storage-agnostic. The epoch manager takes an `Arc<dyn DiagnosticSink>`, defaulting to `NoopSink` for the indexer and tests.

**Panics** are written synchronously through the store rather than the channel, because by the time a panic reaches the hook the writer task may already be gone. Only the first panic is recorded, so a panic raised from inside the store cannot re-enter and deadlock.

## Surface

```toml
[validator_node.diagnostics]
enabled = true
max_events = 10000
max_age_secs = 604800
min_level = "info"
```

- `get_diagnostic_events` — filters on min level, topic prefix and time window; cursor paginated, newest first.
- `clear_diagnostic_events` — deletes what matches the same filters (an empty request clears everything) and records `diagnostics.cleared`.
- A **Diagnostics** page in the validator web UI: level filter, topic filter, per-level counts, expandable context with links to the referenced blocks and transactions, auto-refresh, and a filtered clear behind a confirm.
- The CF is registered in `db_inspector`.

## Testing

- `cargo nextest run -p tari_state_store_rocksdb` — 12 new tests covering append ordering, level/topic/time filtering, cursor pagination, filtered and unfiltered clear, both retention bounds, and that the no-vote rows sharing the column family are untouched.
- `cargo nextest run -p tari_ootle_common_types` — CBOR round-trips including an unknown topic, filter matching, and the `diag_event!` macro.
- `cargo nextest run -p consensus_tests` — 75 passed, confirming the widened hooks trait changes nothing.
- `cargo lints clippy --all-targets` clean over every touched crate; web UI typechecks and builds.

Not verified against a running swarm — left to CI.


sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 8 non-blocking, 9 nits · head `19ba5cd5`

(Comment-only because GitHub refuses approve/request-changes on your own PR. Nothing here blocks; the two I'd want a decision on before this lands on a real node are #1 and #2.)

## What I verified as sound

- **Prefix isolation really does isolate.** `cf.iterator`/`key_iterator` build read options through `create_prefixed_read_opts` (`crates/state_store_rocksdb/src/cf_api.rs:662`), which sets `PrefixRange([46])` — lower bound `46`, upper bound `47`. The prefix-11 no-vote rows are outside the bounds, so they are never handed to the `DiagnosticEvent` value codec (which would otherwise fail the prefix check in `PrefixCodec::decode`). `total_order_seek` is set for the descending case, which is what reverse iteration under a prefix extractor needs. `no_vote_breadcrumbs_in_the_same_cf_are_untouched` covers it.
- **Prune count arithmetic.** Keeping `id > newest - max_events` retains exactly `max_events` rows with no gaps and fewer with gaps, so the bound is an upper bound in both cases; `max_events = 0` collapses the cutoff onto `newest_id` and deletes everything. The age `break` is correct given ids and timestamps rise together, and the comment at `crates/state_store_rocksdb/src/diagnostics.rs:124` states the clock-step-back caveat honestly rather than pretending it away.
- **The writer cannot spin.** `flush` returns before pruning when the batch is empty (`writer.rs:117`), so the 200 ms tick on an idle node does nothing — no write tx, no seek. Prune only ever runs behind a write, which is what makes the "O(deleted)" claim hold in practice.
- **The state machine restructure is behaviour-preserving.** `match (state, &event)` keeps the same arms in the same order; `from` is captured before the match consumes `state`; the only moved field becomes `*target_epoch`; the fallthrough `unreachable!` still fires before the hook. No transition changed.
- **Cursor semantics are exact.** Fetch `limit+1`, derive `has_more` from the overflow, truncate, and take `next_cursor` from the oldest kept row; the query is strictly `id < before_id`. No duplicates, no skipped rows, and "is there a next page" is answered rather than guessed.
- **Nothing is added to a consensus hot path.** The hot hooks (`on_message_received`, `on_local_block_committed`, `on_blocks_committed`, `on_transaction_ready`) have empty bodies in `DiagnosticHooks` and go through a monomorphised `CompositeHook`, so they compile away whether or not diagnostics are enabled.
- Key codec is big-endian `u64`, so insertion order is iteration order. `db_inspector` registers by slugified type name, so the two CFs sharing `cf_names::DIAGNOSTICS` don't collide. No generated artifacts, lockfiles or unrelated files in the diff.

## Non-blocking

**1. `consensus.needs_sync` can never be recorded.** `ConsensusHooks::on_needs_sync` has no call site anywhere in the tree — only the declaration (`crates/consensus/src/traits/hooks.rs:26`) and the four implementors. It was already dead before this PR, but the PR table now advertises a `consensus.needs_sync` topic sourced from it, and `applications/tari_validator_node/src/diagnostics/hooks.rs:66` implements it. An operator filtering on that topic gets silence on a node that is demonstrably behind. Either wire it up — `HotstuffWorker::on_failure` at `crates/consensus/src/hotstuff/worker.rs:846` already branches on `err.is_sync_required()` and has both heights in scope — or drop the row and the handler.

**2. A node that has fallen behind floods the log and evicts everything worth keeping.** `handle_hotstuff_error` (`crates/consensus/src/hotstuff/worker.rs:1452`) calls `on_error` for every inbound message that fails, including `HotStuffError::FallenBehind`, which recurs per proposal for as long as the node is behind. That is an `error`-level row per message. The writer sustains 64 events / 200 ms, so ~30 s of that churns the whole 10 000-row bound. Retention is uniform by id, so the highest-frequency topic evicts `node.started`, the first `consensus.error`, `sync.started` — exactly the rows the feature exists to preserve. A crash loop has the same shape at lower volume: `Sleeping → Resume → Initialising → … → Failure → Sleeping` is ~6 `consensus.state_transition` rows every 5 s, so a node stuck crashing overwrites its own history in a couple of hours. Some coalescing would fix it cheaply — same topic + message within N seconds bumps a `repeat` field on the last row instead of appending — or a per-topic share of the bound.

**3. The panic write can silently overwrite a concurrent append.** `diagnostic_events_append` (`crates/state_store_rocksdb/src/diagnostics.rs:33-42`) is read-max-then-put, and the max comes from a plain `key_iterator`, not `get_for_update`, so it is not conflict-tracked. `CONTEXT.md`'s "Single-writer model" section says at most one write tx mutates at a time and that concurrent writes are a latent risk explicitly out of scope; this adds a second writer (the writer task) and a third (the panic hook, synchronous, on the panicking thread — `panic.rs:173`) to one key family. Two of them computing the same `next_id` means the later commit wins and one event vanishes. The `node.panic` row is the one you can least afford to lose, and `panic.rs`'s own justification for writing synchronously ("the writer task may already be gone") has the mirror case: it may be alive and mid-append. The JSON-RPC `clear` handler is a fourth writer into the same family. A dedicated `next_id` key read with `get_for_update`, or routing the panic through the same store guarded by a mutex, closes it.

**4. Ids restart at 0 after a clear.** `DiagnosticEventRecord` documents `id` as "monotonically increasing, so it doubles as the pagination cursor" (`crates/common_types/src/diagnostics.rs:123`). A clear that removes the newest rows drops `newest_id`, and the next append re-derives `next_id` from what is left. The in-store prune stays correct, but an external client holding a `before_id` across a clear silently pages a different window than it thinks. Persist the counter, or state the caveat where the invariant is claimed.

**5. `install_panic_recorder` ignores `enabled`.** It runs unconditionally at `bootstrap.rs:276`, before `diagnostics::spawn` decides anything. With diagnostics switched off, a panic still writes a row — and nothing ever prunes it, because prune only runs from `Writer::flush`, which never runs. Gate it on `config.diagnostics.enabled`. Separately, it parks a store clone in a process-global `OnceLock`, which keeps the first node's rocksdb handle alive for the process lifetime; the cucumber harness runs several VNs in one process (`integration_tests/src/validator_node.rs:248`), so that first handle outlives its own temp dir.

**6. `bounds` has no consumer.** It costs an extra read tx and two seeks on every `get_diagnostic_events`, is plumbed through `DiagnosticEventStore`, the client response type and the TS binding — and `Diagnostics.tsx` never reads it. Either surface it (an "oldest event" line under the table would be genuinely useful next to the retention bound) or drop it.

**7. Auto-refresh throws away pagination.** `Diagnostics.tsx:240` fires `load()` every 5 s, and `load` issues `before_id: null` and does `setEvents(response.events)`. After the user clicks "Load more" to 150 rows, the next tick snaps the list back to 50 and resets the cursor. Either suspend auto-refresh once `events.length > PAGE_SIZE`, or merge the refreshed head into the existing list by id.

**8. Paging is O(newest → cursor) and decodes every row it skips.** `diagnostic_events_query` (`diagnostics.rs:58-69`) walks the descending `iterator` — which decodes key *and value* — and `continue`s on `id >= before_id`. At the 10 000 bound with 50/page, reaching the tail decodes ~10 000 CBOR values for the last page alone. `range_iterator` with an upper bound at `before_id` would seek straight to the cursor. Minor, but this and `flush`'s write+prune both run synchronously on a tokio worker thread rather than under `spawn_blocking`.

## Nits

9. `classify` arm 2, `(_, _, ConsensusStateEvent::Failure { .. })` (`hooks.rs:121`), is unreachable — the transition table sends every `Failure` to `Sleeping`, which arm 1 already catches.
10. `classify` labels `Syncing --Shutdown--> Shutdown` as `sync.completed` (`hooks.rs:123`). Sync did not complete; the node stopped.
11. `flush` clears `batch` even when `diagnostic_events_append` returned an error (`writer.rs:121-126`), and `take_dropped` was already consumed at the top — so a failed write also loses the record that events were dropped.
12. The `received == 0` exit (`writer.rs:88`) is unreachable: `Writer` holds a `DiagnosticsHandle` clone, so the channel never closes while the task is alive.
13. `recv_many(&mut batch, MAX_BATCH)` appends to a possibly non-empty `batch`, so a batch can reach ~2×`MAX_BATCH` before the `>= MAX_BATCH` check flushes. `MAX_BATCH - batch.len()` matches what the constant says it means.
14. The CBOR round-trip tests (`crates/common_types/src/diagnostics.rs:248,257`) go through `tari_bor` (ciborium/serde), but the store encodes with `DefaultCodec = Minicbor`. The "rows written by an earlier binary always decode" claim rests on the minicbor derive; the rocksdb tests exercise it, these don't.
15. `max_events = 0` is accepted by config and makes every prune delete the batch it just wrote (`prune_with_a_zero_count_keeps_nothing` pins the behaviour). Worth rejecting in config validation.
16. `NoopHooks` re-implements both new defaulted methods with empty bodies (`crates/consensus/src/traits/hooks.rs:163-165`) — redundant now that the trait defaults them.
17. `DiagnosticHooks` builds the event — `format!` plus `to_string` on the error — before `emit` decides to drop it. Free today because none of these are hot, but it stops being free if `on_error` ends up per-message (see #2); an `is_enabled()` guard or a closure-taking `emit_with` costs nothing.

CI: everything green except `integration test` and `test (2/3)`/`test (3/3)`, still pending at this head.

sdbondi:
_Posted by Claude._

Addressed the review in 83d0b53.

**Fixed**

- **Flood / self-eviction (#2).** `handle_hotstuff_error` calls `on_error` unconditionally, including for `FallenBehind`, which recurs per proposal — confirmed. Rather than coalescing (deliberately out of scope for this PR), errors where `is_sync_required()` now record as `consensus.needs_sync` at **warn** instead of `consensus.error`. `on_failure` already treats these as not-an-error and logs them at debug, so recording them at error level was a mislabel as much as a volume problem. A lagging node still writes rows, but it no longer buries the failures an operator filters for at `level=error`. Per-topic retention shares remain a reasonable follow-up.
- **Concurrent append overwrite (#3).** Id assignment is now serialised by a process-wide lock. The wait is bounded (250ms) rather than indefinite, because the panic hook appends from the panicking thread — which may be the thread that panicked while holding the lock. Losing one racing event beats hanging the hook.
- **Panic recorder installed when disabled (#5).** Now gated on `diagnostics.enabled`, so panics no longer write rows nothing will prune.
- **Cursor paging decoded every skipped value (#8).** The query bounds the iterator on a key range instead of `continue`-ing past the cursor, so a deep page seeks straight to its first row.
- **Auto-refresh reset pagination (#7).** It pauses once the reader has loaded older pages, and the control says so.
- **Id monotonicity doc (#4).** Corrected: ids increase with insertion order and restart at zero if the log is emptied. New test `emptying_the_log_restarts_the_id_sequence` pins it.
- **Nits:** `Syncing --Shutdown-->` no longer labelled `sync.completed`; the unreachable `Failure` arm in `classify` is gone.

**Confirmed but not changed**

- **`on_needs_sync` has no call site (#1)** — correct, and it predates this PR. The topic it advertises is now sourced from `on_error` instead, so the table is accurate. `DiagnosticHooks` keeps the `on_needs_sync` impl (it is a required trait method) and it stays correct if the hook is ever wired.
- **`bounds` per query** — one cheap read, kept as part of the response for API clients even though the UI doesn't render it.
- **`recv_many` can grow a batch to ~2×`MAX_BATCH`**, and the `received == 0` arm is unreachable while the writer holds its own handle clone. Both are harmless; the batch cap is a flush trigger, not a limit, and shutdown is the intended exit.
- **`max_events = 0` deletes everything** — an explicit opt-out, and the prune test covers it.

Tests: 13 storage tests, 5 type tests, `consensus_tests` 75/75, clippy clean on every touched crate, web UI typechecks.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 4 non-blocking, 7 nits · 7 resolved since `19ba5cd5` (5 non-blocking, 2 nits) · head `83d0b53a`

Re-verified each fix against the code rather than the commit message. The five I'd called non-blocking are genuinely closed; #2 is narrowed rather than closed, and the append lock introduces one new wrinkle in the exact case it was written for.

## Resolved — what I checked

- **#3, append race.** `APPEND_LOCK` is held across the whole `with_write_tx` (`let _guard`, not `let _`), so the read-max-then-put is serialised. `TryLockError::Poisoned => err.into_inner()` is the right call for a `Mutex<()>` — a panic mid-append must not wedge every later append. The deadlock reasoning holds: `std::sync::Mutex` is not reentrant, the panic hook runs *before* unwinding drops the guard, so a blocking `lock()` on the panicking thread would hang forever. See the new non-blocking item below for what happens after the give-up.
- **#8, cursor decode.** Bounds are right. `rocksdb::IterateBounds for Range<K>` (rocksdb-0.24.0 `src/iter_range.rs:20`) maps to `(Some(start), Some(end))` → `iterate_lower_bound` inclusive, `iterate_upper_bound` exclusive, so `encode_key(0)..encode_key(before)` is exactly the old `id < before_id`. `before_id == 0` gives `start == end`, an empty range — same as the old skip loop. The prefix byte is respected because both bounds come from `encode_key`, which runs the `PrefixCodec` and emits the leading `46`; `range_opts` then sets `total_order_seek` for the descending case, which is what reverse iteration under a prefix extractor needs. The `Box<dyn Iterator<Item = _>>` over a `cf`-borrowing iterator compiles — object lifetime defaults are inferred in expression position, not `'static` (checked with a standalone repro).
- **#5.** `install_panic_recorder` is behind `config.validator_node.diagnostics.enabled` (`bootstrap.rs:277`).
- **#7.** `isPaged = events.length > PAGE_SIZE` is in the effect deps, so the interval is cleared the moment "Load more" lands, and `load()` (Refresh now, a filter change, a clear) drops it back under the threshold and resumes. The label and tooltip say why.
- **#4.** Doc now states the id restarts, and `emptying_the_log_restarts_the_id_sequence` asserts `Some(0)`.
- **Nit, unreachable arm.** Correct to delete: every `Failure` transitions to `Sleeping`, which the arm above already catches.
- **Nit, `Syncing --Shutdown-->`.** New arm sits after the `Sleeping` and `Syncing`-target arms, so the three exits from `Syncing` (Running via `SyncComplete`, `Sleeping` via `Failure`, `Shutdown` via `Shutdown`) land on `sync.completed`, `consensus.crashed` and `consensus.state_transition` respectively. `sync.completed` is still reachable.

## Non-blocking

**1. The give-up path doesn't rescue the panic row in the case it exists for.** If the panicking thread *is* the lock holder, it panicked inside `diagnostic_events_append`, which means it is inside `with_write_tx` with an open rocksdb `Transaction` holding write locks on the keys it already `put`. (That is reachable: `CfApi::encode_key` panics outright on an encode failure — `cf_api.rs:65`.) The hook then spins 250 ms, gives up, and calls `diagnostic_events_append` again on that same thread. The second append derives `next_id` from the highest *committed* key, which excludes the stalled transaction's uncommitted puts, so it tries to `put` a key that transaction already locked — and contends on rocksdb's own pessimistic lock. `TransactionDBOptions::default()` (`store.rs:121`) leaves the lock timeout at the rocksdb default, so that resolves as a timeout error, which `panic.rs:173` swallows with `let _ignore`. Net effect in the target scenario: 250 ms of sleep plus a rocksdb lock wait on the panicking thread, and the row is lost anyway. Cheap fix that is exact rather than heuristic: store the holder's `ThreadId` in an `AtomicU64` next to the mutex and have `acquire_append_lock` return immediately when the current thread is already the holder — O(1), no sleep, and the hook can then skip the append knowingly instead of failing blind. The bounded wait is still right for the *cross-thread* case; it's only same-thread reentry that the timeout can't help.

**2. #2 is narrowed but the same flood still arrives under a different error.** Routing `is_sync_required()` to warn is correct and consistent with consensus's own view — `on_failure` (`worker.rs:846`) uses the identical predicate to decide "not an error", and the three variants it covers (`NeedsSync`, `FallenBehind`, `FutureEpoch`) are all sync-resolvable. Nothing that should stay an error was swallowed. But `handle_hotstuff_error` (`worker.rs:1453`) treats a wider set as the same "we're behind, catch up" condition: `ProposalValidationError::JustifyBlockNotFound` feeds the same catch-up path and is *not* in `is_sync_required()`, so a lagging node whose peers justify against blocks it hasn't got still writes one `consensus.error` per proposal. And the catch-all `ProposalValidationError(err)` arm two blocks down logs at `warn!` with the comment "Failed validations should not crash consensus" — yet `on_error` fired at the top of the function has already recorded it at `error`. Worth either widening the hook's predicate to match what the worker actually treats as recoverable, or keying the level off the worker's own handling rather than off `is_sync_required()` alone. On coalescing: no argument that it's out of scope, but note `handle_hotstuff_error` already does exactly that for `InvalidEpochHash` — `last_epoch_hash_alarm` raises "a single loud alarm per distinct divergence rather than swallowing one warning per rejected proposal" (`worker.rs:1467-1486`). The precedent for the follow-up is 40 lines from the call site.

**3. `on_needs_sync` (#1) — your justification is right, one consequence isn't covered.** Confirmed it is a required trait method (`crates/consensus/src/traits/hooks.rs:26`, no default body), so `DiagnosticHooks` cannot drop the impl. But the topic now has two producers with disjoint field sets: the dead hook emits `consensus.needs_sync` with `local_height`/`remote_qc_height`, and the new `on_error` branch emits the same topic with `error`. If the hook is ever wired, an operator filtering that topic gets two shapes. Cheapest fix is to have the hook's impl delegate, or to give it the same `error` field.

**4. `bounds` (#6).** Fine as an API affordance — noting only that it stays the one part of the response no caller reads.

## Nits

5. The new comment above the `Shutdown` arm says "the transition table only leaves it via `SyncComplete`" (`hooks.rs:135`). That's false on its own reading — `Syncing` is also left via `Failure` and `Shutdown`; it is only true *after* the two arms above have peeled those off. Per the repo's comment rule (present-tense statements that hold for a reader who sees only this file), something like "the arms above have already taken the `Sleeping` and `Shutdown` exits" states the actual invariant.
6. `before_id: Some(0)` is the one cursor edge the new range path changes the mechanism for, and `query_pages_with_the_cursor` stops at `before_id: Some(1)`. One more assert pins it.
7. `acquire_append_lock` can `std::thread::sleep` up to 250 ms, and the writer task calls it from a tokio worker thread. Contention should be nil as the comment says, so this is theoretical — but `flush` was already doing blocking rocksdb I/O there, and this adds a sleep to the same thread.
8. `flush` (`writer.rs:121-126`) still clears `batch` when `diagnostic_events_append` returned an error, and `take_dropped` was consumed at the top — so a failed write also loses the record that events were dropped.
9. The CBOR round-trip tests (`crates/common_types/src/diagnostics.rs:248,257`) go through `tari_bor`, but the store encodes with `DefaultCodec = Minicbor`. The rocksdb tests do cover the real path; these don't test what the forward-compat claim rests on.
10. `NoopHooks` still re-implements both newly defaulted methods with empty bodies (`crates/consensus/src/traits/hooks.rs:163-165`).
11. `DiagnosticHooks` builds the event before `emit` decides to drop it — free today, stops being free if `on_error` ends up per-message (see #2).

No disagreement on the rest of the "confirmed but not changed" list: the `recv_many` over-fill and the unreachable `received == 0` arm are harmless for the reasons you give, and `max_events = 0` is a defensible opt-out that the prune test pins.

CI at this head: `fmt`, `prettier`, `machete`, `file licenses`, `docs build`, `web ui build` and the signed-commit check pass; `clippy`, `ledger app`, `test (1-3/3)` and `integration test` still running.

sdbondi:
_Posted by Claude._

Second round addressed in 890acb6.

**Fixed**

- **The flood was narrowed, not closed (#2).** Correct, and confirmed: `JustifyBlockNotFound` reaches the same catch-up path in `handle_hotstuff_error` without being in `is_sync_required()`, and the catch-all `ProposalValidationError` arm logs at `warn!` ("should not crash consensus") after `on_error` has already recorded it at error. Replaced the single predicate with a `classify_error` that mirrors how consensus itself treats each class:
  - `InvalidEpochHash` / `InvalidProtocolVersion` → `consensus.error` **error** (these stall the node until someone intervenes; consensus raises its own alarm for each)
  - `is_sync_required()` or `JustifyBlockNotFound` → `consensus.needs_sync` **warn**
  - any other `ProposalValidationError` → `consensus.block_validation_failed` **warn**
  - everything else → `consensus.error` **error**

  A lagging node now produces no error-level rows at all, which was the point. PR description table updated to match.
- **The append lock's give-up path didn't rescue the panic row (#1).** Right — waiting cannot help when the holder *is* the panicking thread, and the retry then contended with its own still-open transaction and failed blind after 250ms. A thread-local flag now detects same-thread re-entry and returns immediately; the append returns an error the hook ignores rather than sleeping first. The bounded wait stays for the genuinely-another-thread case.
- **Nits:** the `classify` comment no longer overstates the invariant (it now says the arms above have already peeled off the failure path); added `query_before_the_first_id_returns_nothing` for the `before_id == 0` edge whose mechanism changed.

**Not changed**

- **Two producers for `consensus.needs_sync` with disjoint fields (#3).** Accurate, but `on_needs_sync` has no call site anywhere in the tree, so in practice there is one producer. `fields` is a free-form map by design; when the dead hook is eventually wired, a consumer keying off specific fields would need to handle both shapes either way.
- Coalescing remains out of scope for this PR. Noted that `handle_hotstuff_error` already does it for `InvalidEpochHash` via `last_epoch_hash_alarm` — that's a good template if per-topic rate limiting is added later.
- `bounds` stays as an API affordance even though the UI doesn't render it.

Tests: 14 storage tests, 5 type tests, clippy clean, web UI typechecks.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 2 non-blocking, 10 nits · 4 resolved since `83d0b53a` (2 non-blocking, 2 nits) · head `890acb65`

Both standing items are addressed. The `classify_error` restructure is the right shape, but it surfaces two volume problems one layer up that the `is_sync_required()` version didn't reach.

## Answers to the three questions on the lock

**(a) The flag is never left stuck true.** Walked every path:

| path | flag set? | cleared by |
| --- | --- | --- |
| lock acquired | true | `AppendGuard::Drop` |
| timed out, appends unserialised | true | `AppendGuard::Drop` |
| same-thread re-entry | untouched — returns before the set | outer guard |
| panic inside `with_write_tx` | true | `Drop` during unwind |

The same-thread branch returning *before* the set is the important detail: the flag stays true for the outer append that legitimately owns it. The unserialised path setting it is also right — a panic during an unserialised append must still be caught as re-entry. There is no window between `set(true)` and the `AppendGuard` existing, since only an infallible struct construction sits between them.

One thing worth stating because it is load-bearing and easy to lose in a later edit: the `const { Cell::new(false) }` initialiser is what makes this sound. A `const`-initialised `Cell<bool>` has no lazy init and no destructor, so `IS_APPENDING.with(..)` can never hit `AccessError` — including from `Drop` running during thread teardown, where a panicking `with` inside a `Drop` would abort the process. A non-`const` initialiser would reintroduce that.

**(b) `Err` is the right shape, not `Ok(None)`.** `Ok(None)` already carries a meaning in the trait contract — "`None` if `events` was empty" (`crates/storage/src/diagnostics.rs:58-60`) — so overloading it would make "nothing to do" indistinguishable from "your events were discarded". The two consumers both behave correctly with an `Err`: `panic.rs:173` ignores it (intended — the panic is still on stderr, in the log, and in the panic file written right after `record_panic`), and `writer.rs:121` logs a warning, which is the right outcome for a case the writer can only reach by panicking mid-append.

**(c) Yes, the bounded wait still earns its place.** It is now the sole handler for a *different* thread wedged mid-transaction, where a plain `lock()` would block the panicking thread forever. See nit 5 for what it actually buys in that case.

## Non-blocking

**1. The two variants promoted to `error` are the two consensus deliberately de-duplicates, and the hook fires above the dedup.** `handle_hotstuff_error` calls `self.hooks.on_error(&err)` on its first line (`crates/consensus/src/hotstuff/worker.rs:1452`), unconditionally, *before* the match that handles `InvalidEpochHash` and `InvalidProtocolVersion`. Those two arms exist precisely to suppress repetition — the comment on the first says to "raise a single loud alarm per distinct divergence rather than swallowing one warning per rejected proposal" (`worker.rs:1467-1474`), gated on `last_epoch_hash_alarm`; `alarm_protocol_version_disagreement` does the same with `last_protocol_version_alarm`. Both conditions stall the node indefinitely, so they recur for *every* proposal the committee makes, for as long as the operator hasn't intervened. The diagnostic log bypasses both guards and writes one `consensus.error` row per proposal. The level is right — these genuinely are operator-action conditions — but at that repetition rate the log fills with the same row and evicts the history of how the node got there, which is exactly what someone debugging an epoch-hash divergence needs. This is the flood this PR set out to stop, relocated rather than removed. Cheapest fix: move the `on_error` call in `handle_hotstuff_error` below the match, or into the arms, so it inherits the dedup the worker already implements.

**2. Every proposal validation error now writes two rows, and for the common case both carry the same topic.** `on_receive_local_proposal.rs:285-289` calls `on_block_validation_failed` via `inspect_err` and then *passes the error through*; it propagates out of `dispatch_hotstuff_message` into `handle_hotstuff_error` (`worker.rs:575`, `627`, `1417`), which calls `on_error`. So one validation failure produces:

- row 1 — `consensus.block_validation_failed`, warn, message "A proposal failed validation", field `error`
- row 2 — `classify_error`'s verdict, which for anything that isn't `JustifyBlockNotFound`/`InvalidEpochHash`/`InvalidProtocolVersion` is *also* `consensus.block_validation_failed`, warn, with the error text as the message

Before this commit row 2 was `consensus.error`, so the pair was at least distinguishable. Now an operator filtering that topic sees each failure twice, once generically and once with the detail, at double the storage for the rows this change was meant to make cheaper. This is the ambiguity you asked about, and I'd say it's worth resolving rather than flagging: either drop the `ProposalValidationError` arm from `classify_error` and let `on_block_validation_failed` own that topic (moving the error text into *its* message, which is more useful than "A proposal failed validation" anyway), or give the two producers distinct topics.

On the rest of the mapping: nothing is demoted that shouldn't be. `is_sync_required()`'s three variants plus `JustifyBlockNotFound` all feed the same catch-up path, so warn is consistent with how consensus treats them; the catch-all `ProposalValidationError` arm matches the worker's own `warn!` + "Failed validations should not crash consensus" (`worker.rs:1498-1504`); and everything else falls through the worker's `_ =>` arm, propagates, and crashes consensus into `Sleeping`, so error is right. The one asymmetry — a `ProposalValidationError` reaching `on_failure`'s else-branch is recorded at warn even though it is about to crash consensus — is covered, because `classify` writes `consensus.crashed` at error for the resulting transition.

## Nits

3. The `IS_APPENDING` doc says "Whether this thread is between taking `APPEND_LOCK` and releasing it", and `acquire_append_lock` says "Returns `None` when this thread already holds the lock". Neither holds on the timeout path, where the flag is set without the lock ever being taken. The behaviour is right; the name `IS_APPENDING` is the accurate one and the two doc lines should match it.
4. `StorageError::QueryError` reads as "Query error: diagnostic_events_append re-entered…" for what is a rejected write. `StorageError::general("diagnostic_events_append", "re-entered on a thread that is already appending")` uses the existing helper (`crates/storage/src/error.rs:70`) and renders correctly.
5. The cross-thread fallback is best-effort in the case that triggers it. If the holder is wedged *mid-transaction*, giving up after 250 ms and appending unserialised means computing the same `next_id` and then contending on the rocksdb key lock the stalled transaction holds — resolved by rocksdb's own lock timeout, then swallowed. Worth a word in the comment so a later reader doesn't over-trust the fallback; the 250 ms is still correct as a bound on *blocking*, which is what it's for.
6. `acquire_append_lock` can `std::thread::sleep` up to 250 ms, and the writer task calls it from a tokio worker thread. Theoretical given contention should be nil, but `flush` is already doing blocking rocksdb I/O there.
7. `DiagnosticEvent::new(level, topic, err.to_string()).with_field("error", err)` stores the error text twice, as both `message` and `fields["error"]`. Carried over from the `diag_event!` form, but it doubles the CBOR size of the highest-volume row type — which is the one under retention pressure.
8. `flush` (`writer.rs:121-126`) still clears `batch` when the append failed, and `take_dropped` was consumed at the top, so a failed write also loses the record that events were dropped. More reachable now that `diagnostic_events_append` has a second failure mode.
9. The CBOR round-trip tests (`crates/common_types/src/diagnostics.rs:248,257`) go through `tari_bor`; the store encodes with `DefaultCodec = Minicbor`.
10. `NoopHooks` still re-implements both newly defaulted methods with empty bodies (`crates/consensus/src/traits/hooks.rs:163-165`).
11. `DiagnosticHooks` builds the event before `emit` decides to drop it.
12. The `consensus.needs_sync` two-producer field shapes — agreed this is fine as it stands, keeping it as a nit only so it isn't lost if `on_needs_sync` ever gets wired.

## Resolved, re-verified

- **Append lock give-up.** Re-entry now returns `None` immediately with no sleep and no doomed rocksdb write. The `thread_local!` `Cell<bool>` is a better fit than the `AtomicU64` I suggested — no atomics, no ABA on thread ids, and `Drop` gives exact scoping.
- **`classify` comment.** The new wording states what holds for a reader of this file: the arms above have already taken the failure exit, so shutdown is the only other way out.
- **`before_id == 0`.** `query_before_the_first_id_returns_nothing` pins the empty-range edge, and its comment says why the mechanism changed.

No disagreement with the three items you left unchanged. Coalescing is a reasonable follow-up given `last_epoch_hash_alarm` is the template — though note item 1 above means the two conditions that template covers are currently the *worst* offenders in the log, which strengthens the case for doing it sooner.

CI at this head: `clippy`, `test (1-3/3)` and `integration test` all still running.

sdbondi:
_Posted by Claude._

Third round addressed in f04a528. Both findings were real and I verified the call sites.

- **The stall alarms bypassed consensus's own dedup (#1).** Confirmed: `on_error` is the first line of `handle_hotstuff_error`, above `last_epoch_hash_alarm` and `alarm_protocol_version_disagreement` — guards that exist precisely because those two conditions stall the node and so recur per proposal. So the log was writing an error row per rejected proposal in the one state where its history matters most. `DiagnosticHooks` now draws the same distinction, keyed on the rendered error so it tracks "per distinct divergence" the way the worker's guard does. State is shared across hook clones.

  I chose not to move the `on_error` call below the match in `crates/consensus`, which would be the tidier fix: that changes what `PrometheusConsensusMetrics` counts, and silently altering an existing metric's meaning doesn't belong in this PR. Worth doing on its own.

- **Double row for proposal validation errors (#2).** Confirmed: `on_receive_local_proposal.rs:285` fires `on_block_validation_failed` and passes the error through to `handle_hotstuff_error` → `on_error`, and since both now carry `consensus.block_validation_failed` the pair was an exact duplicate. Resolved the other way round from your suggestion: `on_error` keeps the topic and `on_block_validation_failed` records nothing. `on_error` sees the strict superset — validation failures reaching consensus by paths that never touch the `inspect_err` at 285 would otherwise go unrecorded — and it already carries the full error text, which was your objection to the generic message.

- **Nit taken:** noted that the `const { Cell::new(false) }` initialiser is load-bearing for the `Drop` path. It stays.

On `StorageError::general` vs `QueryError` — leaving as is; re-entry is a programming-error condition, not a query, and neither variant is a natural fit.

Tests: 136 storage + type tests, clippy clean.

sdbondi (commented):
_Posted by Claude._

**Comment only — one blocking finding** — 1 blocking, 0 non-blocking, 10 nits · 1 resolved since `890acb65` (1 non-blocking) · head `f04a528a`

I can't click request-changes on your own PR, so this is comment-only, but item 1 is a blocking finding rather than a refinement: the dedup this commit adds can never fire.

## Blocking

**The stall-alarm key includes the block id, so `is_new_stall_alarm` returns `true` on every proposal.**

Both variants carry a `block_id` and both `Display` impls lead with it (`crates/consensus/src/hotstuff/error.rs:250-259` and `232-241`):

```
"Invalid epoch hash in block {block_id}. Local hash for epoch {epoch} is {local_epoch_hash}, but remote provided {invalid_epoch_hash}"
"Invalid protocol version in block {block_id}: this node's schedule puts epoch {epoch} at {expected_version} but the block was produced under {block_version}"
```

The worker's own guards key on tuples that deliberately *exclude* the block id — `let divergence = (*epoch, *local_epoch_hash, *invalid_epoch_hash)` (`worker.rs:1474`) and `alarm_protocol_version_disagreement(*epoch, *expected_version, *block_version)` (`worker.rs:1494`) — precisely because the whole point is that every rejected proposal is a *different block* carrying the *same* divergence. A stalled node rejects a new block each round, so `err.to_string()` is a different string every time, `last.as_deref() == Some(alarm.as_str())` is never true, and every proposal still writes a `consensus.error` row. The behaviour is identical to `890acb65`; only the appearance of a guard is new, which is worse than no guard, because the next reader will trust it.

The fix is to key on the same fields the worker keys on. Matching the variants directly keeps it honest and lets the two alarms have their own slots, mirroring `last_epoch_hash_alarm` / `last_protocol_version_alarm`:

```rust
fn stall_alarm_key(err: &HotStuffError) -> Option<String> {
    match err {
        HotStuffError::ProposalValidationError(ProposalValidationError::InvalidEpochHash {
            epoch, local_epoch_hash, invalid_epoch_hash, ..
        }) => Some(format!("epoch_hash:{epoch}:{local_epoch_hash}:{invalid_epoch_hash}")),
        HotStuffError::ProposalValidationError(ProposalValidationError::InvalidProtocolVersion {
            epoch, expected_version, block_version, ..
        }) => Some(format!("protocol_version:{epoch}:{expected_version}:{block_version}")),
        _ => None,
    }
}
```

That also collapses `is_stall_alarm` and the key into one function, so the two can't drift apart. Worth a test that two `InvalidEpochHash` errors with the same divergence and different block ids emit one row — that is the assertion that would have caught this.

Separately, while you're in there: the single shared slot lets the two alarm kinds evict each other. If a node were rejecting some proposals for an epoch-hash divergence and others for a version disagreement, they'd alternate and every one would be recorded. The worker uses two slots for the same reason. Unlikely in practice, free to get right alongside the key fix.

## Your question on `on_block_validation_failed` — your reasoning holds

Traced it end to end, and no row is lost. `on_block_validation_failed` has exactly one call site, the tail `inspect_err` in `OnReceiveLocalProposalHandler::handle` (`on_receive_local_proposal.rs:285-289`), which passes the error through rather than consuming it. `handle`'s only caller is `on_proposal_message` (`worker.rs:1371`), via `log_err`, which returns the `Result` unchanged (`worker.rs:1635-1640`). Its `Err` arm calls `handle_hotstuff_error` (`worker.rs:1417`), whose *first* line is `self.hooks.on_error(&err)`, unconditional and before any match. So every error that reaches `on_block_validation_failed` reaches `on_error`.

Your direction is also the better of the two: `on_error` carries the full error text where `on_block_validation_failed` had a fixed "A proposal failed validation", and it covers the entry points at `worker.rs:575` and `627` that the local-proposal `inspect_err` never sees. I'd have lost information with mine.

Agreed on not moving the `on_error` call in `crates/consensus` — changing what `PrometheusConsensusMetrics` counts does not belong in a diagnostics PR, and the hook-local guard is the right place for this once it keys correctly.

Also confirmed: the early return sits ahead of `classify_error`; `is_new_stall_alarm` is only consulted for stall alarms, so the slot can't be thrashed by ordinary errors; `PoisonError::into_inner` avoids a panic-on-poison; the guard is released before `emit`, so there's no nesting with `APPEND_LOCK`; and `derive(Clone)` sharing the `Arc` matches the documented intent.

## Notes — no action needed, recording them so they aren't lost

2. The PR description's event table still lists `consensus.block_validation_failed | warn | on_block_validation_failed`. That topic is now produced by `on_error` via `classify_error`; the hook is a no-op.
3. `on_error` can fire twice for one error on the sync path: `on_proposal_message` routes to `handle_hotstuff_error` (`worker.rs:1417`), which for a future-epoch certificate does `return Err(err)`; that bubbles out of `dispatch_hotstuff_message` into `on_unvalidated_message`'s own `handle_hotstuff_error` call (`worker.rs:571-575`), so `on_error` sees it again. Pre-existing, and it double-counts in the Prometheus hook too — mentioning only because it doubles `consensus.needs_sync` rows for a lagging node.
4. `IS_APPENDING`'s doc ("between taking `APPEND_LOCK` and releasing it") and `acquire_append_lock`'s ("Returns `None` when this thread already holds the lock") don't hold on the timeout path, where the flag is set without the lock being taken.
5. `StorageError::QueryError` for a rejected write; `StorageError::general(..)` fits the existing helper.
6. The cross-thread fallback is best-effort in the case that triggers it — after the 250 ms give-up the unserialised append still contends with the wedged holder's rocksdb key locks.
7. `acquire_append_lock` can `std::thread::sleep` 250 ms on a tokio worker thread.
8. `DiagnosticEvent::new(level, topic, err.to_string()).with_field("error", err)` stores the error text twice, on the highest-volume row type.
9. `flush` (`writer.rs:121-126`) clears `batch` when the append failed, after `take_dropped` was consumed, so a failed write loses the drop record too.
10. The CBOR round-trip tests go through `tari_bor`; the store encodes with `Minicbor`.
11. `NoopHooks` re-implements both newly defaulted methods with empty bodies.
12. `consensus.needs_sync`'s two producers have different field shapes, if `on_needs_sync` is ever wired.

CI at this head: not yet reported.

sdbondi:
_Posted by Claude._

You're right, and thank you for not standing down on it. Fixed in 56aaff5.

**The dedup could never fire.** Confirmed against `crates/consensus/src/hotstuff/error.rs:232-259` — both variants lead their `Display` with `block_id`, and a stalled node rejects a *different block* each round, so `err.to_string()` differed every time and the equality check never matched. Behaviour was identical to the previous commit with a guard that looked like it worked, which is worse than none. That's exactly the answer to the question I asked and I had it backwards.

`stall_alarm(err)` now returns the divergence formatted from the same fields the worker's guards key on — `(epoch, local_epoch_hash, invalid_epoch_hash)` and `(epoch, expected_version, block_version)` — with the block id deliberately absent, and the reason why is in the doc comment so the next reader doesn't reintroduce it. Folded `is_stall_alarm` into it so the predicate and the key can't drift, as suggested.

Also took the second half: each alarm kind gets its own slot, so a node diverging on both the epoch hash and the protocol version records both instead of the two evicting each other.

**Four regression tests** in `diagnostics::hooks::tests`, all of which fail against the previous commit:
- `one_divergence_records_once_however_many_blocks_it_rejects` — three different block ids, same divergence, one row
- `a_different_divergence_records_again`
- `the_two_alarm_kinds_do_not_evict_each_other`
- `errors_sync_resolves_are_not_recorded_as_errors`

Thanks also for tracing the `on_block_validation_failed` path end to end. That was the one I most wanted checked, and knowing `handle`'s only caller reaches `handle_hotstuff_error` with the error unchanged is what makes the direction safe.

The pre-existing double `on_error` on the sync path (`worker.rs:1417` → `571-575`) is worth its own issue — it double-counts in the Prometheus hook too, which is the part that argues for fixing it in `crates/consensus` rather than here.

PR description table was already updated in the previous round and doesn't name the hooks any more.

sdbondi (commented):
_Posted by Claude._

**Comment only — no findings** — 0 blocking, 0 non-blocking, 0 new nits · 1 resolved since `f04a528a` (1 blocking) · head `56aaff5f`

The blocking defect is genuinely fixed. Nothing in this diff regressed. This is my last pass; the carried nits from the earlier rounds stand as they are and none of them need answering.

## The key excludes everything block-varying

Both arms destructure with `..` and format only the worker's own fields:

| variant | `divergence` | worker's guard |
| --- | --- | --- |
| `InvalidEpochHash` | `{epoch}:{local_epoch_hash}:{invalid_epoch_hash}` | `(*epoch, *local_epoch_hash, *invalid_epoch_hash)` (`worker.rs:1474`) |
| `InvalidProtocolVersion` | `{epoch}:{expected_version}:{block_version}` | `(*epoch, *expected_version, *block_version)` (`worker.rs:1494`) |

Exact correspondence, and `block_id` is the only field dropped in each case. I also checked that nothing *retained* varies per proposal while the stall lasts: `local_epoch_hash` is this node's locked hash, `invalid_epoch_hash` and `block_version` are what the committee is proposing (identical across its members), `expected_version` comes from the compiled schedule, and `epoch` is the boundary the divergence sits on. So the key is stable for as long as the divergence is, which is the property the guard needs.

Folding the predicate into the constructor is the right call — `classify_error` and the dedup now read the same function, so a future variant added to one can't be missed by the other.

## The tests — two of the four fail against `f04a528a`, not four

Worked each against the old `err.to_string()` key:

- `one_divergence_records_once_however_many_blocks_it_rejects` — blocks 1/2/3, one divergence. Old key yields three distinct strings → 3 events against an asserted 1. **Fails.** This is the regression test for the bug.
- `the_two_alarm_kinds_do_not_evict_each_other` — four errors on four blocks. Old key yields four distinct strings → 4 against an asserted 2. **Fails**, on the key rather than on the single slot, but it fails.
- `a_different_divergence_records_again` — two divergences on two blocks. Old key yields two distinct strings → 2, which is what's asserted. **Passes against the old code.** It isn't a regression test; it pins the opposite hazard — that the fix doesn't over-dedup and swallow a second, genuinely different divergence. Worth having, just not evidence of the fix.
- `errors_sync_resolves_are_not_recorded_as_errors` — `JustifyBlockNotFound` isn't a stall alarm, so the key never comes into it. **Passes against the old code.** It pins the `classify_error` mapping, which is also worth having.

Not a problem with the tests, only with the claim — flagging it because a later reader treating all four as bug regressions would over-trust two of them. If you want the single-slot bug covered independently of the key, an assertion using the *same* block id for the two kinds would isolate it.

## Nothing else regressed

- The let-chain in `on_error` is right in all three cases: `None` falls through and records, `Some` + new records, `Some` + repeat returns.
- The single-slot → two-slot change is correct — `is_new_stall_alarm` picks the slot by kind and compares and replaces only that one, so neither kind can evict the other.
- Test module checks out: the deps it pulls (`tari_common_types`, `tari_consensus_types`, `tari_engine_types`, `tari_ootle_common_types`, `tokio`) are all already in the VN's `Cargo.toml`; `LeafBlock`'s four fields match the construction; `NumPreshards::P1` exists; `BlockId::new`/`zero` are used the same way elsewhere in the tree; `DiagnosticsHandle::new` is `pub(super)` on `diagnostics::handle`, so `diagnostics::hooks::tests` can reach it; and plain `#[test]` is correct since `emit` uses `try_send` and needs no runtime.
- Only cosmetic thing I noticed, mentioned and dropped: `classify_error` calls `stall_alarm(err).is_some()`, which formats a `String` purely to answer a predicate. Twice per stall-alarm error, on an error path. Not worth a commit.

CI at this head: only the three fast checks have reported so far (signed commits, GitGuardian, check-title — all pass); `clippy`, `test` and `integration test` haven't started reporting.

Standing down here. Good catch on taking the diagnosis but not my mechanism twice over — the `thread_local!` guard and this `StallAlarm` shape are both better than what I proposed.

sdbondi:
_Posted by Claude._

Correction to my previous comment: I said all four new tests fail against `f04a528a`. Two do — `one_divergence_records_once_however_many_blocks_it_rejects` and `the_two_alarm_kinds_do_not_evict_each_other`. The other two pass against the old code: `a_different_divergence_records_again` pins that the fix doesn't over-dedup, and `errors_sync_resolves_are_not_recorded_as_errors` pins the `classify_error` mapping. Both are worth keeping, but only the first two are regression tests for the bug.

sdbondi (commented):
Tested and approved

PR #2645, opened 2026-09-18
feat(stealth)!: a revealed output names its receiver

sdbondi:
Audit finding M8, the last unimplemented item from `engine-audit-2026-09-09`.

## Problem

`stealth_balance_proof64` pins a statement's input commitments, outputs (encrypted data, auth, tags), revealed amounts and witnesses — but `revealed_output_amount` materialises as a bucket in the **submitter's** workspace, and nothing in the statement says who ends up with it.

Key-path inputs are submitter-bound by the spend-key badge. A script-path input (hashlock, timelock, covenant, `AllowAll`) carries its witness inside the statement, so an observer could copy one verbatim into their own transaction and take the revealed funds. Confidential outputs were never at risk: they go where their author encrypted them.

## Change

`StealthOutputsStatement.revealed_output_amount: Amount` becomes `revealed_output: Option<RevealedOutput>`, pairing the amount with the public key authorised to take it:

```rust
pub struct RevealedOutput {
    pub amount: Amount,
    pub receiver: RistrettoPublicKeyBytes,
}
```

The engine requires that receiver's badge in the transaction's auth scope before creating the bucket. The check reads the **base call scope**, exactly as `verify_input_authorizations` does for a key-path input's spend key, so it does not depend on how deeply nested the frame taking the funds is — and is unaffected by #2608's frame-scope revocation.

The type cannot express a revealed amount without a receiver, so key-path and script-path spends need no separate rule. `None` is the only encoding of "no revealed output": a zero amount and a non-canonical receiver are both rejected in `basic_validations`.

`covenant_claims` now enters the balance proof's aux digest. Each claim proves its own partition, but nothing bound the *set*, so a claim could be stripped from a statement whose partition requires one during the pre-seal window in which partial statements circulate. Claim generation moved ahead of proof signing in every builder.

## Keeping the receiver honest

Assembly derives the receiver from a new `seal_public_key`, factored out of `plan_seal` so both go through the same derivation — the key named in the statement is the key the transaction actually carries. The regenerated golden vector shows this: `receiver` equals the fixture's seal-signer key and its auth-signature key.

`finalize_seal` now assigns `is_seal_signer_authorized = true` explicitly instead of resting on the `UnsignedTransactionV1` builder default. That default was load-bearing and undocumented: a promoted stealth seal signer carries no authorization signature of its own, so the flag is what puts its badge in the auth scope — which both its own input's spend and any revealed output naming it depend on.

## Alternatives rejected

- **Tx id / intent commitment in the preimage** — circular. Both hash the instructions, which contain the statement.
- **Signer-set binding** — sound, but binds the statement to a whole signer set the author must know up front, and the SDK's `plan_seal` settles the seal key *after* the balance proof is signed.
- **Rule "script-path ⇒ no revealed output"** — needs no reset, but forces an extra hop and strands a user whose only TARI is script-locked.

## Scope notes

Two items in the original plan turned out not to exist:

- **No retained records to update.** `stealth_balance_proof64_from_parts` still has zero production callers — only a test helper. No retained-record type exists, and the indexer stores no stealth statements.
- **No Ledger change.** `crates/wallet/ledger/{app,common,client}` carry no balance-proof or statement code, only stealth key derivation and seal/auth signing. Putting the receiver in the device's signed message is the parked plan, not this PR.

## Breaking

The statement format and the signed balance-proof message both change. Ships with the testnet reset. Golden vectors regenerated (provenance-only churn on unrelated fixtures reverted).

## Tests

- `a_lifted_statement_cannot_reveal_to_another_submitter` — Alice's statement executed in Mallory's transaction fails naming Alice's key; Mallory's account stays empty.
- Receiver binding, zero-amount rejection, non-canonical receiver rejection, and claim-set binding in `stealth_transfer_statement.rs`.
- SDK-side rejection when a revealed output has no receiver.

Verified: `tari_engine` 92/92 (stealth, spend_script, stealth_fee_estimate); `ootle_sdk_core` + `tari_ootle_wallet_crypto` + `ootle_sdk_ffi_c` 404/404. `cargo lints clippy --all-targets` clean on every affected crate.


sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 2 non-blocking, 1 nit · head `bc9e3183`

Reviewed the whole change at both heads. The core mechanism is sound:

- `check_revealed_output_receiver` (`crates/engine/src/runtime/working_state.rs:2301`) reads `base_call_scope().auth_scope()`, which `TransactionProcessor` seeds from `signers_iter()` — the seal signer's key iff `is_seal_signer_authorized`, plus every auth-signature key. So the badge the statement names is exactly a key that signed the carrying transaction, and it is frame-depth independent as claimed.
- It is the only path that builds the revealed container (`execute_stealth_transfer` is the sole `ResourceContainer::stealth` site downstream of a statement), and it runs before the container is created, so a failure aborts rather than half-applying.
- `basic_validations` runs first in `validate_transfer`, including on the no-inputs/no-outputs early return, so the zero-amount and non-canonical-receiver rejections cover every path.
- `Option<RevealedOutput>` reaches the signed preimage in `StealthBalanceProofParts` and the claim list reaches the aux digest, both under the exhaustive destructuring in `StealthBalanceProofAux`. Claim generation now precedes proof signing in all four producers (`wallet/crypto/src/stealth.rs`, `ootle-rs` generic_impls + `keys/secret.rs`, `ootle_sdk_core/stealth/assemble.rs`) — I checked each.
- The compile break `f77f1b98` left in `tari_consensus` and `tari_transaction_validation` test helpers is fixed by `bc9e3183`; I re-grepped for remaining callers of the changed signatures and found none.

## Non-blocking

1. **`crates/ootle_sdk_core/src/stealth/sign_seal.rs:16-18`** — the module doc still says the flag "is set `true` before building the unsealed tx iff there are zero authorization signatures". `finalize_seal` now sets it unconditionally. The header is the first thing a reader of this file sees about a security-relevant flag, and it now states the opposite of the code.

2. **`crates/ootle_sdk_core/src/stealth/sign_seal.rs:279`** — setting the flag unconditionally also moves `Transaction::main_signer()` (`crates/transaction/src/transaction.rs:166`) for the account-key seal case: it used to return `signatures().first()` (a stealth `c+k` key) and now returns the seal signer (the account key). That value becomes `transaction_signer_public_key` in `TransactionProcessor` (`crates/engine/src/transaction/processor.rs:196`) and reaches templates via `CallerContext`. It looks like the more correct attribution, but it is a behaviour change past what the PR body describes — worth stating, since nothing else in the diff touches it. Related: the invariant is now path-dependent — `crates/ootle_sdk_core/src/tx.rs:73` and the `cosign.rs` module doc still assert the old `auth_signatures.is_empty()` rule for the non-stealth paths.

## Nit

3. **`crates/engine/src/runtime/spend_script_execution.rs:78`** — `SpendScriptContext` still exposes only `revealed_output_amount`, not the receiver. A covenant can now condition on how much is revealed but not on who is authorised to take it, which is the more useful constraint of the two. Not needed for this PR; just noting the gap while the shape is being changed.

CI: `clippy`, the three `test` legs, `integration test` and the fuzz legs were still pending when I looked; everything else green.

crates/wallet/ootle-rs/examples/stealth_transfer.rs
```diff
@@ -151,7 +151,7 @@ async fn main() {
     // statement balances. Two outputs verify well inside the pre-payment credit.
     let (fee_transfer, fee_signers) = StealthTransfer::new(tari_token, &provider)
         .spend_stealth_input(sender_address.clone(), inputs_to_spend[0].commitment())
-        .to_revealed_output(FEE_BUDGET)
+        .to_revealed_output(FEE_BUDGET, *sender_address.account_public_key())
```
sdbondi: Couldnt this be automatically inferred from one of the signers? with an overide if needed.
e.g.

```rust
.to_revealed_output_with_receiver(FEE_BUDGET, *sender_address.account_public_key())
```

sdbondi:
_Posted by Claude._

Thanks — (1) fixed in `634fbbea`. (2) does not hold; (3) agreed and out of scope.

**On (2): `main_signer()` does not move, because the flag's value is unchanged.**

`UnsignedTransactionV1`'s constructors default `is_seal_signer_authorized` to `true` (`crates/transaction/src/v1/unsigned.rs:76,99`), and the old `finalize_seal` only ever *set* it — it had no `else` clearing it. So on the stealth path the flag was already `true` in every seal case before this PR. Making the assignment unconditional is behaviourally a no-op.

The pre-change test says so directly (`sign_seal.rs`, at `f77f1b98^`):

```rust
// (f) is_seal_signer_authorized. Ephemeral (0 auth sigs) ⇒ true. Account-key (1+ auth sigs) ⇒ also
//     true (builder default; the auth sig covers the seal key per the stock semantics).
```

and asserts `decode(&ak).is_seal_signer_authorized()` for the account-key case, which is exactly the "1+ auth signatures" case the finding assumes was `false`. That test is unchanged by this PR and still passes. Nothing reaches `CallerContext` differently.

The reason this matters for the PR rather than being cosmetic is the opposite of what the finding suggests: the flag being `true` with auth signatures present is *load-bearing*, and it was resting on a constructor default with no statement of intent anywhere. A promoted stealth seal signer carries no authorization signature of its own, so the flag is the only thing putting its badge in the auth scope — which its own input's key-path spend already depended on, and which a revealed output naming that key now depends on too. An `else` clause added later in good faith would have broken multi-input stealth spends. Hence the explicit assignment plus the invariant comment.

**On the related note in (2)** — `tx.rs:73` and the `cosign.rs` doc asserting the `auth_signatures.is_empty()` rule: that rule is correct *for those paths*. A public transaction's seal signer is not necessarily an authority, so the flag has to track the signature count there. The stealth path's three seal cases all seal with a key the spender controls, so it does not. The two rules differ because the paths differ, not because one is stale; `634fbbea` says this in the stealth module header so the next reader does not have to rediscover it.

**On (3)** — agreed that a covenant conditioning on the receiver is the more useful constraint, and the gap is real. It needs `SpendScriptContext` to carry the receiver and a predicate to match on it, which is its own change; noting it here rather than widening this PR.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 1 non-blocking, 1 nit · 1 resolved since `bc9e3183` (1 non-blocking) · head `634fbbea`

**Resolved.** Item 1 — `crates/ootle_sdk_core/src/stealth/sign_seal.rs:16-20` now states the invariant `finalize_seal` actually holds (flag `true` in every seal case, because all three seal with a spender-controlled key) and names the public path's differing rule. Read against `finalize_seal` and `crate::tx::seal`; the header and both code paths now agree.

**Still open** (both were questions, neither blocks):

- Item 2 — the flag also moves `Transaction::main_signer()` for the account-key seal case, and so `transaction_signer_public_key` in the engine. Fine to land as-is; I'd just like it stated somewhere that the change is intended rather than incidental.
- Item 3 (nit) — `SpendScriptContext` exposes the revealed amount but not the receiver.

crates/wallet/ootle-rs/examples/stealth_transfer.rs
sdbondi: _Posted by Claude._

Done in `2186463a` — and you were right in a way that caught a bug I'd written.

`to_revealed_output(amount)` now infers the receiver at `prepare()` time from the settled `SealSource`: the account key for an account-key seal, the sealing input's committed spend key for a stealth seal. Both are keys the transaction carries a signature for, so the badge is present by construction. `to_revealed_output_for(amount, receiver)` is the override.

The seal key comes from the UTXO's committed `SpendAuthorization`, so the derivation stays on the key-independent side of the provider boundary — deriving `c+k` from the account secret would have crossed it.

**The bug:** in `examples/adaptor_claim.rs` I had written `.to_revealed_output(FEE, *my_address.account_public_key())` on the *claim* transfer, which spends a script-path input and does not seal with the account key. `PayTo::Conditions` maps to `SpendAuthorization::Script(root)` (`wallet/crypto/src/stealth.rs:93`), so that UTXO commits no spend key and the account key is not automatically a badge there. The engine would have rejected it at execution. Passing a receiver explicitly let a plausible-looking wrong key through; inference does not.

That case is also why two seals now **error** instead of falling back to the account key:

- a script-path sealing input commits a condition root and no spend key, so the satisfying key is known to the caller, not the builder;
- an ephemeral seal is discarded and authorises nothing.

Both name `to_revealed_output_for` in the message. `adaptor_claim`'s claim transfer is the first case and now names the account key explicitly — correct there because that transaction is sealed with `SignatureRequirements::account_key_seal()` chosen outside the builder, which the builder has no way to see.

The other two examples are back to the plain `.to_revealed_output(FEE)`. `ootle-rs` 34/34.

sdbondi (commented):
_Posted by Claude._

**Changes requested** — 1 blocking, 2 non-blocking, 2 nits · 0 resolved since `634fbbea` · head `2186463a`

`2186463a` is a good change in itself — I verified the inference is cryptographically right. `SpendAuthorization::Key(pk)` on a key-path UTXO holds `(c+k)·G` (`pay_to_output_authorization` → `owner_stealth_dh_stealth_address`), which is exactly the public half of the seal secret `plan_seal` derives via `owner_stealth_dh_secret`, and `kdfs.rs:190` asserts that identity directly. `SealSource::AccountKey` → `default_signer_address().account_public_key()` also agrees with the authorizer's own `Seal::AccountKey` branch (`wallet/stealth.rs:70`), since `default_signer_address` delegates to `wallet.default_address()`.

## Blocking

**`crates/wallet/ootle-rs/examples/stealth_spend_conditions.rs:162`** — TX 2 still calls `.to_revealed_output(FEE)`, and it is the same script-path shape you just fixed in `adaptor_claim.rs:167`.

That transfer spends the HTLC output (`SpendAuthorization::Script`, so `spend_key()` is `None` — `unspent_output.rs:83`) and never calls `spend_revealed_input`, so `must_sign_with_account_key` is false and `resolve_inputs` returns `SealSource::StealthInput` with `seal_public_key = None`. `revealed_output()` then takes the script-path arm and returns `UnexpectedError`, and the example's `.prepare().await.unwrap()` panics. `cargo build --examples` will not catch it.

The two examples differ in a way that makes this easy to miss: `adaptor_claim` discards the builder's requirements (`let (claim_transfer, _)`) and seals with `SignatureRequirements::account_key_seal()`, so naming the account key there is right. `stealth_spend_conditions` uses `required_signers` as returned, so its seal is the promoted stealth input and the receiver that would actually work is `stealth_public_key(my_address, nonce)` — not the account key. Either seal it with the account key as `adaptor_claim` does, or name the receiver from `authorizer.seal_public_key()`.

## Non-blocking

1. **`crates/wallet/ootle-rs/src/stealth/builder.rs:317-336`** — the script-path arm rejects a case that works. When a script-path input is promoted to seal, the transaction still seals with a perfectly good badge-bearing key: `WalletStealthAuthorizer::seal_public_key()` (`wallet/stealth.rs:68`) computes it as `stealth_public_key(signer, nonce)`. The key is simply not derivable on `resolve_inputs`' key-independent side. But `prepare()` already crosses that boundary one line later, at `provider.wallet().create_transfer_statement(spec)` — resolving the receiver there would delete both error arms and the `to_revealed_output_for` workaround for this case. As it stands the SDK makes the caller re-derive a key it already knows how to derive, which is how the example above went wrong. Also: no test covers `revealed_output()`'s four arms, and this is the code that decides whether a user's fee bucket is takeable at all.

2. Carried over from `634fbbea` — the unconditional `is_seal_signer_authorized` also moves `Transaction::main_signer()` for the account-key seal case, and so `transaction_signer_public_key` in the engine. Fine to land; I'd just like it stated as intended.

## Nits

3. **`crates/wallet/ootle-rs/src/stealth/builder.rs:197-202`** — the seal key is matched to its input by public nonce alone, while `required_signers` keys on (address, nonce). Two inputs owned by different addresses sharing a nonce is not reachable in practice, but matching the whole `StealthSignerRequirement` costs nothing and removes the question.

4. Carried over — `SpendScriptContext` exposes the revealed amount but not the receiver.

sdbondi:
_Posted by Claude._

Blocking item confirmed and fixed in `ca1891fed`, by taking the route in your non-blocking (1) rather than patching the example — it resolves both at once.

**Blocking — `stealth_spend_conditions.rs` TX 2.** Correct. It spends the script-path HTLC output with no revealed input, so `resolve_inputs` returns `SealSource::StealthInput` with no committed spend key, and `prepare().await.unwrap()` would have panicked. I fixed the identical shape in `adaptor_claim` and missed this one; `cargo build --examples` does not reach it.

**Non-blocking (1) — taken.** `prepare()` crosses the key-provider boundary a line later, so the receiver now comes from `WalletStealthAuthorizer::seal_public_key()`, the same derivation the authorization message is built from. It derives `c+k` from the account secret instead of reading a committed key, so it covers every stealth seal including script-path. Consequences:

- The script-path error arm is gone, and TX 2 needs no change — plain `.to_revealed_output(FEE)` now resolves to `stealth_public_key(my_address, nonce)`, the key that actually seals it.
- `resolve_inputs` no longer captures spend keys or matches them back by nonce, which retires **nit (3)** rather than fixing it.
- `adaptor_claim`'s claim transfer keeps its explicit receiver: it discards the builder's requirements and seals with `SignatureRequirements::account_key_seal()`, which the builder cannot see. That is now the only `to_revealed_output_for` call in the examples, which is the right shape for an override.

The ephemeral arm still errors. Its key is drawn per authorizer and discarded, so a key resolved at `prepare()` would not be the one a later signing pass produces — and such a transfer spends nothing, so it cannot balance a revealed output anyway.

**Tests.** Four cases on `revealed_output()`: stealth seal, account-key seal, named override, and no revealed output. `ootle-rs` 38/38.

**Non-blocking (2) — still does not hold, and I would rather not record it as intended, because nothing changed.** `is_seal_signer_authorized` was already `true` on every stealth path before this PR: the builder defaults it to `true` (`v1/unsigned.rs:76,99`) and the old `finalize_seal` had no `else`. The pre-change test asserts exactly the case in question — `Account-key (1+ auth sigs) ⇒ also true (builder default)` — and passes unmodified. So `main_signer()` returns what it always returned, and `transaction_signer_public_key` is untouched. If you can point at a seal case whose flag value differs between `f77f1b98^` and `ca1891fe`, I will chase it; I could not construct one.

**Nit (4)** — `SpendScriptContext` and the receiver: still agreed, still a separate change.

sdbondi (commented):
_Posted by Claude._

**Approved with reservations** — 0 blocking, 2 non-blocking, 1 nit · 2 resolved since `2186463a` (1 blocking, 1 nit) · head `ca1891fe`

**Resolved.**

- *Blocking* — `stealth_spend_conditions.rs:162` now works unchanged. Routing the receiver through `WalletStealthAuthorizer::seal_public_key` covers the script-path seal: `Seal::StealthInput` derives `c+k` from the account secret at `signer`/`public_nonce`, never reading a committed key, so the promoted HTLC input resolves like any other. Checked the derivation is reproducible for the later real authorizer: `AccountKey` and `StealthInput` are both deterministic in the account secret, and `Ephemeral` — the only case where constructing a throwaway authorizer would draw a different key — is rejected before `stealth_authorizer` is ever called. The ordering is what makes that safe.
- *Nit 3* — the nonce-only seal-key matching is gone with `spend_keys_by_nonce`.

The four new tests cover the resolution outcomes, and `an_unnamed_revealed_receiver_is_the_stealth_seal_key` asserts against `stealth_public_key` directly rather than a hardcoded key, which is the right anchor.

**Still open** (none block):

1. **`crates/wallet/ootle-rs/examples/adaptor_claim.rs:165-166`** — the comment is now stale in its first clause: "the sealing input is script-path, so it commits no spend key for the builder to infer a receiver from" was the reason under `2186463a`, and this commit removed it. The second clause is the one that still holds — that claim is sealed with `SignatureRequirements::account_key_seal()`, outside the builder's own requirements, so the builder would infer the wrong key. Worth trimming to just that, since as written it states a limitation the code no longer has.
2. Carried over — the unconditional `is_seal_signer_authorized` also moves `main_signer()` for account-key seals. Fine to land; stating it as intended is all I'm after.

**Nit**

3. Carried over — `SpendScriptContext` exposes the revealed amount but not the receiver.

The reservation in the verdict is item 1 only. I can't click approve on your own PR; this needs another maintainer.

sdbondi:
_Posted by Claude._

Item 1 fixed in `03d8941da` — good catch, the comment outlived its reason. Trimmed to the clause that still holds: that claim is sealed outside the builder's requirements, so the key the builder would infer is not the one that signs.

Item 2 I'm leaving as it is. I've checked it twice and it does not describe a change: the flag is `true` on every stealth seal both before and after, so `main_signer()` and `transaction_signer_public_key` return what they always did. Recording it as intended behaviour would put a claim in the history that the code does not support. The offer stands — a seal case whose flag differs between `f77f1b98^` and `03d8941d` and I'll chase it — but absent one I'd rather the record stay accurate than tidy.

Item 3 remains a real follow-up: `SpendScriptContext` carries the revealed amount but not its receiver, so a covenant can constrain how much is revealed and not who may take it.

Agreed on approval — this needs another maintainer.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 1 non-blocking, 1 nit · 1 resolved since `ca1891fe` (1 non-blocking) · head `03d8941d`

**Resolved.** Item 1 — `adaptor_claim.rs:165-166` now gives only the reason that still holds: the claim seals with `SignatureRequirements::account_key_seal()` rather than the builder's requirements, so the inferred key is not the one that signs. Checked against the surrounding code; the transfer does discard `required_signers` and build its authorizer from `account_key_seal()`, so the comment and the code agree.

**Still open**, neither blocking and both answerable in a sentence:

- The unconditional `is_seal_signer_authorized` in `finalize_seal` also moves `Transaction::main_signer()` for the account-key seal case, and so `transaction_signer_public_key` in the engine. It reads as a correction, not a regression — I'd just like it on the record as intended.
- Nit: `SpendScriptContext` exposes the revealed amount but not the receiver, so a covenant can constrain how much is revealed but not who may take it.

Nothing else stands. From my side this is ready to land once CI is green — `clippy` and the `test` legs had not yet registered for this head when I looked, so I have not seen them pass on the final code. GitHub won't let me approve your own PR, so this needs another maintainer's click.

sdbondi:
_Posted by Claude._

I went looking for the hole in my own argument this time — I had been reasoning from the constructor default, which leaves open "something clears it later". It doesn't. `is_seal_signer_authorized` has exactly two writers outside `finalize_seal`:

- `cosign.rs:135` — `set_seal_signer_authorized(false)` on an `UnsignedTransactionRecord` for the co-sign path, which never reaches `seal_and_encode_stealth_transfer`.
- `TransactionBuilder::with_seal_signer_authorized` (`builder/mod.rs:319`) — whose only caller in the whole repo is `transaction.rs:565`, a unit test in `crates/transaction`.

So no stealth-path transaction can arrive at `finalize_seal` with the flag `false`, the old `if auth_signatures.is_empty()` had no `else`, and the value is `true` before and after for every seal case. `main_signer()` returns what it always returned and `transaction_signer_public_key` is untouched.

That is why I'm not recording it as intended: there is nothing to intend. If it had been a behaviour change I'd have said so in the commit message — it's the kind of thing that belongs there. Happy to be shown otherwise with a concrete case, but I've now enumerated the writers rather than inferred from the default, and there isn't one.

Nit 3 stays open as a follow-up.

PR #2646, opened 2026-09-18
fix(consensus): admit committee messages only from committee members

sdbondi:
Audit remediation PR A — message admission (H5, H6, M7). Local-only, no consensus rule change.

## H5 — unauthenticated peers drive the collectors

`NewView`, `Vote` and `CatchUpSyncRequest` fell through the catch-all arm of `on_message_validate` with no membership check, so any peer could feed the vote and timeout collectors. All three are now gated on the local committee, as `Proposal` and `CatchUpSyncResponse` already were.

`on_receive_new_view` also verified the reported certificate's 2f+1 signatures *before* establishing that we are the leader for the view it names, giving a sender a quorum verification per message. The leader check now runs first; a NEWVIEW is addressed to that leader and nobody else acts on it.

## H6 — MissingTransactionsResponse accepted from anyone

Responses were matched on a sequential `u32` request id alone. The id was not bound to the peer the request went to, and the "validate that only requested transactions are returned" TODO stood.

`SimpleFixedArray<u32, 32>` is replaced by `MissingTransactionRequests`, which records each in-flight request with its peer and its requested transaction ids. A response must match both the id and the peer asked, and every transaction it carries must be one that was requested — anything else is discarded. The 1000-transaction cap is subsumed by the requested-set size.

## M7 — invalid foreign proposals written to disk

`process_foreign_proposal` persisted a `ForeignProposalRecord` with status `Invalid` for every proposal that failed validation, flagged in-source as debug-only. Any peer could fill the disk with it. No longer persisted.

## Testing

`cargo nextest run -p tari_consensus` — 32/32, including three new tests on the request tracker: a response from a peer we did not ask is not matched, the same id to two peers matches each peer once, and the oldest request is evicted at capacity.


sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 2 non-blocking, 2 nits · head `e144c0fa` (own PR; approve/request-changes unavailable, and `test`/`clippy`/`integration test` still pending)

## Verified sound

- **NEWVIEW leader check reordering.** A NEWVIEW is only ever sent to `next_leader` for `timeout_height` (`on_next_sync_view.rs:69`,`:110`), computed from the same committee and height the receiver uses, and the `high_pc.epoch() != epoch_state.epoch()` guard still runs first. So no legitimate receiver loses the `update_highest` / `FallenBehind`-triggered catch-up that previously ran ahead of the leader check — non-leaders never see the message in the first place.
- **Catch-all arm is still reachable.** `ForeignProposalNotification` and `ForeignProposalRequest` remain, and both legitimately come from foreign committees, so leaving them ungated is correct.
- **`epoch_state` is the right committee.** `msg_relative_view` (`on_inbound_message.rs:445-500`) buffers future-epoch and drops past-epoch NewView/Vote/CatchUpSyncRequest before validation, so the local committee consulted is the one for the message's epoch. Gating `CatchUpSyncRequest` is also safe: every `request_sync` caller picks the peer from the local shard group (`worker.rs:832`,`:1384`,`:1540`).
- **Dropping the receipt-time `Invalid` write orphans nothing.** `ForeignProposalStatus::Invalid` is still set on the three paths that update an already-stored record (`on_receive_local_proposal.rs:253`, `on_receive_foreign_proposal.rs:100`, `on_propose.rs:211`).
- **Eviction test arithmetic.** 33 inserts at capacity 32 evicts id 0 and keeps id 32 — matches `insert`.
- **`SimpleFixedArray` removal is complete**; no other references in the tree, and the trimmed `consensus_models` import leaves nothing unused.

## Non-blocking

1. `crates/consensus/src/hotstuff/on_message_validate.rs:121-134` — the two checks admit duplicates. A peer that was asked for N ids can return N copies of one of them: the count check passes (`N > N` is false) and every id is in the requested set. It is bounded and `get_any_or_build` dedupes downstream, so it is not a DoS, but the invariant is weaker than "only requested transactions are returned" reads. Collecting the ids into a set and requiring `ids.len() == msg.transactions.len()` alongside the subset check would close it exactly.

2. `crates/consensus/src/hotstuff/on_message_validate.rs:178-179` — `insert` runs before `send`, so a send failure leaves a pending entry that occupies one of the 32 slots until it is evicted. Pre-existing, but each slot now also pins a `HashSet<TransactionId>` rather than a `u32`, so the cost of a stuck entry is higher. Inserting after a successful `send` would be strictly better.

## Nits

1. `on_message_validate.rs:95` — the `Proposal` arm's warning still says "Received message from non-committee member"; the new arm names the type via `as_type_str()`. Worth making the three membership warnings read the same way.

2. The PR description carries the "Generated with Claude Code" footer. Repo convention is that PR *descriptions* carry no attribution (comments and reviews carry the AI disclosure instead) — the commit message is correctly clean.

sdbondi:
_Posted by Claude._

All four addressed in b991cf39ea.

**Non-blocking 1 (duplicates)** — the returned transaction ids are now collected into a `HashSet` once and checked two ways: `difference(&request.transactions)` must be empty, and `returned_ids.len()` must equal `msg.transactions.len()`. So each requested transaction comes back at most once and the count check above it is now a real bound on the work one response can ask for. Collecting first also drops the second `calculate_id()` pass the earlier version did.

**Non-blocking 2 (stuck slot)** — `insert` moved after a successful `send`. No race: `handle` and `request_missing_transactions` both take `&mut self` on the worker task, so a response cannot be matched between the send and the insert.

**Nit 1** — the `Proposal` and `CatchUpSyncResponse` warnings now name their message type, matching the new arm.

**Nit 2** — footer removed from the PR description.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 0 non-blocking, 1 nit · 4 resolved since `e144c0fa` (2 non-blocking, 2 nits) · head `b991cf39` (own PR; CI for this head has only registered 3 of the checks so far)

## Resolved, re-verified

1. **Duplicates in a response** (`on_message_validate.rs:127-143`) — ids are collected into a `HashSet`, the subset check runs via `difference`, and `returned_ids.len() != msg.transactions.len()` rejects repeats. The cheap `msg.transactions.len() > request.transactions.len()` check still runs first, so the hashing itself stays bounded by the requested-set size. Both new arms return `Discard`, so their relative order does not matter.
2. **`insert` before `send`** (`on_message_validate.rs:186-199`) — the entry is now recorded only after `send` returns `Ok`. No window opens by doing it in this order: `handle` and `request_missing_transactions` both take `&mut self` on the single worker loop, so a response cannot be validated between the `send` and the `insert`.
3. **Membership warnings** now name `Proposal` and `CatchUpSyncResponse` explicitly, matching the `as_type_str()` arm.
4. **PR description footer** removed.

Both new comments state an invariant in the present tense rather than narrating the change, which is what the repo asks for.

## Nit

`on_message_validate.rs:139-143` — the duplicate guard has no test, while the three peer/eviction rules it sits next to do. A response carrying the same requested id twice is a condition that is easy to pin down directly, and a test for it would fail without the guard.

Nothing else changed; the H5 and M7 conclusions from the previous pass stand unchanged at this head.

sdbondi (commented):
Tested and approved

PR #2647, opened 2026-09-18
fix(consensus): bound the catch-up responder, view buffer and vote store

sdbondi:
Audit item H7: three inbound paths sized the work this node does by what peers send rather than by a local limit. PR B of the consensus audit series (independent of #2646).

The three paths do not share a threat model, and only one of them is reachable by an arbitrary peer:

| path | who can reach it | what bounds it now |
|---|---|---|
| catch-up responder | local committee members (#2646 gates `CatchUpSyncRequest` in `on_message_validate`, and the responder runs only on a `Ready` message) | 5 concurrent responses |
| view buffer | any connected peer — buffering happens in `MessageBuffer::next`, before validation | 64 MiB / 2000 messages, 20-view window |
| vote store | local committee members (`Vote` is gated in the same arm) | votes within 20 views of ours |

### Catch-up requests
`OnSyncRequest::handle` spawned a task per request, and each task walks the block store and streams every block up to our leaf. It now runs through `BoundedSpawn` — the same mechanism the foreign-proposal responder uses, at 5 permits rather than its 20 — and a request arriving with no permit free is dropped with a warning. The task body moved into an associated fn, which is most of the diff in that file.

### Inbound message buffer
The buffer was capped at 1000 messages per view across 100_000 views, and the view a message names is chosen by its sender, so the effective bound was the product. It is now a `ViewBuffer` bounded the way the inbound queue it drains from is bounded: a size budget of 64 MiB as the primary bound (half `max_consensus_messaging_queue_bytes`, whose reservation is released once a message reaches consensus) with a 2000-message count as the secondary guard. A message's decoded size is not measurable here, so each is charged what it could have arrived as — 4 MiB, the wire ceiling, for the variants with unbounded payloads, and a fixed allowance for those whose decoded size is bounded by their shape.

`MissingTransactionsRequest` only qualifies for that fixed allowance because its id set is now capped where it is decoded (`MAX_REQUESTED_TRANSACTIONS`), rather than in `on_message_validate`, which runs after buffering.

Within a budget the furthest view is evicted to admit a nearer one, and an item that is itself the furthest is refused. Because views order epoch-major, eviction alone cannot protect an epoch-rollover proposal from same-epoch messages naming absurd heights, so those are refused at admission: within our epoch only views within 20 of ours are buffered. Heights restart each epoch, so next-epoch messages are admitted whatever height they name.

### Forward vote buffering
`clear_votes_before` only trims below the current view, so votes for future views accumulated. The collector now keeps only votes within 20 views of the current one, and both collectors apply that window before verifying the signature. A node lagging further than that reaches those views by importing blocks, not by certifying them from buffered votes.

### Tests
`cargo nextest run -p tari_consensus -p tari_ootle_p2p` — 49 passed. New unit tests cover the `ViewBuffer` budgets (item count, size, eviction order, refusal of the furthest and of an oversized item), both lookahead windows, and the request decode cap; each guard was confirmed to fail its tests when disabled.


sdbondi (commented):
_Posted by Claude._

**Comment only** — 1 blocking, 3 non-blocking, 4 nits · head `5a00e539`

(Comment rather than a review decision: GitHub refuses approve/request-changes on your own PR.)

## Blocking

**1. The inbound buffer is bounded by message count, not by bytes — the cap admits ~8 GiB.**
`crates/consensus/src/hotstuff/on_inbound_message.rs:81`

`MAX_BUFFERED_MESSAGES = 2_000` counts messages. A `HotstuffMessage` on the direct messaging
protocol may be up to 4 MiB (`networking/libp2p-messaging/src/codec/prost.rs:13`), so the buffer's
worst case is 2000 × 4 MiB ≈ 8 GiB — and it is a *steady* 8 GiB, because at capacity `insert` evicts
to make room rather than refusing, so the buffer stays full for as long as the flood lasts, and
entries for a far-future height are never reached by `discard_before`.

Two things make that reachable by anyone: `push_to_buffer` runs inside `MessageBuffer::next`, i.e.
*before* `on_message_validate`'s committee check (`on_message_validate.rs:160-171`), and a
same-epoch `Proposal` is routed to `Future` on its attacker-chosen `justify().height()` with no
signature verified at that point (`on_inbound_message.rs:369-388`).

The queue this buffer drains from already does it the other way: `message_queue(count, bytes)` with
`max_consensus_messaging_queue_bytes` defaulting to **128 MiB**
(`applications/tari_validator_node/src/bootstrap.rs:160-168`,
`applications/tari_validator_node/src/config.rs:152-157,177-179`), and the comment at
`bootstrap.rs:160-162` says the count there is only "a secondary guard against a flood of tiny
messages". `networking/core/src/message_mode.rs:100-107` spells out the reasoning verbatim: "a count
alone cannot do the job … a count low enough to bound worst-case memory is far too low for ordinary
traffic, while a count high enough for ordinary traffic admits that many maximum-size messages."

So the node hands a message to a downstream buffer whose budget is 62× the transport budget it came
from. `ViewBuffer` wants the same shape as `QueueAccounting`: a byte budget as the primary bound,
with the count kept as the secondary guard. That needs a size accessor on the buffered item, which
is a small trait or a closure passed to `ViewBuffer::new`.

Until that lands, the largest of the three paths this PR is remediating is still an OOM vector,
which is the thing H7 is about. The other two paths are fine.

## Non-blocking

**2. "Furthest view" inverts at an epoch boundary, so the eviction policy protects the wrong thing there.**
`crates/consensus/src/hotstuff/view_buffer.rs` (`View` ordering + `insert`)

`View` is ordered epoch-major, so *every* message for `current_epoch + 1` sorts above *every*
message for `current_epoch`, whatever its height. A peer that floods same-epoch proposals carrying
`justify().height() = u64::MAX` fills the buffer with views consensus will never run; the
epoch-rollover proposal, buffered at `View(current_epoch + 1, pc_height)`
(`on_inbound_message.rs:343-348`), is then the furthest key, so `insert` **refuses** it rather than
evicting the junk.

The doc comment's claim — "a flood aimed at distant views cannot crowd out the nearest views" —
holds within an epoch but not across one. The old per-view cap would have admitted that rollover
proposal (it was a fresh view entry, well under the 100k view cap). The blast radius looks small:
epoch advance is driven by the epoch manager, not by this message, and once `current_epoch`
increments the proposal is `Current` and bypasses the buffer, so this costs a re-request rather than
a wedge. Still, the ordering is worth making explicit — "nearer in the order consensus will consume"
is not what `Ord` on `(epoch, height)` gives you when heights are unbounded.

**3. The vote lookahead runs after the signature has already been verified.**
`crates/consensus/src/hotstuff/vote_collector/collector.rs:50`

Both collectors call `validate_vote` — which does the Schnorr verify — and only then `collect_vote`,
which drops the vote on height (`proposal_collector.rs:74-90`, `timeout_collector.rs:59-70`). The
height comparison needs no crypto. Hoisting it ahead of the signature check (or into `validate_vote`
alongside the existing epoch check) makes the bound save the work it is there to save; as placed it
bounds memory but not CPU.

**4. The 5 catch-up permits have no per-peer share and no timeout.**
`crates/consensus/src/hotstuff/on_catch_up_sync_request.rs:23,195`

A permit is held for the entire stream, and the stream awaits `outbound_messaging.send()` per block.
Five concurrent requests from one committee member — or five sends to stalled peers — occupy every
permit, and every other peer's catch-up is dropped with the warning at line 65. Catch-up is the
recovery path, so starving it is the case worth caring about. `on_receive_foreign_proposal.rs:231`
has the same shape, so this is consistent with the existing convention rather than new; a per-peer
cap of 1–2, or a send timeout, would close it for both.

## Nits

**5.** `on_inbound_message.rs:79-80` — the sizing rationale ("Sized for a committee voting several
views ahead of a lagging local view") doesn't describe what lands in this buffer. Same-epoch votes
for any future height return `Current` (`on_inbound_message.rs:415-417`) and are never buffered.
What is buffered is proposals and next-epoch messages.

**6.** PR description: "the same `BoundedSpawn` the foreign-proposal responder uses (5 permits)" —
that responder uses 20 (`on_receive_foreign_proposal.rs:71`). The mechanism is shared; the figure
isn't. Worth rewording so nobody reads it as parity.

**7.** `on_catch_up_sync_request.rs:56` — `// Spawn: Dont block consensus when serving requests.`
Typo, and it narrates what the next line does.

**8.** `view_buffer.rs` `insert` warns per eviction while `push_to_buffer` warns per refusal. A
generic container emitting WARN is a slightly odd split of responsibility, and under sustained
pressure it is one warn line per inbound message — peer-driven log volume.

## Verified sound

- `ViewBuffer` length accounting: `insert` (including the eviction loop), `pop_front`,
  and `discard_before` all keep `len` consistent, and empty deques are removed on both paths, so
  `pop_back()` in the eviction loop can never fire on an empty deque and underflow `len`.
- `discard_before`'s `split_off` direction is right: `split_off(&view)` returns `>= view`, which
  becomes the new buffer, and the retained `< view` half is what gets counted and dropped.
- `try_spawn` releases the permit on task completion (`bounded_spawn.rs:126-131`), and
  `OnSyncRequest` is constructed once (`worker.rs:253`), so the semaphore isn't reset per epoch.
- The height-only vote lookahead is sufficient: the epoch axis of the vote store is already closed
  by `current_epoch != vote.epoch()` in both `validate_vote` implementations
  (`proposal_collector.rs:135`, `timeout_collector.rs:99`), so votes can't accumulate along epochs.
  With committee size 40 and a 20-view window, the store is bounded at ~800 entries.
- The catch-up task body is a faithful move — no logic changed, only the `spawn` wrapper and rustfmt.
- The new tests key on the condition rather than a rendered message, and each would fail with its
  guard removed.

sdbondi:
_Posted by Claude._

Addressed in `a562132`.

**1 (blocking) — byte budget.** Confirmed: the adapter moves the message out of `InboundMessage` (`applications/tari_validator_node/src/p2p/services/messaging/inbound.rs:74-76`), dropping the permit, so nothing downstream charges the 128 MiB budget. `ViewBuffer` now takes both budgets — 64 MiB primary, 2000 messages secondary — in the shape `QueueAccounting` uses.

One deviation: the size is a charge, not a measurement. The decoded size isn't measurable in this crate and the wire size isn't plumbed through `InboundMessaging`, so each message is charged what it could have arrived as — `UNBOUNDED_MESSAGE_SIZE` (4 MiB, the wire ceiling) for `Proposal`/`CatchUpSyncResponse`/`ForeignProposal`/`MissingTransactionsResponse`, `FIXED_MESSAGE_SIZE` (64 KiB) for the fixed-shape variants, whose decoded size is bounded by their shape whatever the wire carried. That bounds the worst case at the budget and costs nothing to compute. Carrying the true wire length through the trait is the more accurate option and is a bigger change than this PR; say the word if you'd rather have it.

**2 — epoch-major inversion.** Fixed rather than only documented, since the byte budget doesn't help here: a same-epoch flood still fills it and the rollover proposal is still the furthest key. `push_to_buffer` now refuses same-epoch views more than `MAX_VIEW_LOOKAHEAD` (20) ahead, so the `justify().height() = u64::MAX` junk never occupies the buffer. Next-epoch messages are admitted at any height — heights restart each epoch (`BlockHeader::genesis` is `NodeHeight::zero()` per epoch), so the window says nothing about them. The ordering caveat is now stated on `ViewBuffer` as a constraint on its caller.

**3 — vote lookahead after the signature verify.** Fixed: `exceeds_vote_lookahead` runs first in both collectors, ahead of `check_eligibility` and `validate_vote`. The guard stays inside `collect_vote` too — that's the one the memory bound depends on, and it's what the test drives.

**4 — no per-peer share or timeout on the catch-up permits.** Not taken here. It's the same shape as the existing foreign-proposal responder, so it belongs in one change covering both rather than half of it in this PR. Noted for the audit list.

**5, 6, 7, 8** — reworded the count-budget rationale to say what actually lands in the buffer; corrected the "(5 permits)" wording in the PR description; dropped the typo'd comment; and moved the drop/eviction logging out of `ViewBuffer` into `push_to_buffer` at `debug`, so a generic container doesn't log and peer-driven drops don't produce warn volume.

Tests: 45 pass. New cases cover the size budget (many small items, a few large ones, an item larger than the budget) and the view window; each was confirmed to fail with its guard disabled.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 1 blocking, 3 non-blocking, 2 nits · 5 resolved since `5a00e539` (1 non-blocking, 4 nits) · head `a5621329`

(Comment rather than a review decision: GitHub refuses approve/request-changes on your own PR.)

## Blocking — carried forward, narrowed

**1. `MissingTransactionsRequest` is charged `FIXED_MESSAGE_SIZE` but its payload is unbounded at decode.**
`crates/consensus/src/hotstuff/on_inbound_message.rs:95-96,583-596`

The size budget is the right shape and it closes the vector for three of the four variants that can
actually be buffered. It does not close it for the fourth.

`buffered_message_size` puts `MissingTransactionsRequest` in the `FIXED_MESSAGE_SIZE` arm (64 KiB),
described as "a message whose payload is a fixed set of fields". Its payload is
`transactions: HashSet<TransactionId>`, and the decoder applies no cap
(`crates/p2p/src/conversions/consensus.rs:418-432`) — the 1000-id limit is at
`on_message_validate.rs:149-153`, which runs *after* buffering, same as the committee check. So the
message can arrive at the 4 MiB wire ceiling (~120k ids, ~5 MB once it is a `HashSet`) and be
charged 64 KiB.

It is reachable: `msg.epoch > current_epoch` routes it to `Future { epoch, height: zero }`
(`on_inbound_message.rs:457-462`), `exceeds_view_lookahead` always admits a different-epoch view by
construction, and every one of them lands on the same key `View(current_epoch + 1, 0)`. 64 MiB ÷
64 KiB = 1024 admitted → ~4 GiB resident, from any peer.

Two ways out: charge it `UNBOUNDED_MESSAGE_SIZE` like the other three, or cap `transaction_ids` in
the `TryFrom` so the "fixed" claim is true at the point the charge is made. The second is better —
the 1000-id limit already exists and is being enforced a step too late for this buffer — but either
closes it.

Everything else about the budget checks out. I walked the four reachable-`Future` variants:
`Proposal` and `MissingTransactionsResponse` are charged the exact wire ceiling; `Vote` is genuinely
fixed (`messages/vote.rs:11-13`, one `ProposalVote`, one signature) and 64 KiB is a large
over-charge in the safe direction; `NewView`, `ForeignProposal*` and `CatchUpSyncResponse` never
reach the buffer at all (always `Current`), and `CatchUpSyncRequest` is `Discard` off-epoch.

## Non-blocking

**2. (new, residual of the old #2) Inside the next-epoch bucket, an honest rollover proposal is
still the furthest key.** `on_inbound_message.rs:343-348`

`exceeds_view_lookahead` closes the scenario I described — a same-epoch flood at absurd heights is
now refused at admission rather than evicting. Within `current_epoch + 1` there is no such gate, by
design ("heights restart each epoch"), and an epoch+1 proposal is keyed on its justify height, which
is the *old* epoch's leaf height and therefore large. A flood of epoch+1 votes or
`MissingTransactionsRequest`s at low heights sits *nearer* than it, so it is the one refused.

Much smaller than before: it needs the attacker to guess the bucket rather than just spray, and the
cost is a re-request at rollover rather than a wedge. Noting it because the class of problem — the
key order is not "the order consensus will consume" once the epoch axis is in play — survived the
fix for the same-epoch case.

**3. (new) The 20-view same-epoch gate is a hard drop where the old code buffered.**
`on_inbound_message.rs:98-102,582`

A node lagging its committee by more than 20 views within an epoch now discards honest proposals at
admission; previously they were buffered. The commit message's position — such a node recovers by
importing blocks, not from the buffer — is right, and I don't think the buffered proposals were
usable without the intervening blocks anyway. Flagging it as the behaviour to watch in the
consensus_tests lag scenarios rather than as a defect.

**4. (carried forward) The 5 catch-up permits still have no per-peer share and no timeout.**
`on_catch_up_sync_request.rs:22,194`

Unchanged from the last pass, and consistent with `on_receive_foreign_proposal.rs:230`, so fine to
leave for a follow-up. Five concurrent requests from one committee member, or five sends to stalled
peers, occupy every permit and every other peer's catch-up is dropped.

## Nits

**5.** `view_buffer.rs` `insert`: `self.size + size > self.max_size` can overflow when `max_size` is
near `usize::MAX`. Only the tests' `UNLIMITED` reaches that, and they pass `NO_SIZE`, so nothing is
broken today — `saturating_add` removes the trap for the next caller.

**6.** `FIXED_MESSAGE_SIZE`'s doc comment (`on_inbound_message.rs:95-96`) talks about certificates
carrying a signature per committee member, but of the variants routed to that arm only `Vote` is
ever buffered, and it carries exactly one signature. The reasoning describes `NewView`, which is
always `Current`.

## Resolved since `5a00e539`

- **Blocking #1 (count-only budget)** — *not* resolved; see above. The budget exists and is correct
  for `Proposal` / `MissingTransactionsResponse` / `Vote`, which is most of it.
- **Non-blocking #2 (epoch-boundary inversion)** — resolved for the scenario given.
  `exceeds_view_lookahead` refuses same-epoch views beyond 20, so the absurd-height flood cannot
  occupy the buffer at all; verified against the three new unit tests, which cover the window
  boundary on both sides and the always-admit-next-epoch case.
- **Non-blocking #3 (lookahead after signature verification)** — resolved. `exceeds_vote_lookahead`
  is now shared out of `collector.rs:27-30` and both collectors apply it before `check_eligibility`
  and `validate_vote` (`proposal_collector.rs:77-89`, `timeout_collector.rs:63-75`), so a dropped
  vote costs neither the epoch-manager lookup nor the Schnorr verify. The check inside
  `collect_vote` is kept as the invariant, which is the right split.
- **Nit #5 (sizing rationale)** — resolved; the comment now describes proposing, not voting.
- **Nit #6 (description's "5 permits")** — resolved; it now says "at 5 permits rather than its 20".
- **Nit #7 (typo'd narrating comment)** — resolved; the line is gone.
- **Nit #8 (WARN from inside the container)** — resolved; `view_buffer.rs` no longer logs at all
  (the `log` import is gone), eviction is reported through `Inserted { num_evicted }`, and
  `push_to_buffer` logs it at `debug!`. Peer-driven traffic no longer writes WARN lines.

## Verified sound at this head

- `ViewBuffer` accounting with the second budget: `insert`, `pop_front`, `discard_before` and
  `clear` all keep `len` and `size` in step, sizes ride with their item in the deque so an eviction
  refunds the right amount, and the `pop_back()` / `last_entry()` `break` arms are unreachable given
  empty entries are removed on every path.
- `insert`'s early `size > self.max_size` refusal means the eviction loop cannot empty the buffer
  chasing an item that could never fit.
- The eviction loop still terminates: each pass either returns `Err` or decrements both counters.
- `exceeds_view_lookahead`'s epoch guard is correct for the `Future` router — every different-epoch
  case it sees is exactly `current_epoch + 1`, since `next()` discards beyond that at line 169.
- Both lookahead windows are 20 and independent constants, which is right: they bound different
  stores.

sdbondi:
_Posted by Claude._

Addressed in `e2c42a7`.

**1 (blocking) — `MissingTransactionsRequest` charged as fixed.** Confirmed and fixed the second way you suggested. `MAX_REQUESTED_TRANSACTIONS` (1000) now lives with the message type and `TryFrom<proto::consensus::MissingTransactionsRequest>` rejects anything beyond it, so the decoded message is the size the buffer charges it for and the limit is enforced before the message can be buffered rather than after. `on_message_validate` keeps the check against the same constant, for messages that never came off the wire. Two tests cover the boundary.

**2 — residual epoch+1 ordering.** Left as is, because the premise doesn't hold: heights restart each epoch (`CurrentView::enter` says so explicitly, and `create_genesis_block_if_required` writes a height-0 genesis per epoch as leaf and high PC), so the first proposal of epoch N+1 justifies that genesis and is keyed at `View(N+1, 0)` — the *nearest* key in the next-epoch bucket, not the furthest. A flood of epoch+1 votes at higher heights sits above it and is evicted first.

While checking that: an epoch+1 proposal keyed at height 0 looks like it can never be popped either way. Retrieval is at `next_view = (current_epoch, current_height + 1)`, so on entering epoch N+1 at the genesis leaf the first `next()` call does `discard_before((N+1, 1))`, which drops the `(N+1, 0)` entry before anything looks at it. That predates this PR (same keying, same pop in the old buffer) and I haven't traced it past the rollover path, so I've left it alone — but it means early-arriving rollover proposals have always cost a re-request, which matches where your last pass landed.

**3 — the 20-view gate drops where the old code buffered.** Agreed on the reasoning and on where to watch it; the consensus_tests lag scenarios run in CI on this branch.

**4 — catch-up permits.** Still deferred, same reasoning as last pass: one change covering both responders.

**5 — `self.size + size` overflow.** Fixed, `saturating_add`.

**6 — `FIXED_MESSAGE_SIZE` doc.** Fixed; it now describes `Vote`, the only variant in that arm that is ever buffered, rather than `NewView`'s per-member signatures.

Tests: 49 pass across `tari_consensus` and `tari_ootle_p2p`.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 4 non-blocking, 1 nit · 3 resolved since `a5621329` (1 blocking, 2 nits) · head `e2c42a79`

(Comment rather than an approval: GitHub refuses approve on your own PR. Nothing blocking stands —
another maintainer's approval is all this needs. CI is still running; see the end.)

## Resolved since `a5621329`

- **Blocking #1 (`MissingTransactionsRequest` under-charged)** — resolved, and by the better of the
  two routes. `MAX_REQUESTED_TRANSACTIONS` now lives with the message type
  (`messages/request_missing_transaction.rs:11-14`) and the wire conversion rejects beyond it
  (`crates/p2p/src/conversions/consensus.rs:423-430`), so the decoded message really is the size the
  buffer charges it for — the invariant `FIXED_MESSAGE_SIZE` asserts now holds at the point the
  charge is made, rather than being restored a step later. `on_message_validate.rs:157` keeps the
  same check against the shared constant, which is needed for loopback messages that never came off
  the wire, and the constant re-exports cleanly through `messages/mod.rs:18-19`.
  Two decode tests cover the boundary on both sides.
  No honest traffic changes: a request above 1000 was already discarded at validation, so the only
  difference is where it dies.
- **Nit #5 (`self.size + size` overflow)** — resolved, `saturating_add` at `view_buffer.rs:100`.
- **Nit #6 (`FIXED_MESSAGE_SIZE` doc)** — resolved for the claim I raised (it no longer reasons
  about per-committee-member signatures); see the new nit below for what the replacement misses.

## Non-blocking — all carried forward, none new to this head

1. **Inside the next-epoch bucket, an honest rollover proposal is still the furthest key**
   (`on_inbound_message.rs:344-349`). Needs the attacker to aim at `View(current_epoch + 1, low)`;
   costs a re-request at rollover, not a wedge.
2. **The 20-view same-epoch gate is a hard drop where the old code buffered**
   (`on_inbound_message.rs:99-103`). I agree with the position in the commit message; flagging it as
   the thing to watch in the consensus_tests lag scenarios.
3. **The 5 catch-up permits have no per-peer share and no timeout**
   (`on_catch_up_sync_request.rs:22,194`). Consistent with `on_receive_foreign_proposal.rs:230`;
   fine as a follow-up.
4. **(pre-existing, now more visible) `request_missing_transactions` does not chunk.** With
   `MAX_REQUESTED_TRANSACTIONS` promoted to a shared constant, the asymmetry is easier to see: a
   block needing more than 1000 transactions produces a request that the receiver now rejects at
   decode rather than at validation. Same outcome as before this PR — that block's processing
   stalls either way — so nothing here regressed. Worth a separate issue rather than scope creep.

## Nit

**`FIXED_MESSAGE_SIZE`'s new doc says "only `Vote` … is ever buffered"** — `on_inbound_message.rs:94-96`.
`MissingTransactionsRequest` is also buffered, via `Future { epoch, height: zero }` at
`on_inbound_message.rs:459-464` when `msg.epoch > current_epoch`. That is precisely the variant this
commit had to go and cap, so the sentence reads as if the problem it just fixed could not have
existed. Two names instead of one.

## Verified sound at this head

- The decode cap makes the charge honest: 1000 ids at 32 bytes in a `HashSet<TransactionId>` is
  ~2048 buckets, so ~66 KiB held against a 64 KiB charge. I did the arithmetic because the fix turns
  on it; a 3% overshoot on a budget whose other arm is a flat 4 MiB is noise, not a finding.
- `ViewBuffer` accounting still balances with `saturating_add` in place: the loop condition can now
  only become *more* eager to evict, never less, and the `size > self.max_size` early return above
  it still guarantees an item that cannot fit never empties the buffer chasing room.
- All four variants that can reach the buffer are now charged at or above what they can hold:
  `Proposal` and `MissingTransactionsResponse` at the wire ceiling, `Vote` far above its true size,
  `MissingTransactionsRequest` at its newly-enforced decoded maximum.
- The loopback path (`ConsensusInboundMessaging::next_message`, VN `p2p/services/messaging/inbound.rs:64-72`)
  carries `HotstuffMessage` without proto conversion, which is why keeping the validation check
  matters; it is kept.

## CI

`clippy`, `test (1-3/3)`, `integration test` and `ledger app` are pending on this head; the rest are
green. Nothing in my findings depends on them — this is a note, not a reservation.

sdbondi:
_Posted by Claude._

Nit fixed in `3b54046`: the `FIXED_MESSAGE_SIZE` doc now names both variants that reach the buffer — `Vote` and `MissingTransactionsRequest` at its decode cap — rather than reading as if the case this PR had to fix could not arise.

Non-blocking 1–3 stand as answered on the previous passes. On 4 (no chunking in `request_missing_transactions`): agreed it is pre-existing and out of scope here; noted on the audit list rather than fixed in this PR.

CI on `e2c42a79`: everything green except `clippy`, `test (1-3/3)` and `integration test`, still running at that head and re-triggered by this commit.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 4 non-blocking, 0 nits · 1 resolved since `e2c42a79` (1 nit) · head `3b54046e`

(Comment rather than an approval: GitHub refuses approve on your own PR. Nothing blocking stands —
this needs another maintainer's click, plus CI.)

`3b54046e` is the doc line only. The nit is resolved: `FIXED_MESSAGE_SIZE` now names both variants
that actually reach the buffer, and ties `MissingTransactionsRequest`'s bound to the cap that
enforces it. The intra-doc link resolves — `MAX_REQUESTED_TRANSACTIONS` is public through
`messages/mod.rs:18-19` — and `docs build` will confirm.

Nothing else moved, so the four non-blocking items from the last pass stand unchanged and none of
them needs to block a merge:

1. Inside the next-epoch bucket, an honest rollover proposal is still the furthest key
   (`on_inbound_message.rs:344-349`).
2. The 20-view same-epoch gate is a hard drop where the old code buffered
   (`on_inbound_message.rs:99-103`) — the watch item for the lag scenarios.
3. The 5 catch-up permits have no per-peer share and no timeout
   (`on_catch_up_sync_request.rs:22,194`).
4. Pre-existing: `request_missing_transactions` does not chunk above
   `MAX_REQUESTED_TRANSACTIONS`.

The whole series re-verified across four heads: the catch-up bound, the view buffer's two budgets
and both lookahead windows are sound, every buffer-reachable message variant is charged at or above
what it can hold, and each new guard has a test that keys on the condition rather than on a rendered
message.

CI is re-running from scratch on this head; everything was pending when I posted.

sdbondi:
_Posted by Claude._

Correction to this PR's framing, not the code. Two of the three paths are not reachable by an arbitrary peer: #2646 put `CatchUpSyncRequest` and `Vote` in the committee-gated arm of `on_message_validate`, and both the catch-up responder and the vote collectors run only on a `Ready` message. So the catch-up bound and the vote window bound what a *local committee member* can make this node do; only the view buffer, which fills in `MessageBuffer::next` before validation, is reachable by any connected peer.

H7 was written against `8f25994ac`, before that gate existed, and the original title carried its wording forward. Title and description now say which threat model applies to each path. The bounds themselves are unchanged and still worth having — a committee member flooding either path is exactly what non-blocking #4 (no per-peer share on the catch-up permits) is about.

sdbondi (commented):
Tested, reviewed and approved

PR #2648, opened 2026-09-18
test(integration): cover within-epoch catch-up sync

sdbondi:
There was no cucumber scenario for the block-level catch-up path (`CatchUpSyncRequest`/`CatchUpSyncResponse`). The existing ones cover neighbouring mechanisms:

- `state_sync.feature` — a newly registered validator syncing from an epoch checkpoint, a different path.
- `stall_recovery.feature` — the highest-QC probe, which deliberately *suppresses* sync.
- `offline_rollback.feature` — the whole committee restarts onto a fresh genesis, so nobody is behind.
- `leader_failure.feature` — timeouts, not lag.

A validator falling behind inside an epoch and fetching the blocks it missed was covered only by the in-process `consensus_tests` harness (`catch_up_rewind.rs`, `leader_failure.rs::single_shard_node_goes_down_and_catches_up`), never over real networking.

### The scenario
Four validators; one is shut down mid-epoch. The remaining three are still a quorum (`4 - floor(3/3) = 3`), so the chain advances while it is down — blocks for the transactions submitted meanwhile, plus a recovery block for every height where the absent node was leader. It is then restarted on its own data dir with no epoch boundary crossed, so no checkpoint exists for it to state-sync from and the missed blocks have to come from the committee. A transaction after the catch-up confirms it is a full member again.

### The new step
`Then validator node X catches up to validator node Y within N seconds` takes Y's view once and waits for X to reach it, so it measures recovery from a fixed deficit rather than chasing a moving tip. It fails if X's epoch passes the mark — past a boundary a node recovers by state sync, and the step exists to hold the within-epoch path.

Independent of #2647, which bounds the responder that serves these requests; this scenario passes with or without it.


sdbondi (commented):
Approved (ran locally)

PR #2649, opened 2026-09-18
fix(consensus): keep the first vote on equivocation and record the evidence

sdbondi:
Audit findings M3a, M8 and M1 from the single-shard consensus audit. Independent of the merged #2646/#2647; local-only, no state-format or protocol change.

## M3a — an equivocator could erase a vote it had already cast

`VoteStoreInner::save_vote` holds one vote per `(epoch, height, signer)`. On a second, differently signed vote it *removed* the stored one before returning, so a Byzantine committee member could take its own power back out of a quorum an honest committee was forming: vote for A, then send a conflicting vote, and A drops below the threshold.

The stored vote is now kept and the incoming one discarded. This node cannot tell which of the two is honest, and the one it already holds may already be counted.

## M8 — equivocation evidence is now persisted

Both collectors carried `// TODO: store equivocation evidence and punish` and simply logged. A conflicting **proposal-vote** pair is now written as a `VoteEquivocation` record carrying both whole votes, so the record is self-contained: a third party can rebuild each signed preimage and check the signatures without consulting other state. Both votes are signature-verified before the collector sees them, so the pair is proof of misbehaviour rather than a report of it.

What makes a pair evidence is a difference in what is attested — `(block_id, decision)` — not a difference in signature. Signing draws a fresh nonce, so one signer can produce arbitrarily many valid signatures over one message; `VoteEquivocation::from_conflicting_votes` rejects such a pair. **Timeout votes cannot yield evidence at all**: a timeout vote's preimage is `(epoch, height)`, which is exactly the view the vote store buckets it under, so two from one signer at one view always attest to the same thing. The timeout collector logs the duplicate and records nothing.

- One column family, keyed `(epoch, height, signer)`.
- Only the first pair per `(view, signer)` is kept — later pairs prove exactly what the first does, so this bounds what an equivocator can make the node write.
- Records age out in `epoch_cleanup` alongside the blocks of the view they indict.

Nothing consumes the evidence yet — there is no in-protocol penalty (eviction was removed in #2587). It is written for operators and for a future slashing path.

The operator-facing signal is the `consensus.vote_equivocation` diagnostic event (#2644), which the existing `get_diagnostic_events` RPC serves, plus a `consensus_vote_equivocations` Prometheus counter. The stored proof itself is reached through db-inspector, which now lists and browses the column family. There is deliberately no read method on the state store: nothing in the node reads the evidence back, and a slashing path can add the query it actually needs when there is one.

## M1 — a vote was persisted after it was sent

`send_vote_to_leader` sent the vote and *then* wrote `LastSentVote`; `process_block_decision` wrote `LastVoted` after the send returned. A crash in that window left a restarted node with no record of having voted at that height, free to vote again there for a competing block — the exact thing `LastVoted` exists to prevent. Both records are now written in one transaction before the send. The reverse failure only loses a message, which the network may drop anyway and which the next NEWVIEW carries again.

## Tests

Each fix has a test that fails without it (verified by reverting the fix in place):

- `an_equivocating_vote_does_not_erase_the_vote_already_held` — quorum is 3 of 4; the Byzantine member's first vote plus two honest votes must still reach it after the conflicting vote arrives. Without the fix the block reaches only 2 of 3.
- `a_vote_is_persisted_before_it_is_sent` — a new test-harness send observer runs on the sending task with the message in hand and nothing sent yet, so what it reads from the sender's own store is exactly what a crash at that instant would leave behind. Asserts `LastVoted`/`LastSentVote` already match every vote observed leaving a node.
- `crates/state_store_rocksdb/tests/vote_equivocation.rs` — same-attestation pairs rejected, round trip, first-evidence-wins, view/signer scoping, and cleanup pruning past the retention window.

Full `consensus_tests` suite and `tari_state_store_rocksdb` tests pass; `cargo lints clippy --all-targets` is clean.

## Noted, not fixed here

`OnNextSyncViewHandler` persists `LastSentNewView` and never reads it back — `last_sent_new_view_get` has no callers — so the guard against re-signing a NEWVIEW height is in-memory only and does not survive a restart. Pre-existing and unrelated to what this PR records; tracked in the audit's open list.


sdbondi (commented):
_Posted by Claude._

**Changes requested** — 1 blocking, 3 non-blocking, 2 nits · head `7a52190`

M3a and M1 both check out, and both tests target the right condition rather than the rendered error. My one blocking issue is with M8: on the timeout stream the record cannot be evidence, and an honest node will produce one.

## Blocking

### A `Timeout` equivocation record can never prove misbehaviour, and a restart produces one

`TimeoutVote`'s signed preimage is exactly `(epoch, height)` — `timeout_vote.rs:49` hashes nothing else — and the vote store buckets by `(epoch, height)` (`collector.rs:155`). So the two votes in an `EquivocatingVotes::Timeout` record (`vote_equivocation.rs:25`) *always* sign the identical message. They differ only in the signature. Two valid signatures over one message are not equivocation; they attest to exactly the same thing, and any signer can produce arbitrarily many.

What makes it fire is `collector.rs:163`: the equivocation trigger is `prev.signature() != vote.signature()`, not a difference in what the votes say. Signing uses a random nonce (`TariSignatureService::sign` → `rand::rng()`, `signer_service.rs:23`), so re-signing the same view is enough.

That is reachable on the honest path. `OnNextSyncViewHandler::handle` re-signs a fresh `TimeoutVote` on every call (`on_next_sync_view.rs:93`), and the only thing stopping it repeating a `timeout_height` is the in-memory `self.last_sent_new_view` (`on_next_sync_view.rs:64`). `LastSentNewView` is written to the store but never read — `last_sent_new_view_get` has no callers anywhere in the tree — so that guard does not survive a restart. Concretely: a validator sends NEWVIEW at timeout height H, restarts, comes back at `current_height = H-2`, the leader times out again, it signs a second `TimeoutVote` for `(epoch, H)`. A leader that did not restart still holds the first one, and `timeout_collector.rs:125` writes a permanent record naming that validator plus an `error`-level `consensus.vote_equivocation` diagnostic.

This has no consensus-safety consequence today since nothing consumes the record. It matters because of what the record claims to be: `vote_equivocation.rs` describes it as "proof of misbehaviour rather than a report of it", the PR body says a third party can rebuild each preimage and check the signatures, and it is explicitly staged for a future slashing path. For the timeout kind that verification is vacuous — both preimages are the same hash and both signatures verify — so a slashing path built on it would slash restarted honest validators.

Suggested shape: record only when the two votes attest to different things. For `ProposalVote` that is `(block_id, decision)`; for `TimeoutVote` there is nothing that can differ, so the timeout stream cannot yield evidence at all and the `Timeout` variant plus its column family should go. If the duplicate signature is still worth surfacing, it is a "duplicate vote" observation, not equivocation, and does not belong in a record built to be handed to a third party.

Separately worth fixing on its own merits: `OnNextSyncViewHandler` persists `LastSentNewView` and then ignores it on startup.

## Non-blocking

1. **The record is written without checking the votes conflict** — `proposal_collector.rs:153` builds evidence straight from `VoteEquivocationDetected`, which (per above) only means the signatures differ. On the proposal stream `LastVoted` — tightened by this PR's own M1 fix — is what currently keeps an honest node from re-voting at a height, so a false positive needs a bug elsewhere. But combined with first-pair-wins that bug gets expensive: a benign re-sign pair recorded first permanently suppresses the genuinely conflicting pair that follows for that `(kind, view, signer)`. Asserting the conflict at construction would close both.

2. **A write transaction per conflicting vote** — `helpers.rs:59` opens `with_write_tx` on every detection, including the common case where the record already exists and nothing is written. `vote_equivocation_exists` is already on the read trait; using it for the early-out avoids the write tx and gives that method a production caller.

3. **The stored proof has no read path** — `vote_equivocations_get_all_for_epoch` (`reader.rs:2040`) and `vote_equivocation_exists` are used only by `tests/vote_equivocation.rs`. "Written for operators" is true only of the diagnostic event; the record itself is reachable by nothing but a direct RocksDB read. Either expose it (JSON-RPC / CLI) or say in the body that a read path is still to come.

## Nits

1. `column_families/vote_equivocation.rs:15` — the doc comment describing the key layout sits above `macro_rules!`, so it documents the macro. `ProposalVoteEquivocationCf` and `TimeoutVoteEquivocationCf` get no rustdoc.

2. `PrometheusConsensusMetrics` takes the default no-op for `on_vote_equivocation`, so equivocation lands in the event log but is invisible to metrics. A counter here seems worth more than the log line.

## Verified sound

- **M3a**: `save_vote` (`collector.rs:163-186`) now keeps the stored vote. With the store keyed `(epoch, height) -> signer -> vote` and `calculate_threshold_decision` filtering by `agg_key`, the equivocator can no longer withdraw power it already contributed. `an_equivocating_vote_does_not_erase_the_vote_already_held` pins exactly that: quorum 3 of 4, and the byzantine member's first vote has to still be counted afterwards.
- **M1**: `send_vote_to_leader` (`on_receive_local_proposal.rs:703-707`) writes `LastVoted` and `LastSentVote` in one tx before the send, and the old post-send `as_last_voted().set` at the single call site is gone, so nothing regressed in coverage. Setting `LastVoted` even when the send errors is the safe direction. The test harness observer runs synchronously on the sending task before the channel send (`messaging_impls.rs`), which is the right instant to read the sender's own store.
- Evidence is only built after `check_eligibility` and `validate_vote` (signature verification) in both collectors, so the votes in a record are genuinely the signer's.
- The new column families follow `validator_node_epoch_stats.rs` exactly — shared `chain` CF, distinct key prefix, `(EpochCodec, NodeHeightCodec, PublicKeyCodec)` — and `epoch_cleanup` prunes them on the same `up_to_epoch + 1` boundary as blocks and QCs, so retention matches the claim.
- `hooks.clone()` into both collectors is safe: `DiagnosticHooks` shares its alarm state through an `Arc`, and Prometheus handles are registry-backed.

CI: only the title and signed-commit checks have reported so far; the build legs have not started. That is independent of the above.

sdbondi (commented):
_Posted by Claude._

**Changes requested** — 1 blocking, 2 non-blocking, 2 nits · 1 resolved since `7a52190` (1 non-blocking) · head `0d25108`

`0d25108` adds only the db-inspector wiring. The blocking finding is untouched.

## Resolved

- **Operator read path** — `register_all_cfs` and `add_cf_routes!` now both carry `ProposalVoteEquivocationCf` and `TimeoutVoteEquivocationCf` (`server.rs:117-118`, `server.rs:220-221`), and the table arm's column keys — `public_key`, `epoch`, `height`, `detected_at`, `votes` — match `VoteEquivocation`'s serde field names exactly (`tables.rs:210-218`), so the rows will render rather than falling through to the unlabelled default. `VoteEquivocation: Serialize` satisfies the `CF::Value: Serialize` bound on `tables::list`. Checked both lists for the asymmetry that already exists between them (`state_transition`, `foreign_substate_pledge` and `state_tree` are registered but routed by hand); the new pair is in both.

## Still blocking

### A `Timeout` equivocation record can never prove misbehaviour, and a restart produces one

Unchanged at this head. Restating the chain for the record:

- `TimeoutVote`'s signed preimage is `(epoch, height)` and nothing else (`timeout_vote.rs:49`), and the vote store buckets by `(epoch, height)` (`collector.rs:155`). Both votes in an `EquivocatingVotes::Timeout` (`vote_equivocation.rs:25`) therefore always sign the *same* message.
- The detection trigger is `prev.signature() != vote.signature()` (`collector.rs:163`) — signature inequality, not a difference in what the votes say — and signing draws a random nonce (`signer_service.rs:23`).
- `OnNextSyncViewHandler::handle` re-signs a fresh `TimeoutVote` per call (`on_next_sync_view.rs:93`); the only guard against repeating a `timeout_height` is the in-memory `self.last_sent_new_view` (`on_next_sync_view.rs:64`). `LastSentNewView` is persisted but `last_sent_new_view_get` has no callers, so the guard dies with the process.

A validator that sends NEWVIEW at height H, restarts, and is asked for another at H hands a non-restarted leader two valid signatures over one message. `timeout_collector.rs:125` then writes a permanent record naming it and fires an `error`-level `consensus.vote_equivocation`.

The db-inspector page makes this more visible, not less: an operator now has a table headed "Conflicting votes" that, for the timeout CF, can only ever show two votes that agree.

Same suggestion as before — record only where the votes attest to different things (`(block_id, decision)` for proposals; nothing can differ for timeouts, so that variant and its CF have no content to hold). And `OnNextSyncViewHandler` ignoring the `LastSentNewView` it persists is worth fixing regardless of what you decide here.

## Non-blocking (unchanged)

1. **Evidence is built without checking the votes conflict** — `proposal_collector.rs:153` constructs the record straight from `VoteEquivocationDetected`, which only means the signatures differ. `LastVoted`, tightened by this PR's M1 fix, is what currently keeps an honest node from re-voting at a height on the proposal stream. First-pair-wins makes any slip there permanent: a benign re-sign pair recorded first suppresses the genuinely conflicting pair that follows for that `(kind, view, signer)`.

2. **A write transaction per conflicting vote** — `helpers.rs:59` opens `with_write_tx` on every detection, including when the record already exists and nothing is written. `vote_equivocation_exists` would give the early-out — and would give that method a production caller, which it still lacks. `vote_equivocations_get_all_for_epoch` likewise remains test-only now that the inspector reads the column family directly; if nothing is going to call it, it need not be on `StateStoreReadTransaction`.

## Nits (unchanged)

1. `column_families/vote_equivocation.rs:15` — the key-layout doc sits above `macro_rules!`, so it documents the macro; both generated CFs have empty rustdoc.
2. `PrometheusConsensusMetrics` takes the default no-op for `on_vote_equivocation`.

CI: title, signed-commits and GitGuardian pass; no build or test leg has reported at this head.

sdbondi:
_Posted by Claude._

Blocking finding confirmed and fixed in `3014e0c8`. I verified each step rather than taking it on trust:

- `TimeoutVoteMessage::to_signature_message` chains `epoch` and `height` and nothing else (`timeout_vote.rs:86-94`).
- `save_vote` buckets on `(vote.epoch(), vote.height())` (`collector.rs:159-161`).

So the bucket key *is* the whole preimage: two timeout votes from one signer in one bucket necessarily sign the identical message. `grep` for `last_sent_new_view_get` / `LastSentNewView::get` across `crates/consensus/src` and `applications/` returns zero callers, so the restart path you describe is real — the guard is `self.last_sent_new_view` in memory only.

## What changed

**`EquivocatingVotes` and the timeout column family are gone.** `VoteEquivocation` now holds two `ProposalVote`s directly, in one CF (`KeyPrefix::VoteEquivocations`). The timeout collector logs the duplicate and records nothing, with the reason stated where a future reader will hit it.

**`VoteEquivocationDetected` → `DuplicateVoteDetected`.** The store detects a difference in *signature*, not in what is attested — that was the mislabel underneath the bug. The type now says so, and deciding whether a pair is evidence is the caller's job.

**Non-blocking 1 — conflict asserted at construction.** `VoteEquivocation::from_conflicting_votes` returns `None` when both votes carry the same `(block_id, decision)`; the proposal collector returns early on `None`. That closes the suppression hazard you flagged: a benign re-signed pair can no longer take the one slot a `(view, signer)` gets. `votes_attesting_to_the_same_thing_are_not_evidence` covers same-block, differing-block and differing-decision.

**Non-blocking 2 — read before write.** `record_equivocation` now early-outs on `vote_equivocation_exists` in a read tx, so a persistent equivocator costs a point lookup per vote rather than a write transaction. That method has a production caller now.

**Nit 1** — no macro any more, so the key-layout doc sits on `VoteEquivocationCf` itself.

**Nit 2** — `PrometheusConsensusMetrics` gained a `consensus_vote_equivocations` counter.

## Not taken

`OnNextSyncViewHandler` persisting `LastSentNewView` and never reading it is a real bug, but it is not this PR's — it predates it and now has no false-accusation consequence, since the timeout stream records nothing. I have added it to the audit's open list rather than widen this change. Say the word if you would rather it rode along.

Full `consensus_tests` (76) and `tari_state_store_rocksdb` (136) pass; `cargo lints clippy --all-targets` clean.

sdbondi:
_Posted by Claude._

Last non-blocking closed in `0ef3e4f9`, by removing the method rather than adding a caller.

`vote_equivocations_get_all_for_epoch` is gone from the read trait and its rocksdb impl. The reasoning: the operator-facing signal is already live-readable — the `consensus.vote_equivocation` diagnostic event goes through the existing `get_diagnostic_events` RPC, alongside the new `consensus_vote_equivocations` counter — and db-inspector reaches the stored proof by iterating the column family directly. The tests now read it the same way, via a `stored_for_epoch` helper over `vote_equivocation::ByEpochQuery`, so the test path matches the only real read path.

`vote_equivocation_exists` stays; it has its production caller in `record_equivocation`.

That leaves no speculative API: when slashing lands it adds the query it actually needs, rather than inheriting one guessed at now. The PR body says this explicitly so the absence reads as deliberate.

sdbondi (commented):
Tested and approved

PR #2650, opened 2026-09-18
fix(consensus): no-vote on deferrable lock conflicts and assert output existence on every lock grant

sdbondi:
Audit items H3 and H4 (single-shard consensus audit, PR D). Coordinated upgrade, no reset, no state-format change. Both changes alter what a replica votes for, so every node must flip together.

## H3 — replica ignored soft lock conflicts the proposer refuses

When `prepare()` reports lock failures with no hard conflict, the transaction can still succeed once the conflicting lock is released, so the proposer skips it and proposes it in a later block (`on_propose.rs`). The replica's command evaluation never read `lock_status`, so a byzantine leader could sequence exactly what an honest leader refuses and every replica would vote for it.

The condition is now named `LockStatus::is_deferrable_conflict()` and read by the proposer and by both replica evaluations — `evaluate_local_only_command` and `initial_prepare_multishard`, since the proposer's skip precedes its LocalOnly/MultiShard split and covers both. The replica returns a new `NoVoteReason::DeferrableLockConflict`.

## H4 — output-existence assertion only ran on one path

`try_lock` asserted that an OUTPUT lock's version does not already exist only on the early-return path where the substate carries no lock at all. The Read arm (local-only rules, or the same transaction) and the Write arm both grant an OUTPUT lock over an existing lock without it, so a transaction could take an OUTPUT lock on a version that is live and UP it a second time.

The assertion moves into `grant_lock`, which all four grant sites now go through. `try_lock_all` already classifies `SubstateIsUp` as a hard conflict, which is the right outcome: the claimed version is taken permanently, so the transaction aborts rather than being deferred.

## Tests

`crates/consensus_tests/src/substate_store.rs`:

- `an_output_lock_over_a_write_lock_requires_the_version_to_not_exist`, `an_output_lock_over_a_read_lock_requires_the_version_to_not_exist`, `an_output_lock_for_a_live_version_is_a_hard_conflict` — all three fail on `development` and pass here.
- `an_unversioned_lock_conflict_is_deferrable`, `a_versioned_lock_conflict_is_not_deferrable`, `locking_without_failures_is_not_a_deferrable_conflict` — pin the predicate the proposer and the replica now share.
- `it_allows_requesting_the_same_lock_within_one_transaction` took a WRITE lock on v0 and then an OUTPUT lock on v0, which the assertion now rejects. Real transactions lock WRITE on v and OUTPUT on v+1, so the test moves to that shape.

The 82 tests in `consensus_tests` pass, including the three lock-conflict liveness scenarios (`single_shard_input_conflict`, `multishard_unversioned_input_conflict`, `multishard_unversioned_input_conflict_delay_prepare`) that would catch a false no-vote from the H3 guard — the last of those drives a real soft conflict through the Prepare path.

**Not covered:** there is no end-to-end test of the H3 guard firing. That needs a leader that proposes a block its own proposer would refuse, and the harness has no seam for one — a message filter can drop a proposal but not craft a valid substitute. Worth adding a byzantine-proposer hook to the harness as follow-up work; it would serve the rest of the audit's leader-equivocation items too.


sdbondi (commented):
_Posted by Claude._

**Comment only** (own PR; `test`/`integration test` still pending) — 0 blocking, 2 non-blocking, 2 nits · head `4691d170`

Both fixes read correct, and the symmetry argument behind H3 holds under checking.

## What I verified

**H3 — replica/proposer symmetry.** The guard is placed in the two replica entry points that correspond to the proposer's single skip site:
- `prepare_transaction` (on_propose.rs:796) is the only proposer path that turns a `New`-stage pool record into a command (`transaction_pool_record_to_command`, on_propose.rs:315), and its skip precedes the LocalOnly/MultiShard split — so `LocalOnly`, `LocalPrepare` and `LocalAccept`-at-`New` all inherit it.
- `initial_prepare_multishard` (on_ready_to_vote_on_local_block.rs:826) is called from both `evaluate_local_prepare_command` (:961) and the `New`-stage branch of `evaluate_local_accept_command` (:1032), and `evaluate_local_only_command` (:585) covers the third. Those are also the only two `transaction_manager.prepare()` call sites on the replica, so the predicate is applied everywhere `prepare()` can report a lock status.

**No false no-vote.** The replica's lock set is a subset of the proposer's at the same point: the proposer's store additionally carries partial locks from transactions it skipped (`try_lock_all` grants the intents that succeeded before the failing one and nothing rolls them back) and from a foreign proposal whose processing failed after writing (only `change_set` is restored at on_propose.rs:477, not the substate store). A transaction that locked cleanly for the proposer therefore locks cleanly for the replica. Command order is not a divergence source either: the proposer iterates the batch in pool order while the replica replays in `Command` order, but every lock pair that can be granted in one order can also be granted in the other (READ/OUTPUT, WRITE/OUTPUT under local-only rules, READ/READ), and every conflicting pair conflicts in both.

**Empty lock statuses are not tripped.** `EarlyAbort` and both foreign-abort shortcuts (`prepare()` manager.rs:101, `prepare_multishard_from_execution` manager.rs:518, `initial_prepare_multishard`'s own shortcut at :805) carry `LockStatus::new()`, so `is_deferrable_conflict()` is false there and an aborted-by-foreign-block transaction is still evaluated normally. Hard conflicts set `hard_conflict_idx`, so the abort paths are untouched — only unversioned `LockConflict` failures are deferrable, which matches `try_lock_all`'s classification.

**H4 — all grant sites covered.** `add_new_lock` now has exactly one caller (`grant_lock`), and the four grant sites are the no-existing-lock branch and the READ/WRITE/OUTPUT arms. The OUTPUT arm only ever grants READ or WRITE, so the assertion correctly does not fire there. The rewritten no-lock branch is equivalent to the old one: `is_input()` is `!is_output()` (common_types/src/lock_intent.rs:55), and the assertion order is unchanged. `lock_assert_not_exist` is version-specific (`SubstateRecord::exists` takes a `VersionedSubstateIdRef`), so an OUTPUT lock on v+1 over a WRITE lock on v — the shape every real transaction has (block_transaction_executor.rs:103, outputs come from `diff.up_iter()` at the substate's own version) — still passes. The idempotent same-transaction re-request returns before `grant_lock`, so a duplicate OUTPUT request is unaffected.

**Description claims check out**: the proposer-skip-precedes-the-split claim, the `SubstateIsUp`-as-hard-conflict claim (pending_store.rs:432), and the reason `it_allows_requesting_the_same_lock_within_one_transaction` had to change (WRITE v0 + OUTPUT v0 is exactly what the assertion now rejects). No generated artifacts or unrelated files in the diff; `NoVoteReason` is not serialized or exported to bindings, and `as_str()` is updated for the new variant.

## Non-blocking

1. **The sibling of H3 on the `put_diff` paths is still open** — on_ready_to_vote_on_local_block.rs:648, :1446. The proposer also skips a transaction when `put_diff` fails with a lock error (on_propose.rs:853 for LocalOnly, :1061 for Accept), and that skip's existence says the case is reachable. The replica calls `substate_store.put_diff(diff)?` at both matching points, so the same sequencing yields a `HotStuffError` rather than a no-vote — and `handle_hotstuff_error` only swallows `ProposalValidationError`, everything else falls through to `return Err(err)` (worker.rs:1509) and out of the consensus loop. Pre-existing and out of this PR's scope, but it belongs on the same audit list as H3, and it is the one place where the byzantine leader gets something worse than a no-vote.

2. **The no-vote logs carry no conflict detail** — on_ready_to_vote_on_local_block.rs:586, :827. The proposer records the conflicts (`lock_conflicts.add(...)`) so the pool knows what to defer on; the replica logs only the transaction id. A no-vote that fires wrongly stalls the chain and this warn line is the only evidence a node operator will have, so `prepared.lock_status().failures()` (or `.conflicts()`) in the message would pay for itself.

## Nits

3. `it_allows_requesting_the_same_lock_within_one_transaction` (consensus_tests/src/substate_store.rs:172) no longer requests the same lock — it now takes WRITE v0 then OUTPUT v1. The body is right; the name now describes what the middle assertion rejects. Something like `it_allows_an_input_and_output_lock_within_one_transaction` would match.

4. `lock_assert_not_exist` (pending_store.rs:855) returns `Ok` when the pending diff holds a DOWN for exactly that version, so an OUTPUT lock on v is still grantable after an earlier command in the same block DOWNed v (reachable under local-only rules: tx1 WRITE v0, tx2 OUTPUT v0). Honest execution cannot produce an UP at a version the store has DOWNed, so nothing is broken today, but H4 makes this the last remaining way for an OUTPUT lock to name a version that once existed. Worth a follow-up assertion rather than a change here.

sdbondi:
_Posted by Claude._

Applied in 3734dd9, both non-blocking items and both nits addressed or recorded.

**2 (no conflict detail in the no-vote log)** — applied. Both warnings now carry `prepared.lock_status().failures().display()`.

**3 (test name)** — applied. Renamed to `it_allows_an_input_and_an_output_lock_within_one_transaction`.

**1 (`put_diff` lock failures propagate on the replica)** — confirmed and out of scope here, recorded as H8 on the audit list. The two proposer skips (on_propose.rs:847, :1055) and the two replica `?` sites (on_ready_to_vote_on_local_block.rs:648, :1446) line up exactly, and `handle_hotstuff_error`'s catch-all is `_ => return Err(err)` (worker.rs:1509), so the outcome is a worker restart rather than a no-vote. It belongs to the H2 family of unaudited `?` paths and wants the same treatment as H3 — likely its own PR rather than riding this one, since it changes an error path rather than a vote.

**4 (`lock_assert_not_exist` returns `Ok` over a pending DOWN at that version)** — agreed, and recorded as a follow-up. Honest execution cannot produce an UP at a version the store has DOWNed, so tightening it here would be a behaviour change with no reachable case behind it.

sdbondi (commented):
_Posted by Claude._

**Comment only** (own PR; CI re-running on the new head) — 0 blocking, 1 non-blocking, 0 nits · 2 resolved since `4691d170` (0 blocking, 1 non-blocking, 1 nit) · head `3734dd9c`

Re-checked `3734dd9c` against the previous head.

**Resolved**
- Non-blocking 2 (no conflict detail in the no-vote warns): both sites now log `prepared.lock_status().failures().display()` (on_ready_to_vote_on_local_block.rs:588, :831). `failures()` is `&[LockFailedError]`, `Displayable` is implemented for `[T] where T: Display` (common_types/src/displayable.rs:146) and `LockFailedError` derives `Display` via thiserror, so the call resolves and the guard already guarantees the slice is non-empty.
- Nit 3: renamed to `it_allows_an_input_and_an_output_lock_within_one_transaction`; the body is unchanged and still the WRITE v0 / OUTPUT v1 shape.

**Still open (unchanged, non-blocking)**
- Non-blocking 1: the `put_diff` sibling gap — the proposer skips (on_propose.rs:853, :1061) where the replica `?`s (on_ready_to_vote_on_local_block.rs:651, :1449) and the error escapes `handle_hotstuff_error` (worker.rs:1509). Pre-existing, follow-up.

Nothing else moved; the H3/H4 verification from the first pass still stands at this head.

sdbondi (commented):
Approved

PR #2651, opened 2026-09-18
fix(consensus): no-vote when a proposed diff fails to apply with a lock failure

sdbondi:
Audit item H8, the error-path sibling of #2650's H3.

## The bug

The proposer skips a transaction whose `put_diff` fails with a lock failure — `on_propose.rs:847` (LocalOnly) and `:1055` (Accept) both call `err.ok_lock_failed()?` and `return Ok(None)`. The replica called `put_diff` with `?` at the two matching points (`on_ready_to_vote_on_local_block.rs:649`, `:1449`), and `handle_hotstuff_error` only swallows `ProposalValidationError` — everything else falls through `_ => return Err(err)` and out of the consensus loop.

Same shape as H3, worse outcome: a byzantine leader that sequences a transaction an honest proposer would have deferred restarts every replica's worker, repeatably, instead of drawing a no-vote.

Both sites now mirror the proposer's classification: `ok_lock_failed()` keeps a non-lock store error fatal, and a lock failure becomes a no-vote.

The reason is `NoVoteReason::ConflictingTransactionProposed` ("The proposer proposed a conflicting transaction {id}: {details}", where `details` is the store's own failure text), not H3's `DeferrableLockConflict`: the only lock failure `put_diff` can raise is `SubstateNotFound`, and `try_lock_all` classifies that as a *hard* conflict, so calling it deferrable would contradict the store's own classification. The proposer treats both as a skip; the two conditions are still distinct.

### Is it reachable on an honest leader?

No. The replica re-derives the diff by executing locally (execution results are never taken from the leader), from the same parent state and the same command list, so its `put_diff` sequence matches the proposer's. The only way a replica reaches a failing `put_diff` is a leader that included a command its own proposer would have skipped. The no-vote is the whole fix — D's "the replica's lock set is a subset of the proposer's" argument does not need re-checking at this boundary.

### Does the proposer's skip leave the two stores different?

`put_diff` applies every DOWN before any UP and `PendingSubstateStore` has no rollback, so a skip partway through a diff would leave the DOWNs that already succeeded in the proposer's store while the skipped transaction is absent from the block — the two sides would continue from different state, which is what the symmetry argument above assumes away.

The error taxonomy does not rule it out: `assert_is_down` raises `SubstateNotFound`, not only `ExpectedSubstateDown`, when the predecessor version has no record at all (`pending_store.rs:849`), and the fee-withdrawal loop's `not_found` closure raises it too (`:301`) — both skippable, and both running after DOWNs have been applied. What rules it out is the lock rules. Every substate the local part of a diff DOWNs, or UPs over a predecessor, was locked earlier in the same block and `lock_assert_is_up` proved it live then; no other transaction can have spent it since, because WRITE-over-WRITE is rejected even under local-only rules. `filter_diff_for_committee` keeps withdrawals to local substates alongside ups and downs, and every version of a substate shards to the same committee, so a committee never holds `v1` without `v0`.

Validator fee pools are the one part of a diff that is not locked. A withdrawal's `not_found` fires only for a pool with no record at all — `update_in_place` recreates one that was DOWNed — which is not a state a block whose leader fee is being paid can be in. A `put_diff` that could not partially apply would close the question structurally rather than by argument; that is a follow-up, not this PR.

## Also: output locks over a destroyed version

Raised in the #2650 review and carried here. `lock_assert_not_exist` guards every OUTPUT lock grant. Its committed-store branch uses `SubstateRecord::exists`, which is true for a destroyed record, so it rejects any version that has ever existed. Its pending-diff and block-diff branches returned `Ok` on a DOWN at exactly the claimed version.

Reachable shape under local-only rules: tx1 takes WRITE v0 and its diff DOWNs v0 / UPs v1; tx2 then asks for OUTPUT v0, the Write arm grants it, and the assertion sees a pending DOWN and passes. tx2 would UP a version that was already spent.

Honest execution cannot produce it — a v0 output is a fresh id — so this is a tightening, not a live bug. Existence, not liveness, is now the test on all three branches, which is what the committed branch already did.

`LockFailedError::SubstateIsUp` is renamed `SubstateExists` to match what it reports on all three (it already reported "is already UP" for a destroyed committed substate). The rename changes the `FailedToLockOutputs` reject reason string, which is consensus-visible; it rolls out with the coordinated upgrade alongside #2650.

### New storage trait method

`lock_assert_not_exist` runs on every OUTPUT lock grant, and its block-diff branch used `BlockDiff::get_for_versioned_substate`, which reads the `SubstateChange` value — for an UP, the whole substate — to learn only whether one exists.

`StateStoreReadTransaction::block_diffs_contains_versioned_substate` is a new public trait method (with `BlockDiff::contains_versioned_substate` as the model-side wrapper). It answers from the `BySubstateIdQuery` key index and returns on the first key matching the version within the branch: no value read, no ordering pass. It keeps the siblings' `blocks_exists` guard, so an unknown block is still a `QueryError` rather than a silent `false`. `RocksDbStateStoreReadTransaction` is the only implementor.

The three sources in `lock_assert_not_exist` are now checked cheapest first — in-memory pending diff, `SubstateRecord::exists` point lookup, branch scan. They are independent existence tests, so the order does not change the result.

## Tests

`crates/consensus_tests/src/substate_store.rs`:

- `an_output_lock_for_a_destroyed_version_is_rejected` — the tx1 WRITE v0 / tx2 OUTPUT v0 shape above. Fails on `development` (the lock is granted).
- `a_put_diff_over_a_spent_input_is_a_skippable_lock_failure` — pins that a `put_diff` over an input another command already spent classifies as a lock failure, which is what both `return Ok(None)` and both no-votes rest on.
- `a_put_diff_upping_over_a_live_version_is_fatal` — pins the other side: `ExpectedSubstateDown` stays fatal and is not deferred.

Not covered: no end-to-end test of either no-vote firing, for the same reason #2650 gave — it needs a leader that proposes a block its own proposer would refuse, and the harness has no seam for that. A byzantine-proposer hook remains the follow-up.


sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 2 non-blocking, 1 nit · head `fc2f3a55`

Read the four changed files at the head against `development` (`7400eb3cc`). The core change is sound and the description's claims check out.

**Verified**

- All four `put_diff` call sites in the crate are now classified identically: `on_propose.rs:847`/`:1055` skip, `on_ready_to_vote_on_local_block.rs:649`/`:1463` no-vote, and all four route through `err.ok_lock_failed()?` so a non-lock `SubstateStoreError` stays fatal. No fifth site exists (`git grep put_diff` over `crates/`).
- `ok_lock_failed` maps only `LockFailed` and `SubstateNotFound`; `ExpectedSubstateDown` still propagates, which is exactly the split the two new no-votes rest on — and `a_put_diff_upping_over_a_live_version_is_fatal` pins it.
- Both no-vote returns land in the dispatch loop at `:293`/`:334`, which calls `set_no_vote` and returns immediately, so the partially-applied diff and the already-incremented `total_leader_fee`/`total_exhaust_burn` at `:1443`/`:1459` are discarded with the change set. No leak.
- `lock_assert_not_exist` is reached only from `grant_lock` under `lock_type.is_output()`, so the tightening cannot touch READ/WRITE grants.
- The "existence, not liveness" claim holds on all three branches: `get_pending` and `BlockDiff::get_for_versioned_substate` are both keyed on the versioned address, and `SubstateRecord::exists` → `substates_any_exist` takes a `VersionedSubstateIdRef`. The pruning hazard the `SubstateNotFound` doc comment warns about does not apply here — `substates_prune_downed_values` (`state_store_rocksdb/src/writer.rs:1337`) calls `clear_substate_value` and keeps the record, so `exists` stays true after pruning and nodes cannot disagree on `SubstateExists`.
- `an_output_lock_for_a_destroyed_version_is_rejected` genuinely fails on `development`: tx1's WRITE v0 is granted (input path, `lock_assert_is_up`), its diff leaves a pending DOWN at v0, and tx2's OUTPUT v0 reaches `grant_lock` through the Write arm under local-only rules — where the old `change.is_up()` branch returned `Ok`.
- `SubstateIsUp` has no remaining references anywhere in the tree. The rename is consensus-visible through `RejectReason::FailedToLockOutputs(err.to_string())` (`on_propose.rs:1004`, `on_ready_to_vote_on_local_block.rs:1165`), as the description says.
- Let-chain at `:648` is fine on edition 2024 and matches sibling usage (`on_ready_to_vote_on_local_block.rs:437`).

## Non-blocking

**1. The proposer leaves a partially-applied diff behind on skip, which is the one crack in the symmetry argument** — `on_propose.rs:847`, `:1055`

`put_diff` applies every DOWN before any UP and there is no rollback on `PendingSubstateStore`, so when the proposer's `return Ok(None)` fires mid-diff the DOWNs that already succeeded stay in `self.changes`. The skipped transaction is then absent from the block, so the replica never applies them — the proposer and the replica continue from *different* stores, which is precisely what "its `put_diff` sequence matches the proposer's" assumes away.

The divergence is one-directional in the dangerous sense: a later transaction could UP `v1` against a stray DOWN of `v0` on the proposer and hit `ExpectedSubstateDown` on the replica, which these new paths still treat as fatal.

I could not construct a reachable instance — a partial failure needs one DOWN to succeed and a later one to fail, and a transaction cannot hold WRITE on a substate another transaction already spent (the Write arm rejects WRITE-over-WRITE even under local-only rules). So the gap looks closed by the lock rules rather than by the diff application. Worth stating in the PR description rather than fixing here, since the whole no-vote argument leans on the symmetry.

**2. `DeferrableLockConflict` is reported for failures that lock-time classification calls a hard conflict** — `on_ready_to_vote_on_local_block.rs:662`, `:1474`

`ok_lock_failed` can return `LockFailedError::SubstateNotFound`, and `try_lock_all` (`pending_store.rs:428-434`) sets `hard_conflict_idx` for exactly that variant — i.e. at lock time it means ABORT, not defer. Same for the newly-widened `SubstateExists`. The warn text ("a lock failure that an honest proposer defers") is accurate about the proposer's behaviour, but the reason name says something the store's own classification contradicts.

Nothing reads the variant — it is diagnostic only (`no_vote.rs:150` renders the name, nothing branches on it) — so this is naming, not behaviour. It does mean the field will read as "deferrable" in logs for a conflict that can never clear.

## Nit

**`if let Some(diff) = …any_accept()` is a dead `None` arm under a commit decision** — `on_ready_to_vote_on_local_block.rs:648`

The branch is guarded by `pool_tx.current_decision().is_commit()`, and that decision comes from `execution.decision()` a few lines up, so `any_accept()` is `Some` whenever control reaches here. `on_propose.rs:839` treats the same combination as an `InvariantError`. Not worth changing in this PR, but the two sides disagree about whether that state is impossible or merely tolerated.

## Tests

The three new tests exercise what they claim. `a_put_diff_over_a_spent_input_is_a_skippable_lock_failure` reaches `SubstateNotFound` through `assert_is_up` finding a pending DOWN (not the committed branch), which is the shape the proposer actually hits under contention. The `write_diff` helper matches the real WRITE-input diff shape.

The stated gap — no end-to-end test of either no-vote firing — is the same one #2650 documented and needs the same byzantine-proposer seam. Agreed that it does not belong here.

CI: clippy, fmt and all three test shards are green at this head; the integration leg is still pending.

sdbondi:
_Posted by Claude._

Addressed at `efa7e947`.

**Non-blocking 1 — partial `put_diff` on the proposer's skip.** Agreed the symmetry argument needed this stated, and the gap is closed by the lock rules rather than by the diff application. Sharpening your reasoning slightly: a skip only happens for a failure in `ok_lock_failed`'s class, and only the DOWN loop can produce one — an UP failing gives `ExpectedSubstateDown` and a fee-pool failure gives `InvariantError`, both fatal on both sides. So a partial application needs one DOWN to succeed and a later DOWN to fail, i.e. the transaction holds WRITE over a version another transaction already spent, which the Write arm of `try_lock` rejects even under local-only rules. The first failing DOWN is therefore the first change of the diff. Written into the PR description.

**Non-blocking 2 — `DeferrableLockConflict` naming.** Correct, and the case is stronger than stated: at these two sites `SubstateNotFound` is the *only* reachable variant. `put_diff` never locks, so `LockFailed` cannot arise there, which also means `SubstateExists` cannot — it comes only from `grant_lock`. And `try_lock_all` classifies `SubstateNotFound` as a hard conflict, so "deferrable" said the opposite of the store's own classification. New `NoVoteReason::DiffNotApplicable`, and the warn text now says "does not apply, which an honest proposer skips" rather than "defers".

**Nit — dead `None` arm at `:648`.** Left as is. Tightening it to match `on_propose.rs:839` means adding an `InvariantError` path on the replica for a state the replica can reach only if its own execution is broken, which is a separate decision from this PR's; noting it rather than folding it in.

Re-ran: clippy clean, `cargo nextest run -p consensus_tests` 85/85.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 1 non-blocking, 1 nit · 1 resolved since `fc2f3a55` (1 non-blocking) · head `efa7e947`

## Resolved

- **`DeferrableLockConflict` reused for a hard conflict** — fixed by `NoVoteReason::DiffNotApplicable` (`no_vote.rs:107`). Checked that the two prepare-time sites (`on_ready_to_vote_on_local_block.rs:593`, `:850`) still use `DeferrableLockConflict` and only the two diff-application sites (`:662`, `:1474`) moved. The variant is appended last, `reason_name` has its arm, and `NoVoteReason` carries no serde/borsh/ts-rs derive, so adding it is not wire-visible.

## Non-blocking

**The description's new "It cannot happen" argument is not right as stated, though its conclusion holds**

> A skip needs the failure to be in `ok_lock_failed`'s class, and only the DOWN loop produces one (`assert_is_up` → `SubstateNotFound`); an UP failing gives `ExpectedSubstateDown`, which stays fatal on both sides, and a fee-pool failure gives `InvariantError`.

Two of those three are wrong:

1. `assert_is_down` has a `SubstateNotFound` arm, not just `ExpectedSubstateDown` — `pending_store.rs:849`, the `let Some(is_up) = substate_is_up(…) else` branch, which fires when the predecessor version has no record at all. So an UP *can* produce a skippable failure, and by then every DOWN in the diff has been applied.
2. The fee-withdrawal loop does not only raise `InvariantError`. Its `not_found` closure returns `SubstateStoreError::SubstateNotFound` (`pending_store.rs:301-306`) — skippable — and that loop runs after every DOWN *and* every UP. A failure there leaves the entire diff applied.

So "the first DOWN to fail is therefore the first change of the diff, and nothing is left behind" does not follow from the error taxonomy. What does hold is the lock argument on its own: every local substate the diff touches — input, predecessor of an UP, or withdrawal target — was locked earlier in the same block, `lock_assert_is_up` already proved it live, and no other transaction can have spent it since (WRITE-over-WRITE is rejected even under local-only rules). `filter_diff_for_committee` (`common.rs:404`) filters withdrawals by `includes_substate_id` alongside ups and downs, so a foreign fee pool cannot slip into the local diff either — and `SubstateAddress::to_shard` reads only `self.0[0]`, inside the object key, so every version of a substate lands in the same shard and a committee can never hold `v1` without `v0`.

Worth rewriting that paragraph around the lock rules rather than the error classes, since the current version invites a reader to conclude `ExpectedSubstateDown` is the only UP-side failure.

## Nit

Unchanged from the previous pass: `if let Some(diff) = …any_accept()` at `on_ready_to_vote_on_local_block.rs:648` has a `None` arm that cannot be reached under the enclosing commit-decision guard, where `on_propose.rs:839` treats the same state as an `InvariantError`.

## Re-verified at this head

The code change since `fc2f3a55` is the rename plus the new variant; the classification logic, `lock_assert_not_exist` and the tests are byte-identical to what I checked last pass. `git grep DeferrableLockConflict` shows no stale use.

CI: clippy, the integration leg and two test shards are re-running at this head; `fmt`, `machete`, `prettier` and shard 3/3 are green.

crates/consensus/src/hotstuff/substate_store/pending_store.rs
```diff
@@ -852,32 +853,24 @@ impl<'store, TTx: StateStoreReadTransaction> PendingSubstateStore<'store, TTx> {
...
-            }
-            return Ok(());
+        if self.get_pending(&id.to_substate_address()).is_some() {
+            return Err(LockFailedError::SubstateExists { id: id.to_owned() }.into());
         }
 
-        if let Some(change) =
-            BlockDiff::get_for_versioned_substate(self.read_transaction(), self.parent_block.block_id(), id)
-                .optional()?
+        if BlockDiff::get_for_versioned_substate(self.read_transaction(), self.parent_block.block_id(), id)
```
sdbondi: Add a new function that checks this cheaply without decoding etc - this is the hot path

crates/storage/src/consensus_models/no_vote.rs
```diff
@@ -100,6 +100,11 @@ pub enum NoVoteReason {
     BlockExecutionPointsExceeded { total_points: u64, max_points: u64 },
     #[error("Transaction {transaction_id} has substate lock conflicts and must be deferred to a later block")]
     DeferrableLockConflict { transaction_id: TransactionId },
+    /// Locking succeeded but the transaction's state diff does not apply over the pending store. An honest proposer
+    /// skips such a transaction, so it is distinct from [NoVoteReason::DeferrableLockConflict]: the lock rules
+    /// classify the underlying failure as one no later block can clear.
+    #[error("Transaction {transaction_id} state diff could not be applied over the pending substate store")]
+    DiffNotApplicable { transaction_id: TransactionId },
```
sdbondi: This error doesnt really tell you what the specific problem was. Something like "the proposer proposed a conflicting transaction {tx_id}" ?

crates/consensus/src/hotstuff/substate_store/pending_store.rs
sdbondi: _Posted by Claude._

Done at `7a1aadd` — `block_diffs_contains_versioned_substate` answers from the key index (`reader.rs`, reusing the `BySubstateIdQuery` prefix iterator) and returns on the first key matching the version in the branch. No `BlockDiffCf` value read, so an UP no longer costs a decode of the whole substate, and there is no ordering pass.

Also reordered the three sources cheapest first: in-memory pending diff, then the `SubstateRecord::exists` point lookup, then the branch scan. All three are independent existence tests, so the order does not change the result.

crates/storage/src/consensus_models/no_vote.rs
sdbondi: _Posted by Claude._

Done at `7a1aadd`. Renamed to `NoVoteReason::ConflictingTransactionProposed` and it now reads:

```
The proposer proposed a conflicting transaction {transaction_id}: {details}
```

`details` is the store's own failure text (e.g. `Substate component_… v0 is not found or DOWN`), so the reason names both the situation and the specific conflict.

sdbondi:
_Posted by Claude._

`7a1aadd` — two review comments plus the non-blocking correction.

**The "It cannot happen" paragraph was wrong.** Checked both counterexamples and they hold: `assert_is_down` has a `SubstateNotFound` arm at `pending_store.rs:849` for a predecessor with no record at all, and the fee-withdrawal `not_found` closure returns `SubstateNotFound` at `:301`. Both are skippable, and both run after DOWNs have been applied, so the error taxonomy does not close it. Rewritten around the lock rules, as suggested.

One correction back: validator fee pools are not locked — the DOWN and UP loops skip them and they go through `update_in_place` — so "every local substate the diff touches was locked" does not cover the withdrawal target. That case rests on something weaker: `update_in_place` recreates a DOWNed pool, so `not_found` fires only for a pool with no record whatsoever, which is not a state a block whose leader fee is being paid can be in. Both are in the description now, along with the note that a `put_diff` which cannot partially apply would settle this structurally instead of by argument. Follow-up, not this PR.

**`lock_assert_not_exist` cost** (review comment). `BlockDiff::contains_versioned_substate` → `block_diffs_contains_versioned_substate` reuses the `BySubstateIdQuery` key iterator and returns on the first key matching the version in the branch — no `BlockDiffCf` value read, so an UP no longer costs a decode of the whole substate, and no ordering pass. The three sources are now checked cheapest first: in-memory diff, `SubstateRecord::exists` point lookup, branch scan.

**No-vote reason** (review comment). Renamed to `ConflictingTransactionProposed`, message `The proposer proposed a conflicting transaction {transaction_id}: {details}`, where `details` is the store's own failure text.

**Nit, `:648`** — still left. Tightening it to `on_propose.rs:839`'s `InvariantError` adds a fatal path on the replica for a state only its own broken execution can produce; that is a decision about replica error policy, not about this fix.

clippy clean, `cargo nextest run -p consensus_tests` 85/85.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 2 non-blocking, 1 nit · 1 resolved since `efa7e947` (1 non-blocking) · head `7a1aaddb`

## Resolved

- **The "It cannot happen" argument** — the description now states it correctly. Re-checked both citations: `assert_is_down`'s third branch does raise `SubstateNotFound` when the predecessor has no record (`pending_store.rs:849`), and the withdrawal `not_found` closure raises it too (`:301`), and the paragraph now rests on the lock rules rather than the error classes. The fee-pool carve-out is right — `update_in_place` recreates a DOWNed pool, so `not_found` needs a pool with no record at all.

## New in this head

The reorder and the new existence query are sound. Verified:

- `block_diffs_contains_versioned_substate` (`state_store_rocksdb/src/reader.rs:1162`) filters on exactly the same two predicates as `block_diffs_select_last_change` (`:370`) — `key.version == v` and `applicable_blocks.contains(&key.block_id)` — and since only existence is wanted, returning on the first match instead of folding for the last is equivalent. `query_prefix_range_key_iterator` yields keys, so the comment's claim that the change value is never read holds.
- Reordering the three branches in `lock_assert_not_exist` cannot change the outcome: all three now return the same error, so only the cost order moves. `SubstateRecord::exists` → `substates_any_exist` → `cf.exists` is a real `get_pinned_cf` (`cf_api.rs:150`), not `key_may_exist`, so it is a deterministic point lookup and cannot false-positive across nodes.
- `StateStoreReadTransaction` has exactly one implementor, so the new trait method has no other impl to update. `ConflictingTransactionProposed` is appended last with its `reason_name` arm, and `NoVoteReason` still carries no serde/borsh/ts-rs derive, so the added `details: String` is not wire-visible.

## Non-blocking

**1. The description no longer matches the code it describes**

It still says `NoVoteReason::DiffNotApplicable`; this head renamed that to `ConflictingTransactionProposed` and gave it a `details` field (`no_vote.rs:106`, used at `on_ready_to_vote_on_local_block.rs:662` and `:1475`).

The description also does not mention this head's other half at all: a new public method on `StateStoreReadTransaction` (`state_store/mod.rs:230`), its `BlockDiff::contains_versioned_substate` wrapper, the RocksDB implementation, and the reordered branches in `lock_assert_not_exist`. A new method on a public storage trait is worth a line in a PR whose stated subject is a consensus no-vote.

**2. `block_diffs_contains_versioned_substate` drops the block-existence guard its two siblings have** — `state_store_rocksdb/src/reader.rs:1162`

`block_diffs_get_last_change_for_substate` and `block_diffs_get_change_for_versioned_substate` both open with `if !self.blocks_exists(block_id)` → `QueryError`. The new method does not, and `get_pending_chain_until` returns an empty set for a block absent from the pending chain index (`:220`), so an unknown `block_id` yields `Ok(false)`.

That flips behaviour at the call site. The old code was `BlockDiff::get_for_versioned_substate(…).optional()?`, and `optional()` only swallows `NotFound`/`NotFoundDbAdapter` (`storage/src/error.rs:79`) — a `QueryError` for a nonexistent parent block propagated and aborted the lock. Now it is indistinguishable from "the branch records nothing", and the OUTPUT lock is granted.

The parent block always exists in practice, so this is not a live bug, but the trait doc at `state_store/mod.rs:228` claims "Selection follows the same branch rules as `block_diffs_get_last_change_for_substate`", and on this point it does not. Either add the guard or drop that sentence.

## Nit

Unchanged: `if let Some(diff) = …any_accept()` at `on_ready_to_vote_on_local_block.rs:648` has a `None` arm unreachable under the enclosing commit-decision guard, where `on_propose.rs:839` treats the same state as an `InvariantError`.

CI: `fmt`, `machete`, `prettier`, `docs build` and the web/ledger legs are green at this head; clippy and all three test shards are still running.

sdbondi:
_Posted by Claude._

Both fixed.

**1. Description.** Updated: the reason is named `ConflictingTransactionProposed` with its message, and there is a new "New storage trait method" section covering `StateStoreReadTransaction::block_diffs_contains_versioned_substate` and the `BlockDiff::contains_versioned_substate` wrapper — why it exists, what it avoids reading, the `blocks_exists` guard, and the cheapest-first reordering in `lock_assert_not_exist`.

**2. Missing `blocks_exists` guard** — `ad6306ff`. Confirmed: `get_pending_chain_until` returns an empty set for a block not in the pending-chain index, so an unknown block was `Ok(false)`, indistinguishable from a genuine absence, where both siblings and the `.optional()?` path it replaced raise `QueryError`. Guard added in the same form, so the trait doc's "same branch rules" now holds.

clippy clean, `cargo nextest run -p consensus_tests` 85/85.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 0 non-blocking, 1 nit · 2 resolved since `7a1aaddb` (2 non-blocking) · head `ad6306ff`

Everything I raised across the four passes is resolved. This reads as approvable; `gh` will not take `--approve` on your own PR, so this is a comment — it needs another maintainer's click.

## Resolved

- **Missing block-existence guard** — `block_diffs_contains_versioned_substate` now opens with the same `blocks_exists` → `QueryError` guard as `block_diffs_get_last_change_for_substate` and `block_diffs_get_change_for_versioned_substate` (`state_store_rocksdb/src/reader.rs:1167`), so the trait doc's "same branch rules" claim now holds and an unknown block can no longer be read as "the branch records nothing". The guard was already in this code path at the original head — the old call went through `get_for_versioned_substate`, which carries it — so the existing store tests already establish that the fixture's zero block satisfies it.
- **Stale description** — the body now names `ConflictingTransactionProposed` with its `details` field, documents the new trait method, its `BlockDiff` wrapper, the single implementor and the retained guard, and states the cheapest-first ordering rationale.

## Still open

One nit, carried from the first pass and not worth another push on its own: `if let Some(diff) = …any_accept()` at `on_ready_to_vote_on_local_block.rs:648` has a `None` arm that cannot be reached under the enclosing commit-decision guard, where `on_propose.rs:839` treats the same state as an `InvariantError`. The two sides disagree about whether that state is impossible or merely tolerated.

## CI

Green apart from one pre-existing integration flake. `clippy`, `fmt`, `docs build`, `machete`, all three test shards and the `integration test` leg pass at this head. The aggregate `Test Results (CI)` check reports `1 fail, 2 460 pass`: `Scenario: All committee members restart after consensus stalls before an epoch boundary` (`tests/features/stall_recovery.feature:7:3`).

That same scenario at that same line fails on this PR's base, `7400eb3cc` on `development` (check run 105598268007, `1 fail, 2 457 pass`). It is not this change. The earlier heads' runs showed `All 2 427 tests pass` only because their integration legs were cancelled by the next push, so the cucumber suite never contributed to the count.

## What was verified over the series

- All four `put_diff` sites classify identically through `ok_lock_failed`; `ExpectedSubstateDown` stays fatal on both the proposer and the replica.
- Both no-vote returns reach `set_no_vote` and return immediately, so the partial diff and the already-incremented fee accumulators die with the change set.
- `lock_assert_not_exist` is reached only from `grant_lock` under `is_output()`; all three branches test the versioned address; reordering them cannot change the outcome; `SubstateRecord::exists` bottoms out in `get_pinned_cf`, not `key_may_exist`, so it is deterministic across nodes.
- `substates_prune_downed_values` clears the value and keeps the record, so "existence, not liveness" survives pruning and nodes cannot disagree on `SubstateExists`.
- The new existence query filters on the same two predicates as `block_diffs_select_last_change` and reads keys only.
- `NoVoteReason` carries no serde/borsh/ts-rs derive, so the new variant and its `details` string are not wire-visible; the `SubstateExists` rename is consensus-visible through `FailedToLockOutputs` and rides the coordinated upgrade, as the description says.
- `an_output_lock_for_a_destroyed_version_is_rejected` genuinely fails on `development`.

sdbondi (commented):
Tested and approved

PR #2652, opened 2026-09-18
feat(consensus): skip suspended validators in leader selection

sdbondi:
A validator that stops proposing currently costs the network a pacemaker timeout every time its slot
in the round-robin comes up. This makes the committee skip that slot instead. It is purely a liveness
accelerator: no punishment, no proof, self-healing, and the validator's votes keep counting towards
every quorum throughout.

Three commits, each mergeable on its own.

### 1. Record per-epoch validator liveness state

`ValidatorConsensusStats` grows `probation_failures`, `votes_since_suspended` and
`suspended_at_height`. The state — Normal, Suspended, Probation — is a pure function of those,
`missed_proposals` and the committed height it is read at; no enum is stored. Missing
`missed_proposal_suspend_threshold` proposals suspends a validator. It gets one slot back once either
`probation_base_votes` of its votes have reached the certificate of a committed block or
`probation_base_blocks` have committed since it was suspended; both waits double per failed probation
up to `probation_max_backoff_exp`. Proposing a block that commits clears every counter.

Leader selection must read this state as of a height every node has committed, not as of the latest
commit, so a new append-only column family keyed `(epoch, validator, committed_height)` records the
counters whenever a commit moves one, and answers "the state as of height H". It is pruned with its
epoch.

Nothing consumes the state in this commit.

### 2. Skip suspended validators in leader selection

The leader of a view becomes the first validator from that view's round-robin position onwards that
is not suspended. Views are remapped, not skipped: the effective leader collects the votes for the
block at its view and proposes the next one, so a skipped validator costs no dummy block, no timeout
certificate and no NEWVIEW.

Every node must pick the same leader for a view, so the state is read as of the block the view's
certificate commits — three views below the justify — which is exactly what every node that can act
on the view has committed, and what a node that has since run ahead still answers from the log. The
same anchor holds after a leader failure, where the proposal carries an older certificate: a
validator regenerating the dummy chain derives the same attribution, and so the same block ids, as
the proposer did. Dummy blocks are attributed to the effective leader of the view they fill, so the
missed proposal committed with them is charged to the validator that actually had the slot.

Checking the proposer against the leader of the view below it is no longer stateless, so it moves
from `check_block` to `validate_local_proposed_block`, where the justify block that anchors it is on
hand.

**CONSENSUS BREAKING.** Nodes on either side of this disagree about who may propose for a view in
which a validator is suspended. It ships with the testnet reset.

### 3. Withhold our vote from a proposer we caught equivocating

A node holding evidence from #2649 that a proposer signed two conflicting votes for one view does not
sign that proposer's block. It does everything else with it: validates, processes, locks and commits
it once the rest of the committee certifies it.

That is the full extent of what local evidence may drive. Only the leader of a view receives the
votes for it, and #2649 records what it finds without sharing it, so two nodes routinely disagree
about whether a validator equivocated — feeding that into leader selection would split the committee
over an equivocator and turn one misbehaving validator into a liveness fault for everyone. If enough
nodes do hold evidence, the slot fails to reach a quorum and the dummy block charges a missed
proposal through the deterministic path in commit 1, which is what suspends it.

### Two deltas from the plan, both found by the tests

- **Probation needs a second path.** With only the vote-based path, the returning validator in
  `a_returning_validator_is_restored` never recovered: a vote counts only once it is in a certificate,
  and a leader stops collecting at a quorum, so a validator that is even slightly slower than its
  committee — exactly the population that gets suspended — can have every vote arrive too late. It
  would then never propose again for the rest of the epoch. `probation_base_blocks` is what guarantees
  the slot comes back; the vote path just makes it faster for a validator that is clearly back.
- **Withholding is a signature, not a no-vote.** Implemented first as a `NoVoteReason`, which drops
  the block's change set and left the accuser's transaction pool out of step with a chain that went on
  to commit the block — it then no-voted everything that followed and never voted again (verified over
  40 blocks). Withholding only the vote message keeps the node in consensus, which the test now
  asserts.

### Testing

- Transition-table unit tests on the liveness state machine.
- `state_store_rocksdb` tests for the as-of query, per-validator and per-epoch isolation, and which
  events write a log entry.
- `consensus_tests`: an absent validator is suspended and stays suspended; its slot is skipped and it
  is charged nothing while skipped, with the committee still committing; a returning validator goes
  Suspended → Probation → Normal; two adjacent suspended validators are both walked past; a node
  holding equivocation evidence withholds its signature from that proposer, the block is certified
  without it, and the node goes on taking part.
- Full `consensus_tests`, `tari_ootle_storage` and `tari_state_store_rocksdb` suites pass (217 tests).

### Constants

`missed_proposal_recovery_threshold` described the eviction NEWVIEW behaviour removed with node
eviction and is replaced by `probation_base_votes`. Defaults: suspend after 5, 5 votes, 100 blocks,
backoff cap 2^6 — worth tuning on esmeralda.


sdbondi (commented):
_Posted by Claude._

**Comment only** (own PR — cannot `--request-changes`; treat as changes requested) — 1 blocking, 3 non-blocking, 2 nits · head `fe1243c3`

The anchoring argument holds up where I traced it. `LIVENESS_ANCHOR_LAG = 3` matches the commit rule in `BlockStore::update_nodes`: processing a block at height `H` with justify `J = H-1` commits `J-2`, so a node that has processed the justify block has committed exactly `J-3` and has written every liveness log entry at or below it — the anchor is the newest height it can read without racing a peer. All five selection sites agree on it in the steady state: proposer (`worker.rs:984`), validator (`on_receive_local_proposal.rs:959`), dummy attribution (`common.rs:306`), voter (`on_receive_local_proposal.rs:428`) and vote collector (`on_receive_vote.rs:99`) all land on `block_height - 3`. The write path is deterministic (`ValidatorConsensusStats::apply` is a pure function of stored counters, thresholds and the committed height), and `LivenessState` is monotone in `as_of`, so a validator granted a probation slot at an anchor is still Probation when the dummy that charges the miss commits. I also checked that the probation failure from a missed slot is logged at a height below the next slot's anchor for any committee of four or more, so the backoff can't be skipped.

---

## Blocking

**1. `on_beat` anchors the skip set on a certificate the proposal will not carry — `crates/consensus/src/hotstuff/worker.rs:984`**

`highest_justified` comes from `get_highest_seen_justified_view`, which is `max(high_pc.height(), high_tc.height())` (`common.rs:487`). The block this path proposes takes its justify from `HighPc::get()` (`on_propose.rs:178`). After a timeout certificate forms at height `T` above the high PC at height `P`, the proposer decides its own leadership at anchor `T-3`, while every validator runs `check_proposed_by_leader` against the proposal's justify at anchor `P-3` (`on_receive_local_proposal.rs:959-966`).

Failure scenario: `P = 40`, a TC forms at `T = 43`, and validator `V` has a liveness log entry at height `38` that suspends it. Anchor `40` says `V` is suspended; anchor `37` says it is not. The node that is round-robin leader of view 43 and the node the walk remaps to disagree about which of them may propose — either the proposal is dropped as `NotLeader` by the whole committee, or the validator the committee expects also declines (it runs the same `on_beat` computation) and nobody proposes. Either way the view is lost to another pacemaker timeout, which is the cost this PR exists to remove.

`on_force_beat` already gets this right — `worker.rs:1097` deliberately reads `HighPc::get(...).height()` for the anchor with a comment explaining why. `on_beat` needs the same. The comment at `worker.rs:982-983` ("The block this would propose carries a certificate for `highest_justified`") is only true when `high_tc <= high_pc`.

## Non-blocking

**2. NEWVIEW routing can split the timeout quorum — `crates/consensus/src/hotstuff/on_next_sync_view.rs:77`**

The sender anchors on its own `HighPc`; the receiver re-derives from the `high_pc` carried in the message (`on_receive_new_view.rs:99`). That is self-consistent per message — no NEWVIEW is misrouted — but it is not consistent *across* senders. At the moment of a leader failure, replicas that received the failed view's block hold a high PC one height above those that did not, so their anchors differ by one. If a liveness log entry falls on that boundary, the committee addresses its NEWVIEWs to two different leaders, neither collects `2f+1` timeout votes, and no TC forms. Before this PR the leader for a timeout view was a pure function of `timeout_height` and could not split this way.

It is narrow and self-healing (the next view re-rolls the anchor), but it is the one selection site where the anchor is genuinely a local value rather than a committed one. Worth either anchoring it on something every sender agrees on, or noting in the comment that a one-height split is accepted here.

**3. The skip set is reloaded per message, not per anchor — `crates/consensus/src/hotstuff/leader_skip.rs:44`, `on_receive_vote.rs:96`**

`load_for_justify` does one `range_iterator` seek per committee member. `check_leader_for_vote` runs it for every vote received, so the vote-collecting leader of a view performs O(committee²) seeks per block, plus one read transaction per vote. The `LeaderSkipSet` docstring ("Loaded once per anchor and then consulted for every view it decides") describes a cache that does not exist — the object is rebuilt at each call site. An anchor-keyed one-entry cache on the worker would both cut the cost and make the doc accurate.

**4. `ValidatorConsensusStats` gains three non-optional CBOR fields — `crates/storage/src/consensus_models/validator_stats.rs:169-181`**

`#[n(2)]`–`#[n(4)]` have no `Option` or default, so any record written by the current release fails to decode after the upgrade rather than reading as `Default`. That is consistent with "ships with the testnet reset", but nothing in the code says so and the failure mode is a decode error at commit time. Worth confirming the reset wipes the state store, or giving the new fields a decode default.

## Nits

**5. Misattributed doc comment — `crates/state_store_rocksdb/src/writer.rs:1959`**

`validator_liveness_log_for_epoch` was inserted between the comment `/// Equivocation evidence is retained for as long as the blocks of the view it indicts, so an operator reading the record can still fetch the blocks it refers to.` and `vote_equivocations_for_epoch`, which it now documents instead. The liveness log's own retention rationale ("pruned with its epoch") is in the PR description but not in the source.

**6. The bounded walk assumes round-robin — `crates/consensus/src/hotstuff/leader_skip.rs:88`**

`effective_leader` walks `0..committee.len()` offsets and falls back to the raw leader if every candidate is skipped. That covers the whole committee only because `calculate_leader` is `height % len`; for any `LeaderStrategy` that is not a bijection over `len` consecutive heights the fallback would silently return a leader that *is* skipped. `RoundRobinLeaderStrategy` is the only impl, so this is a documentation gap rather than a bug — worth stating the assumption on the function or the trait.

---

CI: `test (1/3)` and `integration test` still pending at this head; everything else green. Branch is 3 ahead / 0 behind `development`.

sdbondi:
_Posted by Claude._

All six addressed, pushed as `c16b6070` (fixups squashed into the commits they belong to, so it is still three commits).

**1. `on_beat` anchor — confirmed and fixed.** `get_highest_seen_justified_view` is `max(high_pc, high_tc)` (`common.rs:487`) and `on_propose` takes the justify from `HighPc::get()` (`on_propose.rs:178`), so the two diverge for exactly as long as a TC sits above the high PC. `on_beat` now reads `HighPc::get(...).height()` for the anchor and keeps `highest_justified` as the *view* — the view to propose for and the certificate the proposal carries are two different heights, which the old comment conflated. Same shape as `on_force_beat`.

**2. NEWVIEW split — accepted, now documented.** The alternatives are worse: anchoring on a view-derived height makes routing disagree with the proposer's own leadership check (it uses the high PC, so a node could be sent 2f+1 NEWVIEWs and then decline to propose), and routing by the raw round-robin leader sends them to the validator that is most likely the one that is down. There is nothing every sender agrees on at the moment of a timeout, so the comment now states the one-height split, why it is bounded to one extra timeout, and why the other two options were not taken.

**3. Per-message reload — fixed where it was quadratic.** `OnReceiveVoteHandler` now keeps the set for the view it last answered for, so a leader collecting `n` votes for one block loads it once rather than `n` times. The other sites answer once per block, so they keep loading per call. The docstring no longer claims a cache the type does not have; it says what loading costs and who should hold on to the result.

**4. CBOR fields — fixed, with a test that fails without it.** `#[cbor(default)]` on the three new fields, and `stats_without_the_liveness_fields_decode_as_normal` encodes the old two-field shape and decodes it as a validator in good standing. Verified it fails with the attribute removed. The reset makes this moot, but a decode error at commit time is not the way to find out that someone skipped it.

**5, 6. Fixed.** The equivocation retention comment is back on `vote_equivocations_for_epoch` and the liveness log has its own; `effective_leader` now states that the `committee.len()` bound covers every member because `calculate_leader` is a bijection over that many consecutive heights, and that a strategy which is not would need a different bound.

Local run after the fixes: 218 tests across `consensus_tests`, `tari_ootle_storage` and `tari_state_store_rocksdb`, all passing.

On the anchoring walkthrough — the monotonicity argument for `LivenessState` in `as_of` is the one I would most want a second pair of eyes on, since the whole backoff rests on it.

sdbondi (commented):
_Posted by Claude._

**Comment only** (own PR — cannot click; CI also still pending) — 0 blocking, 1 non-blocking, 1 nit · 6 resolved since `fe1243c3` (1 blocking, 3 non-blocking, 2 nits) · head `c16b6070`

The blocking item is genuinely fixed. Both new items below come out of the cache added for item 3.

## Resolved

- **Blocking 1 — `on_beat` anchor.** `worker.rs:965` now reads `HighPc::get(...).height()` alongside `get_highest_seen_justified_view`, and `worker.rs:988` anchors the skip set on the high PC while still selecting for view `highest_justified`. That is exactly the pair the validator uses: `check_proposed_by_leader` runs at `parent_height = next_height - 1 = highest_justified` against a skip set anchored on `justify_block.height()`, which `on_propose.rs:178` takes from `HighPc`. Verified `HighPc` is always present for a started epoch (`worker.rs:1652` sets it when the epoch genesis is created), so the non-`optional()` read cannot fail the beat path.
- **Non-blocking 2 — NEWVIEW split.** Documented as accepted at `on_next_sync_view.rs:77`, which was one of the two options. The comment states the condition and the cost accurately.
- **Non-blocking 3 — per-message reload.** `OnReceiveVoteHandler` now keeps a one-entry `(epoch, view_height, LeaderSkipSet)` cache. Every vote for one block asks about the same view, so the hit rate is right and the key is correct — same view height means same anchor. See the new item below.
- **Non-blocking 4 — CBOR decode.** `#[cbor(default)]` on the three new fields plus `stats_without_the_liveness_fields_decode_as_normal`. Confirmed the value codec really is minicbor (`codecs/mod.rs:87`, `DefaultCodec<T> = Minicbor<T>`), so the attribute applies to what is persisted, and `#[cbor(default)]` is already used this way elsewhere in the tree.
- **Nit 5 — doc comment.** Moved back onto `vote_equivocations_for_epoch`; `validator_liveness_log_for_epoch` got its own.
- **Nit 6 — walk bound.** The bijection assumption is stated on `effective_leader`.

## Non-blocking

**A stale skip set can now stick for a whole view — `crates/consensus/src/hotstuff/on_receive_vote.rs:115`**

`load_skip_set_for_view` caches whatever the store answers, and `is_leader_for_vote` runs *before* `check_and_collect_vote`, so it runs before any lookahead or eligibility check on the vote.

`validator_liveness_counters_as_of` returns the last entry at or below the anchor and reports nothing when the node has not yet committed it — a node behind the anchor gets a silently stale answer rather than an error. The anchor for a vote on block `H` is `H-3`, which the leader of view `H` has committed *once it has processed block `H`*. If the first vote for `H` is handled before the proposal for `H` (the proposer broadcasts to everyone at once, so a replica near the leader can get its vote in first), the load happens at a height below the anchor.

Before the cache that misread cost one vote and the next vote re-read correctly. Now it is latched for every remaining vote of that view: if the stale set names a different effective leader, the node discards its own quorum and the view is lost to a timeout. Same narrow trigger as the NEWVIEW case — a liveness entry landing exactly on the boundary height — but the failure is no longer self-correcting.

Cheapest guard is to not populate the cache when the anchor is above what the node has committed (`LastExecuted`), leaving those calls on the uncached path.

## Nit

**The new doc comment overclaims — `crates/storage/src/consensus_models/validator_stats.rs:171`**

"decodes as a validator in good standing" is only true when `missed_proposals` is below the threshold. The pre-liveness record in the new test carries `missed_proposals: 9` against a threshold of 5, so it decodes to `Suspended` (or `Probation`, depending on `as_of` and `probation_base_blocks`) — which the test does not assert either way. What the defaults actually buy is that the record decodes at all instead of failing at commit time; the first half of the sentence already says that.

---

CI at this head: `clippy`, all three `test` legs and `integration test` pending; everything else green.

sdbondi:
_Posted by Claude._

Both addressed in `888a8882`.

**Stale skip set latched by the cache — confirmed, fixed differently.** The race is real: a vote for block `H` can be handled before this node has processed `H`, and `validator_liveness_counters_as_of` answers from below the anchor without saying so, which the cache then held for the rest of the view.

Rather than skipping the cache when the anchor is above `LastExecuted`, the cache is now keyed by what this node had committed when the set was loaded, alongside the epoch and view. Committing more is exactly what makes a previously-stale answer authoritative — `LastExecuted` is set to the commit node's height in `BlockStore::update_nodes` — so the entry loaded before block `H` was processed is dropped the moment it is, and the reload happens once rather than per vote. It also costs one point read per vote instead of the conditional's read plus a full uncached load for every early vote, and it does not need the anchor lag spelled out a second time at the call site.

**Doc overclaim — fixed.** The struct comment now says the counters decode at zero, the test is `stats_without_the_liveness_fields_decode_with_the_counters_at_zero`, and its doc says what the record then means is up to the thresholds like any other.

218 tests still pass locally; clippy clean.

sdbondi (commented):
_Posted by Claude._

**Comment only** (own PR — GitHub refuses approve on it; this is a sign-off, another maintainer has to click) — 0 blocking, 0 non-blocking, 0 nits · 2 resolved since `c16b6070` (1 non-blocking, 1 nit) · head `888a8882`

Nothing stands. Both remaining items are fixed and re-verified.

## Resolved

- **Stale skip set latching for a whole view — `on_receive_vote.rs:125`.** The cache key is now `(epoch, view_height, committed)` with `committed` read from `LastExecuted`. That closes it at the right place: the failure needed a set loaded while the node sat below the anchor to survive the node catching up, and committing past it now changes the key and forces a reload. `LastExecuted` is the right notion of committed height here — `BlockStore::update_nodes` sets it to the committed block, and it is what the liveness tests read. The hit rate is unaffected, because nothing commits between the votes of one view.
- **Overclaiming doc comment — `validator_stats.rs:171`.** Now "with those counters at zero", and the test renamed to match. The added sentence about the thresholds deciding what the record then means is accurate.

## CI

Green for this change. The one red mark is the aggregate `Test Results (CI)` check reporting 1 of 2 485 tests failed: the cucumber scenario `All committee members restart after consensus stalls before an epoch boundary` (`stall_recovery.feature:7:3`), which panics with

```
Could not connect to storage: IO error: lock hold by current process ... rocksdb/LOCK: No locks available
```

That is the harness restarting VN1 before the previous process has released the RocksDB lock, not a consensus failure. The same scenario fails with the identical error on every recent `development` run I checked (35340393981, 35337321073, 35329994221, 35323111972, 35320801685), so it is pre-existing and unrelated to this branch. Every other check, including all three unit-test legs and clippy, passes.

## What I verified across the series

The anchoring invariant, which is what the whole design rests on. `LIVENESS_ANCHOR_LAG = 3` matches the commit rule in `BlockStore::update_nodes` — processing a block at height `H` with justify `J = H-1` commits `J-2`, so a node that has processed the justify block has committed exactly `J-3` and has written every liveness log entry at or below it. All selection sites agree on that anchor: proposer (`worker.rs:988` and `worker.rs:1097`), validator (`on_receive_local_proposal.rs:966`), dummy attribution (`common.rs:306`), voter (`on_receive_local_proposal.rs:437`) and vote collector (`on_receive_vote.rs:110`). The one site whose anchor is local rather than committed is the NEWVIEW path, which is now documented as an accepted one-timeout cost.

The write path is deterministic — `ValidatorConsensusStats::apply` is a pure function of the stored counters, the thresholds and the committed height — and `LivenessState` is monotone in `as_of`, so a validator granted a probation slot at an anchor is still in Probation when the dummy that charges the miss commits. The probation failure from a missed slot is logged at a height below the next slot's anchor for any committee of four or more, so the backoff cannot be skipped.

crates/consensus/src/hotstuff/leader_skip.rs
```diff
@@ -0,0 +1,215 @@
...
+        tx: &TTx,
+        epoch: Epoch,
+        justify_height: NodeHeight,
+        committee: &Committee<TAddr>,
+        thresholds: &LivenessThresholds,
+    ) -> Result<Self, StorageError> {
+        let anchor = justify_height.saturating_sub(LIVENESS_ANCHOR_LAG);
+        let mut skipped = Vec::new();
+        for public_key in committee.public_keys() {
+            let counters = ValidatorConsensusStats::liveness_counters_as_of(tx, epoch, public_key, anchor)?;
```
sdbondi, line 56: nit(perf): N + 1, add query that loads counters (if exists) for each member

sdbondi (commented):
Reviewed, tested and approved

PR #2660, opened 2026-09-21
feat(validator-node): compile templates in the background before they are needed

sdbondi:
Third PR in the WASM artifact cache series, after #2609 (header CRC + fsync), #2614 (one cache per
directory) and #2619 (byte-bounded LRU eviction).

## The problem

Compiling a template is unpriced work on the critical path. The first transaction in a block that
calls a template this process has not seen pays its Cranelift compile inside consensus execution —
around 140 ms per MiB of binary, on the thread the block is waiting on. The disk cache removes the
*second* such compile per node, but never the first, and an eviction puts the first one back.

The pricing question is closed in both directions: charging unconditionally is unaffordable
(~0.47 TARI per template per transaction), and charging only on a miss is non-deterministic. So the
compile has to move off the critical path rather than be paid for.

## What this does

A bounded pool of worker threads compiles a template as soon as the node learns it will need one.

**Two triggers.**

- Mempool admission (`handle_new_transaction`), after the validator chain passes and gated on `is_involved`. Only validated transactions, and only ones this node will actually execute — otherwise it is free CPU for anyone who can gossip.
- The commit of a `PublishedTemplate` substate, as a `ConsensusHooks::on_blocks_committed` impl. The commit rather than the publishing execution: the address a template will hold is only settled once its substate commits, and a publish may abort. From the commit on, any transaction may call it and this node has the source but not the compiled form.

**Admission gates the handoff to consensus.** The leader executes inline while building a proposal
(`prepare_transaction` at the `New` stage), so a transaction handed to consensus cold puts the
compile inside the proposal path — where prewarming has bought nothing. `handle_new_transaction`
therefore waits for the prewarm it queued before calling `notify_new_transaction`. No leader can
propose a transaction it has not warmed up for, while every transaction already warm is handed over
with no wait at all.

Three things keep that a latency decision rather than a liveness one:

- **The wait is bounded** at `min(cold_unique_templates * 512ms, 2s)`. 512 ms is about twice the ~260 ms a 1 MiB binary — the largest publishable — takes to compile (measured: 52 ms at 151 KiB, 142 ms at 530 KiB), which absorbs a loaded machine. The 2 s ceiling is the real bound, since nothing caps the distinct templates one transaction may call; the rate only shapes the ramp. It scales linearly although N workers finish K templates in about `ceil(K/N)` compiles' time — deliberately generous, because this is a starvation valve rather than a latency budget. On timeout, on a full queue, or on a failed compile the transaction is handed over regardless, degrading to the pre-prewarm behaviour.
- **The timeout never cancels the compile.** The worker runs on, so the executor that follows joins the in-flight compile through the provider's per-address semaphore rather than starting its own. This is what makes the bound cheap: a wait that expires still leaves the compile most of the way done.
- **Locally introduced transactions are propagated before the wait.** Every involved validator then starts its own compile at the same moment, so the leader's own warm-up doubles as a network-wide timer — by the time it can propose, replicas on comparable hardware are warm too. Leaving the forward at the end of the function would have serialised what is meant to happen in parallel and added compile latency to propagation at the introducing hop.
- **The wait is not taken on the mempool's own task.** That service is one task draining gossip, mempool requests and consensus events from a single `select!`, and its inbound gossip queue drops on overflow rather than pushing back — so a wait taken inline would be paid by every message queued behind it, and a burst of cold templates would cost transactions rather than latency. A transaction with nothing to wait for is handed over inline and in arrival order; one that waits does so on a task of its own, under a permit bounding how many can be held that way at once.

Not covered: `process_requested` in `on_receive_new_transaction.rs`, which inserts straight into the
pool from a `MissingTransactionsResponse` and never reaches the mempool service. A node catching up
still executes cold; a prewarm can be queued there but cannot gate anything, since the transaction is
already in a block.

**Through the shared provider.** The pool holds a clone of the same `MemoryCacheTemplateProvider`
the executor uses, so its existing per-address semaphore is what coalesces a prewarm with an
execution wanting the same template. No new single-flight machinery.

**A memo, never an authority.** A prewarm that has not finished, or never ran, is a cache miss —
`get_template` compiles inline exactly as it does today. Nothing here can change the outcome of an
execution, only its latency.

**Resolution.** `CallFunction` names its template. `UpdateComponentTemplate` names the template that
will replace the component's, which is the only one `update_component_template` loads. `CallMethod`
names a component, so the component is read at its latest version and the template taken from it; a
component in another shard group is absent and resolves to nothing. That read happens on the caller
rather than on a worker, which is what makes the wait exact — the caller blocks on the compiles its
transaction needs and on nothing else. Builtins and templates already resident are never queued.

## Why OS threads and not `spawn_blocking`

A compile is synchronous and CPU-bound, so it can never run on a runtime worker. Of the two ways
off the runtime, a dedicated pool is the one that bounds itself:

- The thread count *is* the concurrency budget. `spawn_blocking`'s pool is 512 threads by default and shared with every other blocking call in the process, so prewarm could hold an arbitrary number of the provider's 100 `CONCURRENT_ACCESS_LIMIT` permits and starve executor lookups. Workers are `(cores / 4).clamp(1, 4)`.
- The workers are long-lived loops parked on `recv()`, which is what `spawn_blocking` asks you not to submit.
- Backpressure is explicit: a bounded queue that drops on full, and a target already queued (or in flight) is not queued again.

## The semaphore fix comes first

`ConcurrentMapSemaphoreGuard::drop` removed its dashmap entry unconditionally, so a guard could
remove an entry a *later* `acquire` had just created. A waiter blocked on the old mutex and an
arrival holding the new one then entered the critical section together — two threads compiling one
template. Rare today; routine once a worker pool races the executor for the same address, which is
why it is the first commit. The entry is now released only by the last guard referencing it, counted
under the shard lock `acquire` also takes.

`cmap_semaphore::tests::an_entry_outlives_every_guard_but_the_last` carries that claim: it takes two
guards on one key, drops the first, and asserts both that the entry survives and that the surviving
guard holds the same `Arc` a later arrival would get. Against the old unconditional `remove` the
entry is gone and it fails, while the two tests either side of it — that keys are independent, and
that the critical section admits one holder — still pass.

## Metrics

`template_prewarm_{enqueued,dropped,compiled,already_resident,failed}` and a `queue_depth` gauge.
`already_resident` counts prewarms whose template was resident by the time a worker reached it —
execution usually got there first, which is the honest "too late" number. `not_found` is counted
apart from `failed`, since a template this node has not synced yet is what catching up looks like
and a load error is not. A true hit rate would need the execution path instrumented and is not in
here.

## Not in this PR

Moving the *cache write* off the execution path. `WasmModuleCache::store` fsyncs a multi-MB tempfile
before renaming, inline, so the first store of a template in a process blocks execution on a flush.
That belongs to the cache in `tari_engine` rather than to this pool, and prewarm already makes it
rarer by doing the store on a worker. It follows as its own PR; the durability is load-bearing (a
torn body reaches `deserialize_unchecked`, past the CRC's coverage) so the write relocates, the sync
does not go.

## Verification

`cargo lints clippy --all-targets` and `cargo nextest run` clean on `tari_validator_node` and
`tari_ootle_template_provider` (51 tests). The prewarm module carries 17 of them, covering the queue
(builtins and resident templates never queued, duplicates collecting a waiter rather than a slot, a
full queue dropping rather than blocking), the wait (completes when its compile does, gives up on
its timeout without cancelling it, scales with the cold templates and stops at the ceiling), the
worker (a resident template is not loaded again, a drained template leaves the queue), and
resolution (a `CallFunction`'s template, a `CallMethod`'s component, a component this node does not
hold, an `UpdateComponentTemplate`'s replacement, one template reached two ways waited on once, and
a wholly resident transaction waiting for nothing).


sdbondi:
_Posted by Claude._

Pushed `0dadaa2` — admission now **gates the handoff to consensus** rather than enqueueing and falling through. The PR body is updated to match.

The reason for the change: the leader executes inline while building a proposal, so a transaction handed over cold pays its Cranelift compile inside the proposal path, which is where prewarming was supposed to have bought something. `handle_new_transaction` now waits on the prewarm it queued before calling `notify_new_transaction`, so no leader proposes a transaction this node has not warmed up for.

Three properties keep that from becoming a liveness question:

- **Bounded at 500 ms.** Sized off the worst compile a transaction can require, not a typical one: the largest publishable binary is 1 MiB (`EngineLimits::max_template_binary_size_bytes`), which is ~140 ms on the faster of the two machines this was measured on and ~265 ms on the slower, doubled to leave room for one compile sitting behind another in the pool. On timeout, on a full queue, or on a failed compile the transaction is handed over anyway.
- **The timeout abandons the wait, never the compile.** The worker runs on, so the executor joins the in-flight compile through the provider's per-address semaphore instead of starting its own. That is what makes an expired wait cheap rather than wasted.
- **`is_local` propagation moved ahead of the wait.** Every involved validator then starts its own compile at the same moment, so the leader's warm-up doubles as a network-wide timer. Leaving the forward at the end of the function would have added compile latency to propagation at the introducing hop and serialised what is meant to happen in parallel.

One design change came with it: resolving a `CallMethod`'s component to its template moved from the worker to the caller. A queued-but-unresolved component would have made the caller wait for a worker to reach it even when the template behind it was already resident; resolving first means the wait covers exactly the compiles the transaction needs, and an all-resident transaction waits for nothing. The cost is a point lookup per referenced component on the mempool task, alongside the reads `transaction_exists` already does there.

Not covered, and called out in the body: `process_requested` inserts straight into the pool from a `MissingTransactionsResponse` and never reaches the mempool service, so a node catching up still executes cold. A prewarm can be queued there but cannot gate anything — the transaction is already in a block.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 1 blocking, 3 non-blocking, 1 nit · head `5d3fd807`

Reviewed the four commits at `5d3fd807` against `development`. No build or test run locally — CI's clippy/test legs are still pending.

## Blocking

**1. The prewarm wait stalls the whole mempool service, not just the transaction waiting on it.** `p2p/services/mempool/service.rs:326`

`handle_new_transaction` is awaited inline inside the mempool service's `select!` loop (`service.rs:107` and `service.rs:117`), and the service is one task. `prewarm_transaction(..).wait().await` therefore parks that loop for up to `PREWARM_WAIT_CEILING`, and while it is parked the service does nothing else at all:

- no other gossiped transaction is validated or has its verdict reported;
- no `MempoolRequest` is served — including `RemoveTransactions`, which is how consensus prunes the mempool after a commit (`node.rs:167`);
- no `HotstuffEvent::EpochChanged` is handled, so `gossip.subscribe()`/`unsubscribe()` at `service.rs:132` is deferred too.

This contradicts what both the PR description and `template_prewarm/mod.rs:207` state outright — *"what a waiting transaction costs is its own admission delay and nothing else's"*. The ceiling is sized as if that were true (2 s is defensible as one transaction's admission latency; it is not defensible as a global mempool freeze), so the bound and the claim it rests on come apart.

Two concrete consequences:

- Inbound transaction gossip is a **lossy** bounded queue. `gossip_queue` (`bootstrap.rs:170`) hands back a byte-budgeted sender that drops on overflow rather than applying backpressure (`networking/core/src/message_mode.rs:159-181`). Slowing the drain by up to 2 s per admission turns a burst of distinct cold templates into *dropped transactions* — under exactly the load this PR exists to improve.
- `service.rs:280-292` already reasons that "gossipsub only holds a message in its validation cache for a few heartbeats", which is why the verdict is reported early. That reasoning applies to every message queued behind the stalled one, whose validation handles age through the stall.

The gating idea itself is sound — I agree a leader should not be handed a cold transaction. What needs to change is where the wait lives. Moving just the `wait().await` + `notify_new_transaction` into a spawned task (`ConsensusHandle` is already `Clone`, `tx_id` is already inserted into `self.transactions` before the wait, so dedup still holds) keeps the property the PR wants while letting the loop keep draining. That does make consensus admission order non-deterministic with respect to arrival order — worth a sentence on why that's acceptable, or a small bound on concurrent waits if it isn't.

## Non-blocking

**2. A panicking worker is neither replaced nor unwound out of `queued`.** `template_prewarm/mod.rs:361-385`

The address is removed from `queued` only after `prewarm` returns normally. If `get_template` panics — the provider chain reaches `self.map_mutex.lock().unwrap()` at `cmap_semaphore.rs:60`, which panics on a mutex poisoned by some *other* thread's panic, and the disk tier deserializes artifacts — the worker thread unwinds and the entry stays in the map forever. Every later `enqueue` of that address then takes the `Entry::Occupied` branch (`mod.rs:146-149`) and hands back a receiver that nothing will ever complete, so each caller pays the full timeout.

It partly self-heals, because `enqueue` checks `is_resident` before the map and an ordinary execution will eventually compile the template. The thread does not: `worker_count()` is `(cores / 4).clamp(1, 4)` (`mod.rs:345-348`), so it is **1** on any machine with fewer than 8 cores, and one panic there silently disables the pool while every admission keeps paying the wait until the 2048-slot channel fills.

Cheap fix: remove the entry from a scope guard rather than after the call, so an unwinding worker still releases its waiters. Respawning (or at least `error!`-logging the loss, since today it is invisible) would be worth having too.

**3. The component-resolution path — the only genuinely new resolution logic here — has no test.** `template_prewarm/mod.rs:474-481`, `mod.rs:262-287`

`FakeComponents` is only ever built via `Default` (`mod.rs:490`), so `template_of` never returns `Some` in any test, and `prewarm_transaction` is never called at all. That leaves untested: the `CallMethod` → component → template lookup, the dedup across `named.chain(instantiated)` (`mod.rs:113-124`), and `StateStoreComponentLookup` in full.

Relatedly, the description claims tests for "a component target loads the template it instantiates" and "an unheld component is no work" and says there are seven. The tree has eleven, and neither of those two is among them.

**4. `UpdateComponentTemplate` warms the template being replaced, not the one that will run.** `crates/transaction/src/v1/unsigned.rs:193-196`

`as_referenced_components` yields the component for `UpdateComponentTemplate` as well as `CallMethod`, so `template_of` resolves that component's *current* template — precisely the one the instruction discards. The template execution will actually load is `new_template` (`crates/transaction/src/v1/instruction.rs:138-145`), and `referenced_templates_iter` matches `CallFunction` only, so it is never queued.

Not a correctness problem — the cache is a memo — but it is a guaranteed-wrong compile, a queue slot, and a real wait charged for it. `new_template` is sitting right there in the instruction.

## Nit

**5. `failed` conflates "not known to this node" with "compile error".** `template_prewarm/mod.rs:403-412`

`Ok(None)` and `Err(e)` both increment the same counter. A template this node has not synced yet is an expected outcome during catch-up; a compile that errors is not. Collapsing them makes the series hard to alert on, and the split costs one more counter.

## Sound

- **The `cmap_semaphore` fix is correct.** `acquire` clones the `Arc` while holding the dashmap shard lock (`entry().or_insert_with().clone()`, `cmap_semaphore.rs:26-30`), and `remove_if` evaluates its predicate under that same shard lock, so the count cannot be observed stale. During `Drop::drop` the guard's own `Arc` is still alive, making `strong_count == 2` exactly "the map and the last guard"; with two live guards the first sees 3 and leaves the entry, the second sees 2 and removes it — so neither the double-entry race nor a leak is possible. The new test does discriminate: under the old unconditional `remove`, the first guard's drop deletes the entry, the arrival's `acquire` creates a fresh mutex, and both it and the waiter sit in the critical section together.
- **Hook timing is right.** `ConsensusHooks::on_blocks_committed` is documented as firing with "the ancestor blocks whose substates have just been written to the state store" (`crates/consensus/src/traits/hooks.rs:17`) and is called from `on_receive_local_proposal.rs:442` after the decision is persisted, so `hooks.rs:30-41` cannot queue a template whose substate is not yet readable.
- **The shared receiver does not serialise the pool.** `Worker::run` (`mod.rs:363-366`) drops the receiver lock before compiling, so a parked worker only blocks the *handoff*, not the compile — the comment at `mod.rs:312-314` is accurate.
- **The mempool restructure is behaviour-preserving** on the non-involved path: the early return at `service.rs:304-311` propagates for `is_local` and skips the insert/notify exactly as the old fall-through did.
- **`is_resident` is correctly memory-only** (`memory_cache.rs:166-170`) — builtins plus the moka cache, deliberately not the disk tier — which is what keeps the "resident" check from hiding a deserialize.

sdbondi:
_Posted by Claude._

All five actioned in `191ad216`. No pushback; the blocking one was right and I had talked myself out of it during the original design.

## 1 — blocking, the wait stalls the whole service

Fixed. What I had reasoned, and what was wrong with it: I considered the inline await early on, decided it was acceptable because "the wait is zero in the common case" and because blocking the loop gave "natural back-pressure matching the pool's capacity". The first half is true and irrelevant; the second is simply false, and your citation of `message_mode.rs:159-181` is why — the inbound gossip sender drops on overflow, so there is no back-pressure to be had, only loss. A stall does not slow admission down, it deletes transactions. That inverts the argument.

The shape is the one you suggested, with one refinement: a transaction whose templates are all resident still hands over **inline**, so the common case keeps arrival order and keeps this function's error path (`ConsensusChannelClosed` still propagates). Only a transaction that actually has something to wait for goes to a task, which is also what bounds how many tasks exist — `PrewarmWait::is_empty()` is the discriminator.

On the two open questions you raised:

- **Ordering.** Non-determinism is acceptable and I have said why in the code rather than only here: consensus holds its pool as a set and a proposal orders from it by its own rules, so arrival order decides which transaction a leader sees first, not what any block contains.
- **Bounding concurrent waits.** Added, because the natural bound I first assumed is not one: `enqueue` returns a receiver for a template *already* queued, so N transactions calling one cold template spawn N waits. `MAX_CONCURRENT_PREWARM_WAITS` is 64 owned permits; without a permit the transaction is handed over immediately, the same degradation a full prewarm queue produces.

The claim at `mod.rs:207` that you flagged as contradicted is now true rather than deleted, and says why it is true.

## 2 — a panicking worker

`prewarm` runs under `catch_unwind`, so the release from `queued` happens on the unwind path too and the waiters are freed. The panic is `error!`-logged, which it was not before. I also took the cause you named at the source: `ConcurrentMapSemaphoreGuard::access` now recovers a poisoned mutex instead of unwrapping. That mutex guards `()`, so there is no half-written state to protect and propagating the poison only converted one panicking load into a panic for every later caller of that key.

## 3 — the resolution path was untested, and the description claimed tests that do not exist

Both true, and the second was my error: those two tests existed against the earlier implementation and I removed them when resolution moved from the worker to the caller, without updating the body. Six tests added covering exactly the gap — a `CallFunction`'s template, a `CallMethod` through `FakeComponents`, a component this node does not hold, the `UpdateComponentTemplate` case from finding 4, one template reached both ways deduplicated to a single wait, and a wholly resident transaction waiting for nothing. `prewarm_transaction` is now exercised over real sealed transactions. The body's verification section is rewritten to describe the 17 tests that are there.

## 4 — `UpdateComponentTemplate`

Confirmed at `processor.rs:446-464`: `update_component_template` loads `new_template` and never the template it replaces. `prewarm_transaction` no longer goes through `as_referenced_components`, which conflates the two instruction shapes; it walks the instructions itself and takes `new_template` for an update and the component only for a `CallMethod`.

## 5 — `failed`

Split. `not_found` for `Ok(None)`, `failed` for `Err`.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 2 non-blocking, 1 nit · 5 resolved since `5d3fd807` (1 blocking, 3 non-blocking, 1 nit) · head `191ad216`

Re-reviewed `191ad216` against `5d3fd807`. Everything I raised is genuinely addressed; the two items below are new, and neither blocks.

## Resolved

**The mempool loop no longer pays the wait.** `service.rs:343-377`. Verified the three paths: an empty wait is handed over inline (`:347`) so the common case keeps arrival order and the `Result` path; a wait with no permit available is handed over inline (`:357-363`); only a wait that has both work and a permit moves to a spawned task. `self.transactions.insert(tx_id)` still precedes all of it (`:325`), so dedup holds for the whole wait window, and `prewarm_waits` at 64 bounds both the held transactions and the task count. The ordering consequence is stated and argued at `:380-385` rather than left implicit, which is what I was after. `mod.rs:217-225` no longer claims the property the old code didn't have.

**Worker panics no longer strand waiters.** `mod.rs:396-403`. `catch_unwind` sits inside the loop and above the `queued.remove`, so the release happens on either path and the worker keeps running. `AssertUnwindSafe` is defensible here — the closure touches `self.provider`, whose shared state is a moka cache and the dashmap semaphore, both internally synchronised, and a panic mid-`get_template` can at worst leave a cache entry unwritten.

The `cmap_semaphore.rs:57-63` change is a better fix than the one I suggested and stands on its own: the mutex guards `()`, so there is no invariant a panic can break, and propagating the poison would have converted one panicking load into a permanent panic for every later caller of that key.

**The resolution path is now tested.** `mod.rs:688-772`. Six tests over `prewarm_transaction` covering `CallFunction`, `CallMethod` through a held component, an unheld component, `UpdateComponentTemplate`, the same template reached by two routes deduping to one wait, and an all-resident transaction waiting for nothing. `FakeComponents::holding` is what was missing before. `StateStoreComponentLookup` is still untested, which I'm fine with now that the seam is the only thing between it and covered logic.

**`UpdateComponentTemplate` warms the right template.** `mod.rs:125-141`. Checked the engine rather than taking the comment's word: `update_component_template` (`crates/engine/src/transaction/processor.rs:445-555`) calls `get_template(&new_template)` and nothing else — the component's current template is never loaded, even when a `migrate` function runs, since the migration frame executes against the target template. The comment at `mod.rs:128-129` is accurate. Rewriting the walk as a per-instruction `template_for` also reads better than the `named.chain(instantiated)` split it replaces.

**`not_found` split from `failed`.** `metrics.rs:48-58`, `mod.rs:433-441`.

## Non-blocking

**1. `catch_unwind` cannot fire in any shipped binary.** `Cargo.toml:337-339`, `mod.rs:396-403`

`[profile.release]` sets `panic = 'abort'`, and `ci-build` inherits it. A panicking worker therefore aborts the process; the catch is live only in `dev`/`test`. That's not a bad outcome — for a validator, abort is arguably the honest one — but `mod.rs:396-399` states "Catching leaves the pool running and lets the release below happen, which is what keeps an unwind from stranding every waiter on this address for the life of the process", and that is not a property the release binary has. Per the repo's comment rule it should be a true present-tense invariant: either scope the claim to where unwinding is enabled, or drop the catch and let the abort speak. The poison fix is worth keeping either way.

**2. One read transaction per `CallMethod`, on the mempool loop.** `mod.rs:281-290`, `service.rs:343`

`StateStoreComponentLookup::template_of` opens a fresh `with_read_tx` per component and passes `iter::once(&id)` to `substates_get_any_max_version` — an API that takes an iterator precisely so a batch can share one snapshot. A transaction with many `CallMethod` instructions now creates that many RocksDB read transactions synchronously on the `select!` loop. Individually cheap, and the loop is no longer the thing I was worried about, but this is the one part of the admission path that grew a per-instruction store cost. Widening the seam to `templates_of(&[ComponentAddress]) -> Vec<TemplateAddress>` would make it one snapshot per transaction.

## Nit

**3. A closed consensus channel is now a `warn!` for a waiting local transaction.** `service.rs:369-374`

On the inline paths a send failure still becomes `MempoolError::ConsensusChannelClosed` and reaches the JSON-RPC submitter. On the spawned path it can only be logged, so `submit_transaction` returns success for a transaction consensus never received. Only reachable during shutdown, so this is a note rather than a request.

## On approving

Nothing blocking stands at this head. Holding the click until `clippy`, `test (1-3/3)` and `integration test` report — the new tests pull `tari_crypto`, `rand`, `TransactionBuilder` and `args!` into the module for the first time, and I have not built locally. I checked each against the tree (`rand` 0.10 at `Cargo.toml:263` with `rand::rng()` used elsewhere, `builder_localnet`/`build_and_seal`/`update_component_template` all present with these signatures, `args` re-exported in both namespaces as the wallet crates already rely on), but that is not a substitute for the compiler.

sdbondi:
_Posted by Claude._

Both non-blocking items actioned in `676903fb`, plus a flake of my own that surfaced while verifying them. The nit I am leaving, with reasoning.

## 1 — `catch_unwind` cannot fire in a shipped binary

Confirmed: `Cargo.toml:337-339` sets `panic = 'abort'` for `release`, and `ci-build` inherits it. So the catch was live only in `dev`/`test` while its comment asserted a property of the running validator.

I took the second option you offered and removed it rather than scoping the claim. A branch that cannot execute where the comment points, kept alive by a comment explaining where it does execute, is more words and more code for no production effect — and abort is the honest outcome for a Cranelift compile that panics. The poisoned-mutex recovery stays: it is one token, it matches what `cache.rs` and this module already do with `PoisonError::into_inner`, and it addresses a different thread's panic rather than this one's.

## 2 — one read transaction per `CallMethod`

Right, and the iterator parameter was the tell. `ComponentTemplateLookup` is now `templates_of(&[ComponentAddress]) -> HashMap<ComponentAddress, TemplateAddress>`: `prewarm_transaction` collects every component the transaction calls into, resolves them in one `with_read_tx`, then walks the instructions against that map. One snapshot per transaction, whatever its instruction count. The trait doc says why it takes the whole set, so the next implementor does not put it back.

`template_for` and `called_component` are free functions now that the map is passed in, which also made the two instruction shapes easier to keep apart.

## 3 — a closed consensus channel is only logged on the spawned path

Leaving it, and here is why rather than just agreeing it is minor. `submit_transaction` reports *admission* — validated, deduplicated, propagated, and accepted into this node's mempool — and all of that has happened before the spawn. Consensus handoff is guaranteed from that point except during shutdown, which is the only way the channel closes. Making the RPC wait for the handoff would mean waiting for the prewarm too, which is what the blocking finding was about.

## A flake I introduced, caught on the way

`cmap_semaphore`'s test staged three threads with `sleep(50ms)` to get one blocked inside the critical section before another dropped its guard. It failed once in a full `-p tari_validator_node -p tari_ootle_template_provider` run and passed in isolation — a coin flip under a loaded runner, which is the worst possible test to ship alongside a concurrency fix.

Replaced with assertions on the property itself, single-threaded and deterministic: a guard released while another holds the key leaves the entry behind, the surviving guard and the next arrival hold the same `Arc`, and the last guard leaves nothing. Plus one that two keys are independent and one that the critical section admits a single holder. I checked they still discriminate — against an unconditional `self.map.remove(&self.key)`, `an_entry_outlives_every_guard_but_the_last` fails and the other two pass, which is the right split.

53 tests green across both crates, clippy clean.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 0 non-blocking, 1 nit · 2 resolved since `191ad216` (2 non-blocking) · head `676903fb`

Re-reviewed `676903fb` against `191ad216`. Both open items are closed and the code is in a state I'd approve; the only thing left is the PR description.

## Resolved

**One snapshot per transaction.** `mod.rs:301-326`. `templates_of(&[ComponentAddress])` replaces the per-component seam, and `StateStoreComponentLookup` now builds the whole id list and passes `ids.iter()` to a single `with_read_tx`. Checked the pieces rather than assuming:

- `slice::Iter` satisfies the `ExactSizeIterator` bound on `substates_get_any_max_version`.
- The rocksdb impl (`state_store_rocksdb/src/reader.rs:1485-1512`) `continue`s past an id missing from the head index and makes no assumption about result length, so a component this node does not hold simply drops out of the map, and a component named by two `CallMethod`s yields a duplicate record that the `collect` into a `HashMap` collapses. Both are the behaviours the new doc comment claims.
- The demultiplexing is the part that had to change and did: the old single-id code ended in `find_map(|record| record.into_substate_value())`, which over a batch would have attributed the first record's template to every component. Keying each record by `record.substate_id.as_component_address()` is correct, and the partial borrow is fine because `as_component_address` returns an owned `ComponentAddress` before `into_substate_value(self)` moves the record.

Hoisting `called_component`/`template_for` out to free functions reads better than the `&self` method they replace, now that the lookup is no longer per instruction.

**`catch_unwind` removed.** `mod.rs:413`. Given `panic = 'abort'` at `Cargo.toml:337-339` this is the honest resolution — nothing was being caught in a release build, and the comment no longer claims a property the binary lacks.

**The semaphore tests are better than what I signed off on.** `cmap_semaphore.rs:76-130`. I checked that the replacement still discriminates against the bug, since a deterministic rewrite easily stops doing that: `an_entry_outlives_every_guard_but_the_last` takes two guards on one key, drops the first, and asserts the entry survives — under the old unconditional `remove` that entry is gone and the `expect` fails. It also asserts `Arc::ptr_eq` against the surviving guard's mutex, which is the actual invariant, and drops the dashmap `Ref` before dropping the second guard, which it has to do or `remove_if` would deadlock against the shard lock it holds. Drop order in `one_key_admits_one_holder` works out too (`other`, then `_access`, then `guard`), so the last guard releases with the mutex already free. Trading two `thread::sleep(50ms)` synchronisation points for three deterministic tests is a straight win.

## Nit

**The description still describes the previous two heads.** It names `cmap_semaphore::tests::an_arrival_contends_with_a_waiter_that_took_the_mutex_before_it` as the test that fails against the old `drop` — that test no longer exists, and `an_entry_outlives_every_guard_but_the_last` is what carries the claim now. "Seven new tests" is also stale (there are more, and they cover different things). Worth a pass over the description before merge, since it is the record of why the semaphore commit leads.

## On approving

Nothing blocking or non-blocking stands. Holding the click only because the workflow for this head has just started — `clippy`, `test (1-3/3)` and `integration test` have not registered yet. I'll approve once they are green.

sdbondi:
_Posted by Claude._

Description updated.

The semaphore section named a test I deleted two heads ago. It now names `an_entry_outlives_every_guard_but_the_last` and states the discriminating check the way I actually ran it: two guards on one key, drop the first, assert the entry survives *and* that the surviving guard holds the `Arc` a later arrival would get — failing against the unconditional `remove` while the two tests either side of it keep passing.

The "seven tests" line was already gone; the verification section was rewritten alongside `191ad216` to describe the 17 that are there and what each group covers.

Thanks for checking the rewrite still discriminates rather than taking it on the commit message — that was the risk in swapping a threaded test for a deterministic one, and you're right that the `Ref` drop before the second guard is load-bearing: holding it across `remove_if` deadlocks on the shard lock. I hit that while writing it.

CI at this head: 8 green, 7 pending, nothing red.

sdbondi:
_Posted by Claude._

`compiled` fixed in `3377b475`. The log-target half is #2663, since it is engine code that has nothing to do with this branch.

## What `compiled` was counting

Everything `get_template` returned `Ok(Some(_))` for — so the tariswap control, which only read an artifact the disk cache already held, incremented it exactly as a cold Cranelift compile would. About 1.2 ms of work and about 80 ms of work, indistinguishable in the series meant to show how much of the second this pool moved off the critical path.

## Why not just split the counter

A worker asks the outer provider for a template and is not told which tier answered. Getting that answer means a new method on `TemplateProvider` or on the residency trait, threaded down through `MemoryCacheTemplateProvider` to `DiskCachedWasmTemplateProvider`, so the pool can be told something only a metric consumes. That is the trade #2662 lost on, and it is a worse deal here.

The two are told apart by how long they take instead, which needs nothing new from the provider:

- `compiled` → `loaded`, which is what it counts: a prewarm that made a template resident, by whichever route.
- `template_prewarm_load_seconds`, a histogram of how long a worker spent on one template. The modes are two orders of magnitude apart, so a deserialize lands in the first two buckets and a compile around the seventh. Buckets double from 1 ms, so the fast mode is resolved rather than piled into the bottom bucket.

That also gives what a counter split would not: total compile time the pool absorbed, rather than a count of compiles whose cost still has to be guessed at.

53 tests green, clippy clean.

sdbondi (commented):
Tested and approved

PR #2661, opened 2026-09-21
perf(engine): write cached artifacts on the cache's own thread

sdbondi:
Fourth PR in the WASM artifact cache series, after #2609 (header CRC + fsync), #2614 (one cache per
directory), #2619 (byte-bounded LRU eviction) and #2660 (background compilation). Independent of
#2660 — it touches only `crates/engine/src/wasm/cache.rs`.

## The problem

`WasmModuleCache::store` did all its work on the caller's thread: serialize the wasmer artifact,
write it to a tempfile, `sync_all`, rename, update the index, evict. Its caller is an execution —
on a validator the first call to a template inside consensus, on the indexer a dry run — so the
first store of a template in a process put a multi-megabyte flush to the device inside that
execution.

The durability is not optional. `store` publishes by rename, which can expose contents still sitting
in the page cache, and the artifact body lies past the header CRC's coverage: a torn body reaches
`deserialize_unchecked`. So the sync stays where it is and the write moves off the caller.

## What this does

`WasmModuleCache::open` spawns one writer thread per cache. `store` now offers the artifact to that
thread over a bounded queue and returns.

- **Best-effort throughout, as before.** A full queue drops the artifact and a failed write is logged. Either way the caller's compiled module is valid and the only cost is a recompile the next time that template is wanted cold — the same cost the cache already accepted for an eviction.
- **A queued artifact is cheap.** It is already held in memory by whoever compiled it, so a queue slot costs a pointer. 64 of them.
- **A failed write no longer ends the writer.** `the_writer_outlives_a_write_it_cannot_complete` makes a write fail (the directory is gone) and asserts the next artifact still lands.
- **One writer per cache orders the writes.** Two stores of one address could previously race for its filename through separate tempfiles; now they are serialized.

The writer holds a handle onto the same directory and index whose own queue sender is disconnected.
A full clone would keep the queue's sender alive from inside the thread draining it, so the queue
would never close and the thread would never end.

## What callers see

`store`'s contract changes from "written when this returns" to "written before it is wanted again",
which is the next transaction to call the template at the earliest. Nothing reads a cache entry it
has just written — the compiled module is returned directly to the caller and held by the in-memory
cache above it.

Tests that assert on the file after a store now `flush()` first, a test-only barrier through the
same queue.

## Why this comes before keeping the publish artifact

The next step in the series has `publish_template` hand the artifact its compile already produced to
the cache instead of discarding it, so the same binary is not compiled a second time by prewarm or
on first call. That puts a `store` directly in the publish execution path, which is exactly where an
inline flush would be worst. This lands first.

## Verification

`cargo lints clippy --all-targets -p tari_engine --features wasm-cache` and
`cargo nextest run -p tari_engine --features wasm-cache` clean; 16 cache tests, two of them new.


sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 3 non-blocking, 3 nits · head `56dd57f8`

Reviewed the whole of `crates/engine/src/wasm/cache.rs` at this head plus both consumers. No blocking findings. The core mechanism is right; what I want on the record is that a cache that stops working is now nearly invisible in the logs.

## Verified sound

- **The barrier really is a barrier.** `store`'s `try_send` and `flush`'s `send` go through the same `SyncSender` and a single `mpsc::Receiver` drains in offer order, so "every artifact queued before this call has been written" holds by construction rather than by timing.
- **`clone_without_writer` is the right shape, for the stated reason.** A full clone would keep a sender alive inside the thread draining the queue, so the queue would never close. The `sync_channel(0)` whose receiver is dropped at the end of the function is inert, and nothing on the writer's handle calls `store`.
- **Nothing reads back what it just stored**, which is what the "what callers see" section rests on. `DiskCachedWasmTemplateProvider::get_template` (cache.rs:820) and `TemplateManager::load_template` (applications/tari_indexer/src/template_manager/manager.rs:129-134) both return the compiled module directly. `has_template`'s `path_for(..).exists()` is a fast path only — a miss falls through to the inner provider, so a not-yet-written artifact costs it the shortcut, not the answer.
- **One thread per process, not per request.** `WasmModuleCache::open` is called once in each consumer (applications/tari_validator_node/src/bootstrap.rs:344, applications/tari_indexer/src/bootstrap.rs:337) and the handle is cloned downstream.
- **`the_writer_outlives_a_write_it_cannot_complete` discriminates.** If the writer had died on the failed write, `flush`'s `send` would fail, `done.recv()` would return at once, the second artifact would never be written and the final assert would fail. The test earns its name.
- **The durability argument is unchanged and still correct** — `write_durable` keeps `sync_all` before the rename, which is the only thing covering the body the header CRC cannot reach.
- The index/eviction interleavings between a caller-thread `discard` and the writer's `record` are the ones the inline `store` already had; the index is documented as a tally over an authoritative directory, and the worst outcome is still an unaccounted file.

## Non-blocking

1. **`store` conflates a full queue with a dead writer, and both are debug-level** — cache.rs:630-634. `try_send` fails on `Full` *and* `Disconnected`. Disconnected means the writer is gone, and from that point every store for the life of the process drops silently with a message saying the writer "is behind", which it is not. Spawn failure does `warn!` at open, but a panic in the writer leaves no warn-level evidence at all. Matching on `TrySendError` and warning once on `Disconnected` would also bring the message in line with naming the situation rather than the condition.

2. **Sustained drops are invisible** — cache.rs:630. Even the honest `Full` case: on a device where the writer stays behind, every store drops, every template recompiles on every cold call, and the only trace is a debug log. Before this change a slow disk showed up as slow execution; now it shows up as a cache that quietly does nothing. A rate-limited `warn!`, or a dropped count folded into the `info!` that already reports the cache's size, would make that legible to an operator.

3. **`WRITE_QUEUE_CAPACITY = 64`, and the doc reasons away the cost that is actually there** — cache.rs:284-286. The clone itself is genuinely cheap: `wasmer::Module` is a single `Arc<Artifact>` (wasmer 7.4.0, `src/backend/sys/entities/module.rs:25`), as is `Engine`, and `template_def` is already behind an `Arc`. But "already held in memory by whoever compiled it" is what stops being true the moment the execution that compiled it returns — the queue slot is then the artifact's last owner and is extending its lifetime, not sharing it. This module's own docs put artifacts at "hundreds of KiB to tens of MiB", so a full queue can pin ~64 artifacts the executor has already let go. The queue's job here is to absorb a burst, not to buffer a cache's worth of compiled code; 8 would decouple the caller just as well, or bound it by bytes the way the cache itself is. At minimum, say what the bound is rather than implying there isn't one.

## Nits

4. **Double `flush()`** — cache.rs:971-972, 977-978, 1069-1070. One barrier is sufficient by construction (see above); the second is a no-op. The pair at 977-978 is after a cache *hit*, where no store happened at all. Two in a row reads like the barrier is only probabilistic, which undersells it.

5. **The tempfile comment is now stale for the same-process case** — cache.rs:650-652: "a cache directory is shared by concurrent threads and by whatever other processes point at it, so two `store`s of one address can run at once". One writer per cache is precisely what stopped concurrent threads racing here — the PR body says so. The surviving reason for a per-call tempfile name is the other-process case.

6. **`enum Write` at module scope** — cache.rs:289, in a file where `write_durable` does `use std::io::Write` (cache.rs:740). It compiles because that import is function-local, but it is a trap for the next edit that needs `io::Write` higher up. `WriteRequest` reads no worse.

CI: all legs green except `test (1/3)` and `integration test`, still pending at the time of writing.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 4 non-blocking, 3 nits · 0 resolved since `56dd57f8` · head `bdc59537`

`bdc59537` is docs-only and does not touch any of the six items from the previous pass; all still stand as written. It adds one new non-blocking finding of its own.

## New: the comment on `impl TemplateProvider` states a premise that is already false

cache.rs:784-790 says the two lookups are sound "only while every artifact in the cache came from a successful `inner.get_template`", and warns that "a writer that puts an artifact into this cache from anywhere else breaks that".

Such a writer already exists. The indexer opens **one** `WasmModuleCache` and hands a clone to both consumers — applications/tari_indexer/src/bootstrap.rs:337, with the comment "One instance serves every consumer of the directory". One of those consumers is `TemplateManager`, whose `load_template_with_cache` stores into it (applications/tari_indexer/src/template_manager/manager.rs:133) from its own `templates` table, never through the dry-run provider's `inner`.

The conclusion survives, but for a longer reason than the one given: `fetch_cached_template` only returns `Active`/`Deprecated` rows (manager.rs:166), those rows are written by `add_and_load_template` only after `substate_manager.get_substates` resolved the template substate (manager.rs:290-305), and templates are immutable and never destroyed. Builtins never reach the cache from either side — the provider bypasses them on address, and `load_template_with_cache` returns from `build_in_template_cache` first (manager.rs:126).

As written, the comment sends the next reader to grep for writers, find manager.rs:133, and conclude the invariant is already broken. It should name that writer and say why it satisfies the requirement, which is what makes the requirement legible to the writer after it.

Minor, and separable: this comment is about the cache-hit-versus-`inner` invariant, not about moving the write off the caller's thread. It would read more naturally in its own change.

## Still standing from `56dd57f8`

Re-checked each against the new head; the code behind all six is unchanged.

**Non-blocking**
1. `store` conflates a full queue with a dead writer, both at `debug!` — cache.rs:637-641.
2. Sustained drops leave no warn-level trace, so a slow device turns the cache into a silent no-op — cache.rs:637.
3. `WRITE_QUEUE_CAPACITY = 64` can pin ~64 artifacts past their caller's lifetime; the "costs a pointer" note reasons away the clone, not the retention — cache.rs:284-286.

**Nits**
4. Double `flush()` where one barrier is sufficient by construction — cache.rs:978-979, 984-985, 1076-1077.
5. The per-call-tempfile comment still cites concurrent threads, which one writer per cache is precisely what stopped — cache.rs:657-659.
6. `enum Write` at module scope against the function-local `use std::io::Write` — cache.rs:289, cache.rs:747.

CI at this head: `clippy`, `integration test`, `test`, `web ui build` and the signed-commits check still pending; nothing red.

sdbondi:
_Posted by Claude._

Line-number correction to the previous comment: items 1, 2, 5 and 6 sit before `bdc59537`'s insertion point, so they did not shift. At head `bdc59537` they are:

- 1 & 2 — `store`'s `try_send` and its `debug!`: cache.rs:630-634 (not 637-641).
- 5 — the per-call-tempfile comment: cache.rs:650-652 (not 657-659).
- 6 — `use std::io::Write` in `write_durable`: cache.rs:740 (not 747).

Items 3 (cache.rs:284-286), 4 (cache.rs:978-979, 984-985, 1076-1077) and the new finding (cache.rs:784-790) were correct. The findings themselves are unchanged.

sdbondi:
_Posted by Claude._

All seven actioned — the six from the first pass and the new one on the comment. No pushback on any of them.

**The new finding is right, and the comment was worse than imprecise.** `TemplateManager::load_template_with_cache` (manager.rs:129-134) does store into the same `WasmModuleCache` the dry-run provider reads from — the indexer opens one and clones it to both consumers (bootstrap.rs:337). A reader following my comment would have gone looking for a second writer, found that one, and concluded the invariant was already broken.

I checked the chain you gave rather than taking it: `fetch_cached_template` refuses anything but `Active`/`Deprecated` (manager.rs:166), and the `Active` rows come from `add_and_load_template` inside the loop over `fetched_templates` (manager.rs:291-305), each of which is a template substate `substate_manager.get_substates` resolved. So the artifact corresponds to a template that exists on chain, which is the property that actually matters.

The comment now states that property as a fact about the *directory* rather than about this type, names both writers and how each establishes it, and puts the obligation on a third writer. It stays its own commit.

**1 & 2 — `store` conflates a full queue with a dead writer, both quiet.** Now matched on `TrySendError`, and both arms `warn!`. `Disconnected` says the writer has stopped and nothing further will be cached for the life of the process, which is what it means; `Full` says the writer is behind and names the consequence. Both are warnings because your framing of (2) is the point: a slow device used to show up as slow execution and now shows up as a cache that quietly does nothing. The volume is bounded by how often a cold template is loaded, which is not often.

**3 — `WRITE_QUEUE_CAPACITY`.** Cut to 8, and the doc no longer reasons away the wrong cost. It now says the clone is cheap *and* that the queue outlives the execution that compiled it, so a full queue is the last owner of that many artifacts at this module's own "hundreds of KiB to tens of MiB" each. You were right that the retention, not the clone, is what the number is buying.

**4 — double `flush()`.** Both pairs gone, and the one after the cache *hit* in `round_trip_compile_then_deserialize` is gone entirely rather than halved: no store happens there, so there was nothing to barrier on.

**5 — stale tempfile comment.** Rewritten to the surviving reason: another process pointing at the same directory. The pid separates this process's writes from theirs, the counter separates this call from the next.

**6 — `enum Write`.** Renamed `WriteRequest`.

`cargo lints clippy --all-targets -p tari_engine --features wasm-cache` clean, 16 cache tests green, full `tari_engine` suite green (476/476 at `61890566`).

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 0 non-blocking, 1 nit · 7 resolved since `bdc59537` (4 non-blocking, 3 nits) · head `61890566`

`61890566` closes every item from both previous passes. Re-verified each against the code rather than the commit message.

## Resolved

1. **Full vs disconnected, and both audible** — cache.rs:629-653. `try_send`'s result is now matched on `TrySendError`, and the two cases say different things: one names a writer that is behind, the other one that has stopped and will not cache anything again this process. Both at `warn!`. This closes non-blocking 1 and 2 together. I checked the rate before worrying about noise: `store` runs only after a disk-cache miss and a full cranelift compile, so the warn is bounded by the compile rate, not by lookups — it will not flood.
2. **Queue capacity and the note over it** — cache.rs:280-287. 64 → 8, and the comment now says the thing it previously reasoned away: the clone is cheap, but the queue outlives the execution that compiled the artifact, so a full queue is the last owner of that many. Sizing it for "decouple a caller from one slow write" rather than for buffering is the right frame, and 8 makes the drop-costs-a-recompile tradeoff the one actually being taken.
3. **Double `flush()`** — cache.rs:1004, 1099. Down to one barrier each, and both flushes after the *second* `get_template` in `round_trip_compile_then_deserialize` are gone, which is right: that call is a cache hit and stores nothing. Every remaining assert still sits behind a barrier that covers the store it depends on.
4. **Tempfile comment** — cache.rs:670-672. Now cites the cross-process case only, which is the reason that survives one writer thread per cache.
5. **`Write` → `WriteRequest`** — cache.rs:292, and the four use sites. No shadowing against the function-local `use std::io::Write` at cache.rs:760.
6. **The `impl TemplateProvider` doc** — cache.rs:804-816. Rewritten to put the property on the directory, name both writers, and state what a third owes. I confirmed both writers as described: this provider stores only after `inner.get_template` returned `Some` (cache.rs:848), and `TemplateManager::load_template_with_cache` stores from rows that `add_and_load_template` writes after `substate_manager.get_substates` resolved the substate (applications/tari_indexer/src/template_manager/manager.rs:133, 290-305).

Nothing else moved, and I re-read the writer, `clone_without_writer`, `write_now` and `write_durable` at this head — the ordering and durability arguments from the first pass are unchanged and still hold.

## Nit

**One exception to the new doc's supporting clause** — cache.rs:810-812 says the indexer's rows "reach `Active` only through `add_and_load_template`". `TemplateManager::initialize` also inserts builtin rows at `TemplateStatus::Active` (manager.rs:80). It is harmless — `load_template_with_cache` returns from `build_in_template_cache` before it ever reaches the cache (manager.rs:126), and this provider bypasses builtin addresses outright — so no builtin ever becomes an artifact and the load-bearing sentence above it stays true. But it is the same trap the rewrite was for: a reader greps `TemplateStatus::Active` in that file, finds line 80, and has to reconstruct why it does not count. Naming the builtin short-circuit in the same breath would settle it.

CI at this head is still running (clippy, test, fuzz, integration, ledger); nothing red so far. Not clicking approve — this is your own PR, so GitHub only takes a comment from me.

sdbondi:
_Posted by Claude._

Nit taken, fixed in `870c34f2`.

You are right that `TemplateManager::initialize` inserts builtin rows at `TemplateStatus::Active` (manager.rs:72-90), so "reach `Active` only through `add_and_load_template`" was false as written, and false in exactly the way the rewrite was meant to prevent — a reader greps the status, finds line 80, and is left to work out why it is not a counterexample.

The clause now says where those rows go instead: `load_template_with_cache` answers a builtin from `build_in_template_cache` before it reaches the wasm cache (manager.rs:126), and this provider resolves builtin addresses without consulting the cache at all — with the reason attached, since a builtin's address is a constant rather than a hash of its binary, so an artifact filed under one would outlive the binary it came from.

Verified both before writing it, not from the review.

`cargo lints clippy --all-targets -p tari_engine --features wasm-cache` clean, 16 cache tests green. Docs-only change on top of `61890566`, so the full suite result from that head stands.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 0 non-blocking, 0 nits · 1 resolved since `61890566` (1 nit) · head `870c34f2`

`870c34f2` closes the last item. Verified all three claims it adds at cache.rs:811-816:

- `load_template_with_cache` returns a builtin from `build_in_template_cache` before reaching the cache — applications/tari_indexer/src/template_manager/manager.rs:126-128.
- This provider short-circuits builtin addresses in both lookups, so neither consults the cache: `get_template` at cache.rs:825 and `has_template`'s `!is_builtin_template_address(..)` guard at cache.rs:861.
- The reason given — constant addresses rather than content hashes, so an artifact filed under one outlives its binary — is the same one the bypass in `get_template` already states, which is where a reader following the claim lands.

Nothing outstanding. The change I reviewed across all four heads holds up: `store` hands off over a bounded queue and returns, the sync stays ahead of the rename where it has to, no caller reads back what it just stored, a write that never lands is now audible at `warn!`, and the test barrier orders every assertion that depends on one.

CI: everything green except the four `fuzz` legs, still running. This is your own PR, so my posting a review can only be a comment — with CI settled, it is an approve from my side.

sdbondi:
_Posted by Claude._

`WRITE_QUEUE_CAPACITY` is 16, not 8, as of `e9dec045`. Reopening the number briefly because the reasoning behind it was thinner than the surrounding notes.

Your retention argument for cutting 64 holds, but it is weaker than it read: on the validator path the artifact is co-owned, not pinned. `MemoryCacheTemplateProvider::get_template` inserts a clone into its moka cache immediately after `inner.get_template` returns, and the store was queued inside that call, so the queue shares ownership with a cache that has its own 1 GiB bound. The queue is the sole owner only if moka evicts the entry while the write is still pending.

What actually sizes this is the producer side. Every `store` follows a full Cranelift compile — 40-265 ms — while the write is a serialize plus one `fsync`, comparable or faster. One executor thread therefore cannot build a backlog at all; depth only accumulates from compiles running at once, which is bounded by cores in practice rather than by `CONCURRENT_ACCESS_LIMIT`. 16 sits above that, and because the policy on full is to drop rather than block, anything beyond it buys delay before a drop rather than avoiding one — at one recompile per drop.

The note over the constant now says that, so the number reads as derived rather than picked.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 0 non-blocking, 0 nits · 0 resolved since `870c34f2` (nothing of mine was outstanding) · head `e9dec045`

`e9dec045` is your own tuning of the queue depth, 8 → 16, not a response to a finding. Checked it anyway and it holds:

- The premise that a backlog can only form from compiles running at once is right — `store` is reached only on a disk-cache miss (cache.rs:848), which is always behind a full `load_template_from_code`, so the arrival rate is the concurrent-compile rate and nothing else.
- Sizing above the core count is the right shape for that. A node with more cores than 16 doing a cold-start burst across distinct templates can still fill it, which drops and now warns — the documented tradeoff, and the case the `Full` arm exists for.
- No tension with the PR's premise, which is that the flush is extra latency *inside* an execution rather than that it dwarfs the compile. Both readings of cache.rs:285-291 are consistent.

Nothing outstanding from any of the five heads I've looked at.

CI is re-running at this head; 9 green, the rest pending, nothing red.

sdbondi (commented):
Tested and approved

PR #2662, opened 2026-09-21
perf(engine): keep the artifact a published template's compile produces

sdbondi:
Stacked on #2661, which has to land first: this puts a cache write inside the publish execution
path, and until #2661 that write was an inline `fsync`.

Fifth PR in the WASM artifact cache series (#2609, #2614, #2619, #2660, #2661).

## The waste

`publish_template` runs a full Cranelift compile of the binary being published — that is what
`WasmModule::validate_code` is — and `charge_template_compile` bills the transaction for it, so
every involved validator performs and is paid for that compile. Then `compile_prevalidated` called
`loaded.into_template_def()` and threw the compiled module away, keeping only the ABI.

The same binary is then compiled a second time on that node: by the prewarm pool when the publish
commits, or by the first transaction to call the template. The network pays for one compile and
performs two.

## What this does

Keeps the artifact the publish already produced.

- `WasmModule::compile_prevalidated` returns the `LoadedTemplate` instead of its `TemplateDef`. `validate_code` takes the def from it, so its callers are unchanged.
- `RuntimeInterface::publish_template` returns the `PublishedTemplateAddress` it derived. The processor cannot compute that address — it is `H(seal signer, binary)` and only the runtime holds the signer.
- `TemplateProvider` gains `offer_compiled(&self, address, &Self::Template)` with a **default no-op body**, so no implementor changes. `MemoryCacheTemplateProvider` forwards it to its inner provider, and `DiskCachedWasmTemplateProvider` keeps it.

## An offered artifact is not evidence that its template exists

The thing that makes this more than a one-line change. Every artifact the disk cache held until now
came from a successful `inner.get_template`, so the template's substate existed by construction, and
both `get_template` and `has_template` took a cache hit as proof of existence without asking the
store.

An offered artifact breaks that. It is compiled during execution, which is before commit, and the
publishing transaction may abort — on a fee failure, a lock conflict, anything. Left unguarded, a
node holding such an artifact would serve `get_template` for a template that does not exist, while
every other node returns `TemplateNotFound` for the same call. `call_function` resolves templates
through the provider and nothing else, so the two nodes produce different execution results for the
same transaction. That is a consensus divergence, and reaching it costs an attacker one publish
crafted to abort plus one call.

So keeping offers is opt-in, and a provider that keeps them gives up the shortcut:

- `DiskCachedWasmTemplateProvider::new` ignores offers. `keeping_offers()` opts in, and only the validator node does.
- A provider keeping offers confirms existence with `inner` before serving a cached artifact, and `has_template` always asks `inner`. The cache stays an authority on bytes; the state store remains the authority on what exists.

The check is `inner.has_template`, which on a validator is a RocksDB point read, and it only runs on
a disk-cache hit — the first load of a template per process, since the in-memory cache absorbs the
rest.

The indexer stays on the default for two independent reasons: its `inner` answers existence with a
network fetch, so the check would be expensive there, and its executions are speculative dry runs
over unauthenticated input, where a live `offer_compiled` would let anyone fill the node's cache
with artifacts for templates that will never exist.

Builtins are never kept, matching the lookup-side bypass: a builtin's address is a constant rather
than a hash of its binary, so an artifact filed under one would outlive the binary it came from.

## Cost of an artifact for a transaction that never commits

Bounded disk, reclaimed by the LRU from #2619. It cannot be served — that is the guard above — and
its key is `H(author, binary)`, so it is exactly right if the same publish is retried.

## Verification

`cargo lints clippy --all-targets` clean on `tari_engine` (with `wasm-cache`), `tari_ootle_common_types`,
`tari_ootle_template_provider`, `tari_validator_node` and `tari_indexer`; `cargo nextest run -p tari_engine
--features wasm-cache` green. Three new cache tests: an offered artifact is kept but does not make its
template callable, a provider that ignores offers keeps nothing, and an offered builtin is never kept.


sdbondi:
_Posted by Claude._

**Merge order:** this branch is cut from #2661, so its diff carries that commit until #2661 lands. #2661 first, then this.

The reason the two cannot be reordered: this PR puts a `WasmModuleCache::store` directly inside the publish execution path, and until #2661 that store serialized the artifact and `fsync`'d it on the calling thread — which here is consensus executing `PublishTemplate`.

One thing worth a reviewer's attention, because it contradicts the note this work was planned against. The plan asserted that an artifact written for a publish that never commits "cannot be read before the template exists". That is not true of the code as it stands: `DiskCachedWasmTemplateProvider::get_template` returned a `try_load` hit without consulting `inner`, and `has_template` treated the cache file's presence as the answer. `call_function` resolves a template through the provider and nothing else, so a node holding the artifact of an aborted publish would execute a call that every other node aborts with `TemplateNotFound` — different execution results for the same transaction. An attacker reaches it with one publish crafted to abort after the `PublishTemplate` instruction, plus one call.

Hence `keeping_offers()`: opt-in, validator-only, and it buys the offer at the price of an existence check on `inner` before any cached artifact is served. On a validator that is a RocksDB key-existence read, and only on a disk-cache hit — the first load of a template per process, since the in-memory cache absorbs the rest. `an_offered_artifact_is_kept_but_does_not_make_its_template_exist` is the test that pins it.

The indexer stays on the default for two independent reasons, either of which would be enough: its `inner` answers existence with a network fetch, and its executions are speculative dry runs over input nobody has authenticated.

sdbondi:
Closing this because the complexity it adds is not worth the gain

sdbondi:
_Posted by Claude._

Closing. Recording why, since the idea will look attractive again.

**What changed under it.** This was designed when the publish compile was the only thing standing between a newly published template and an inline Cranelift compile during execution. #2660 closes that independently: Trigger B fires on the commit of the `PublishedTemplate` substate and compiles the binary on a prewarm worker, and a transaction that calls the template afterwards is withheld from consensus until its compile finishes. So what this PR removes is no longer a stall — it is a background recompile, roughly 100–140 ms of one worker thread, once per published template per involved validator, that nothing is waiting on.

**What it cost to get that.** A permanent weakening of the cache's strongest property: that every artifact in it corresponds to committed state. That property is what lets `get_template` and `has_template` answer from a cache hit without consulting `inner`, and it is what makes "eviction has no consensus reach" true. Trading it for CPU nobody is blocked on is the wrong side of the deal.

**Why offer-on-commit is not the fix.** It would need the artifact held from execution until commit — ~10× the binary, so up to ~10 MiB per pending publish, across branches that may never commit. That is a second bounded store with its own eviction policy plus a fork-discard path, to save a compile that is already off the critical path.

**Why the two modes could not be collapsed.** `keeping_offers` gated two things, and the second — confirming existence with `inner` before serving a cache hit — was load-bearing rather than defensive. Without it a node holding the artifact of an aborted publish executes a call every other node aborts. Keeping it unconditionally would charge the indexer a network round trip per disk-cache hit, because its `inner` resolves templates over the network. The mode existed because the guard's cost is chain-dependent; removing the offer removes both.

**What survives.** The invariant this work surfaced. `get_template` and `has_template` serve a cache hit without asking `inner`, and that is sound only because every artifact in the cache came from a successful `inner.get_template`. Nothing in the file said so, and this PR is the shape of change that breaks it silently. It is going into the lookup path in #2661.

#2661 is unaffected: the inline `fsync` it moves off the caller is on the ordinary store path — the first call to any template — not on publishes.

PR #2663, opened 2026-09-21
fix(engine): log under the prefix the engine appender routes

sdbondi:
Found while running the swarm against #2660: a run with dozens of stores, an eviction pass and a
startup directory scan produced **zero** lines from the WASM artifact cache.

## What happens

`log4rs_sample.yml:210` routes `tari::ootle::engine` to the `engine` appender at `debug`. Four files
in the engine crates log under a prefix that matches no logger:

| File | Target | |
|---|---|---|
| `crates/engine/src/wasm/cache.rs` | `tari::engine::wasm::cache` | |
| `crates/engine/src/runtime/working_state.rs` | `dan::engine::runtime::working_state` | pre-rename leftover |
| `crates/engine_types/src/crypto/covenant.rs` | `tari::engine::crypto::covenant` | |
| `crates/engine_types/src/stealth/transfer.rs` | `tari::engine_types::stealth::transfer` | |

None of those is a prefix of any configured logger, so they fall through to `root`, which is
`level: warn` with `stdout` alone (`log4rs_sample.yml:128-131`). Everything below `warn` was
discarded before it reached an appender, and the `warn`s went to stdout and not to the engine log.

Note it is specifically the `ootle` segment that is missing: `tari::ootle::wasm` in `process.rs` is
fine, because the `tari::ootle` logger at `:151` catches it even though nothing names it directly.

## Why the cache one matters

#2619's review asked for the startup tally — how many artifacts the directory holds, how many bytes,
against which cap — to be logged unconditionally, so an operator who upgraded without touching their
config could see what changed underneath them. That line has never been emitted. Same for every
eviction, every stored artifact, and the two `warn!`s added in #2661 for a writer that falls behind
or stops, which is the case an operator most needs to see: past it the cache silently does nothing
and every cold call recompiles.

## The fix

All four move to `tari::ootle::engine::…`. `engine_types` goes under the same prefix rather than a
new `tari::ootle::engine_types`, since a new prefix nothing configures would land back at `root`.

No behaviour change beyond where the lines go.

## Found, not fixed

The same class of mismatch exists outside the engine, and I have left it alone rather than widen
this: `tari::validator::migrations{,::v1,::v2}`, `tari::comms::logging::sqlite_message_log`,
`auth::tari::dan::wallet_sdk::storage_sqlite::writer`, `wallet::sdk::api`,
`ootle_rs::wallet::provider::input_resolver`, and `tari::identity`. Each wants checking against the
`log4rs_sample.yml` of the application that loads it, which is a different set of files from these.


sdbondi (commented):
ACK

PR #2666, opened 2026-09-21
fix(bor)!: bound decode nesting and reject integers the encoder cannot emit

sdbondi:
Description
---

Two decode-side defects in `tari_bor`, both of which abort a validator process. They share a root cause — the decode side enforced nothing by default — so they land together.

**1. Nesting depth was a per-type opt-in.**

`tari_bor::Value` and `RestrictedAccessRule` hand-write a bounded `Decode`. `tari_template_abi::Type` does not, and it is self-recursive through `Vec`/`Tuple`/`Option` with *both* of its decoders derived — a derive cannot thread a depth counter. A published template's `tari_tdef` custom section is decoded straight into `TemplateDef`, so a few hundred bytes of nested `Type::Vec` run the decoder off the end of the stack. That is a guard-page SIGSEGV: not a catchable `Err`, not a catchable unwind. It fires on the publish/validate path as well as at call time, and the poisoned template is consensus state, so restart-and-replay reproduces it. `serde_codec::from_slice` is a second live entrance to the same bomb through the serde derive.

Nesting is a property of the bytes, not of the target type, so the bound is applied to the input before a typed decode runs: an iterative head-walk that rejects anything nested deeper than a caller-supplied `max_depth`. A per-type patch would close `Type` and leave the next self-recursive type unbounded; this covers every type.

**The bound is the caller's, not the codec's.** `tari_bor` states the mechanism and nothing else — `check_nesting_depth(input, max_depth)`, `decode_with_max_depth`, `decode_exact_with_max_depth`, `serde_codec::from_slice_with_max_depth`, `RawCbor::decode_with_max_depth`. `decode`, `decode_exact`, `decode_with`, `decode_exact_with` and `from_slice` are unchanged, so the addition to that crate is purely additive and nothing a template compiles against moves. Only the caller knows whether its input is untrusted and what stack it is decoding on: a guest's stack overflow is a trap its embedder reports, a validator's is a guard-page abort of the process.

The figure is **`tari_engine_types::limits::MAX_CBOR_NESTING_DEPTH`** (256), beside the other limits that decide what a validator accepts. Every trust boundary applies that same one — a payload one node accepts and another rejects is a consensus split. The boundaries are where bytes this node did not produce first reach a native decoder:

| Boundary | Why it needs its own bound |
| --- | --- |
| `static_template_def` and `wasm/module.rs` — a published template's `tari_tdef` section | PANIC-01's route. Attacker-authored, decoded at publish *and* at call |
| `p2p::encoding::decode_from_slice` | Every gossiped or RPC-carried message |
| `Substate`/`SubstateValue::from_bytes` | A peer's substate bytes arrive as an opaque protobuf field, outside any outer CBOR walk |
| `EngineArgs::get_opt` | An argument is a blob inside the transaction; the transaction's own decode does not descend into it |
| `validator_node_rpc::client` — a peer's `ExecuteResult` | PATH-05's route, decoded before any content validation |
| `wasm/process.rs` — an engine op's payload | Read out of the guest's own WASM memory, decoded on the validator's stack |
| `p2p::conversions::rpc` — a peer's `CommandCommitProof` | An RPC-carried message that takes the serde route rather than `decode_from_slice` |
| `CommittedBlockProof::from_bytes` | A peer's commit proof, same serde route |
| `SubstateValueProof` | A peer's proof, decoded *before* `verify_inclusion` runs over it |

256 sits well above anything a legitimate payload nests to, because rejecting a valid payload is the worse failure. The semantic limits are the types' own and far lower — 50 for a materialised `IndexedValue`, 64 for the `tari_bor::Value` it comes from — but those count from the top of their own subtree, while this counts from the top of the whole payload.

**2. `Value`'s decoder admitted integers its own encoder refuses.**

minicbor classifies `3B FF..FF` as `Type::Int` and `From<Int> for i128` yields `-1 - u64::MAX`, so `Value::Integer(-18446744073709551616)` was stored verbatim; the encoder later does `i64::try_from` and returns `Err`. One engine argument reaches this — `args.assert_one_arg()` is inferred as `tari_bor::Value` and stored verbatim into `ComponentBody.state` — and the `Err` lands on the substate hasher, which takes borsh serialization to be infallible. Panic during finalize. The size gate does not filter it either: `encoded_len` is the infallible `minicbor::len_with` path.

`INTEGER_RANGE` is now stated once and every route into `Value::Integer` goes through `Value::integer`: the minicbor decoder, serde's `visit_i128` / `visit_u128`, the `@cbor:int` sentinel (which `deserialize_any` reaches from CBOR bytes, not only from JSON) and `parse_json_arbitrary_number`.

The decoder now accepts exactly the range the encoder can write back. That also closes the secondary `encode(self).unwrap()` consumers in `engine_types::substate` and `transaction_manifest::value`, which could not be fixed individually — the panic that fires first is an `.expect` inside `tari_hashing`.

Motivation and Context
---

Both are chain halts reachable from attacker-controlled input: an arbitrary published template for (1), a single transaction argument for (2). Found by a security review of the execution engine.

`MAX_CBOR_NESTING_DEPTH` lives in `tari_engine_types::limits` — the lowest crate all five boundaries already depend on, so nothing gains a dependency for it — and is deliberately set well above anything a legitimate payload nests to (256). It is a stack-safety bound, not a semantic one: the semantic limits stay with the types that own them (`MAX_DECODE_DEPTH` = 64 for the dynamic `Value` tree, `MAX_VISITOR_DEPTH` = 50 for the materialised one), and those count from the top of their own subtree while this one counts from the top of the whole payload, so reusing 64 here would have risked rejecting valid input.

How Has This Been Tested?
---

Every guard has a test that was confirmed to fail before the fix:

- `tari_engine_types::static_template_def` — **the original defect, end to end**: a real `tari_tdef` custom section in a real WASM binary, whose function argument's type nests 100,000 deep. Without the bound it aborts with `fatal runtime error: stack overflow` (SIGABRT under nextest); with it, a decode error.
- `tari_template_abi` — the same payload through both of `Type`'s derives, minicbor and serde. Before the fix both aborted the same way.
- `tari_template_abi` — the deepest payload the bound admits decodes on a thread with a **2 MiB stack**, the size of a tokio worker, for both routes. This is what makes 256 a bound rather than a number: the test asks `check_nesting_depth` for the deepest accepted payload rather than hard-coding a wire-level count, so it stays honest if the framing changes.
- `tari_bor::depth` — scalars and flat containers pass; each container kind passes at the bound and is rejected one level deeper; a 1 MB nesting bomb is rejected without recursing; trailing bytes past the first item are not walked; malformed input is left for the decoder.
- `tari_bor::value` — `3B FF..FF` is rejected at decode (this test failed before the fix), and `i64::MIN` / `u64::MAX` still round-trip.

```
cargo nextest run -p tari_bor -p tari_template_abi --all-features        # 91 passed
cargo nextest run -p tari_engine_types --features extract-template-def   # 98 passed
cargo nextest run -p tari_ootle_p2p -p tari_validator_node_rpc \
                  -p tari_ootle_storage                                  # 194 passed (with engine_types)
cargo nextest run -p tari_engine                                         # 460 passed
cargo nextest run -p tari_template_lib_types -p tari_ootle_transaction   # 166 passed
cargo nextest run -p tari_engine_types -p tari_transaction_manifest      # 121 passed
cargo lints clippy -p tari_bor -p tari_template_abi --all-targets --all-features
```

---

Revert either guard and re-run the tests above — the nesting tests abort the test process with a stack overflow rather than failing an assertion, which is the point.

Worth a close look:
- **The boundary list.** This is the part that decides whether the CRITICAL is actually closed: `decode`/`decode_exact` are unbounded, so a boundary I missed is a boundary with no bound. The five in the table are what I could find; a second pass over them is the most valuable thing a reviewer can do here.
- `walk()` in `crates/tari_bor/src/depth.rs` — the frame accounting for definite vs. indefinite containers, maps counting key and value at the same depth, and tags counting as a level (matching what `Value`'s own decoder does).
- Whether `MAX_CBOR_NESTING_DEPTH = 256` is comfortably above the deepest legitimate payload. The worker-stack tests in `tari_template_abi` establish the other side of it: the deepest payload the bound admits still decodes on a 2 MiB stack, the size of a tokio worker.

Breaking Changes
---

- [ ] None
- [x] Other - Please specify

**Coordinated upgrade required.** Both changes narrow the set of bytes that decode successfully, so nodes must upgrade together. Neither is gated by protocol version.

Template execution pricing does **not** move: nothing is added to the decoders a template compiles against, so its metered compute is unchanged. `tari_template_abi`, `tari_template_lib` and `tari_template_macros` are untouched and unbumped.

The asymmetry is worth noting: the depth bound only rejects payloads that would have aborted the process anyway, so nothing that previously *completed* a decode now fails. The integer rejection genuinely narrows the accepted set.

`tari_bor` 0.16.0 → 0.16.1. No dependent republish is required (`^0.y` pins pick it up), per `scripts/crate_versioning.py impact tari_bor`.


sdbondi (commented):
_Posted by Claude._

**Comment only** (would be: changes requested — own PR, cannot click) — 1 blocking, 2 non-blocking, 2 nits · head `dcaad15`

## Blocking

**1. The guard is compiled into templates, and it moves metered execution cost. `tari_engine::fees::underfunded_compute_rejects_as_a_fee_shortfall` fails at this head and passes at the merge base.**

`crates/tari_bor/src/lib.rs:152,160,173` puts `check_nesting_depth` in front of `decode` / `decode_with` / `decode_exact_with`. Those are the guest's decoders too, not just the host's:

- `crates/template_macros/src/template/dispatcher.rs:171` — one `decode_exact` per call argument, in the generated dispatcher.
- `crates/template_lib/src/engine.rs:68` — `RawCbor::decode` → `decode_exact` on the result of *every* engine call.

So every in-guest decode now pays a heap allocation (`crates/tari_bor/src/depth.rs:47`, `vec![Some(1)]` → a dlmalloc call in wasm) plus a full head walk, and all of it is metered. CI shows that crossing a boundary:

```
FAIL [21.791s] tari_engine::fees underfunded_compute_rejects_as_a_fee_shortfall
panicked at crates/engine/tests/fees.rs:512:27:
Transaction failed: Insufficient fees paid: At instruction #5: Insufficient fees to fund
native verification requiring 196984 points: 706524 of 700000 allowance points already consumed
```
(run 35591506637, `test (1/3)`.)

706524 against a 700000 allowance — the transaction now runs out where it previously did not, and the test's expected outcome flips from commit-with-failure to reject. Metering is deterministic, so this is not a flake, and CI on the merge base (`7400eb3c`, run 35340393981) is green.

The bound buys nothing in the guest to pay for this: guest-side recursion that runs off the wasm stack is a trap the engine already catches and charges for, not the guard-page SIGSEGV the PR is about. The abort risk is entirely host-side. Options, in the order I'd try them:

- Skip the walk in the guest (`#[cfg(not(target_arch = "wasm32"))]` — deterministic, every guest is wasm32), leaving the host entry points bounded.
- Failing that, at least drop the allocation: the stack can never exceed `MAX_NESTING_DEPTH + 1`, so a fixed-size array works and costs no allocator call.

Either way the fee figures need to be re-baselined deliberately, and the PR should say that template execution cost moved — it is consensus-visible pricing for every template compiled against 0.16.1, and right now the description does not mention it.

## Non-blocking

**2. `Value`'s serde `Deserialize` still admits the integers the encoder cannot write back** — `crates/tari_bor/src/value_serde.rs:156` (`visit_i128`), `:247` (the `@cbor:int` sentinel's `s.parse::<i128>()`), `:284` (`parse_json_arbitrary_number`). `serde_codec`'s `deserialize_any` routes CBOR `Type::Int` and the bignum tags straight to `deserialize_i128`/`u128` (`crates/tari_bor/src/serde_codec.rs:577-583`), so `serde_codec::from_slice::<Value>(&[0x3b, 0xff × 8])` still yields a `Value` that `encode` rejects — the exact value `value.rs:281` now refuses.

I traced the consensus paths and none of them reach `Value` through serde: engine args go through `decode_exact` (`crates/engine/src/runtime/engine_args.rs:34`), substates through `SubstateValue::from_bytes` → `decode_exact`, p2p through `decode_from_slice` → `decode_exact` (`crates/p2p/src/encoding.rs:15`). So I don't think it is reachable from the network today. But the PR's claim is that the decoder now accepts exactly the encodable range, and that holds on one of the two routes; mirroring the check in `visit_i128` and the sentinel parse is a couple of lines and makes it true of the type.

**3. The RocksDB read path builds a `Decoder` directly and bypasses the guard** — `crates/state_store_rocksdb/src/codecs/minicbor.rs:49`. Nothing self-recursive reaches it unguarded today (a published template's `TemplateDef` is re-decoded from the wasm custom section through `tari_bor::decode`, `crates/engine/src/wasm/module.rs:347`), so this is not a live hole. Raising it because the PR's own threat model is "the poisoned template is consensus state, so restart-and-replay reproduces it" — replay reads from here, and this is the one read path the bound does not cover.

## Nits

**4. The upper half of the new range check is unreachable.** `crates/tari_bor/src/value.rs:286` tests `i > u64::MAX`, but minicbor only reports `Type::Int` for `0x3b` with the high argument bit set (`minicbor-2.3.0/src/decode/decoder.rs:785`); positives are always `Type::U*` and handled on the line above. Fine as a belt-and-braces bound, but the comment reads as though both halves can fire.

**5. An empty container at the bound is rejected one level before a scalar is.** `depth.rs:82` checks after the frame is pushed, so `0x81 × 256` + `0x80` fails while `0x81 × 256` + `0x00` passes. Inconsequential at 256 — noting it only so it reads as a choice.

## Verified sound

- Entry-point coverage: `decode`, `decode_with`, `decode_exact` (via `decode_exact_with`), `serde_codec::from_slice`, and transitively `RawCbor::decode`, `from_value`/`to_value`, p2p ingest, substate ingest, and the `tari_tdef` section decode. The only direct `minicbor::decode` calls outside `tari_bor` are two test fixtures.
- `walk()` frame accounting: traced definite and indefinite arrays and maps, tags, empty containers, an indefinite container nested in a definite one, break handling, and trailing bytes. The decrement-before-read ordering is right, `saturating_mul(2)` for map frames is right, the stack is capped at 257 entries, and `Decoder::skip` is iterative in minicbor 2.3 (`decoder.rs:550`), so "immune to the recursion it rejects" holds — the walk only calls it on scalars.
- Treating a walk error as in-bound is safe: the walk and the typed decode traverse the same structure in the same order, so bytes malformed for one are malformed for the other.
- The accepted integer range matches the encoder exactly (`value.rs` `Encode` uses `u64::try_from` / `i64::try_from`).
- `tari_template_abi`'s `serde` feature enables `tari_bor/serde`, so the `serde_route` test module compiles under `--all-features`.
- Versioning: `scripts/crate_versioning.py impact tari_bor` prints exactly the patch bump the description claims, with no dependent republish.
- No unrelated files in the diff; the `Cargo.lock` hunk is only the version line.

crates/tari_bor/src/lib.rs
```diff
@@ -147,12 +149,14 @@ pub fn from_value<T: for<'b> Decode<'b, ()>>(val: &Value) -> Result<T, BorError>
 
 /// Decode a single value from a byte slice (unit context). Extra trailing bytes are ignored.
 pub fn decode<T: for<'b> Decode<'b, ()>>(input: &[u8]) -> Result<T, BorError> {
+    check_nesting_depth(input)?;
```
sdbondi: This function is used for all decodes, including from trusted/pre-verified sources (e.g rocksdb codec) and, worse, when decoding across the engine abi boundary, so this will needlessly increase WASM fees. We need a way to only perform this check when needed on the engine side.

sdbondi: Max depth should be configurable per caller (the max shouldnt be in the tari_bor crate itself)

sdbondi:
_Posted by Claude._

All five addressed in `e0fdc4c`. Every claim checked before acting on it.

**1 (blocking) — confirmed, and the diagnosis was right.** `tari_bor` is linked into every template: `dispatcher.rs:171` decodes each call argument and `template_lib/src/engine.rs:68` decodes every engine-call result, both through `decode_exact`. `underfunded_compute_rejects_as_a_fee_shortfall` fails at `dcaad15` and passes at the merge base.

Taken the first option: `check_nesting_depth` is a no-op on `wasm32`. Running off the stack inside wasm is a trap the embedder reports as an error — the abort the bound exists for is native-only — so the guest paid for something it could not use. The cfg is on `target_arch`, which is the axis that decides the failure mode rather than a build-time choice, and every guest is wasm32, so it stays deterministic across validators.

That makes the re-baselining moot: template execution costs what it did before, and nothing in the PR's pricing is consensus-visible any more. `cargo nextest run -p tari_engine` is 460/460 at this head, `fees` included.

Did not take the second option. The allocation was worth removing only in the guest; a fixed `[Option<u64>; 258]` is 4 KiB written on entry to every native decode, which is worse than one 16-slot `Vec` for the shallow payloads that are almost all of them. The `Vec` now starts at capacity 16 so it does not re-grow at any realistic depth.

**2 — confirmed and fixed, and it was reachable from CBOR, not only from JSON.** `ValueVisitor::visit_map` parses the `@cbor:int` sentinel out of a *map*, and `serde_codec` will hand it one, so `{"@cbor": "int", "value": "<out of range>"}` in CBOR bytes reaches `parse_i128_sentinel` — a second route past `visit_i128`. Every route now goes through `Value::integer`, which is the single statement of `INTEGER_RANGE`; that covers `visit_i128`, `visit_u128`, the sentinel and `parse_json_arbitrary_number`. Regression test on the CBOR route in `value_serde`.

One knock-on: `integer_outside_i64_uses_sentinel` asserted a JSON round trip of a value outside the range, which no longer decodes. Split into `integers_outside_i64_but_inside_the_cbor_range_round_trip` (`u64::MAX`, `i64::MAX + 1` — the values that actually exercise the sentinel-adjacent paths) and `an_integer_outside_the_cbor_range_is_refused_on_the_way_back`.

**3 — confirmed and fixed.** `codecs/minicbor.rs:49` now checks the bound before decoding. Agreed it is not a live hole today; replay is exactly the path the PR's own threat model turns on, so it should not be the one read that skips the check.

**4 — the nit dissolved rather than being answered.** The upper half is unreachable *from `Type::Int`*, but `visit_u128` can carry a value above `u64::MAX`, so with the check moved into `Value::integer` both halves fire. The comment is now on `INTEGER_RANGE` and states the range rather than the arm.

**5 — fixed, and it was not a deliberate choice.** The check moved from where a container is opened to where an item is read, so depth is the frames open around the item being read. `0x81 × 256` followed by `0x80`, `0xa0` or `0x00` now all pass, and `0x81 × 257` still fails. Test added.

sdbondi (commented):
_Posted by Claude._

**Comment only** (would be: changes requested — own PR, cannot click) — 1 blocking, 0 non-blocking, 1 nit · 4 resolved since `dcaad15` (1 blocking, 1 non-blocking, 2 nits) · head `26a3511`

## Blocking

**1. `crates/state_store_rocksdb/src/codecs/minicbor.rs:51` still calls the one-argument `check_nesting_depth`. This does not compile.**

`e0fdc4c` added the guard there; `26a3511` changed the signature to `check_nesting_depth(input: &[u8], max_depth: usize)` (`crates/tari_bor/src/depth.rs:31`) and updated every other call site — `lib.rs:166,174,196,205`, `serde_codec.rs:128`, `template_def.rs:229` — but not this one:

```rust
tari_bor::check_nesting_depth(bytes).map_err(|e| RocksDbStorageError::DecodeError {
```

`codecs/mod.rs:7` declares the module unconditionally, so this is an unconditional build failure in a workspace member. It wants `MAX_NESTING_DEPTH` as the second argument — stored bytes are replayed into the types the network wrote them as, which is the bound they came in under.

## Nit

**2. `INTEGER_RANGE` is not reachable from outside the crate.** It is `pub` in the private `value` module and `lib.rs:36` re-exports only `{MAX_DECODE_DEPTH, Value}`, while `Value::Integer`'s doc comment links to it. `Value::integer()` is reachable, so nothing is blocked by this — it just means the named range a caller is pointed at cannot be read.

## Resolved since `dcaad15`

- **In-guest metering (blocking).** The opt-out is now an explicit `max_depth` argument rather than a target-arch no-op, which I prefer: `decode_with_max_depth` / `decode_exact_with_max_depth` / `from_slice_with_max_depth` / `RawCbor::decode_with_max_depth`, with `None` meaning no bound. All four opt-out sites are guest paths and none of them is reachable on a native stack: `dispatcher.rs:137,155,171,183` is generated template code; `template_abi/src/abi/mod.rs:79` sits behind `tari_engine`, which is `unimplemented!()` on non-wasm (`abi/non_wasm.rs:25`); `template_lib/src/args/result.rs:36` decodes what `call_engine` returned, and the host side only ever calls `InvokeResult::encode`. Host entry points keep the bound by default. The workspace-pin bumps are needed and correct — the generated dispatcher now needs `tari_template_lib` 0.32.1's re-export, and template authors get the macro through `tari_template_lib`, which pins `tari_template_macros` 0.23.1.
- **`Value`'s serde route (non-blocking).** `Value::integer()` is now the single constructor and all four routes go through it — `visit_i128`, `visit_u128`, the `@cbor:int` sentinel, `parse_json_arbitrary_number` — with a test that the CBOR serde route refuses `3B FF..FF` and one that an out-of-range value serialized to the sentinel does not come back.
- **Dead upper bound (nit).** Replaced by `INTEGER_RANGE.contains`.
- **Empty container off-by-one (nit).** The check moved to before each item is read (`depth.rs:71`), so an empty container at the bound is accepted exactly as a scalar is, with a test over `0x00`/`0x80`/`0xa0`. I re-traced the loop at the new placement: an item at depth 256 is accepted, its frame pushes to 258 at most and the next item is rejected, nothing skips the check, and every iteration still pops, consumes or returns.

Not resolved: the RocksDB read-path bypass I raised as non-blocking is addressed in intent, but the call site does not build — see above.

CI is pending on this head; the fee test that failed at `dcaad15` has not re-run yet.

sdbondi:
_Posted by Claude._

Follow-up on two points from @sdbondi, in `26a3511`. Both were right and both changed the shape of the fix.

**The `wasm32` no-op was the wrong mechanism.** It put the decision in a `cfg` no call site could see, and said the bound was a property of the target rather than of what the caller is decoding and where its stack overflow lands. The bound is now an argument:

- `check_nesting_depth(input, max_depth)`
- `decode_with_max_depth`, `decode_exact_with_max_depth`, `serde_codec::from_slice_with_max_depth`, `RawCbor::decode_with_max_depth` — all taking `Option<usize>`, where `None` is no bound at all
- `decode`, `decode_exact`, `decode_with`, `decode_exact_with` and `from_slice` keep `MAX_NESTING_DEPTH`, so a call site added later is bounded unless it says otherwise

The three guest decodes now ask for `None` where they stand: `call_engine`'s reply, `InvokeResult::decode`, and each argument the generated dispatcher reads. `tari_template_abi`, `tari_template_lib` and `tari_template_macros` are bumped so a template built against the published crates gets those rather than paying the walk per argument. `tari_engine::fees` is green, so template pricing is still unmoved.

**The RocksDB check is reverted.** Bytes reach that store having already been decoded through a bounded decoder, so checking them again on the way out prices every database read for a threat the store does not have — the opposite direction to the one the finding was reaching for. Finding 3 stands as a note about where replay reads from, not as something to guard.

The other three (the serde integer range, the unreachable upper bound, the empty-container off-by-one) are unchanged from `e0fdc4c`.

sdbondi:
_Posted by Claude._

Agreed on all three, done in `dc77c38`. The layering was wrong and the `Option` was a symptom of it.

**The bound is now the caller's.** `tari_bor` states the mechanism and nothing else: `check_nesting_depth(input, max_depth)`, `decode_with_max_depth`, `decode_exact_with_max_depth`, `serde_codec::from_slice_with_max_depth`, `RawCbor::decode_with_max_depth` — each taking the bound outright, no `Option`. `decode`, `decode_exact`, `decode_with`, `decode_exact_with` and `from_slice` are byte-for-byte what they were before this branch, so the change to that crate is purely additive.

**The figure is `tari_engine_types::limits::MAX_CBOR_NESTING_DEPTH`**, beside the limits that decide what a validator accepts, and documented against the two semantic depth limits it is not (`MAX_VISITOR_DEPTH` 50, `Value`'s `MAX_DECODE_DEPTH` 64 — both count from the top of their own subtree, this one from the top of the payload).

**This also deletes the part of the PR I liked least.** With nothing added to the decoders a template compiles against, there is no guest cost to compile out, no `target_arch` cfg, and no reason to touch `tari_template_abi`, `tari_template_lib` or `tari_template_macros` — they are back to `development` and their version bumps are gone. `tari_bor` 0.16.1 is the whole publish surface.

**The trade I want on the record.** `decode`/`decode_exact` now being unbounded means the fix is exactly as good as the boundary list, and a boundary added later is unbounded unless its author says otherwise — the failure mode this PR is about, moved from per-type to per-call-site. What makes that acceptable rather than a wash is that the boundaries are few, named, and visible at the call site instead of hidden in a `#[derive]`. Here is the list:

| Boundary | Why it needs its own bound |
| --- | --- |
| `static_template_def`, `wasm/module.rs` — the `tari_tdef` section | PANIC-01's route; attacker-authored, decoded at publish *and* at call |
| `p2p::encoding::decode_from_slice` | Every gossiped or RPC-carried message |
| `Substate`/`SubstateValue::from_bytes` | Peer substate bytes arrive as an opaque protobuf field, outside any outer CBOR walk |
| `EngineArgs::get_opt` | An argument is a blob the transaction's own decode does not descend into |
| `validator_node_rpc::client` — a peer's `ExecuteResult` | PATH-05's route, decoded before content validation |

Two things I checked rather than assumed while drawing that up: `serde_bridge` and `RawCbor` do *not* need their own bounds, because both skip a real CBOR item in the parent stream, so their contents are inside the parent's walk. Engine arguments and substate bytes do, because both are opaque byte strings the parent walk steps over.

**And the test is now the real one.** `static_template_def` gets a genuine WASM binary with a `tari_tdef` custom section whose function argument's type nests 100,000 deep — the actual publish-path payload, not a shaped approximation. `SIGABRT` without the bound, decode error with it. The `tari_template_abi` tests stay as the proof that both of `Type`'s derives are covered and that the deepest accepted payload still decodes on a 2 MiB stack.

sdbondi (commented):
_Posted by Claude._

**Comment only** (would be: approved — own PR, cannot click; CI still pending) — 0 blocking, 3 non-blocking, 2 nits · 1 resolved since `26a3511` (1 blocking) · head `dc77c38`

The inversion is the right shape. `tari_bor` now states the mechanism and nothing else, the figure lives beside the other things that decide what a validator accepts, and the guest pays nothing: `git diff development..HEAD -- crates/template_lib crates/template_macros crates/template_abi crates/template_lib_types` is a test module and nothing else, so generated dispatcher code and every in-guest decode are byte-identical to `development`. That closes the metering regression at the source rather than by exemption, and `tari_bor` 0.16.1 stays a correct patch bump because the crate's change is purely additive.

The cost of opt-in is that a boundary nobody enumerated is silently unbounded, so the findings below are all about the boundary list.

## Non-blocking

**1. Three peer-bytes call sites sit outside the table, all of them `serde_codec::from_slice`.** The table's second row reads "Every gossiped or RPC-carried message" for `p2p::encoding::decode_from_slice`, but these three are RPC-carried messages that take the serde route instead and get no bound:

- `crates/p2p/src/conversions/rpc.rs:181` — a peer's `CommandCommitProof`, out of an `EpochCheckpoint` response.
- `crates/storage/src/consensus_models/committed_block_proof.rs:31` — a peer's commit proof.
- `crates/storage/src/state_store/substate_value_proof.rs:148` — a peer's `SubstateValueProof`, decoded *before* `verify_inclusion` runs over it.

None of those types is self-recursive, so I don't think any is exploitable today — the point is that under the new model that is a property of the types, checked by hand, not something the boundary keeps true. `from_slice_with_max_depth` is already there for them.

**2. `crates/engine/src/wasm/process.rs:474` decodes guest-supplied bytes on the host with no bound.** This is the outer engine-op payload read straight out of WASM memory, the sibling of `EngineArgs::get_opt` which the table does cover — the same attacker (a published template), the same native stack, one level up. No current op arg type is self-recursive, so again this is safe by inspection of the types rather than by the boundary.

**3. The description claims a guard the code no longer has.** "The RocksDB codec, which builds a `Decoder` directly, checks the nesting bound too — replay reads a poisoned template's bytes back through it." `02d7849` removed that check and it has not come back; `crates/state_store_rocksdb/src/codecs/minicbor.rs` is identical to `development`. Dropping it is fine by me — replay reads bytes that passed a bounded ingest — but the sentence should go, or the bound should.

## Nits

**4.** The description still calls the figure `MAX_NESTING_DEPTH` in the "Motivation and Context" section; it is `MAX_CBOR_NESTING_DEPTH` now, and in `engine_types::limits` rather than `tari_bor`.

**5.** `INTEGER_RANGE` is `pub` in the private `value` module and `lib.rs:36` re-exports only `{MAX_DECODE_DEPTH, Value}`, while `Value::Integer`'s doc points at it. `Value::integer()` is reachable, so nothing is blocked.

## Resolved since `26a3511`

- **The build break (blocking).** `check_nesting_depth`'s one-argument call in the RocksDB codec is gone, by removing the call rather than fixing it. Re-checked: `MAX_NESTING_DEPTH` has no remaining referents outside `template_def.rs`'s own local const, `decode_with_max_depth` / `decode_exact_with_max_depth` / `from_slice_with_max_depth` / `RawCbor::decode_with_max_depth` all take a plain `usize` and delegate to the unbounded entry point after the check, and `tari_p2p` and `tari_validator_node_rpc` both already depend on `tari_engine_types` for the new import.

## Still verified sound at this head

The walk itself is untouched since `26a3511` — iterative, stack capped at 258 frames, frame accounting traced across definite and indefinite arrays and maps, tags, empty containers and trailing bytes. `Value`'s integer range is enforced on both routes through the single `Value::integer` constructor. The new `static_template_def` test builds its payload from a real encode and asserts the decode errors rather than aborting, which is the PANIC-01 route end to end.

CI is pending on this head.

crates/engine_types/src/limits.rs
```diff
@@ -358,6 +358,24 @@ pub const ENGINE_LIMITS: EngineLimits = EngineLimits {
...
+/// hundred bytes reaches the end of the stack — a guard-page abort of the validator process, which
+/// no caller can catch. Every boundary applies the same figure, because a payload one node accepts
+/// and another rejects is a consensus split.
+///
+/// It sits generously above anything a legitimate payload nests to, because rejecting a valid
+/// payload is the worse failure. The semantic depth limits are the types' own and are far lower —
+/// 50 for a materialised `IndexedValue` tree, 64 for the dynamic `tari_bor::Value` it comes from —
+/// but those count from the top of their own subtree, while this counts from the top of the whole
+/// payload. What this figure has to guarantee is that the deepest accepted input still decodes
+/// within the smallest stack untrusted decode runs on: a 2 MiB tokio worker.
```
sdbondi, line 373:
```suggestion
/// It sits generously above anything a legitimate payload nests to, because rejecting a valid
/// payload is the worse failure. This figure guarantees that the deepest accepted input still decodes
/// within the smallest stack untrusted decode runs on: a 2 MiB tokio worker.
```

sdbondi:
_Posted by Claude._

All five correct, all five fixed in `0532e66`. Findings 1 and 2 are the useful ones: they are exactly the failure mode this design trades for, found by doing the thing the design asks for.

**1 and 2 — four boundaries, not three.** Checked each, all peer- or guest-supplied and all now bounded:

| Site | Source |
| --- | --- |
| `p2p/src/conversions/rpc.rs` — `CommandCommitProof` | A peer's `EpochCheckpoint` RPC response |
| `storage/.../committed_block_proof.rs:31` | A peer's commit proof |
| `storage/.../substate_value_proof.rs:148` | A peer's proof, decoded before `verify_inclusion` |
| `engine/src/wasm/process.rs:474` | An engine op's payload, read from guest memory onto the validator's stack |

You are right that none of the four types is self-recursive today, and right that this is the point: that was a property of the types, held by inspection, and it is now held by the boundary. The engine-op one is the one I should have caught unaided — `EngineArgs::get_opt` was in the table and this is its outer envelope, same attacker, same stack, one frame up.

**3 — the description claimed a guard the code does not have.** Sentence removed. The RocksDB check is gone for good: bytes reach that store having passed a bounded ingest, so re-checking on the way out prices every database read for a threat the store does not have.

**4 and 5 — both fixed.** The description says `MAX_CBOR_NESTING_DEPTH` and names where it lives; `INTEGER_RANGE` is re-exported from `lib.rs`, so the range `Value::Integer`'s doc points at can now be read.

**On placement, since it came up separately:** no crate gained a dependency for this. `git diff development -- '*/Cargo.toml'` on the whole branch is the `tari_bor` version bump and nothing else — `p2p`, `storage`, `validator_node_rpc` and `engine` all already depended on `tari_engine_types`, which sits below `common_types` in the graph and is the lowest crate the boundaries share. One figure rather than a per-crate constant is deliberate: a payload one node accepts and another rejects is a consensus split, so the boundaries should not be free to drift apart.

`cargo nextest run -p tari_engine` is 460/460 at this head, and 194 across bor/engine_types/p2p/storage/validator_node_rpc.

crates/engine_types/src/static_template_def.rs
```diff
@@ -123,6 +126,61 @@ mod tests {
         })
     }
 
+    // The `tari_tdef` section is attacker-controlled — anyone may publish a template — and `Type`
+    // is self-recursive through `Vec`/`Tuple`/`Option` with both of its decoders derived, so
+    // neither can thread a nesting bound of its own. Without the bound this decoder applies, the
+    // payload below recurses once per level and the stack overflow aborts the process instead of
+    // returning an error.
```
sdbondi:
```suggestion
    // The `tari_tdef` section is attacker-controlled and `Type` is self-recursive. This test asserts that an maliciously deep template definition encoding does not decode.
```

crates/tari_bor/src/depth.rs
```diff
@@ -0,0 +1,172 @@
...
+/// Nesting is a property of the bytes, not of the target type: a `#[derive(Decode)]` on a
+/// self-recursive type descends one native stack frame per level and has no way to thread a counter
+/// of its own, so a bound reached this way is the only one that covers every type. What `max_depth`
+/// should be is the caller's, because only the caller knows whether its input is untrusted and what
+/// stack it is decoding on — a guest's overflow is a trap its embedder reports, a validator's is a
+/// guard-page abort of the process. `tari_engine_types::limits::MAX_CBOR_NESTING_DEPTH` is the
+/// figure this project's own trust boundaries use.
+///
+/// The walk is iterative and reads only item heads, so it costs a fraction of the decode it guards
+/// and is itself immune to the recursion it rejects.
```
sdbondi, line 14:
```suggestion
/// Rejects input nested deeper than `max_depth`.
///
/// The walk is iterative and reads only item heads, so it costs a fraction of the decode it guards
/// and is itself immune to the recursion it rejects.
```

crates/engine_types/src/limits.rs
sdbondi: _Posted by Claude._

Applied in `05b8835`.

crates/engine_types/src/static_template_def.rs
sdbondi: _Posted by Claude._

Applied in `05b8835`.

crates/tari_bor/src/depth.rs
sdbondi: _Posted by Claude._

Applied in `05b8835`.

sdbondi: _Posted by Claude._

Correcting my reply above: this one landed in `a819290`, not `05b8835` — I answered all three threads before I had read this one, which was wrong of me.

Applied, and it is the right call for a reason beyond length: `check_nesting_depth` was naming `tari_engine_types::limits`, a crate that depends on this one. The codec documents the mechanism; where the figure comes from belongs to whoever supplies it. I also took the same sentence out of `decode`, `decode_exact` and `from_slice`, which pointed here for an explanation that no longer exists.

sdbondi (commented):
Finally happy with this. Tested and approved

PR #2667, opened 2026-09-21
fix(engine)!: remove the raw-pointer Runtime interface

sdbondi:
Description
---

`Runtime` held a `NonNull<Box<dyn RuntimeInterface>>` and handed out `&mut dyn RuntimeInterface` from it. A nested call rebuilt a second `Runtime` from that same pointer while the outer frame's `&mut` was still live:

```
invoke_resource_access_hook(&mut self)            // outer &mut, live across the call below
  └─ invoke_component_method(&mut self)
       └─ get_call_runtime(&self)                 // Runtime::from_pointer(self.runtime_pointer)
            └─ call_method(&mut call_runtime)
                 └─ call_runtime.interface_mut()  // second &mut to the same box
```

Two live `&mut` to one object is undefined behaviour under both Stacked and Tree Borrows, and it was expressible entirely from safe code: `from_mut` and `from_pointer` were safe `pub fn`s, and the `unsafe` block was hidden inside the accessors. `from_mut` additionally took the address of a *local* `Box` handle in `execute`, pinned by nothing stronger than declaration order.

The realised consequence, if the optimiser ever acted on the aliasing, is the auth-hook frame losing its `FrameWriteMode::OwnComponent` restriction and running in `Full` — which is what stops it acting on the vaults reachable through the `AuthHookCaller` argument it is handed.

**The fix is structural, not a cast.** A nested call runs against the same execution state as the frame that made it, so the sharing belongs in the state rather than in a `&mut` every frame has to hold alive across the re-entry:

- `StateTracker` owns its mutability. Its methods take `&self` and borrow internally, and every borrow closes before the statement that took it ends — which is what keeps a nested call from finding one open.
- `RuntimeInterfaceImpl` holds shared handles to the tracker, the entity id provider and the spend-script context, and `for_nested_call` hands out another interface over the same state instead of aliasing this one.
- `RuntimeInterface` is `&self` throughout. `Runtime` is an `Rc` handle with an `interface()` accessor and no `interface_mut`.
- `NonNull`, `from_mut`, `from_pointer`, `as_pointer`, `set_runtime_pointer` and the `runtime_pointer` field are all gone. `RuntimeModule` takes `&StateTracker`.

The only `unsafe` left in `crates/engine/src/runtime/` is the pair of `Send`/`Sync` impls wasmer requires of a `FunctionEnv`'s data. `Runtime` satisfies neither structurally, so those impls need an invariant, and it has to be one a reader can check against the code. `FunctionEnv<T>` is `#[repr(transparent)]` over a `StoreHandle`, so cloning one clones a handle — but the value itself lives in the `Store`'s `StoreObjects`, and the `Store`'s drop drops it. What holds is narrower: the `Store` is created and dropped inside `TransactionProcessor::invoke_template`, on the thread executing the transaction, so every refcount touch — wasmer's own drop included — happens there. The SAFETY comments say that.

Everything a `Runtime` reaches is `Rc`, and the confinement is what the crate boundary keeps rather than what a comment asks for: `Runtime::new` and `RuntimeInterfaceImpl::initialize` are `pub(crate)`, so safe code outside this crate cannot build a handle and move it — `Runtime` is `Send`, which is the only reason that would be possible.

An earlier revision made `Runtime::interface` an `Arc`, on the grounds that a stray cross-thread clone under `Rc` is a use-after-free where the raw pointer this PR removes made it a stale read. It came back out: the atomic was only ever partial, since the tracker, entity-id provider, spend-script context and blobs are `Rc` whatever the outer handle is, so it bought a cheaper failure on one field while the others kept the sharp one. Uniform `Rc` puts the thread-confinement in the types, and `pub(crate)` is what actually enforces it.

The handles `TransactionProcessor` is given from outside — the template provider, the modules collection and the claim proof verifier — stay `Arc`. Those are built at bootstrap and held by services that are `Clone` and spawned (the indexer's as axum state), so they genuinely cross threads; the engine's own state does not.

**The frame write restriction is now a parameter.** It used to be a one-shot `restricted_frame_pending` flag set on the interface, left set across the nested call, and consumed by whichever `push_call_frame` happened next. Restrictions are per-frame, so `call_function` / `call_method` take the restriction and pass it to the frame it restricts.

Motivation and Context
---

Found by a security review of the execution engine, as the one UNSOUND-with-a-realised-consequence finding. The review rated it HIGH on consequence rather than CRITICAL because no miscompilation was demonstrated.

How Has This Been Tested?
---

```
cargo nextest run -p tari_engine                 # 460 passed
cargo lints clippy -p tari_engine -p tari_template_test_tooling --all-targets
```

The whole engine suite is the verification here: it is what exercises nested calls (`cross_template`, `reentrancy`, `caller_badge`, `access_rules`, `spend_script`, `recall`, `freeze`) and therefore what would surface a borrow left open across a re-entry as a panic. 460/460 pass, unchanged from the merge base.

For the restriction parameter I checked the test fails without it: passing `None` where `invoke_resource_access_hook` passes `Some(FrameWriteMode::OwnComponent)` turns `caller_badge::auth_hook_cannot_spend_the_depositors_caller_badge` from a pass into

```
Expected reject reason "Write to resource_0094c9... attempted in a resource auth hook,
which may only modify its own component state"
but got "... Resource Auth Hook Denied Access for action native.resource.Deposit: Access Denied: native.resource.Burn"
```

**What no test here establishes:** the aliasing itself. It is a static reading against Stacked/Tree Borrows, and Miri cannot run wasmer, so there is no execution that distinguishes the old code from the new. What the suite establishes is that the restructuring preserves behaviour; the soundness argument is the absence of the second `&mut`, which is now a type-level property.

---

- `grep -rn "interface_mut\|from_mut\|from_pointer\|as_pointer\|set_runtime_pointer" crates/` returns nothing.
- `grep -rn unsafe crates/engine/src/runtime/` returns only the two `Send`/`Sync` impls.
- The change that most deserves scrutiny is `StateTracker`'s move to interior mutability: a borrow held open across a nested call would be a `BorrowMutError` panic rather than a compile error. Every access goes through `read_with` / `write_with` / `with_chargeable_state`, each of which closes its borrow within the statement, and no nested call is made from inside one. The engine suite covers the re-entrant paths.
- `for_nested_call` builds a new interface over the same state — worth checking that nothing which must be per-frame ended up shared, or vice versa.

Breaking Changes
---

- [ ] None
- [x] Other - Please specify

**Rust API only**, and the engine's own: `RuntimeInterface` is `&self`, `push_call_frame` takes a restriction, `Runtime::{from_mut, from_pointer, as_pointer, interface_mut}` are gone, `RuntimeModule` takes `&StateTracker`, and `StateTracker::chargeable_state` is now `with_chargeable_state`. No wire format, no state layout and no consensus behaviour changes, and template execution is unaffected.


sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 2 non-blocking, 1 nit · head `ec6f5d2ab`

I would approve this. The restructuring is the right shape: there is no longer any route from safe code to a second `&mut` on the interface, and the property is type-level rather than a convention. Details below, with what I checked.

## What I verified

- **The raw-pointer route is gone repo-wide.** `interface_mut`, `from_mut`, `from_pointer`, `as_pointer`, `set_runtime_pointer` and `WasmEnv::state_mut` return nothing at this head; the only `unsafe` left under `crates/engine/src/runtime/` is the `Send`/`Sync` pair at `crates/engine/src/runtime/mod.rs:327-333`.
- **Re-entrancy vs. the new `RefCell`s** — this was the change most worth attacking, since a borrow left open across a nested call is now a runtime panic rather than a compile error. All 13 `borrow()`/`borrow_mut()` sites close within their statement, and no `read_with`/`write_with`/`with_chargeable_state` closure re-enters the tracker: only two closures mention `self` at all (`crates/engine/src/runtime/impl.rs:3610` and `:3937`), and both only read the immutable `seal_signer_public_key`. The interesting case, `stealth_transfer`, runs `verify_input_authorizations` — which can execute a WASM predicate — between its two `write_with` blocks, not inside either (`crates/engine/src/runtime/impl.rs:4097`).
- **`with_chargeable_state` holds the right borrow for the right span.** The `RefMut` temporary at `crates/engine/src/runtime/tracker.rs:587-594` lives to the end of the tail expression, so it covers `f` and releases on return — which is what the `&mut WorkingState` inside needs. Neither `FeeModule` nor `TrackCallsModule` reaches back into the tracker from inside it.
- **`for_nested_call` shares exactly what must be shared.** Tracker, entity-id provider, spend-script context and blobs are `Rc`; modules are `Arc<[Box<dyn RuntimeModule>]>`; only `seal_signer_public_key` is copied, and it is immutable for the transaction. Nothing per-frame is left on the interface once `restricted_frame_pending` moved out, so the clone cannot drift from the original. No `Rc` cycle: nothing inside `RuntimeInterfaceImpl` holds a `Runtime`.
- **The restriction parameter is equivalent to the old flag.** The only two setters were `invoke_resource_access_hook` and `evaluate_spend_script`, and both now pass it explicitly; I checked no `push_call_frame` occurs between the old set and the old take on either path (`load_component`/`track_template_loaded`/`resolve_args`/`lock_component`/`next_entity_id` push no frame). Every other `push_call_frame` site passes `None`, matching a flag that was unset there.
- **The breaking-change scope in the description is accurate.** `git grep RuntimeInterface` outside `crates/engine/src` at this head returns only a doc-comment mention in `crates/engine_types/src/limits.rs:415`.
- **CI**: the one red leg is Test Results (CI), from `tests/features/stall_recovery.feature:7` — the known RocksDB `LOCK` flake, consensus-side and untouched by this PR. 2482/2483 otherwise.

## Non-blocking

1. **`crates/engine/src/runtime/mod.rs:327-330` — the SAFETY comment's load-bearing claim is not quite true.** "it never clones or drops the value itself" holds for clone (`FunctionEnv<T>` is `#[repr(transparent)]` over a `StoreHandle`, and its `Clone` clones the handle — wasmer-7.4.2 `src/backend/sys/entities/function/env.rs:16,111`), but not for drop: the value lives in the `Store`'s `StoreObjects` as a `VMFunctionEnvironment`, so dropping the `Store` drops it, and that drop glue is wasmer's. What actually makes the reference counts safe is narrower and stronger: the `Store` is created and dropped inside `invoke_template` (`crates/engine/src/transaction/processor.rs:978-980`), on the thread executing the transaction, so every refcount touch — including the drop — happens there. Since this comment is the whole justification for two false auto-trait impls, it is worth stating the invariant that is actually checkable.

2. **`crates/engine/src/runtime/mod.rs:323` — `Rc` under a manual `Sync` is a new hazard class, not just the old one relocated.** Before, `Runtime` was a bare `NonNull` and `Clone` was a pointer copy, so a `Runtime` that escaped to another thread could not corrupt anything by itself. Now `Runtime: Clone + Sync` means two threads cloning from one `&Runtime` race a non-atomic refcount, which is a use-after-free rather than a stale read. The single-thread invariant holds today and I could not find a path that breaks it (the cache writer thread at `crates/engine/src/wasm/cache.rs:586` moves serialized bytes, never a `Store`). Making just this field `Arc<dyn RuntimeInterface>` costs one atomic per nested call and per `invoke_template`, and removes the cheapest route from "invariant broken" to "memory corruption". The inner `Rc`s can stay: they are only cloned in `for_nested_call`, which is unreachable from another thread without the invariant already being violated.

## Nit

3. **`crates/engine/src/runtime/mod.rs:316-321`** — "every frame of a transaction holds one over the same interface" is not what `for_nested_call` does: it builds a *new* `RuntimeInterfaceImpl` (`Rc::new(self.clone())`), so nested frames share the *state*, not the interface. Direct `runtime.clone()` in `invoke_template` does share the interface, so both happen. Given that the point of the PR is precision about what is shared, the doc could say "shares this transaction's state" rather than "the same interface".

sdbondi:
_Posted by Claude._

All three taken, in `cd512d6`.

**1 — the SAFETY comment's claim about drop.** You are right and the comment was wrong. `FunctionEnv<T>` is `#[repr(transparent)]` over a `StoreHandle<VMFunctionEnvironment>`, so cloning one clones a handle and drops nothing — but the value itself sits in the `Store`'s `StoreObjects`, and the `Store`'s drop drops it. "wasmer never drops the value" was the load-bearing half of a justification for two false auto-trait impls, and it was false.

Replaced with the invariant that can actually be checked against the code: the `Store` is created and dropped inside `TransactionProcessor::invoke_template`, on the thread executing the transaction, so every refcount touch — wasmer's own drop of the `FunctionEnv` data included — happens on that thread. The comment now says that, and says which structural properties it is standing in for (`RefCell` state, `Rc` handles).

**2 — `Rc` under a manual `Sync`.** Agreed, and taken. This was the part of the change that made something genuinely worse: the old `NonNull` made a stray cross-thread clone a stale read, and `Rc` makes it a use-after-free. Defending a cheap failure with an expensive one is the wrong direction for a PR whose point is removing a soundness hazard.

`Runtime::interface` is now `Arc<dyn RuntimeInterface>`. Three lines — the field, `Runtime::new`, and the two construction sites (`processor.rs:206` and `for_nested_call`). The inner handles stay `Rc` for the reason you give: they are only cloned in `for_nested_call`, which is not reachable from another thread unless the invariant is already gone. The cost is one atomic per nested call and per `invoke_template`, against a WASM instantiation.

**3 — the doc's "same interface".** Fixed, and worth more than a wording change, since `for_nested_call`'s whole design is that frames *don't* share an interface. The doc now says frames share that transaction's state, and names both routes: `Runtime::clone` shares the interface, `for_nested_call` builds a new `RuntimeInterfaceImpl` over the same tracker, entity-id provider and spend-script context.

```
cargo nextest run -p tari_engine
    Summary [379.268s] 460 tests run: 460 passed (14 slow), 0 skipped
cargo lints clippy -p tari_engine --all-targets   # clean
```

On the CI red: agreed, `stall_recovery.feature:7` is the known flake. Checked it rather than assumed — `development` at this PR's merge base `b946ac2` fails the same single scenario on its own run, 1 of 2 483.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 0 non-blocking, 1 nit · 3 resolved since `ec6f5d2ab` (2 non-blocking, 1 nit) · head `cd512d6fc`

All three items from the previous pass are genuinely fixed. Re-verified rather than taken from the commit message:

- **SAFETY comment** (`crates/engine/src/runtime/mod.rs:331-339`) — now names the invariant that is actually checkable ("the `Store` holding it is created and dropped inside `TransactionProcessor::invoke_template`, on the thread executing the transaction") and explicitly includes wasmer's own drop of the `FunctionEnv` data in what happens there. That matches `crates/engine/src/transaction/processor.rs:978-980` and wasmer-7.4.2's ownership model, where the value lives in the `Store`'s `StoreObjects`.
- **`Rc` → `Arc` on the outer handle** (`crates/engine/src/runtime/mod.rs:329`, `:345`, `crates/engine/src/runtime/impl.rs:526`, `crates/engine/src/transaction/processor.rs:206,214`) — all four construction/coercion sites moved together, `mod.rs` no longer references `Rc` outside comments, and the inner `Rc`s are correctly left alone (they are only cloned in `for_nested_call`). The comment now states the reason for the asymmetry, which is the part a future reader would otherwise undo.
- **The sharing doc** (`crates/engine/src/runtime/mod.rs:316-326`) — now separates the two routes correctly: `Runtime::clone` shares the interface, `for_nested_call` builds a new `RuntimeInterfaceImpl` over the same state. The three fields it names (tracker, entity-id provider, spend-script context) are exactly the mutable ones.

Nothing else changed: the diff against `ec6f5d2ab` is confined to these three files and touches no control flow, so the earlier verification of the re-entrancy, restriction-threading and `for_nested_call` sharing still stands unaltered.

## Nit

1. `crates/engine/src/runtime/mod.rs:322` — `RuntimeInterface::for_nested_call` names the wrong type; `for_nested_call` is an inherent method on `RuntimeInterfaceImpl`, not on the `RuntimeInterface` trait. It is in plain backticks so rustdoc does not fail on it, but it is a path a reader would grep for and not find.

CI on this head is still running; the only red leg on the previous head was the known `stall_recovery.feature:7` RocksDB `LOCK` flake.

sdbondi:
_Posted by Claude._

Fixed in `fd1509e`. Confirmed before changing it: `for_nested_call` is defined in the inherent `impl RuntimeInterfaceImpl<TStore, TTemplateProvider>` block at `crates/engine/src/runtime/impl.rs:222`, not in the `RuntimeInterface` trait — the path I wrote pointed at a method that does not exist on that type. The doc now says `RuntimeInterfaceImpl::for_nested_call`.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 0 non-blocking, 0 nits · 1 resolved since `cd512d6fc` (1 nit) · head `fd1509eec`

`fd1509eec` corrects the doc path to `RuntimeInterfaceImpl::for_nested_call`, which is where the method is defined. That was the last item; nothing stands against this head.

I would approve this, and am posting as a comment only because GitHub does not accept an approving review on your own PR — it needs another maintainer's click.

CI on `fd1509eec`: everything green except Test Results (CI), whose single failure is `tests/features/stall_recovery.feature:7` — the known RocksDB `LOCK` flake, consensus-side and untouched by this PR. Notably `clippy` passes, which is the leg the `Rc` → `Arc` change was most likely to disturb.

What this review verified across the three heads, for the record:

- No route from safe code to a second `&mut` on the interface remains; `interface_mut` / `from_mut` / `from_pointer` / `as_pointer` / `set_runtime_pointer` / `WasmEnv::state_mut` are absent repo-wide, and the only `unsafe` under `crates/engine/src/runtime/` is the documented `Send`/`Sync` pair.
- The interior-mutability conversion is re-entrancy-safe: every `borrow`/`borrow_mut` and every `read_with` / `write_with` / `with_chargeable_state` closure releases within its statement, and no closure re-enters the tracker. `stealth_transfer` runs `verify_input_authorizations` — the path that can execute a WASM predicate — outside both of its `write_with` blocks.
- `for_nested_call` shares exactly the mutable state (tracker, entity-id provider, spend-script context, blobs, `Arc<[Box<dyn RuntimeModule>]>`) and copies only the immutable `seal_signer_public_key`; nothing per-frame survives on the interface, and there is no `Rc` cycle.
- The frame-restriction parameter is behaviourally equivalent to the old one-shot flag on both setter paths, and every other `push_call_frame` site passes `None` where the flag would have been unset.
- The breaking change really is engine-internal: no user of `RuntimeInterface` exists outside `crates/engine/src`.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 1 non-blocking, 0 nits · 1 resolved since `fd1509eec` (0 blocking) · head `1c7609b46`

Rebased onto development (picking up #2666 and #2669) plus one new commit, `1c7609b46`, which reverts the handle to `Rc` and rewrites the SAFETY comment to justify it. Taking the `Arc` suggestion back is your call and the reasoning it leaves behind is a fair one — `Arc` on the outer handle was only partial anyway, since the tracker, entity-id provider, spend-script context and blobs are `Rc` regardless, and keeping all of them `Rc` at least makes the thread-confinement visible in the types rather than papering over it with atomics. I have no objection.

Re-verified on this head:

- The PR's file list is unchanged from the first pass — the same 9 files, so the rebase swept nothing in. The net delta to the PR's own code since `ec6f5d2ab` is the `mod.rs` doc and SAFETY comment; the `Rc` → `Arc` → `Rc` round trip lands exactly where it started, on code I have already checked. (The `decode_exact_with_max_depth` lines now visible in `engine_args.rs` and `process.rs` come from #2666 in the base, not from here.)
- Every claim in the new SAFETY comment holds. `tracker`, `entity_id_provider` and `blobs` are all built inside `execute` (`crates/engine/src/transaction/processor.rs:150,182,204`); the handles that come from outside it — `template_provider`, `modules`, `claim_burn_proof_verifier` — are all `Arc`; and `Runtime` has exactly the two construction sites the doc describes (`crates/engine/src/transaction/processor.rs:216` and `crates/engine/src/runtime/impl.rs:526`).

## Non-blocking

1. **`crates/engine/src/runtime/mod.rs:344` — `Runtime::new` is `pub`, so the comment's "its whole lifetime is one execution" is only guaranteed in-crate.** `runtime` is a `pub mod`, `RuntimeInterfaceImpl` is re-exported with a `pub fn initialize` (`crates/engine/src/runtime/mod.rs:27`, `crates/engine/src/runtime/impl.rs:225`), and `Runtime` is `Send`, so an outside crate can build one in safe code and move it to another thread — the same "expressible entirely from safe code" shape the PR's description sets out to close, relocated to the crate boundary. Nobody does today (no user of any of this exists outside `crates/engine/src`), and it was equally true of `from_pointer`/`from_mut` before, so this is not a regression. But `pub(crate) fn new` is one keyword, both call sites are in-crate, and it turns the invariant the `unsafe impl`s rest on into something the crate boundary enforces rather than something the comment asks for. `RuntimeInterfaceImpl::initialize` is worth the same look while you are there.

CI on this head is still running; `cargo_deny`, `fmt`, `docs build` and the rest of the fast legs are green.

sdbondi:
_Posted by Claude._

Taken, in `7557207`. Both constructors, as you suggested.

`Runtime::new` and `RuntimeInterfaceImpl::initialize` are `pub(crate)`. Verified the call sites before narrowing them: `Runtime::new` at `crates/engine/src/transaction/processor.rs:216` and `crates/engine/src/runtime/impl.rs:526`, `initialize` at `processor.rs:206`, and nothing else in the workspace. Confirmed by compiling rather than by grep — `tari_template_test_tooling`, `tari_ootle_app_utilities`, `tari_indexer` and `tari_validator_node` are all clean, and those are the crates that would have noticed.

Your framing is the right one: this was the same "expressible entirely from safe code" shape the branch exists to remove, sitting at the crate boundary instead of inside the module. A comment asking callers to keep an invariant that two `unsafe impl`s depend on is worth one keyword to stop asking.

On the `Arc` coming back out — for the record, since it was your suggestion I took and then reverted. The argument against it in the end was the one you make here: the tracker, entity-id provider, spend-script context and blobs are `Rc` whatever the outer handle is, so the atomic hardened one field and left the sharp failure on the rest. It bought a cheaper symptom rather than a smaller surface. Uniform `Rc` plus `pub(crate)` gets the actual property: the types say the thing is thread-confined, and the crate boundary makes it so.

The PR description's paragraph arguing for the atomic is rewritten to match, and says the `Arc` was tried and why it came out, so the next reader does not re-derive it.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 0 non-blocking, 0 nits · 1 resolved since `1c7609b46` (1 non-blocking) · head `7557207d8`

`7557207d8` closes the last item and closes it wider than I asked: `Runtime::new` *and* `RuntimeInterfaceImpl::initialize` are both `pub(crate)` now, so the only two ways to obtain a `Runtime` are inside the crate. Re-verified: both `Runtime::new` call sites (`crates/engine/src/transaction/processor.rs:216`, `crates/engine/src/runtime/impl.rs:526`) and the one `initialize` call site (`crates/engine/src/transaction/processor.rs:206`) are in `crates/engine`, and nothing outside `crates/engine/src` names `RuntimeInterfaceImpl` at all — the single hit is a doc-comment mention in `crates/engine_types/src/limits.rs:430`. The doc on `new` states the reason rather than just the modifier, which is the part that stops it being widened again by accident.

That leaves the confinement the `unsafe impl`s rest on enforced by the crate boundary instead of asked for in a comment — which is the same move the PR makes for the aliasing itself, so it is the right note to end on.

Nothing stands against this head. I would approve; comment only because GitHub does not accept an approving review on your own PR, so it needs another maintainer's click.

Full record of what this review verified, across all five heads:

- No route from safe code to a second `&mut` on the interface remains. `interface_mut` / `from_mut` / `from_pointer` / `as_pointer` / `set_runtime_pointer` / `WasmEnv::state_mut` are absent repo-wide, and the only `unsafe` under `crates/engine/src/runtime/` is the documented `Send`/`Sync` pair.
- The interior-mutability conversion is re-entrancy-safe. Every `borrow`/`borrow_mut` and every `read_with` / `write_with` / `with_chargeable_state` closure releases within its statement, and no closure re-enters the tracker. `stealth_transfer` runs `verify_input_authorizations` — the path that can execute a WASM predicate — outside both of its `write_with` blocks.
- `for_nested_call` shares exactly the mutable state (tracker, entity-id provider, spend-script context, blobs, `Arc<[Box<dyn RuntimeModule>]>`) and copies only the immutable `seal_signer_public_key`. Nothing per-frame survives on the interface, and there is no `Rc` cycle.
- The frame-restriction parameter is behaviourally equivalent to the old one-shot flag on both setter paths, and every other `push_call_frame` site passes `None` where the flag would have been unset.
- The breaking change is engine-internal: no user of any of this exists outside `crates/engine/src`, and after this commit no outside user could construct one.

sdbondi:
Tested and approved

PR #2668, opened 2026-09-21
fix(engine)!: authenticate the whole cache file and derive the engine fingerprint

sdbondi:
Description
---

Two problems with the on-disk wasm module cache, both in the same file.

**1. The integrity tag covered the header and stopped there.**

`try_load` checked a CRC32 over the 40 header bytes, then handed everything past `HEADER_BYTES` to `Module::deserialize_unchecked`. I read wasmer 7.4 to find out what that actually validates, since the file's own comments were the only source for the claim and the review could not confirm it:

- `ArtifactBuild::is_deserializable` checks the body starts with `b"wasmer-universal"`. If it does **not**, the bytes are *not rejected* — `Artifact::deserialize_unchecked` falls through to `deserialize_elf` and then `deserialize_object`, i.e. wasmer's native-object loaders (`wasmer-compiler-7.4.0/src/engine/artifact.rs:395-414`).
- If it does, `MetadataHeader::parse` checks 8-byte alignment, the magic `WASMER\0\0` and ABI version 24 (`wasmer-types-7.4.0/src/serialize.rs:43-69`).
- Everything after those 32 bytes goes to `SerializableModule::archive_from_slice` → `rkyv::access_unchecked` (`wasmer-compiler-7.4.0/src/serialize.rs:125`). No validation.

So the comments were right that a magic and version check exists, and wrong to treat it as cover. Every byte at or past `HEADER_BYTES` was a byte nothing checked — and no test in the file ever mutated one (the "corrupt artifact" test writes 29 bytes and returns at the short-header branch, 19 bytes before the CRC).

The tag now covers the artifact body, and the file's identity with it — the template address and the engine fingerprint. Bit rot, truncation and a stale or restored artifact end as a miss and a recompile instead of as undefined behaviour, and a file only answers for the name it was found under.

**A checksum is not an authenticator, and this PR does not pretend otherwise.** A process that can write the cache directory can write a matching tag. What addresses *that* is the directory's mode, which `open` now takes away from group and other: whoever can write here chooses the native code this process executes. The tag is for damage; the mode is for the writer. The mode is hardening, so it reports rather than fails — a directory owned by another uid cannot be chmod'd, and a node that recompiles every template is a better outcome than one that will not start.

**2. `ENGINE_FINGERPRINT` was `"v5"` plus a seven-item bump list that nothing enforced — and the list did not name the metering cost tables.**

The per-operator cost table and the bulk-metering constants are baked into the artifact at compile time, and they price a consensus fee. A cost-table edit without a fingerprint bump leaves warm-cache nodes metering at old prices while cold nodes compile fresh: a consensus divergence reached by a release rather than by an attacker.

It is now derived, in a `const` block, over:

- **the sources that decide what an artifact contains**: `engine_config.rs`, `module_shape.rs`, `metering.rs`, `bulk_metering.rs`, `limiting_tunable.rs`;
- **the header layout**, because the order of the fields decides which count a hit reads each slot into. Swap two at an unchanged field count and every constant the digest could otherwise name is unmoved, while a warm node reads a count into the wrong `ModuleShape` slot and prices `instantiation_points` off it;
- **every field of `WASM_LIMITS`**, destructured so a limit added later does not compile until it is named. Not only the two the artifact bakes in: `validate_module_structure` enforces `max_tables` and `max_globals` at compile time, and a hit goes straight to `finalize_loaded_module` without it, so tightening either would leave a warm node executing a template a cold node refuses.

**A digest over a file is only as precise as the file is narrow**, which is why the first two are files at all. `create_engine` and `validate_module_structure` used to sit in `module.rs` alongside the loader, and the header layout in `cache.rs` alongside eviction, logging and 493 lines of tests — so a comment or a log line orphaned every artifact on every node for a change no artifact can observe. Hashed text is 44 KiB over five narrow files rather than 133 KiB over five broad ones, and neither `module.rs` nor `cache.rs` is in the preimage.

The header layout is hashed as values rather than text, since it is configuration: `HEADER_FIELDS` names each slot, `store` iterates it to emit and `try_load` iterates it to parse. The order hashed is therefore the order the file is written and read in, rather than a restatement that could drift from it.

Measured on this branch: adding a comment to `cache.rs` leaves the fingerprint at `54d3c81c9a4e9776`; flipping `canonicalize_nans` in `engine_config.rs` gives `3020173ad23a6803`, and swapping two `HEADER_FIELDS` entries gives `e5ee96bd19404e82`.

The hash is FNV-1a with an avalanche finalizer rather than Blake2b. Nothing outside this build chooses the input — every part is this project's own source and constants — so the property wanted is separation rather than collision resistance, and the Blake2b it replaces was truncated to the same 64 bits for the filename anyway. Being a `const` also means no runtime work and no `LazyLock`, and `include_bytes!` keeps rustc's own change tracking, which a build script would not.

The wasmer version is the one input still outside it, because a crate cannot read its dependencies' resolved versions and a published crate cannot reach the lockfile. Two things stand in: wasmer's `MetadataHeader` carries an ABI version it refuses to load across, so a format change ends as a miss; and `tari-wasmer-middlewares` pins `wasmer`, `wasmer-types` and `wasmer-vm` at `=7.4.0`, so the resolved version moves only when a manifest does. A wasmer bump that keeps the format and changes codegen is what this leaves to the upgrade itself — and that is now the only item left to remember, down from seven.

Motivation and Context
---

From a security review of the execution engine. The cache findings were rated MEDIUM: a template author controls the compiler's input, never its output, so this is not template-reachable. The realised attackers are a local process that can write the cache directory, and non-adversarial bit rot.

How Has This Been Tested?
---

```
cargo nextest run -p tari_engine --features wasm-cache
cargo lints clippy -p tari_engine --features wasm-cache --all-targets
```

New tests, both confirmed to fail before the fix:

- `a_flipped_artifact_byte_falls_back_to_recompile` — flips a bit at `HEADER_BYTES`, in the middle of the artifact, and at its last byte. With the tag header-only, the middle-of-artifact case reaches `deserialize_unchecked`: `a damaged artifact at 2459252 is a miss` fails. This is the region the review flagged as having no coverage at all.
- `an_artifact_under_the_wrong_address_falls_back_to_recompile` — a whole artifact copied to another template's filename under this build's fingerprint. Without the address in the tag it loads.

`the_stored_tag_covers_the_fields_and_the_body` (was `stored_header_crc_covers_the_fields`) now asserts the stored tag against `integrity_tag` over the identity, the fields and the body.

---

- The wasmer claim above is worth re-reading at the cited lines rather than taken on my word — the ELF fall-through in particular, since it is the part the file's comments did not mention.
- `integrity_tag`'s inputs: fingerprint, address, header fields, body. Anything a hit serves verbatim should be inside it.
- The fingerprint's source-text inputs, as a set. The question to ask of each file is whether a change to it can alter what a cached artifact means without altering a constant the digest already names.
- `restrict_dir_to_owner` is unix-only and a no-op elsewhere. It warns and continues on every failure path, so the worst case is a node that logs and recompiles — worth confirming that is the right trade for a hardening measure.

Breaking Changes
---

- [ ] None
- [x] Other - Please specify

**Existing cache files are orphaned**, both by the fingerprint changing and by the tag's new input. No operator action is needed: `build_index` reaps artifacts under a superseded fingerprint at startup, and each template is recompiled once on its next cold load. No wire format, state layout or consensus behaviour changes.


sdbondi (commented):
_Posted by Claude._

**Comment only** (GitHub refuses a changes-request on your own PR; I would request changes) — 2 blocking, 3 non-blocking, 2 nits · head `8832a5ab`

The wasmer claim in the description checks out at the cited lines: `Artifact::deserialize_unchecked` falls through `is_deserializable` to `deserialize_elf` and `deserialize_object`, and `SerializableModule::archive_from_slice` hands the rest to `rkyv::access_unchecked`. Body bytes past the 32-byte prefix were genuinely covered by nothing, and `integrity_tag`'s input set — fingerprint, address, header fields, body — is every byte a hit serves. Two things stand in the way.

## Blocking

**1. This file's own header layout is not an input to the derived fingerprint** — `cache.rs:103-111`

The doc this PR deletes listed "the order or meaning of the header's fields at an unchanged `HEADER_BYTES`" as a mandatory bump trigger. The derivation that replaces it hashes `module.rs`, `metering.rs`, `bulk_metering.rs` and `limiting_tunable.rs`, but not `cache.rs` — and the header's field order lives only in `cache.rs`.

Swap two fields in `write_now`'s array (`cache.rs:770-776`) and in `try_load`'s `read_field` calls (`cache.rs:621-626`) — say `data_segment_count` and `element_segment_entries`. `HEADER_FIELD_COUNT` stays 5, `HEADER_BYTES` stays 48, so the fingerprint does not move and the filename does not move. The tag is computed over bytes, so it matches. A warm node then reads the old file's counts into the wrong `ModuleShape` slots and prices `instantiation_points` off them, charging a different fee for the same transaction than a node that compiled fresh.

That is the exact divergence the derivation exists to make unforgettable, and it is the one item on the old list that this PR leaves less protected than it found it: before, a human was told to bump; now nobody is told and nothing derives it.

`include_str!("cache.rs")` inside `cache.rs` is legal and closes it, on the same "coarse in the safe direction" rationale the doc already argues at `cache.rs:76-79` — and it subsumes the `HEADER_FIELD_COUNT` / `HEADER_BYTES` entries.

**2. `restrict_dir_to_owner`'s failure refuses to start, which is the opposite of what the description says it does** — `cache.rs:401`

`open` propagates with `?`, and both callers propagate out of bootstrap: `applications/tari_indexer/src/bootstrap.rs:337` and `applications/tari_validator_node/src/bootstrap.rs:343-348`. `fs::set_permissions` needs ownership of the directory. A `wasm_cache` owned by another uid — created by an earlier root-run container against the same mounted data dir and then run as non-root with group write, or any mount without working chmod — returns `EPERM`, and the node does not boot.

The description says: "It repairs the mode rather than refusing to start, on the grounds that a node whose umask made the directory 0755 should not fail to boot over it." The code refuses. Since the mode is a hardening measure and the tag/miss path does not depend on it, a failed repair should be a `warn!` and a continue, not a bootstrap error.

Two more consequences of the same function worth deciding on deliberately rather than inheriting:

- On an ordinary umask-002 host the directory is `0775`, so this fires on every node, drops group write, and logs a `warn!` each boot for a completely standard configuration.
- If a deployment shares the data dir between two service accounts through a group, the second one silently loses its writes after the first one starts: `store` fails, and every load recompiles.

## Non-blocking

**3. The `WASM_LIMITS` fields that gate compile-time-only checks are still outside the fingerprint** — `cache.rs:103-111`

The description says the wasmer version is "the only item left to remember, down from seven". Not yet. `validate_module_structure` enforces `max_tables` (`module.rs:471`) and `max_globals` (`module.rs:483`), and it is not re-run on a hit: `load_template_from_serialized` goes straight to `finalize_loaded_module` (`module.rs:124-133`), which runs `validate_functions`, `validate_module_exports` and `Instance::new` but never the structure walk. Those two constants live in `tari_engine_types::limits`, which no `include_str!` reaches.

Meanwhile the two limits that *are* hashed as values (`cache.rs:105-106`) are the two that need it least — `LimitingTunables` re-applies `max_memory_pages` and `max_table_elements` at `Instance::new` on every hit, so a change to them is enforced against a cached artifact anyway.

Tighten `max_tables` from 4 to 2 and a warm node keeps executing a 3-table template that a cold node refuses. Hashing every field of `WASM_LIMITS` rather than two closes the gap and makes the "one item left" claim true.

**4. The `unsafe` block still justifies itself with the claim this PR disproves** — `cache.rs:636-644`

> The cache directory is node-local and not attacker-controlled in any sane operational setup. ... a deserialize failure simply triggers the recompile fallback.

The second sentence is what the commit message calls false — a body that fails wasmer's prefix check is handed to an ELF loader, it does not "simply fail". The corrected account went into the docs on `TAG_OFFSET` and `integrity_tag`; the block that performs the unsafe operation still carries the old one, and that is the comment someone auditing the `unsafe` reads. The first sentence is also now contradicted by `restrict_dir_to_owner`, which exists precisely because the directory *can* be attacker-controlled.

**5. `write_durable`'s rationale is stale** — `cache.rs:841-845`

"The artifact body lies past the header CRC's coverage, so it must reach the device before the rename names it." The body is inside the tag's coverage now, so the stated reason no longer holds — a torn body ends as a tag mismatch and a recompile. The `sync_all` is still worth keeping; it needs a different reason or none.

## Nits

**6.** `try_load`'s doc (`cache.rs:544-548`) still describes a hit as "a single `mmap` syscall (and the page faults wasmer's deserializer triggers as it walks the artifact)". The tag now faults every page of the file unconditionally, ahead of wasmer. Cheap — roughly 0.2 ms for the 2.4 MB artifact the new test reports — but no longer what the comment describes.

**7.** The four `include_str!` bodies are concatenated with no separator or length prefix (`cache.rs:90-102`), so text moved between two of them digests identically. Not an adversarial input, so it costs nothing to leave; one `hasher.update(&(part.len() as u64).to_le_bytes())` in the loop closes it.

## Verified sound

- Orphan reaping handles the old `v5` files: `parse_orphan_name` strips `.bin`, splits on the last `_`, and `v5` never equals the new hex, so `build_index` removes them (`cache.rs:463-472`). The "no operator action needed" claim in Breaking Changes holds.
- `a_flipped_artifact_byte_falls_back_to_recompile` hits the three positions that matter (first body byte, middle, last), and `an_artifact_under_the_wrong_address_falls_back_to_recompile` genuinely exercises the address binding — the file is written outside the index and `try_load` discards it on the tag, not on a deserialize error.
- `HEADER_BYTES` is unchanged at 48, so the 16-byte alignment assert and the rkyv-alignment reasoning above it still hold.
- Every `ENGINE_FINGERPRINT` call site was updated to the deref, and the only consumer outside this file is the `wasm::mod` re-export.
- No unrelated files in the diff.

sdbondi:
_Posted by Claude._

All seven addressed in `0167592`. Both blocking findings were real and the first was the more serious of the two.

## Blocking

**1 — the header layout was not in the fingerprint.** Confirmed exactly as described: `HEADER_FIELD_COUNT` and `HEADER_BYTES` were hashed, the field *order* was not, and it lives only in `write_now`'s array and `try_load`'s `read_field` calls. Swap two and nothing the digest names moves, the filename does not move, and the tag matches because it is over bytes — while a warm node reads a count into the wrong `ModuleShape` slot and prices `instantiation_points` off it. You are right that this was the one item on the deleted list that came out less protected than it went in: a human used to be told to bump, and then nobody was.

`include_str!("cache.rs")` is now an input, which subsumes the `HEADER_FIELD_COUNT` and `HEADER_BYTES` entries — both removed. Same "coarse in the safe direction" rationale the file already argues.

**2 — `restrict_dir_to_owner` could stop a node booting.** Confirmed: `?` in `open`, propagating out of `bootstrap.rs` in both the indexer and the validator node. `set_permissions` needs ownership, so a `wasm_cache` owned by another uid returns `EPERM` and the node does not start — over a hardening measure the tag and miss paths do not depend on.

It now returns `()` and warns on every failure path, including a failed `metadata`. The function doc says why: a node that recompiles every template is a better outcome than one that will not start, and a group-shared data dir losing the second account's writes is the point rather than a side effect.

One correction on the third consequence, since it changes how noisy this is: **the `warn!` fires once, not each boot.** `create_dir_all` does not touch an existing directory's mode, so on an umask-002 host the repair takes the directory from `0775` to `0755` on the first boot after upgrade and every boot after that takes the `mode & 0o022 == 0` early return. The other two consequences stand as you describe them.

## Non-blocking

**3 — the compile-time-only limits.** Confirmed against the code: `finalize_loaded_module` runs `validate_functions`, `validate_module_exports` and `Instance::new`, and never `validate_module_structure` — so `max_tables` and `max_globals` were enforced on a cold path only and were outside the fingerprint, while `max_memory_pages` and `max_table_elements` are re-applied by `LimitingTunables` at every `Instance::new` anyway. The two that were hashed were the two that needed it least.

Every field of `WASM_LIMITS` is hashed now, and destructured rather than read one by one, so a limit added later does not compile until it is named here. That makes "the wasmer version is the only item left to remember" true rather than aspirational, which is what the description claimed.

**4 — the `unsafe` block's justification.** Replaced. It rested on "a deserialize failure simply triggers the recompile fallback", which is the claim this branch exists to disprove, and on a directory that is not attacker-controlled, which `restrict_dir_to_owner` contradicts by existing. It now rests on what is actually true at that point: the tag over fingerprint, address, fields and body has been checked, and only a writer holding the directory can produce a matching one.

**5 — `write_durable`.** The `sync_all` stays, with the reason it actually has now: a torn body is a tag mismatch and a recompile, so the flush buys latency rather than safety — without it a crash leaves a file that is found, mmap'd and hashed in full before being discarded.

## Nits

**6.** `try_load`'s doc now says a hit costs one `mmap` and one pass over the mapping to verify the tag, which faults in every page ahead of the deserializer.

**7.** Parts are length-prefixed.

## Verification

```
cargo nextest run -p tari_engine --features wasm-cache
    Summary [357.698s] 478 tests run: 478 passed (1 slow), 0 skipped
cargo lints clippy -p tari_engine --features wasm-cache --all-targets   # clean
```

sdbondi (commented):
_Posted by Claude._

**Comment only** (GitHub refuses approve on your own PR; I would approve) — 0 blocking, 0 non-blocking, 1 nit · 7 resolved since `8832a5ab` (2 blocking, 3 non-blocking, 2 nits) · head `01675927`

Everything from the last pass is genuinely closed. Re-verified at the new head:

**Blocking 1 — header layout inside the fingerprint.** `include_str!("cache.rs")` is now a hashed part (`cache.rs:101-103`), so the field order in `write_now` (`:815`) and the `read_field` calls in `try_load` are covered by the same text that defines them. `HEADER_FIELD_COUNT` and `HEADER_BYTES` were dropped from the value list, which is right — they are inside the included text, and hashing them separately was the weaker version of the same thing.

**Blocking 2 — `restrict_dir_to_owner` no longer refuses to start.** Returns `()`, both failure paths warn (`:310-319`, `:331-339`), and `open` calls it unconditionally (`:447`). Neither bootstrap can fail on it any more. The group-shared-directory consequence is now stated as intended behaviour at `:301-303` rather than inherited by accident — your call, and a defensible one.

**Non-blocking 3 — `WASM_LIMITS`.** All seven fields hashed, and the destructuring at `:110-118` makes a field added later a compile error until it is named. That is a stronger guarantee than the finding asked for: the gap cannot silently reopen.

**Non-blocking 4 — the `unsafe` block's SAFETY comment** (`:683-689`) now rests on the tag and the directory mode and describes the ELF fall-through, instead of the "a deserialize failure simply triggers the recompile fallback" claim this branch disproved.

**Non-blocking 5 — `write_durable`** (`:884-891`) no longer justifies itself by the CRC's coverage. The replacement is accurate: a torn artifact is a tag mismatch, so the `sync_all` buys the latency of not mmap-and-hashing a doomed file rather than safety.

**Nits 6 and 7** — `try_load`'s doc (`:589-594`) now says the tag's pass faults in every page ahead of the deserializer, and the parts are length-prefixed (`:105`).

## Nit

`restrict_dir_to_owner`'s new rationale names a consequence the code does not produce — `cache.rs:299-301`

> a node that has to recompile every template is a better outcome than one that will not start

When the chmod fails nothing makes the node recompile: it warns and then loads from the unrestricted directory exactly as before. The real trade being made is "run with a directory this process does not exclusively own" versus "do not start", which is still the right call — the sentence just describes the wrong alternative.

Related, in the same function: the "restricting it to {:#o}" warn at `:324-330` fires before the attempt, so a failed chmod logs that it restricted the directory and then logs that it could not. Moving it after the `set_permissions` (or folding the two into one message) keeps the log honest.

## Still outside the fingerprint, as documented

The wasmer version, unchanged and correctly argued: `MetadataHeader`'s ABI version turns a format change into a miss, and the middleware fork's `=7.4.0` pins mean the resolved version moves only when a manifest does. A wasmer bump that keeps the format and changes codegen remains the upgrade's own responsibility.

CI is pending at this head — `test (1..3/3)`, `clippy`, `integration test` and the fuzz legs have not reported. That does not change the read of the code; it is the only reason I would hold rather than click, if clicking were available on your own PR.

sdbondi:
_Posted by Claude._

Both nits fixed in `6f19c16`, and both were right.

**The rationale named an alternative the code does not produce.** A failed chmod leaves the node loading from the directory exactly as it found it — nothing recompiles. The sentence described the consequence of a *different* design. It now names the trade actually being made: running against a directory this process does not exclusively own, versus not starting at all.

**The warn fired before the attempt.** So the failure path logged "restricting it to `0755`" and then "could not restrict it" — the first line asserting something that had not happened yet and then did not. `set_permissions` is now matched on: the success arm says it restricted the directory, the failure arm says it could not and that the node is continuing anyway. The two failure paths (unreadable mode, failed chmod) share one message, since what an operator needs from either is the same fact.

```
cargo nextest run -p tari_engine --features wasm-cache --lib
    Summary [2.755s] 43 tests run: 43 passed, 0 skipped
cargo lints clippy -p tari_engine --features wasm-cache --all-targets   # clean
```

The full suite passed 478/478 at `0167592`; this commit touches a doc comment and two log lines.

On holding for CI — agreed, and that is the right instinct on a change whose whole subject is what a node executes. It is green through `clippy`, `fmt`, `machete`, `prettier` and the fuzz legs at the previous head, with `test (1..3/3)` and `integration test` re-running now.

sdbondi (commented):
_Posted by Claude._

**Comment only** (GitHub refuses approve on your own PR; I would approve) — 0 blocking, 0 non-blocking, 0 nits · 1 resolved since `01675927` (1 nit) · head `6f19c166`

Both halves of the remaining nit are closed, re-verified at this head:

- `cache.rs:299-304` now names the alternative the code actually takes — "running against a directory this process does not exclusively own" versus not starting — instead of a recompile that never happens.
- The success message moved into the `Ok` arm as past tense (`:325-331`), and both failure paths share `warn_unrestricted` (`:308-317`), so a failed chmod logs once and does not first claim it restricted anything.

Nothing new introduced by the change. I have no findings against this head.

For the record, what the branch ends up resting on: the tag covers fingerprint, address, header fields and body, so every byte a hit serves is inside it; the fingerprint derives from `module.rs`, `metering.rs`, `bulk_metering.rs`, `limiting_tunable.rs` and `cache.rs` as length-prefixed text plus all seven `WASM_LIMITS` fields and `MAX_WASM_POINTS_PER_CALL`, with the destructure making a future limit a compile error until it is named; and the wasmer version is the one input left to the upgrade itself, which `MetadataHeader`'s ABI check and the `=7.4.0` pins bound.

CI at this head is still running — 3 of 7 reported, with `test`, `clippy`, `integration test` and the fuzz legs outstanding. Worth waiting for the `test` legs specifically, since `a_flipped_artifact_byte_falls_back_to_recompile` compiles a template three times over and is the slowest thing this file has ever asked of them.

sdbondi (commented):
_Posted by Claude._

**Comment only** (GitHub refuses approve on your own PR; I would approve) — 0 blocking, 0 non-blocking, 0 nits · 0 resolved since `6f19c166` (no new commits) · head `6f19c166`

CI has settled. Every leg passed except `Test Results (CI)`, which reports 2 fail / 2 483 pass. I chased both; neither is this branch.

- `stall_recovery.feature:7` — the standing red. It failed identically on the last three `development` runs (`b946ac2a`, `04c12830`, `031ed3cb`), and is the only failure on each of them.
- `transfer.feature:62` "Transfer tokens to existing account" — died in `Given a network with registered validator VN and wallet daemon WALLET_D`, before any Ootle code ran: `Wallet MINOTARI_WALLET did not have at least 20000000 uT after 100 seconds (balance: 0 uT, pending: 203090087002 uT)`. The funds were mined and sitting unconfirmed when the 100 s gate expired — the L1 wallet funding timeout that usually lands on `indexer.feature:125`, on a different scenario this time.

The three `test` legs, `clippy`, `fmt`, `machete` and the four fuzz targets are green, which is where the new `a_flipped_artifact_byte_falls_back_to_recompile` and `an_artifact_under_the_wrong_address_falls_back_to_recompile` actually run.

Nothing outstanding from me on this branch.

crates/engine/src/wasm/cache.rs
```diff
@@ -58,31 +61,83 @@ const LOG_TARGET: &str = "tari::ootle::engine::wasm::cache";
...
+        // derivation of the shape counts a hit serves verbatim.
+        include_str!("module.rs").as_bytes(),
+        // The per-operator cost tables the middlewares bake into the emitted instrumentation.
+        include_str!("metering.rs").as_bytes(),
+        include_str!("bulk_metering.rs").as_bytes(),
+        // The memory and table bounds applied when an instance's storage is created.
+        include_str!("limiting_tunable.rs").as_bytes(),
+        // This file: the order and meaning of the header fields a hit reads the shape counts out
+        // of live here and nowhere else, so swapping two of them is a fee change.
+        include_str!("cache.rs").as_bytes(),
```
sdbondi: Maybe create a hash at compile time and use that in the preimage

sdbondi: _Posted by Claude._

Doable — a build script rather than `const fn`, since `blake2` isn't `const` and the const-capable hash crates would be a new dependency for no gain. build.rs hashes the files, writes `OUT_DIR/fingerprint.rs` with a `pub const ENGINE_FINGERPRINT: &str`, and this file `include!`s it. The `LazyLock<String>` and every `*ENGINE_FINGERPRINT` deref go away, and the source text stops being embedded in the binary — 133 KB of it today.

Two things worth weighing before doing it here.

**It doesn't reduce how often the cache is invalidated.** A comment edit in this file still moves the fingerprint; it just moves it at compile time instead of on first use. Same orphaning, computed earlier.

**`include_str!` self-tracks and a build script doesn't.** rustc records `include_str!` in dep-info, so cargo rebuilds when a hashed file changes. A build script only reruns for the paths it declares, so if the hashed-file list and the `rerun-if-changed` list drift, the fingerprint goes stale — two nodes on different source computing the same value, which is the exact divergence this derivation exists to prevent. Containable by driving both from one array in the script, but it is a new way to get it wrong, and the current form cannot.

What does cut the invalidation is narrowing the preimage, and it splits cleanly:

- **This file can leave the hash entirely.** The only thing in it the fingerprint needs is which `ModuleShape` field each 8-byte header slot holds, and that is data. A `const HEADER_FIELDS: [HeaderField; 5]` that `write_now` iterates to emit, `try_load` iterates to parse, and the fingerprint hashes. Eviction logic, logging, docs and the 493 lines of tests then stop invalidating every node's cache. The descriptor has to *drive* both sites rather than describe them, or it becomes the bump list this PR deleted.
- **`module.rs` mostly can too.** `create_engine` is nothing but literals — opt level, `canonicalize_nans`, the thirteen feature flags, the page and table-element limits, the metering limit — all hashable as values. `validate_module_structure` is genuine logic, since its counts are served verbatim from a hit and priced, so it moves to its own small file and that gets hashed. `validate_functions`, `validate_module_exports` and `reject_disallowed_custom_sections` then stop mattering, and `validate_functions` re-runs on every hit anyway.
- **`metering.rs` / `bulk_metering.rs` stay as text.** The cost table is a ~500-arm `match` over `Operator`; the only exact value form is enumerating every variant by hand, which is the forget-to-add-it failure this PR removed. They also change rarely, and when they do the invalidation is correct — it is a fee change.
- **`limiting_tunable.rs` stays.** `memory_style` and `table_style` delegate to base *after* `adjust_memory`/`adjust_table`, so the clamping reaches codegen through the style decision.

That takes the hashed text from ~133 KB across five files to ~39 KB across three, and the two files under active edit stop touching the fingerprint at all.

Suggest doing the narrowing first and treating the build script as optional polish afterwards: it is then guarding three stable files instead of five active ones, which is also where the staleness risk is cheapest. Happy to do either here or as a follow-up — say which.

sdbondi: _Posted by Claude._

Done in `74fc50e`, both the compile-time hash and the narrowing.

**The hash is a `const`.** FNV-1a with an avalanche finalizer, in `const_hash.rs`. `ENGINE_FINGERPRINT` is a `&'static str` rather than a `LazyLock<String>`, so there is no runtime work and no `Blake2b` on this path. You were right that a const hasher beats the build script I suggested: `include_bytes!` keeps rustc's own change tracking, so the staleness hole I raised against build.rs — hashed-file list drifting from `rerun-if-changed` — does not exist here. Non-cryptographic is also the right call and is not even a narrowing: the Blake2b digest was already truncated to 64 bits for the filename.

**The preimage is narrower.** A digest over a file is only as precise as the file is narrow, so the two parts of `module.rs` that decide what an artifact contains moved into files of their own — `create_engine` to `engine_config.rs`, `validate_module_structure` to `module_shape.rs`. `cache.rs` leaves the preimage entirely; what it contributed was the header layout, which is configuration rather than logic. `HEADER_FIELDS` names each slot, `store` iterates it to emit and `try_load` iterates it to parse, so the order hashed is the order the file is written and read in rather than a restatement beside it.

133 KiB over five broad files → 44 KiB over five narrow ones, with `module.rs` and `cache.rs` both out.

Measured on the branch rather than argued:

| change | fingerprint |
| --- | --- |
| baseline | `54d3c81c9a4e9776` |
| a comment added to `cache.rs` | `54d3c81c9a4e9776` — unchanged |
| `canonicalize_nans(true→false)` | `3020173ad23a6803` |
| two `HEADER_FIELDS` entries swapped | `e5ee96bd19404e82` |

The last row is the blocking finding from the first pass, which is what put `cache.rs` in the preimage to begin with — the descriptor keeps that property while dropping the file.

481/481 engine tests pass with `--features wasm-cache`, clippy clean. One correction to something I asserted in chat while working on this: the hashed text is dropped at final link, not from the intermediate rlib — `libtari_engine.rlib` still contains it, and only a linked binary is free of it. That is the binary-size side, which was the lesser half of this; the invalidation behaviour is the table above.

sdbondi (commented):
_Posted by Claude._

**Comment only** (GitHub refuses approve on your own PR; I would approve) — 0 blocking, 0 non-blocking, 2 nits · 0 resolved since `6f19c166` (nothing was outstanding) · head `74fc50e7`

The branch was force-pushed onto a `development` that now carries #2666/#2667/#2669; the PR's own diff is the six files. New commit is `perf(engine): narrow the fingerprint's preimage and derive it at compile time`, which is a real redesign rather than a touch-up, so I re-read all of it.

**I recomputed the digest independently** — FNV-1a in Python over the domain separator, the five files' bytes with their length prefixes, the five header-field names, then the eight limits as LE `u128` — and got `54d3c81c9a4e9776`, matching the figure in the description. That confirms the preimage is exactly what is documented, in that order, and that `cache.rs` is genuinely out of it.

## What I checked that could have been wrong

**`reject_disallowed_custom_sections` staying in `module.rs`, now unhashed, is correct — not a regression.** It was covered incidentally by the old `include_str!("module.rs")`, and the obvious worry is that tightening the allowed-section set would leave a warm node serving a template a cold node refuses. It does not, because the cache-*miss* path does not run it either: `load_template_from_code` (`module.rs:87-90`) calls `validate_module_structure` alone, and only `prevalidate_code` / `validate_code` — the registration path — reject sections. `module.rs:68-69` says as much. Warm and cold agree either way, so it is not fingerprint-relevant.

Everything else left in `module.rs` is in `finalize_loaded_module`, which both paths run: `validate_template_abi_version`, `validate_functions`, `validate_module_exports`, `load_template_def_from_custom_section`, `Instance::new`. Correctly excluded. The two halves that are path-asymmetric — `create_engine` and `validate_module_structure` — are the two that moved into files of their own. The split is complete.

**`HEADER_PAD_BYTES` / `HEADER_BYTES` dropping out of the preimage is safe.** Changing either moves the artifact's start, so a reader computes the tag over a different split of the same file and misses at 1 − 2⁻³² rather than parsing old bytes under a new layout. Weaker than hashing the value, but it fails closed.

**`static ENGINE_FINGERPRINT: &str = match str::from_utf8(&ENGINE_FINGERPRINT_HEX)` compiles.** Reading another static's value is forbidden in a `const` initializer but allowed in a `static` one; I built the pattern standalone to be sure rather than assume. `ModuleShape` derives `Default`, so `ModuleShape::default()` in `try_load` is fine, and in `store` the `chunks_exact_mut(8).zip(HEADER_FIELDS)` writes five of the six chunks and leaves the tag slot to the explicit write below. No `*ENGINE_FINGERPRINT` derefs survive the type change.

## Nits

**1. `HeaderField::name` is the fingerprint's only witness to what a slot means, and nothing ties it to `read_from` / `write_into`** — `cache.rs:150-180`

Repurposing a slot means editing the two match arms; the digest moves only if the name string is edited too. The doc says the name exists "so that repurposing a slot moves the digest even if its position does not", which is true only as long as someone keeps the three in step. It is a far smaller thing to remember than the seven-item list this replaces — one local correspondence in one `impl` — but it is still a thing to remember, in a PR whose whole argument is that those do not survive. Worth a line in the `impl` saying the name must describe what `read_from` returns, if nothing cheaper ties them.

**2. The commit message's "44 KiB" is off** — the five hashed files are 49 266 bytes (48.1 KiB). The "133 KiB" before figure matches the old five at 133 509 bytes read as KB, so the two numbers are not in the same unit either. The claim's direction and rough magnitude hold — roughly a third of the text it was.

## Sound, re-verified at this head

- The order `HEADER_FIELDS` declares is the order `store` writes, `try_load` parses and the fingerprint hashes, driven off the one array. That closes the swap-two-fields hazard at its source rather than by hashing the file around it.
- All seven `WASM_LIMITS` fields plus `MAX_WASM_POINTS_PER_CALL` are still hashed as values, and the destructure still makes a field added later a compile error until it is named.
- `const_hash`'s length prefixes are real: `a_part_boundary_is_not_a_byte_boundary` fails without them, and `init` is a fixed domain followed only by prefixed parts, so there is no ambiguity at the front either.
- FNV-1a plus a splitmix finalizer is the right tool here — every input is this build's own source, so there is no party to search for a collision, and the Blake2b it replaces was truncated to the same 64 bits for the filename anyway.
- The integrity tag is unchanged and still covers fingerprint, address, header fields and body.

CI is running at this head; `clippy`, the three `test` legs, `integration test` and the fuzz targets have not reported yet.

sdbondi (commented):
Please move the cache-related modules into a `cache` module

sdbondi:
_Posted by Claude._

Done in `414716a`, along with the two nits from the previous pass.

**`wasm/cache/{mod.rs, const_hash.rs}`**, moved with `git mv` so the history follows. `const_hash` is now private to `cache`, which is where it belongs — the fingerprint is its only user.

`engine_config.rs` and `module_shape.rs` stayed outside. They are not cache concerns: `module.rs` calls `create_engine` and `validate_module_structure` directly, so filing them under `cache` would point the dependency backwards. They are separate files because the fingerprint needs narrow ones, and the `../` in the `include_bytes!` paths now shows at a glance that the preimage reaches outside the cache module.

**Nit 1 — the name was the only witness.** Fixed at the source rather than with a comment. The slot names are gone; the fingerprint now hashes what the slots *answer*: a `ModuleShape` carrying a different value in every field, read back through `HeaderField::read_from`. Reordering permutes the sequence, repurposing replaces a value, and the sequence is produced by the accessor a hit runs rather than restated beside it, so there is nothing left to keep in step.

Measured on the branch, including the case the name list would have missed:

| change | fingerprint |
| --- | --- |
| baseline | `0876af45ee8b9a2f` |
| comment added to `cache/mod.rs` | `0876af45ee8b9a2f` — unchanged |
| two slots swapped | `4450fb067d597e65` |
| a slot repurposed, name untouched | `dde6182482199faa` |

Your nit also pointed at something I had not covered: the witness constrains `read_from` and says nothing about `write_into`. I checked whether an existing test caught a mismatch and it does not — `round_trip_compile_then_deserialize` asserts the template name, code size and function count, never the shape. `every_header_slot_reads_back_the_field_it_wrote` now does.

`const_hash::values` is length-prefixed like `part` while I was there, so a value moving between two adjacent runs cannot slip past the boundary.

**Nit 2 — the sizes.** You are right and both figures were wrong, not just the unit: it is **48.1 KiB over five files, from 130.4 KiB**. Corrected in the description and in this commit's message.

482/482 engine tests passed on the commit before the move; the post-move run and CI are going now, and I will report anything that is not the known cucumber flakes.

sdbondi (commented):
Reviewed tested and approved

sdbondi (commented):
_Posted by Claude._

**Comment only** (GitHub refuses approve on your own PR; I would approve) — 0 blocking, 0 non-blocking, 2 nits · 1 resolved since `74fc50e7` (1 nit) · head `414716ab`

`refactor(engine): let each header slot witness its own meaning` closes the nit, and closes it better than the line of documentation I suggested.

**Resolved — the slot's meaning is no longer restated alongside the accessor, it is read out of it.** `header_layout()` (`cache/mod.rs:206-213`) calls `HeaderField::read_from` against a witness shape whose every field is a distinct number, so what the fingerprint hashes is `[0x55, 0x11, 0x22, 0x33, 0x44]` — produced by the same accessor a hit uses. Repurpose a slot and `read_from` returns a different witness value; reorder `HEADER_FIELDS` and the sequence permutes. Either way the digest moves, with nothing to keep in step by hand. I recomputed it: `23a36aa3cb4bc31a`.

`write_into` is the half the digest cannot witness, and `every_header_slot_reads_back_the_field_it_wrote` (`cache/mod.rs:1210-1221`) is what pins it to `read_from`. The test has real teeth, because it zeroes everything else first: a `write_into` that crossed to a neighbouring field leaves the slot's own field at zero and the read-back fails. None of the witness values is zero, so no case is vacuous. Between the two, `read_from` answers to the fingerprint and `write_into` answers to `read_from`.

Two details in the same change that are load-bearing and easy to miss:

- `const_hash::values` now length-prefixes its run (`cache/const_hash.rs:58-67`). It has to: `header_layout()` and the limits are two adjacent `u128` runs now, and without the prefix a value moved from one to the other would leave the digest unchanged. The old `names` had its prefix per part; this keeps the property after the switch.
- `LAYOUT_WITNESS_SHAPE` is written out field by field rather than from a default (`cache/mod.rs:192-197`), so adding a field to `ModuleShape` fails to compile until someone visits this file — which is where they need to be, since a new shape field with no header slot would be recomputed on a cold load and left at zero on a warm one.

The module move to `cache/mod.rs` + `cache/const_hash.rs` is mechanical; the five `include_bytes!` paths and the `mod` declarations all follow it correctly.

## Nits

**1. The description's measured fingerprints are stale** — PR body, "Measured on this branch"

It quotes `54d3c81c9a4e9776` as the baseline, with `3020173ad23a6803` and `e5ee96bd19404e82` for the two perturbations. That was the `names()` preimage. This head's baseline is `23a36aa3cb4bc31a`, and the other two will have moved with it. The paragraph invites a reviewer to check exactly these numbers, so it is worth a re-run before merge.

**2. Still standing from the last pass:** `74fc50e7`'s commit message says the hashed text is "44 KiB"; the five files are 49 266 bytes (48.1 KiB), and the "133 KiB" before figure is 133 509 bytes read as KB — the two are not in the same unit. History, so only worth fixing if the branch is being rewritten anyway.

## Unchanged and still sound

The preimage's membership, which is the part worth being sure of: `engine_config.rs` and `module_shape.rs` are the two path-asymmetric halves of the old `module.rs`; `reject_disallowed_custom_sections` stays out because the cache-miss path does not run it either; everything in `finalize_loaded_module` stays out because both paths do. All seven `WASM_LIMITS` fields and `MAX_WASM_POINTS_PER_CALL` are hashed as values, with the destructure still forcing a new limit to be named. The integrity tag is untouched.

CI is running at this head.

PR #2669, opened 2026-09-21
chore(ci): run the supply-chain policy on every pull request

sdbondi:
Description
---

`deny.toml` had never been run. Two independent reasons:

1. It was on a cargo-deny **v1** schema — `[advisories] vulnerability/notice`, `[licenses] unlicensed/copyleft/allow-osi-fsf-free`. The current tool refuses to parse it: `error[unexpected-value]: expected '["all", "workspace", "transitive", "none"]'`.
2. Nothing under `.github/` invoked it, and `audit.yml`'s entire `push:` block — including its `**/Cargo.toml` / `**/Cargo.lock` path filters — was commented out, leaving a nightly cron and `workflow_dispatch`.

So the `[bans]`, `[sources]` and `[licenses]` policies were **unverified, not clean**. (The security review could not run it either: the installed cargo-deny 0.13.9 panics in `krates-0.13.0/src/builder.rs:751` before any check starts, which is its own reason the file went stale.)

**What running it actually found.** Each of these is a supply-chain fact that was true and unrecorded:

- **A git source.** 27 crates come from `https://github.com/tari-project/rust-libp2p.git`. `allow-git` was `[]` and `unknown-git` was `"warn"`, so the policy neither allowed it nor objected. It is now named in `allow-git`, with `unknown-git = "deny"` — a crate from anywhere other than crates.io is now a decision that has to be written down.
- **Two licences off the allow list.** `libbz2-rs-sys` (`bzip2-1.0.6`, via the integration tests) and `webpki-root-certs` (`CDLA-Permissive-2.0`). Both permissive; both now listed with a note on what they are.
- **A crate with no licence at all.** `libp2p-peersync` — which is this repository's own crate at `networking/libp2p-peersync/`, not one of the fork's. Its siblings `libp2p-messaging` and `libp2p-substream` both inherit the workspace licence; this one was never wired up. It now carries `license.workspace = true` (BSD-3-Clause) like they do.
- **Wildcard dependencies on 31 crates.** A path dependency has no version, so cargo records it as `*`. `allow-wildcard-paths` exempts that — but only for crates marked `publish = false`, because crates.io disallows path dependencies and a publishable crate carrying a `*` really would accept whatever the registry serves next. None of these crates is in `scripts/publish_crates.py::CRATES`, and none said so.

Every workspace member that is not in the publish set now says `publish = false` — 45 of the 74 packages `cargo metadata` reports. The 31 the wildcard check flagged, plus 14 more it did not: those escape it only because their wildcard path dependencies are dev-only or absent (`consensus_tests` has six of them under `[dev-dependencies]`), which makes them exactly as publishable by accident and exactly as unintended.

A new `.github/workflows/cargo-deny.yml` runs `cargo deny check bans licenses sources` on the pinned `EmbarkStudios/cargo-deny-action@v2.1.1`, triggered by `pull_request:` and by `push:` on the `**/Cargo.toml` / `**/Cargo.lock` / `deny.toml` paths — the latter restricted to `development` and `main`, since a pull request branch pushed to this repository would otherwise run the workflow twice. `audit.yml` keeps the nightly cron and `workflow_dispatch` it already had and carries the advisory report alone.

**Advisories are deliberately not in the gate, and the config says why.** The tree carries a live advisory backlog — rkyv, memmap2, crossbeam-epoch, h2, quinn-proto, quick-xml, bincode and more. Gating on it today means either blocking every pull request or filling `[advisories] ignore` with entries nobody has assessed, and an unassessed ignore reads exactly like a clean check. `rustsec/audit-check` already reports them nightly and opens an issue per finding — and because it fails rather than files on push and pull-request events, it stays in `audit.yml` on the cron. Splitting the two workflows is what keeps it there without a skipped job on every pull request. Moving `advisories` into the gate is what closing that backlog is for, and it is a follow-up, not this PR.

The first CI run also surfaced two yanked crates in the lockfile — `chacha20` and `spin` — which an offline `cargo audit -n` cannot see. `[advisories] yanked = "deny"` will catch them the day advisories join the gate.

**One thing this PR cannot do from the diff, and the title no longer claims it does.** `development`'s required status checks are `["clippy","check stable","fmt","file licenses","machete","prettier","test"]`. `cargo_deny` has to be added to that list in branch protection for a red run to stop a merge; until then the job reports. That is a repository setting rather than a file, so it is the follow-up that turns this from a check into a gate.

The diff does make that follow-up a one-line change with nothing else to discover. `cargo_deny` lives in its own `cargo-deny.yml` so the job id — the context branch protection names — is stable, and its `pull_request:` trigger carries **no path filter**. That last part matters: a workflow a path filter skips leaves its check *pending* rather than absent, so requiring the context while filtering the trigger would wedge every pull request that touches no manifest. The run is under a minute against `clippy`'s ten and the integration tests' fifty.

`push` keeps its filter and covers `development` and `main` — the branches a merge lands on — so a pull request branch pushed to this repository runs the workflow once rather than twice, and nothing requires a context there.

**`publish = false` is not a decision about anyone's roadmap.** It records that a crate is not in `scripts/publish_crates.py::CRATES` today. `ootle_sdk_core` and `ootle_sdk_ffi_c` are the two where that could be misread — `ootle_sdk_core` carries a `version = "0.41"` pin in the workspace dependency table alongside crates that are published. If either joins the publish set, the line comes out with the same commit that adds it.

Motivation and Context
---

Finding SCAN-05 of the engine security review — *"the supply-chain gates do not gate anything"* — plus the review's own coverage gap: cargo-deny never ran at any stage of it, so those three policies were reported as unverified rather than as passing.

How Has This Been Tested?
---

Locally, with cargo-deny 0.20.2 (0.13.9 replaced):

```
$ cargo deny check bans licenses sources
bans ok, licenses ok, sources ok
```

Each of the findings above was reproduced as a failure and then resolved; the run above is what is left. `cargo deny check advisories` still fails, by design.

`[graph] all-features` is `true`, because `cargo-deny-action` defaults its `arguments` input to `--all-features` and a dependency reachable only under a feature is still one this repository ships. The workflow now passes `--all-features` explicitly so that the graph CI resolves, the graph `deny.toml` describes and the graph a reviewer reproduces are the same one.

Two entries that were carried over from the v1 file are gone rather than repaired: the `ring` clarification never fired (its `LICENSE` hash stopped matching at 0.17.14, which declares `Apache-2.0 AND ISC` in its manifest — both already allowed), and `Unicode-DFS-2016` matched nothing and warned on every run.

Also checked that `publish = false` changes nothing about what gets published: `python3 scripts/publish_crates.py --list` prints the same 29 crates as before, none of which is among the 45 marked.

---

```
cargo install cargo-deny --locked   # 0.13.9 panics before it checks anything
cargo deny check bans licenses sources
```

Worth a second opinion:

- **`unmaintained = "workspace"`** rather than `"all"`. A transitive unmaintained crate is a fact about a dependency's choices, not something this repository can act on. If you would rather see all of them, it is a one-word change — but that check is not in the gate either way.
- **The 45 `publish = false` lines.** I derived the set from `cargo metadata` minus `publish_crates.py::CRATES`, but a second pair of eyes on "is anything here meant to be published one day?" is worth having.
- **Adding `cargo_deny` to `development`'s required checks.** That is the setting that turns this from a report into a gate, and it needs someone with admin on the repository.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

No code changes. `publish = false` affects `cargo publish` only, and only for crates that were never published.


sdbondi:
_Posted by Claude._

First CI run said something worth recording, and the PR now accounts for it.

**`cargo_deny` passed in CI** — bans, licences and sources, on the real tree. That is the first time those three policies have ever run in this repository.

**`security_audit` failed**, as it was always going to: re-enabling the `push` trigger put `rustsec/audit-check` on every pull request, where the advisory backlog fails it whatever the change does. `0aea72f` keeps that job on `schedule` and `workflow_dispatch`, which is where it was, and lets `cargo_deny` take the per-change triggers. A check that is red before you start is one nobody reads — and it would have made this PR's own gate unmergeable.

The run also turned up two things the review's offline `cargo audit -n` could not see, because yanked-crate detection needs the registry index: **`chacha20` and `spin` are both yanked** in the current lockfile. Not addressed here — `[advisories] yanked = "deny"` is in `deny.toml` and will catch them the day advisories join the gate — but they are now on the record rather than invisible.

sdbondi (commented):
_Posted by Claude._

**Changes requested** — 3 blocking, 3 non-blocking, 2 nits · head `6d1c66d` (posted as a comment; GitHub refuses `--request-changes` on one's own PR)

Verified locally with cargo-deny 0.20.2 against a detached worktree at `6d1c66d`: `bans ok, licenses ok, sources ok` reproduces, and the 31 `publish = false` lines are load-bearing — deleting them produces exactly 30 `error[wildcard]` diagnostics, one per crate the PR marked (plus `transaction_generator`, which was never flagged). The `allow-git` entry, the two added licences and the `Zlib`/`Unicode-3.0` additions all check out.

## Blocking

**1. `pull_request:` puts the advisory check into the gate — the one thing the description says it doesn't.** `.github/workflows/audit.yml:15`

The PR argues at length that advisories stay out of the pull-request gate because the tree carries a live backlog. But the new `pull_request:` trigger applies to the whole workflow, and `security_audit` is a job in it. `rustsec/audit-check` annotates and **fails** on push/PR events (it only opens issues on `schedule`). On this very PR it is red:

```
##[error]Critical vulnerabilities were found, marking check as failed
```
<https://github.com/tari-project/tari-ootle/actions/runs/35598505931/job/106328765104>

So every future PR touching `**/Cargo.toml` or `**/Cargo.lock` gets a red X from the advisory backlog — exactly the "blocking every pull request" outcome the description rejects, just without the merge block. It also trains people to ignore a red Security audit, which is the failure mode the whole PR is written against.

Either gate `security_audit` on the event (`if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'`) or move `cargo_deny` into its own workflow with the push/PR triggers and leave `audit.yml` on the cron.

**2. `[[licenses.clarify]]` for `libp2p-peersync` asserts the wrong licence for a first-party crate.** `deny.toml:73-76`

`libp2p-peersync` is not from the fork. It is this repository's own crate at `networking/libp2p-peersync/` — `Cargo.lock:6338` has no `source` line, and the PR already edits that manifest to add `publish = false`. Its siblings `networking/libp2p-messaging/Cargo.toml:7` and `networking/libp2p-substream/Cargo.toml:7` both carry `license.workspace = true`; `libp2p-peersync` is the one that was never wired up (it also hardcodes `version`/`edition` instead of inheriting).

So the clarification records **MIT** for a crate whose workspace licence is **BSD-3-Clause** (`Cargo.toml:6`), and the comment — "the field belongs in the fork's manifest, and this entry goes when it lands" — points at a repository that does not own the file. A supply-chain policy file stating the wrong licence for a first-party crate is the specific thing this PR exists to prevent.

The fix is one line in `networking/libp2p-peersync/Cargo.toml`, and the clarify block then comes out entirely. Verified: adding `license.workspace = true` and deleting the clarify still gives `licenses ok`.

**3. `cargo_deny` is not a required status check, so it still gates nothing.** `.github/workflows/audit.yml:47`

`development`'s protection currently requires `["clippy","check stable","fmt","file licenses","machete","prettier","test"]`. Neither `cargo_deny` nor `security_audit` is in it, so a red `cargo_deny` does not stop a merge. The title says "make the supply-chain gate gate something"; as it stands the job reports, it does not gate. Branch protection is out of the diff's reach, but the PR shouldn't land claiming a gate without that setting changed — please add `cargo_deny` to the required contexts (or say explicitly in the description that it's a follow-up).

## Non-blocking

**4. The `ring` clarification is inert, and it silently gained a licence that isn't on the allow list.** `deny.toml:66-69`

`expression` went from `MIT AND ISC` to `MIT AND ISC AND OpenSSL`, and `OpenSSL` is not in `[licenses] allow`. That passes only because the clarify never fires: the `LICENSE` hash `0xbd0eed23` no longer matches ring 0.17.14, so cargo-deny ignores the entry and falls back to the manifest. Proof — replacing the expression with `AGPL-3.0-only` still yields `licenses ok`.

The comment above it ("states its licensing in prose rather than in machine-readable metadata") is also no longer true: ring 0.17.14's manifest declares `license = "Apache-2.0 AND ISC"`, both already allowed. Dead config in a file the PR is otherwise rewriting for accuracy — drop the block.

**5. CI checks a different dependency graph than the documented repro.** `.github/workflows/audit.yml:57` vs `deny.toml:9`

`cargo-deny-action`'s `arguments` input defaults to `--all-features`, so CI runs `cargo deny --all-features check bans licenses sources`. `[graph] all-features = false` in `deny.toml` is overridden by the command line and is dead in CI, and the "How can a reviewer verify" command (`cargo deny check bans licenses sources`) resolves a smaller graph than the gate does. Both pass today, but a feature-gated dependency can make them disagree. Pass `arguments: ''` to match the config, or set `all-features = true` and say so in the repro.

**6. "Every crate in the workspace that is not published now says `publish = false`" isn't accurate.** commit message / description

13 workspace members are neither in `scripts/publish_crates.py::CRATES` nor marked: `applications/tari_watcher`, `clients/base_node_client`, `crates/consensus_tests`, `crates/state_tree`, `crates/template_provider`, `crates/wallet/sdk_services`, `crates/wallet/sdk_tests`, `crates/ootle_sdk_core`, `crates/ootle_sdk_ffi_c`, `networking/rpc_macros`, `networking/sqlite_message_logger`, `utilities/generate_ristretto_value_lookup`, `utilities/traffic-sim` (plus `networking/proto_builder`). They escape the wildcard check only because their wildcard path deps are dev-only or absent — `consensus_tests` has six of them under `[dev-dependencies]`. The accidental-publish guard therefore covers 31 of ~44. Either mark the rest or narrow the claim to "every crate the wildcard check flags".

## Nits

**7.** `"Unicode-DFS-2016"` (`deny.toml:49`) is no longer matched by anything and produces a `warning[license-not-encountered]` on every run. `Unicode-3.0` supersedes it — drop the old entry so a clean run is actually clean.

**8.** The `push:` and `pull_request:` path lists are byte-identical duplicates (`audit.yml:6-13` / `16-21`). For branches pushed to this repo rather than a fork, both fire and the workflow runs twice per push. A YAML anchor, or dropping `push:` in favour of `pull_request:`, avoids both.

sdbondi (commented):
_Posted by Claude._

**Changes requested** — 2 blocking, 3 non-blocking, 2 nits · 1 resolved since `6d1c66d` (1 blocking) · head `0aea72f` (posted as a comment; GitHub refuses `--request-changes` on one's own PR)

## Resolved

**1 (blocking) — advisories out of the per-change gate.** `audit.yml:32` now carries `if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'`. Checked the outcome rather than the commit message: on this head `security_audit` reports **skipping** and `cargo_deny` **passes** ([run 35599431190](https://github.com/tari-project/tari-ootle/actions/runs/35599431190)). The `if` is on the job, so the nightly cron and manual dispatch still reach `rustsec/audit-check` with `issues: write` intact. Correct fix.

## Still blocking

**2. `[[licenses.clarify]]` for `libp2p-peersync` asserts the wrong licence for a first-party crate.** `deny.toml:73-76` — unchanged at this head.

Restating the evidence, since nothing in the new commit touches it: `libp2p-peersync` is this repository's own crate at `networking/libp2p-peersync/`. `Cargo.lock:6338` has no `source` line, and this PR already edits that manifest to add `publish = false`. Its siblings `networking/libp2p-messaging/Cargo.toml:7` and `networking/libp2p-substream/Cargo.toml:7` both carry `license.workspace = true`; this one was never wired up.

The clarify records **MIT** for a crate whose workspace licence is **BSD-3-Clause** (`Cargo.toml:6`), and its comment sends the reader to a fork that does not own the file. Verified fix: add `license.workspace = true` to `networking/libp2p-peersync/Cargo.toml` and delete the block — `licenses ok` still holds.

**3. `cargo_deny` is not a required status check.** `development` requires `["clippy","check stable","fmt","file licenses","machete","prettier","test"]`. `cargo_deny` isn't in it, so a red run doesn't stop a merge and the title's claim isn't met yet. Branch protection is outside the diff — add the context, or say in the description that it's the follow-up.

## Still open (non-blocking)

**4.** The `ring` clarification (`deny.toml:66-69`) is inert — the `LICENSE` hash no longer matches ring 0.17.14, proved by swapping the expression for `AGPL-3.0-only` and still getting `licenses ok`. So the newly added `OpenSSL` term, absent from `[licenses] allow`, goes unnoticed, and the comment's premise is stale: ring 0.17.14's manifest declares `Apache-2.0 AND ISC`. Drop the block.

**5.** `cargo-deny-action`'s `arguments` input defaults to `--all-features`, so CI resolves a larger graph than `[graph] all-features = false` and than the repro command in the description. Pass `arguments: ''`, or set `all-features = true` and update the repro.

**6.** "Every crate in the workspace that is not published now says `publish = false`" is off by 13 members (`applications/tari_watcher`, `clients/base_node_client`, `crates/consensus_tests`, `crates/state_tree`, `crates/template_provider`, `crates/wallet/sdk_services`, `crates/wallet/sdk_tests`, `crates/ootle_sdk_core`, `crates/ootle_sdk_ffi_c`, `networking/rpc_macros`, `networking/sqlite_message_logger`, `utilities/generate_ristretto_value_lookup`, `utilities/traffic-sim`, plus `networking/proto_builder`). They escape the wildcard check only because their wildcard path deps are dev-only or absent.

## Nits

**7.** `"Unicode-DFS-2016"` (`deny.toml:49`) is unmatched and emits `warning[license-not-encountered]` on every run; `Unicode-3.0` supersedes it.

**8.** The `push:` and `pull_request:` path lists (`audit.yml:6-13` / `16-21`) are identical, so a branch pushed to this repo runs `cargo_deny` twice per push. A YAML anchor, or dropping `push:`, avoids it.

sdbondi:
_Posted by Claude._

All eight addressed in `6dc04a7`, six by changing the code and two by changing what the PR claims.

**2 (blocking) — `libp2p-peersync` is ours.** Confirmed: no `source` line in `Cargo.lock`, the crate is at `networking/libp2p-peersync/`, and its two siblings inherit the workspace licence. The clarify block asserting MIT is gone and the manifest now says `license.workspace = true`. `licenses ok` holds without the clarification.

**3 (blocking) — required status checks.** Verified: `development` requires `["clippy","check stable","fmt","file licenses","machete","prettier","test"]`, so a red `cargo_deny` stops nothing today. Branch protection is a repository setting rather than a file, so the description now says so under its own heading — the job reports until `cargo_deny` is added to that list, and adding it is the follow-up that finishes the change. Not something to slip in silently, so I have left it to a decision rather than making it.

**4 — the `ring` clarification is inert.** Confirmed at the source: `~/.cargo/registry/src/*/ring-0.17.14/Cargo.toml` declares `license = "Apache-2.0 AND ISC"`, both already on the allow list, and the `LICENSE` file is prose pointing at `LICENSE-other-bits` and `LICENSE-BoringSSL`. The hash no longer matches, so the block never fired and the unlisted `OpenSSL` term went unnoticed. Removed rather than repaired — nothing needs it.

**5 — the graph CI resolves.** Confirmed from the action's own manifest at `v2.1.1`: `arguments` defaults to `--all-features`. Resolved the other way to the one you suggested: `[graph] all-features = true` and the workflow passes `--all-features` explicitly. A dependency reachable only under a feature is still one this repository ships, so the wider graph is the one worth checking — and now the config, the workflow and the repro command all describe it.

**6 — the `publish = false` claim.** Your 13 plus `networking/proto_builder`, which is a workspace member by way of being a path dependency rather than by being listed. All 14 are marked, so the claim is now accurate rather than narrowed: 45 of the 74 packages `cargo metadata` reports. They were as publishable by accident as the 31 — they just had no wildcard path dependency outside `[dev-dependencies]` to give them away.

**7 — `Unicode-DFS-2016`.** Dropped. No `license-not-encountered` warnings left.

**8 — duplicate triggers.** Restricted `push` to `development` rather than reaching for an anchor: that keeps the check on the base branch after a merge, where no `pull_request` event fires, and stops the double run on a branch with an open pull request.

Re-verified with cargo-deny 0.20.2 on the new head, with `all-features = true` and both clarifications gone:

```
bans ok, licenses ok, sources ok
```

The only remaining diagnostics are the `duplicate` warnings (`multiple-versions = "warn"`, deliberate) and one `no-license-field` warning for `stringmetrics 2.2.2`, which is a warning because cargo-deny resolved it from the crate's licence files. That one is feature-gated, so it was already in CI's `--all-features` run before this commit.

sdbondi (commented):
_Posted by Claude._

**Changes requested** — 1 blocking, 0 non-blocking, 2 nits · 6 resolved since `0aea72f` (1 blocking, 3 non-blocking, 2 nits) · head `6dc04a7` (posted as a comment; GitHub refuses `--request-changes` on one's own PR)

## Resolved — each re-verified against a detached worktree at `6dc04a7`, not taken from the commit message

**2 (blocking) — the `libp2p-peersync` licence.** `networking/libp2p-peersync/Cargo.toml:6` now carries `license.workspace = true`, and both `[[licenses.clarify]]` blocks are gone from `deny.toml`. The crate now reports the workspace's BSD-3-Clause rather than an asserted MIT. `cargo deny check licenses` → `licenses ok`.

**4 (non-blocking) — the `ring` clarification.** Removed. ring 0.17.14's own `Apache-2.0 AND ISC` is on the allow list, so nothing was resting on it: licenses still passes without it.

**5 (non-blocking) — graph divergence.** `deny.toml:12` is now `all-features = true` and `audit.yml:67` names `arguments: --all-features`. CI and a bare `cargo deny check bans licenses sources` now resolve the same graph — the config drives the local run and the flag matches it. Both pass on the wider graph: local run gives `bans ok, licenses ok, sources ok`, and [CI's `cargo_deny`](https://github.com/tari-project/tari-ootle/actions/runs/35605087952) is green. The only remaining licence diagnostic is the informational `no-license-field` for `stringmetrics 2.2.2`, which resolves Apache-2.0 from its `license-file`.

**6 (non-blocking) — the `publish = false` claim.** Now true. Checked all 70 workspace members programmatically: every one is either in `scripts/publish_crates.py::CRATES` or marked `publish = false` — 46 marked, 0 left over. Also re-ran the reverse check: no crate in `CRATES` depends on a marked crate in any section, including dev and build. `publish_crates.py --list` still prints the same 29.

**7 (nit) — `Unicode-DFS-2016`** dropped; `warning[license-not-encountered]` no longer fires.

**8 (nit) — duplicate triggers.** `push:` is now scoped to `branches: [development]` (`audit.yml:8`), so a pull-request branch pushed to this repo runs the workflow once rather than twice. `security_audit` correctly reports **skipping** on this head while `cargo_deny` runs.

## Still blocking

**3. `cargo_deny` is not a required status check.** `development` still requires `["clippy","check stable","fmt","file licenses","machete","prettier","test"]`. The job runs on every dependency change and is green, but a red one would not stop a merge — so the title's "gate" is still a report. This is the one item the diff cannot contain. Add `cargo_deny` to the required contexts, or narrow the description to say the branch-protection change is the follow-up. Everything else here is ready.

## New nits

**9.** `ootle_sdk_core` and `ootle_sdk_ffi_c` are now `publish = false`, but `ootle_sdk_core` carries a `version = "0.41"` pin in the workspace dependency table (`Cargo.toml:154`) alongside the crates that are published. If those two are on a path to crates.io, the marking will need lifting when they join `publish_crates.py::CRATES` — worth a word in the description so the next person doesn't read `publish = false` as a decision that was made.

**10.** Scoping `push:` to `development` also means pushes to `main` no longer trigger the workflow. Harmless while pull requests target `development` and the nightly cron covers the rest, but if `main` ever carries a dependency change directly it goes unchecked until the next cron.

sdbondi:
_Posted by Claude._

Both nits fixed in `1c58a5e`; the blocking item is now the title's problem rather than the description's.

**3 (blocking) — the gate that doesn't gate.** You are right that narrowing the description was not enough while the title said otherwise. The title is now *"run the supply-chain policy on every dependency change"*, which is what the diff does. The description keeps the branch-protection follow-up under its own heading, with the current required contexts spelled out so the next reader can see exactly what is missing.

I have deliberately not added `cargo_deny` to `development`'s required contexts myself — that setting binds every open pull request in the repository the moment it lands, including ones whose authors have never seen this change, and it is not mine to make on the way past. It wants a maintainer pressing the button knowingly.

**9 — the two SDK crates.** Added to the description: `publish = false` records that a crate is not in `publish_crates.py::CRATES` today, not a decision about where it is headed. `ootle_sdk_core` and `ootle_sdk_ffi_c` are called out by name as the two where that could be misread, since `ootle_sdk_core` carries a `version = "0.41"` pin in the workspace dependency table. If either joins the publish set, the line comes out in the same commit.

**10 — `main` went uncovered.** Real gap, and the one place it bites is exactly where it matters: a release merge into `main` carries dependency changes. `push` is now `branches: [development, main]` — the branches a merge lands on — which still leaves a pull-request branch running the workflow once rather than twice.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 1 non-blocking, 0 nits · head `4bf0fd0` (GitHub refuses a verdict on one's own PR)

The split landed: `cargo_deny` runs on this head in 55s and `security_audit` no longer appears in the check list at all. Everything raised earlier in this thread stays resolved — `deny.toml`'s cross-references were repointed at the new file, and the now-vestigial `**/audit.toml` filter (no such file in the tree) is gone with it.

One thing I should have caught earlier, and it bears on the follow-up this description commits to.

## Non-blocking — the path filters and the required-check follow-up don't compose

`cargo-deny.yml`'s `pull_request:` trigger is filtered to `**/Cargo.toml`, `**/Cargo.lock` and `deny.toml`. That is right for cost, but it means the workflow does not run at all on a pull request that touches none of those paths — which is most of them.

From GitHub's own docs on skipped workflows:

> If a workflow is skipped due to path filtering, branch filtering or a commit message, then checks associated with that workflow will remain in a "Pending" state. A pull request that requires those checks to be successful will be blocked from merging.

So the moment `cargo_deny` is added to `development`'s required contexts — the follow-up named in the description — every pull request that doesn't touch a manifest will sit on `Expected — Waiting for status to be reported` and cannot merge. The check has to report on every pull request before it can be required.

Two ways out:

- **Drop the `paths:` filter from `pull_request:`** and let it run on every PR. It costs 55s of an `ubuntu-latest` runner, against `clippy` at ~10m and `integration test` at ~52m in the same run. Simplest, and it makes the required-check follow-up a one-line branch-protection change with no other moving parts.
- **Keep the filter and add a path-independent job** that reports the same context, the usual "always-run stub that short-circuits" pattern. More machinery for the same outcome.

I'd take the first. Either way it's worth saying in the description, so whoever makes the branch-protection change doesn't discover it by wedging the merge queue.

This does not block anything as the PR stands, since `cargo_deny` is not required yet.

sdbondi:
_Posted by Claude._

Taken, in `f854d28`, and this was the right catch — it is the kind of thing that only shows up after the branch-protection change, as a merge queue that has quietly stopped.

Confirmed the mechanism rather than taking the doc quote at face value: a workflow skipped by a path filter does not report its check as neutral or absent, it leaves the context `Expected — Waiting for status to be reported`. Branch protection treats that as unsatisfied, so `cargo_deny` required + `pull_request:` filtered to `**/Cargo.toml` would have blocked every pull request that touches no manifest — which is most of them, and none of which this policy has anything to say about.

Went with your first option. `pull_request:` carries no `paths:` at all now, with the reason written above it so nobody re-adds the filter as an optimisation. `push:` keeps its filter: it still covers `development` and `main`, and nothing requires a context there.

The description now says this under the follow-up heading, so whoever makes the branch-protection change sees why the trigger is shaped the way it is before they touch it.

On the split in `4bf0fd0` — the job id staying `cargo_deny` is what keeps the follow-up a one-line change, and `security_audit` dropping out of the pull-request check list entirely is better than the `skipping` row it used to leave.

PR #2670, opened 2026-09-22
chore(deps): close the advisory backlog and put advisories in the gate

sdbondi:
Description
---

`cargo deny check advisories` reported **15 findings**. It now reports none, and it runs in the gate.

**Eleven of the twelve were a bump away** — ten in the lockfile alone, and the two `quick-xml` DoS advisories by taking `cucumber` 0.21 → 0.23, which moves `junit-report` 0.8 → 0.9 and quick-xml 0.31 → 0.41. That one needed a manifest change and crosses breaking releases in `gherkin`, `typed-builder` and `synthez`, but the step definitions need no edits: `cargo check -p integration_tests --all-targets` passes unchanged. Running the suite is CI's job.

**The lockfile ten.** `crossbeam-epoch`, `h2`, `memmap2`, `quinn-proto`, `rkyv` (three advisories), `rustls`, and the two yanked crates the first `cargo_deny` run surfaced, `chacha20` and `spin`. Fifteen packages move in `Cargo.lock`, all semver-compatible; no manifest changes.

`rustls` is worth a note for whoever hits this next. `cargo update -p rustls` stops at 0.23.43, and `cargo info rustls` reports 0.23.43 as the latest — but 0.23.45 is published and unyanked, which the sparse index shows plainly. Taking `cargo info` at face value would have filed a fixable TLS advisory as unfixable. `--precise 0.23.45` takes it, pulling `aws-lc-rs`/`aws-lc-sys` forward with it.

**The remaining three carry an assessed reason and a way out**, which is what `[advisories] ignore`'s own comment has always asked for.

Reachability below is **per binary** — `cargo tree -p <bin> -i <crate>` — rather than across the workspace. That distinction did real work here: `--all-features` over every member unions graphs that are never linked together, so a crate the workspace "depends on" can be in nothing this project ships.

It ranks an advisory; it never on its own keeps one. Each entry also has to say why the fix is out of reach — `quick-xml`'s was not, and it is now fixed rather than filed.

| crate | advisory | in a shipped binary? |
| --- | --- | --- |
| `bincode 2.0.1` | RUSTSEC-2025-0141, unmaintained | **yes, all of them** |
| `rsa 0.9.10` | RUSTSEC-2023-0071, Marvin timing | no — `integration_tests` only |
| `tracing-subscriber 0.2.25` | RUSTSEC-2025-0055, ANSI in logs | no — dev-dependencies only |

- **bincode** is unmaintained by its authors' decision rather than defective; they consider 1.3.3 complete. It is in every binary, and it decodes exactly one thing: the `tari_mmr::MerkleProof` inside a burn claim, in the v1 format the base layer wrote it in — `claim_burn_proof_verifier.rs` pins `bincode::config::legacy()`. The encoding belongs to the L1, so moving off the crate is a base-layer change.
- **rsa** has no patched release at all. `pgp` is an *optional* dependency of `tari_p2p`, enabled only by its `auto-update` feature, which only the base layer's own `minotari_node` and wallet binaries turn on; it reaches the graph because `integration_tests` runs those to stand up a local L1. What `tari_p2p` does with it is verify detached signatures on release hash files against maintainer public keys — no private-key operation, so no timing oracle for Marvin to use.
- **tracing-subscriber 0.2** is reachable only through dev-dependencies: `ark-groth16`'s dev-deps pull `ark-relations`, which still depends on the 0.2 line. `cargo tree -i tracing-subscriber@0.2.25 -e normal` is empty.

**Advisories are now in the gate.** `cargo-deny.yml` runs `check advisories bans licenses sources`. The consequence is deliberate and worth stating plainly: a newly published advisory against anything in the tree turns **every** pull request red until someone bumps the crate or writes an entry — including pull requests that had nothing to do with it. That is the cost of the check meaning something. `rustsec/audit-check` still reports nightly and opens an issue per finding, so a new advisory is normally visible before a pull request runs into it.

Motivation and Context
---

The follow-up #2669 named: *"Moving `advisories` into the gate is what closing the backlog is for."* This closes it.

How Has This Been Tested?
---

```
$ cargo deny check
advisories ok, bans ok, licenses ok, sources ok
```

Two compile checks, for the two bumps with teeth:

- `cargo check -p tari_networking -p tari_ootle_walletd` — `rustls` 0.23.45 pulls `aws-lc-sys` 0.41 → 0.45, and that one compiles C.
- `cargo check -p integration_tests --all-targets` — the cucumber major bump, zero errors and no source changes.

---

```
cargo deny check advisories
cargo tree -p tari_validator_node -i rsa@0.9.10      # absent
cargo tree -p integration_tests   -i rsa@0.9.10      # present
```

The three ignore entries are judgement calls and are the thing to disagree with. In particular:

- **Is `bincode` acceptable?** It is the only ignored crate that is actually linked into what we ship. The advisory is a maintenance notice, not a defect, but "unmaintained and in every binary" is a position worth being deliberate about rather than inheriting.
- **Is the `rsa` reasoning right?** It rests on `pgp` being used only for public-key signature verification in `tari_p2p::auto_update`. I read that code to check; a second pair of eyes on the claim would be worth having, since it is what turns a live advisory into an accepted one.

Something this surfaced, not addressed here
---

`wasmer-compiler` requires `rkyv ^0.8.13`, so **rkyv's resolved version moves independently of wasmer's** — as it does in this PR, 0.8.16 → 0.8.18. The wasm module cache hands artifacts to `rkyv::access_unchecked` through wasmer, and `ENGINE_FINGERPRINT` hashes neither version, so a rkyv release that changed the archived layout would let a warm node read an old artifact under new rules. It is the same limitation that leaves the wasmer version outside the fingerprint — a crate cannot read a dependency's resolved version — and it is safe here, because 0.8.x promises archived-format stability and the three fixes are in the checked `access`/`from_bytes` paths rather than the unchecked one the cache uses. Worth its own issue rather than a fix smuggled into a dependency bump.

Breaking Changes
---

- [x] None
- [ ] Other - Please specify

Lockfile and CI configuration only; no source changes.


sdbondi (commented):
ACK

PR #2671, opened 2026-09-22
fix(walletd)!: gate webauthn enrolment on the wallet, not the username

sdbondi:
## The defect

With `auth.method = webauthn` — the *secure* configuration — any process that could reach the wallet's RPC port could enrol its own credential and mint itself an Admin token. No hardware authenticator was needed: `webauthn-rs` requests `AttestationConveyancePreference::None`, so a software credential is accepted.

Two things combined:

**No wallet-level enrolment gate.** `webauthn.reg_start` and `reg_finish` are unauthenticated by design — they are the path that creates the wallet's first credential — and the only check they made was `assert_user_not_registered(username)`, a `COUNT` filtered by the username *the caller supplies*. Enrol as `attacker` and the check never fires. Nothing asked whether the wallet already had a credential.

**Caller-chosen permissions.** `reg_finish` passed `request.requested_permissions` straight into `generate_auth_claims`, which only stamps an expiry. The requested set was never clamped.

The minted token is a user-session token, so it also satisfies `authorize_user_only` and reaches `auth.create_api_key` — the attacker could leave a persistent API key behind that outlives the JWT.

The wallet is single-user. There is no need for any username beyond the one owner, so refusing enrolment once the wallet is enrolled is the fix.

## The change

- **`WalletStoreReader::webauthn_has_any_registration`** — an `EXISTS` over `webauthn_registrations` with no username filter. It replaces `webauthn_is_user_registered`, which no longer had a caller.
- **The check runs in the same write transaction as the insert.** This is load-bearing and is the actual boundary. The old check and the insert were in different transactions, so two concurrent `reg_finish` calls could both pass and both insert — with a first-run-only rule that TOCTOU *is* the attack, since it lets an attacker race the owner's legitimate first enrolment. `WebauthnService::finish_registration` now decides it. The handler keeps an early rejection for a clearer error, but it is not what separates the two callers.
- **The bootstrap token carries a fixed permission set** (`Admin`, what the web UI already asked for) decided by the daemon. `requested_permissions` is gone from `WebauthnFinishRegisterRequest`, so the unauthenticated path no longer lets the caller shape the token it mints.
- **`webauthn.already_registered` answers the wallet-level question** — "is this wallet set up yet?", which is what a single-user wallet's UI is asking. It also stops an unauthenticated caller probing which usernames exist.

## Tests

`applications/tari_walletd/src/services/webauthn.rs`, keyed on the condition rather than the rendered error. Each was confirmed to fail with the guard removed:

| Test | Without the guard |
| --- | --- |
| `enrolment_under_another_username_is_refused` | **fails** — the second enrolment succeeds |
| `second_enrolment_is_refused_within_the_write_tx` | **fails** — both enrolments land |
| `first_enrolment_succeeds` | passes (as it must) |

`cargo nextest run -p tari_ootle_walletd`: 43 passed. `cargo lints clippy --all-targets` clean on the touched crates.

## Breaking

- `WebauthnFinishRegisterRequest` no longer carries `requested_permissions` (TS bindings and web UI updated; the UI sent exactly `["Admin"]`, which is what the daemon now issues).
- `WebauthnAlreadyRegisteredRequest` no longer carries `username`, and the response's meaning changes from "is this username registered" to "is this wallet set up".
- `WalletStoreReader::webauthn_is_user_registered` → `webauthn_has_any_registration`.
- All three library crates touched are published, so this needs `tari_ootle_wallet_sdk` 0.43.1 → 0.44.0 and `tari_ootle_wallet_storage_sqlite` / `tari_ootle_walletd_client` 0.43.0 → 0.44.0 at release time (per `scripts/crate_versioning.py impact tari_ootle_wallet_sdk --breaking`). Recent breaking PRs carry no Cargo.toml bumps, so this is a note for whoever cuts the release.

## Decisions this leaves open

1. **Adding a second device, and recovering a lost one.** Refusing all further enrolment means the owner cannot add a second passkey — and an owner who loses their only authenticator has no in-band way back: the routes out are `auth.method = None` in the config, or deleting the row by hand. Both wants are the same feature: a separate *authenticated* "add credential" flow. The unauthenticated `reg_start`/`reg_finish` pair must not be it.
2. **The bootstrap race on a genuinely fresh wallet.** First-run-only enrolment means whoever enrols first wins, and on a fresh daemon that can be an attacker. Binding first enrolment to loopback, or to a one-time secret printed to the daemon console at first start, would close it. This is the residual risk after this fix, not something the fix removes.
3. **Existing wallets with more than one registration.** The gate only blocks *further* enrolment, so any already-enrolled credentials stay usable. No migration is included.

Closes SCAN-01 (choke point CP-1) from the wallet security review. It breaks PATH-01/02/03/04/06 as *zero-credential* attacks but does not fix the underlying defects those paths exploit — each stays reachable by a legitimately authenticated caller.


sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 2 non-blocking, 4 nits · head `0c08da7e`

Comment rather than approve: this is the author's own PR, and `integration test` is still pending. Nothing blocks on the code.

## What I verified

The fix holds. Tracing it end to end:

- `finish_registration` (`applications/tari_walletd/src/services/webauthn.rs:126`) runs the count and the insert inside one `with_write_tx`, and `WriteableWalletStore::with_write_tx` (`crates/wallet/sdk/src/storage/mod.rs:47`) commits only on `Ok` and rolls back on `Err`, so the `AlreadyEnrolled` return leaves nothing behind.
- That transaction is genuinely exclusive: `SqliteWalletStore` is a single `Arc<Mutex<SqliteConnection>>` and `create_write_tx` holds the guard inside the `WriteTransaction` for its whole lifetime (`crates/wallet/storage_sqlite/src/lib.rs:73`), with `Drop` rolling back and releasing it. Check-and-insert is therefore atomic against every other wallet-store caller in the process. See non-blocking (2) — this, not the test, is what closes the race.
- `WriteTransaction: DerefMut<Target = ReadTransaction>`, so `tx.webauthn_has_any_registration()` inside the write tx resolves to the reader without a second connection. No self-deadlock.
- `assert_not_enrolled` opens and drops its read tx entirely inside `is_enrolled`, before `finish_registration` takes the write tx — no lock overlap, and no `MutexGuard` held across an `await`.
- No leftovers: `webauthn_is_user_registered` / `is_user_registered` / `requested_permissions` have no remaining references anywhere in the tree, and `ReadTransaction` is the only `WalletStoreReader` impl, so the rename has one call site to move.
- `DEFAULT_PERMISSIONS` is still used by `Login.tsx` and `AuthNone.tsx`, so dropping it from `Registration.tsx` leaves no dead export.
- The hand-edited binding matches what ts-rs emits for an import-less type (header comment, blank line, then the type) — same shape as `WebauthnStartRegisterRequest.ts`.

On the description's privacy claim for `webauthn.already_registered`: it holds, and the obvious escape hatch is closed too. `handle_start_auth` passes an explicit `Some(UserVerificationPolicy::Required)` into `start_passkey_authentication`, which means `new_challenge_authenticate_builder` never hits the `CredentialNotFound` branch for an empty credential list — an unknown username gets a normal challenge, not an error. So `auth_start` is not an enumeration oracle either.

## Non-blocking

**1. The authoritative check returns a worse error code than the courtesy one.** `assert_not_enrolled` (`applications/tari_walletd/src/handlers/webauthn.rs:45`) builds `invalid_request(..)`, which carries `ApplicationErrorCode::InvalidRequest` (400). `WebauthnServiceError::AlreadyEnrolled` (`services/webauthn.rs:19`) is a bare `std::error::Error` by the time it reaches `resolve_any_error`, so it falls through every `downcast_ref` and lands in the `GeneralError` (500) arm — with the same message text. The caller who actually loses the race, i.e. the one the boundary exists for, is the one that gets told "internal error". Mapping it at the call site keeps the two paths consistent:

```rust
webauthn_service
    .finish_registration(request.session_id, passkey)
    .await
    .map_err(|e| match e {
        WebauthnServiceError::AlreadyEnrolled => invalid_request(e),
        e => e.into(),
    })?;
```

`WebauthnServiceError` is already re-exported through `services::*`.

**2. `concurrent_enrolment_leaves_one_credential` does not run concurrently.** `tokio::join!` (`services/webauthn.rs:256`) polls both futures on one task; it gives concurrency at await points, not parallelism, and `#[tokio::test(flavor = "multi_thread")]` does not change that for futures joined this way. `finish_registration` has exactly one await — `SessionStore::remove`, a `tokio::RwLock::write()` that is uncontended here and so completes without yielding — and `with_write_tx` is fully synchronous. The first future therefore runs to completion, DB write included, before the second is polled. The test does fail with the guard removed (both inserts land, the `filter(..).count() == 1` assertion trips), so it is a valid regression test for the in-transaction check; it just isn't evidence about a race. Two options, and I'd take the second:

- `tokio::spawn` both halves with a barrier before the call — but the connection mutex serialises them anyway, so this still won't produce a real interleaving.
- Rename it to something like `second_enrolment_is_refused_within_the_write_tx` and say in the comment that atomicity comes from `SqliteWalletStore` holding the connection mutex for the transaction's lifetime, not from SQLite's own locking. That is the invariant a future reader needs, and it's the one that would break if the store ever moved to a connection pool — at which point the deferred `BEGIN` on line 75 would let two readers both see zero.

## Nits

**3. `WebauthnAlreadyRegisteredRequest.username` is now ignored but still required.** `handle_already_registered` takes `_request` (`handlers/webauthn.rs:73`), yet serde still demands the field. The Breaking section lists the `requested_permissions` removal but not this — the response's meaning changed from "is this username registered" to "is this wallet set up". Since the release is breaking either way, dropping the field would be cleaner than leaving a required argument that does nothing; at minimum it belongs in the Breaking list.

**4. `.limit(1)` on a `COUNT(*)` is a no-op** (`crates/wallet/storage_sqlite/src/reader.rs:1593`) — the aggregate always returns one row. Carried over from the old shape. `diesel::select(diesel::dsl::exists(webauthn_registrations::table))` says what's meant and stops at the first row.

**5. Lost-authenticator recovery.** With enrolment refused for the lifetime of the wallet, an owner who loses their passkey has no in-band way back — the routes out are `auth.method = None` in the config or deleting the row by hand. That's acceptable for an operator-controlled daemon, but it's a consequence of this change and belongs next to decision 1 in "Decisions this leaves open".

**6. Release-time, not this PR:** all three library crates touched are published and this breaks their public API. `scripts/crate_versioning.py impact tari_ootle_wallet_sdk --breaking` asks for `tari_ootle_wallet_sdk` 0.43.1 → 0.44.0 plus `tari_ootle_wallet_storage_sqlite` and `tari_ootle_walletd_client` 0.43.0 → 0.44.0 (both re-expose the SDK). Recent breaking PRs (#2667, #2668) carry no Cargo.toml bumps, so this is a note for whoever cuts the release rather than a change request here.

## Reservation I'm recording rather than raising

Decision 2 in the description — an attacker who reaches a genuinely fresh daemon before the owner wins the wallet outright, and now permanently, since the owner can no longer enrol over them. This PR narrows the window from "forever" to "until first enrolment", which is the right direction, and the loopback-or-one-time-secret binding it proposes is the thing that actually closes it. Worth an issue so it doesn't live only in this description.

sdbondi:
_Posted by Claude._

All six addressed in `4bccf104`. Verified each before acting on it.

**(1) 500 on the authoritative rejection — confirmed and fixed.** `resolve_any_error` (`server.rs:347`) downcasts to `HandlerError`, `JsonRpcError` and `AuthError`; a bare `WebauthnServiceError` matches none of them and lands in the `GeneralError` arm. Mapped at the call site exactly as suggested, so the loser of the race now gets the same 400 as the early rejection.

**(2) The test is not concurrent — confirmed.** `SessionStore::remove` (`webauthn_session_store.rs:88`) is the single await and takes an uncontended `RwLock::write()`, which completes without yielding, so `tokio::join!` runs the two in sequence. Took the second option: renamed to `second_enrolment_is_refused_within_the_write_tx`, dropped the misleading `multi_thread` flavor, and recorded the invariant — atomicity comes from `SqliteWalletStore` holding its connection mutex for the transaction's lifetime, and a connection pool would break it via the deferred `BEGIN`.

**(3) `username` on `already_registered`.** Dropped the field rather than documenting a required argument nothing reads, and added it to the Breaking list along with the change in what the response means.

**(4) `.limit(1)`.** Now `diesel::select(exists(webauthn_registrations::table.select(id)))`. It needs the explicit `.select()` — `exists` over a bare table doesn't satisfy `SelectQuery`.

**(5) Lost authenticator.** Folded into decision 1 in the description: adding a second passkey and recovering from a lost one are the same missing feature, an authenticated "add credential" flow.

**(6) Version bumps.** Added to the Breaking list as a release-time note, matching how #2667/#2668 left it.

On the recorded reservation — agreed, and I'd rather it lived in an issue than only in this description. Leaving that to the maintainer to file, since it is a design decision (loopback binding vs. a one-time secret) rather than a follow-up to this change.

sdbondi (commented):
_Posted by Claude._

**Comment only** — 0 blocking, 0 non-blocking, 0 nits · 6 resolved since `0c08da7e` (2 non-blocking, 4 nits) · head `4bccf104`

Everything I raised is genuinely fixed. Comment rather than approve only because this is the author's own PR; CI on this head is still running, so the approval gate isn't met either way.

## Resolved, re-verified

**Error code for a refused enrolment** (`handlers/webauthn.rs:117`) — `AlreadyEnrolled` is now mapped to `invalid_request`, so it downcasts to `JsonRpcError` in `resolve_any_error` and comes back as `InvalidRequest` (400) like the early check, instead of falling through to the `GeneralError` (500) arm. The `e => e.into()` arm keeps `SessionStore`/`WalletStorage` on the same anyhow path they took before, so nothing else changes shape. `WebauthnServiceError` reaches the handler through `services`' `pub use webauthn::*`.

**The concurrency test** (`services/webauthn.rs:243`) — renamed to `second_enrolment_is_refused_within_the_write_tx`, dropped to a plain `#[tokio::test]`, and the doc comment now names the real invariant: atomicity comes from `SqliteWalletStore` holding its single connection's mutex for the transaction's lifetime. The added note that a connection pool would leave the deferred `BEGIN` letting two readers both count zero is the right thing for the next reader to know. The assertions are unchanged, so it still fails with the in-transaction check removed.

**`WebauthnAlreadyRegisteredRequest.username`** — removed from `clients/wallet_daemon_client/src/types.rs:1477` and threaded all the way through: `useWebauthnAlreadyRegistered()` drops the parameter and the `queryKey` element, `webauthnAlreadyRegistered()` sends `{}`. `grep` finds no remaining reference to the field. The binding's `Record<string, never>` is what ts-rs emits for an empty named-field struct — it matches every existing pair in the tree (`AuthListSessionsRequest`, `AccountGetDefaultRequest`, `SettingsSetResponse`), and `{}` is assignable to it. `APP_NAME` is still used by `Login.tsx:42` and `Registration.tsx:81`, so the export doesn't go dead.

**The `COUNT` query** (`storage_sqlite/src/reader.rs:1591`) — now `SELECT EXISTS (SELECT id FROM webauthn_registrations)`. `id` is the table's primary key in `schema.rs:322`, `get_result` infers `bool` from the signature, and the `map_err` tail types correctly without a `?`.

**Lost-authenticator recovery** — folded into decision 1 with the two actual routes out (`auth.method = None`, or deleting the row). That's the thing an operator needs to know before this ships.

**Release-time crate bumps** — recorded in the Breaking section with the versions and the script that produced them, and correctly framed as a release note rather than something this PR should carry.

## Standing reservation, not a finding

Decision 2 still stands on its own: an attacker who reaches a genuinely fresh daemon before the owner wins the wallet, and now permanently, since the owner can no longer enrol over them. This PR moves that window from "forever" to "until first enrolment", which is the right trade, and the loopback-or-one-time-secret binding it describes is what closes it. Worth an issue so it outlives this description.

Nothing blocks. Once `clippy`, the three `test` shards and `web ui build` come back green on `4bccf104`, this is good to merge — it needs another maintainer to click approve.

sdbondi (commented):
Tested and approved

PR #2672, opened 2026-09-22
fix(engine)!: close the security review's panic and pricing residue

sdbondi:
Description
---

The tail of the engine security review, after #2666–#2669 closed the four choke points. No CRITICAL or HIGH remained; this is the LOW residue plus the two PLAUSIBLE items that were never verified either way. Three commits, independent of each other.

**1. `RISTRETTO_MSM` priced one of the two arguments it decodes** (PANIC-04)

`price` derived the term count from `args.arg_encoded_len(0)` — the point list. `dispatch` decodes argument 0 **and** argument 1 before it compares their lengths, so the scalar list is decoded unpriced. The old comment asserted the opposite: *"a caller cannot get free work out of an argument this never looks at."* It does look at it, three lines before the check that would have rejected it.

Declaring an empty point list and a megabyte of scalars was charged the flat `PER_RISTRETTO_MSM` base and nothing more. The charge now reads the longer of the two encoded lengths, which is what the decode work is proportional to. A well-formed call carries the same number of terms in both lists, so it pays exactly what it paid before.

**2. Whether an amount operator wrapped or panicked was decided by the build profile** (PANIC-03)

`op_impl!` expanded to the raw operator on the inner value, for both `Amount` and `PrecisionAmount`. The review rated this LOW on the reasoning that `Amount` is checked at every reachable call site, and that is true — every `-=` in `resource_container.rs` is guarded by an insufficiency check immediately above it, and `validator_fee.rs` uses `min`. But the finding is aimed at the wrong type, and understates what it found.

- **`Amount` is `u128`, and this workspace sets `overflow-checks = true` on release.** It already panicked rather than wrapping. What is wrong is that the guarantee is the profile's, not the type's: `tari_template_lib_types` is a published crate, and a template author compiles their own WASM under their own profile.
- **`PrecisionAmount` is backed by `bnum`, and `bnum` gates its overflow check on `debug_assertions`** (`bnum-0.13.0/src/int/ops.rs:258-266`), which `overflow-checks = true` does not set. So every `PrecisionAmount` operator wrapped silently in **release**, including in a validator. The type exists for AMM arithmetic; a wrapped amount there is value creation. The in-tree caller, the `liquidity_pool` builtin, uses `checked_*` throughout and is unaffected — but nothing made that a requirement rather than a habit.

Both types now check for themselves. The panic messages are the ones Rust emits for the primitive operators, since these operators are meant to behave as those do (a divisor takes a second form that names the by-zero case separately, because the checked operator answers `None` to both and they are not the same mistake). The compound-assignment forms are generated from the binary ones so they cannot drift apart, `Sum` folds through the checked add, `pow` goes through `checked_pow`, and `PrecisionAmount::neg` through a new `checked_neg`. `checked_rem` is added to both to back `%`.

`tari_template_lib_types` goes 0.32.0 → 0.32.1, per `scripts/crate_versioning.py impact`. No API is removed; `checked_rem` and `checked_neg` are additive.

**3. Four panic sites on or beside the execution path** (PANIC-06, PANIC-07, SCAN-11, SCAN-07 + UNSAFE-15)

The release profile sets `panic = "abort"`, so each of these aborts a validator process rather than failing a transaction. Two are unreachable today and are removed because being unreachable is not a property the code states; two are cheap to state properly.

- `ObjectIds` incremented the bucket/proof and uuid counters with a bare `+= 1`. **Settled as unreachable**: exhausting a `u32` needs 4.3e9 allocations against `MAX_WASM_POINTS_PER_TRANSACTION = 250_000_000`, and each allocation costs at least one metered engine crossing. Naming the exhaustion costs nothing and makes `new_bucket_id`/`new_proof_id` fallible like every other allocator on the type.
- `WorkingStateStore::down_utxo` and `down_confidential_output` asserted the agreement between a substate's address kind and its value with `expect`, immediately below a kind check that already returns `InvariantError`. They now return that same error. `new_vaults` filtered on `is_vault()` and then unwrapped both halves; it filters on both instead.
- `validate_bullet_proof` unwrapped `RistrettoAggregatedPublicStatement::init`. **Settled**: `init` has exactly one `Err` arm — a statement count that is not a power of two (`tari_crypto-0.23.3/src/extended_range_proof.rs:157-164`) — and the padding immediately above establishes that count. The review flagged this as a possible chain halt if any other arm existed. There is no other arm. It is now an `InvariantError`.
- `ZERO_SECRET_KEY` fabricated a `RistrettoSecretKey` by `transmute`ing `[0u8; 32]`, as a `pub const`, resting on the layout of two foreign `repr(Rust)` newtypes that guarantee none. It has no callers anywhere in the tree, so it is removed rather than rewritten.

Motivation and Context
---

Three review items are closed here without code, and are worth recording so nobody re-opens them:

- **SCAN-08 (wasmparser coherence) rests on a premise that does not hold.** The engine has no standalone `wasmparser` dependency — `module.rs`, `module_shape.rs`, `metering.rs` and `bulk_metering.rs` all import through `wasmer::wasmparser`, so validation and compilation cannot split, today or after a wasmer bump. The only other copy in the tree is `tari_engine_types`' own, behind the `extract-template-def` feature, which only the wallet daemon enables. It is not on the validate-then-compile path.
- **PATH-06-residual was addressed by #2668**, as an argued invariant in the doc comment at `crates/engine/src/wasm/cache/mod.rs:993-1009`: every artifact in the directory is one whose template substate was resolved before it was written, and templates are immutable, so an artifact that was right when written stays right. The comment also states the obligation a third writer would owe.
- **SCAN-02/03/04/10 and the two yanked crates were closed by #2670.** `cargo deny check advisories` is clean and `yanked = "deny"` is in the gate.

`PANIC-05` (manifest parser panics) remains confirmed not validator-reachable.

How Has This Been Tested?
---

```
cargo nextest run -p tari_engine -p tari_engine_types -p tari_template_lib_types --all-features
cargo lints clippy --all-targets -p tari_engine -p tari_engine_types -p tari_template_lib_types
cargo deny check advisories
```

694 tests pass. New tests, each confirmed to fail before its fix:

- `intrinsics::tests::msm_prices_every_argument_it_decodes` — no points, 1000 scalars. Before: `1000 unpriced scalars cost the same 100000 points as none`.
- `operators_reject_unrepresentable_results` on both amount types — add, sub, mul, div-by-zero, rem-by-zero, a compound-assign form, `Sum` and `pow`. Sixteen tests; all sixteen failed before, the `Amount` ones on the message and the `PrecisionAmount` ones on `bnum`'s.
- `encoded_term_len_is_at_least_the_minimum` (was `encoded_point_len_...`) now pins the 33-byte floor for scalars as well as points, since the price divides by it for both.

---

- **The `bnum` claim is the load-bearing one** and is worth reading at `bnum-0.13.0/src/int/ops.rs:258-266` rather than taken on my word: `#[cfg(debug_assertions)] strict_add` / `#[cfg(not(debug_assertions))] wrapping_add`. `overflow-checks` and `debug_assertions` are separate flags, and the release profile sets only the first.
- **The MSM pricing question** is whether `max(len0, len1)` bounds the decode work from above. Both a point and a scalar encode to at least 33 bytes, `dispatch` decodes both lists, and each is at most `max/33` terms.
- **The two "settled as unreachable" claims** — the `u32` points budget and `init`'s single `Err` arm — are the ones to disagree with if any of this is wrong. Both are stated above with their citations.
- Worth checking that the `IdProvider` signature change did not turn an infallible path fallible in a way that changes when a transaction aborts: the six call sites all already sat behind `state.id_provider()?`.

Breaking Changes
---

- [ ] None
- [x] Other - Please specify

Rust API only, no wire format, state layout or consensus behaviour change.

- `IdProvider::new_bucket_id` and `IdProvider::new_proof_id` return `Result<_, IdProviderError>`; `ObjectIds::next_bucket_id`, `next_proof_id` and `next_uuid_id` likewise. `IdProviderError` gains a `CounterExhausted` variant.
- `tari_engine_types::crypto::ZERO_SECRET_KEY` is removed.
- `tari_engine_types` is tier-3, so it moves with `workspace.package.version` at release rather than bumping here — but the next release of it must be a **minor**, not a patch.
- `Amount` and `PrecisionAmount` operators panic on an unrepresentable result. For `Amount` under this workspace's release profile that is already what happened; for `PrecisionAmount` it replaces a silent wrap.


sdbondi (commented):
_Posted by Claude._

**Comment only** (own PR — GitHub refuses approve on it; CI still pending) — 0 blocking, 5 non-blocking, 4 nits · head `71c5ab08`

I re-derived the three load-bearing claims rather than taking them on the description's word, and all three hold.

## Verified

- **`bnum` gates its operators on `debug_assertions`.** `bnum-0.13.0/src/int/ops.rs` `trait_fillers!` is `#[cfg(debug_assertions)] strict_add` / `#[cfg(not(debug_assertions))] wrapping_add`, and the same shape for `mul`. `overflow-checks = true` does not set `debug_assertions`, so every `PrecisionAmount` operator did wrap in release. The `checked_*` family it now routes through is genuinely checked — `bnum-0.13.0/src/bint/checked.rs:62-150` builds them from `overflowing_*`, not from the gated fillers — so the fix is real and not circular.
- **`RistrettoAggregatedPublicStatement::init` has exactly one `Err` arm.** `tari_crypto-0.23.3/src/extended_range_proof.rs:157-164`: `if !statements.len().is_power_of_two()`, then `Ok`. Nothing else. `crates/engine_types/src/crypto/range_proof.rs:69` is correct to call the remaining case an invariant.
- **The MSM under-charge was real and the fix reproduces the old price on the legitimate path.** `dispatch` decodes argument 0 at `intrinsics/mod.rs:156` and argument 1 at `:157`, and only compares lengths at `:158` — the old comment's "an argument this never looks at" was wrong. On the `max`: `RistrettoPublicKeyBytes` and `Scalar32Bytes` are both `#[cbor(transparent)]` over `[u8; 32]` with `#[cbor(with = "minicbor::bytes")]` and the same `fixed_hex` serde helper (`crypto/ristretto.rs:15-24`, `crypto/scalar.rs:9-19`), so a well-formed call has `len0 == len1` exactly and pays what it paid before. No fee movement on any well-formed call.

Also checked and sound:

- `ZERO_SECRET_KEY` has no remaining reference anywhere in the tree.
- All 14 `new_bucket_id`/`new_proof_id` call sites are updated, and `next_bucket_or_proof_id` bumps *before* returning, so an exhausted counter refuses rather than handing the same id out twice.
- `PrecisionAmount` has no native host caller — the only mention outside `template_lib_types` is a comment in the `liquidity_pool` template. The new panics therefore cannot reach `panic = "abort"` in a validator; `Amount`'s are what `overflow-checks = true` already produced.
- The `ops::Add<u64>`/`Sub`/`Mul`/`Div` impls at `amount.rs:433-463` and `precision/amount.rs:543-573` delegate to the binary operators, so they inherit the check without needing their own arm.
- `crates/template_builtin/compiled/*.wasm` are checked-in artifacts and unchanged here, so no builtin template's behaviour moves in this PR.
- `Cargo.lock` carries only the version bump; the workspace pin is `"0.32"` (caret), so nothing else needs a pin update.

## Non-blocking

1. **`op_impl!` is exported public API and its arity changed in a patch release.** `crates/template_lib_types/src/amount/macros.rs:73` — `#[macro_export]`, no `#[doc(hidden)]`. A downstream `op_impl!(MyAmount, Add, add)` compiles against 0.32.0 and stops compiling on 0.32.1, which `^0.32` will pull in silently. The sibling macros in that file are already unusable outside the crate (`partial_eq_impl` calls the private `into_inner_value`, `impl_try_from` writes `Self(..)`), so `#[doc(hidden)]` on the lot is the honest label and costs nothing — otherwise this wants 0.33.0 rather than 0.32.1.

2. **`CounterExhausted` ships without a test.** `crates/engine_types/src/id_provider.rs:129`. The other two commits each carry a test that fails without the fix; this one carries the argument only. `ObjectIds`' fields are private but the `tests` module is in the same file, so `ObjectIds { bucket_or_proof_id: u32::MAX, .. }` followed by asserting `next_bucket_id()` errors pins both halves — that it refuses, and that it did not hand `u32::MAX` out a second time.

3. **The new error surfaces under a name that contradicts it.** `crates/engine/src/runtime/error.rs:196` is `TooManyOutputs(#[from] IdProviderError)` with `"Transaction generated too many outputs: {0}"`, so exhausting the bucket/proof counter reads as `Transaction generated too many outputs: Ran out of bucket or proof ids`. Buckets and proofs are transient handles, not outputs. The conversion predates this PR, but the new variant is the first one where the wrapper is plainly wrong about what happened.

4. **`msm_prices_every_argument_it_decodes` pins only half of what it is named for.** `crates/engine/src/intrinsics/mod.rs:350` asserts `lopsided > empty` with 0 points and 1000 scalars. That passes just as well if `max(len0, len1)` is later reduced to `arg_encoded_len(1)` alone — the point list would then be the free one, and no test would notice. The mirror case (1000 points, 0 scalars) closes it.

5. **`new_vaults` drops silently where its two siblings in the same commit raise.** `crates/engine/src/runtime/state_store.rs:297`. `down_utxo` and `down_confidential_output` turn "address kind and value kind disagree" into `InvariantError`; here the same disagreement makes the entry vanish from the iterator. Its only caller is the dangling-locked-value check at `working_state.rs:601`, so a dropped entry is a skipped safety check rather than a visible error. Same class of impossible state, opposite treatment.

## Nits

6. `crates/template_lib_types/src/precision/amount.rs:318` — `checked_div_ceil` rounds up with `div.add(I192::ONE)`, which is the exact bnum filler this commit's own argument says wraps in release, under a doc line promising `None` on overflow. Unreachable (`div == MAX` needs `d == 1`, which forces `rem == 0`), and `amount.rs:222` has the same shape on the raw `u128`. Worth closing while the reasoning is fresh, or stating it.
7. `crates/template_lib_types/src/precision/amount.rs:281` — `checked_rem`'s doc says `None` only for a zero divisor; bnum also answers `None` for `MIN % -1`. The operator handles that case correctly (falls through to the overflow panic, matching the primitive); only the doc is short.
8. `crates/template_lib_types/src/amount/ops.rs:17` — the `"attempt to divide with overflow"` arm is unreachable for `Amount`: `u128::checked_div` fails only on zero, and the by-zero guard runs first. Harmless, and the shared macro makes it hard to avoid.
9. `crates/engine_types/src/id_provider.rs:165` — `next_id` keeps the bare `+= 1`. It is genuinely bounded (`max_ids` is `max_substate_outputs = 1000`, checked immediately above), so this is only an observation that the commit's rule reached two of the three counters.

sdbondi:
_Posted by Claude._

All nine re-derived and all nine hold. Seven are addressed in `cae485d7`; two are left as they are, with reasons.

**5 — `new_vaults` dropped where its siblings raise.** Right, and the caller makes it worse than cosmetic: a dropped entry skips the dangling-locked-value check instead of reporting anything. It now returns `Result<Vec<(VaultId, &Vault)>, RuntimeError>` with the same `InvariantError`. The split matters — the address kind stays a filter (a non-vault address is simply not a vault), and only a vault-addressed value that is not a vault raises.

**3 — the wrapper was wrong about what happened.** `TooManyOutputs` is now `IdAllocation` and `#[error(transparent)]`; `IdProviderError`'s own messages already say which limit was hit. It was matched nowhere, so the rename is a rename only.

**2 — `CounterExhausted` has a test.** Writing it corrected me on the semantics: `bump` runs before the return, so a counter refuses on the call that *cannot advance it* and the last id is never served. That is the property worth pinning, so the test asserts `u32::MAX - 1` is served and the next call errors — losing one id is exactly what keeps one from being served twice. My first draft asserted `u32::MAX` was served and failed.

**4 — the mirror case.** Both directions now, 1000-and-0 each way, driven off one `msm` helper. As written it would indeed have survived a reduction to `arg_encoded_len(1)`.

**6 and 7.** `checked_div_ceil` rounds up through a checked add on both types — you are right that it was the one bnum filler this commit's own argument condemns, sitting under a doc line promising `None`. Unreachable either way, but it is a strange thing to leave in this particular PR. `checked_rem`'s doc now names `MIN % -1`.

**1 — the `op_impl!` arity break.** Real, and the honest label is what I took: `#[doc(hidden)]` on all six, plus a module doc saying they are exported because that is how a `macro_rules!` reaches the crate's other modules, not because they are callable. Correcting one detail: two of them are unusable outside the crate (`partial_eq_impl` / `partial_ord_impl` call the private `into_inner_value`), not "the lot" — `impl_from!` and `impl_try_from!` would both compile against a foreign type.

That leaves the version question genuinely open, and it is a maintainer's call rather than mine: `crate_versioning.py impact tari_template_lib_types --breaking` wants 0.33.0 and drags the entire tier-3 cohort to 0.42.0 plus ten independent crates, for an undocumented internal macro. I kept 0.32.1. If you would rather be strict about it, the bump is mechanical and belongs in the release rather than here.

**8 and 9 — left as observations, and both correct.** The `"attempt to divide with overflow"` arm is dead for `Amount` (`u128::checked_div` fails only on zero, and the by-zero guard runs first) but live for `PrecisionAmount` (`MIN / -1`), and the arm is shared. `next_id` keeps its bare `+= 1` because it is bounded by the `max_ids` check on the line above — the rule reached the two counters that had no bound, which is where it was needed.

695 tests pass; clippy clean on the three crates.

sdbondi (commented):
_Posted by Claude._

**Comment only** (own PR; CI still pending) — 0 blocking, 0 non-blocking, 2 nits · 7 resolved since `71c5ab08` (5 non-blocking, 2 nits) · head `cae485d7`

Re-verified each item against the new commit rather than the commit message.

## Resolved

1. **`op_impl!` arity** — `#[doc(hidden)]` on all five macros plus a module doc that says why (`amount/macros.rs:4-7`). That is the label; 0.32.1 is now the right bump.
2. **`CounterExhausted` untested** — `id_provider.rs:194-214`. It pins the half that matters: seeding `u32::MAX - 1`, `next_bucket_id` serves `u32::MAX - 1` and advances to `u32::MAX`, then `next_proof_id` errors, so `u32::MAX` is never served and never served twice. The same for `uuid`. Fails without `bump`.
3. **Misnamed wrapper** — `TooManyOutputs(#[from] IdProviderError)` → `IdAllocation`, `#[error(transparent)]` (`runtime/error.rs:195-196`). `MaxIdsExceeded` and `CounterExhausted` now speak for themselves. No reference to the old name or its message survives anywhere in the tree, including the feature files.
4. **Half-pinned MSM test** — `intrinsics/mod.rs:347-364` now runs both lopsided directions against the same `empty` baseline, so reducing `max(len0, len1)` to either argument alone fails it.
5. **`new_vaults` dropping silently** — `state_store.rs:295-309` now filters on the address kind only and raises `InvariantError` when the value disagrees, matching `down_utxo`/`down_confidential_output`. The caller takes it at `working_state.rs:601`. It collects into a `Vec` rather than staying lazy, which is bounded by `new_substates` and borrows `&self` for the whole of `validate_finalized` — no conflict there.
6. **`checked_div_ceil` rounding up unchecked** — both now go through `checked_add` (`amount/amount.rs:219-225`, `precision/amount.rs:316-322`), so the "or if the result overflows" in both doc lines is true rather than only unreachable.
7. **`checked_rem` doc** — `precision/amount.rs:280-281` now names `MIN % -1`.

## Standing nits

Both were flagged as non-issues last time and still are; noting only that they are unchanged, not that they need a commit.

- `amount/ops.rs:17` — the `"attempt to divide with overflow"` arm stays unreachable for `Amount` (`u128::checked_div` fails only on zero, and the by-zero guard runs first). The shared macro makes it hard to avoid and it costs nothing.
- `id_provider.rs:165` — `next_id` keeps the bare `+= 1`, bounded by the `max_ids` check immediately above it (`max_substate_outputs = 1000`).

Nothing blocking. Holding the approve click only because CI is mid-run (10 pass / 10 pending) — and because GitHub will not take an approval on your own PR regardless; this needs another maintainer.

sdbondi (commented):
Reviewed, Tested and approved

